PR #80, opened 2014-12-15
add versioning with versioneer

talumbau:
Adding support for `versioneer` a nice way of doing versioning. The installed package will be able to report its version, which will be useful for reporting results in the webapp.


MattHJensen:
@hdoupe, is versioneer.py still used for reporting versions on the webapp? 

cc @martinholmer

MattHJensen:
@hdoupe, sorry I wasn't very clear. What I meant was, does the webapp need Tax-Calculator to have versioneer.py. Otherwise @martinholmer and I would like to delete this file. Is it ok for me to delete versioneer.py from Tax-Calculator?

hdoupe:
Ohh, I see. Nope, the webapp doesn't require that Tax-Calculator have versioneer, but Tax-Calculator should report it's version in some way that can be accessed programmatically.

martinholmer:
@MattHJensen and @hdoupe,
After working some on PR #2087, it is now clear that `versioneer.py` (and some other files I don't understand) are necessary for the repository.  So, now I know what's necessary and what's not, but several of the necessary files are mysteries to me.

PR #85, opened 2014-12-19
rename to enforce the calc/apply convention

talumbau:
As we proceed with the using @jit on the various functions, the code is more clear we follow the convention:

```
def foo():
    # call to foo_apply(...args)

def foo_apply():
    # call to foo_calc(...args)
```

This PR just enforces the foo->foo_apply->foo_calc style for TaxGains


talumbau:
this is no longer relevant. closing.

PR #86, opened 2014-12-19
added time_it.py timing file

jlyons871:
Added a script, similar to test.py, but for timing instead of accuracy usage.

### DIFF
--- a/time_it.py
+++ b/time_it.py
@@ -1,0 +1,79 @@
+"""
+Timing file for calculate.py
+"""
+
+from pandas import DataFrame, concat
+from taxcalc.calculate import *
+from taxcalc.puf import *
+from taxcalc.parameters import *
+import taxcalc.parameters as parameters
+
+
+
+def run(puf=True, data=None):
+    """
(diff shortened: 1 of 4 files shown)


MattHJensen:
@jlyons871, what is the status on this PR? I still view this as a potentially useful tool, so I am hesitant to close it. 

From @martinholmer  in #322 
>The last commit to the func_timer branch was on Dec 19, 2014.
Is it really still active? If not, perhaps it should be deleted?

MattHJensen:
@talumbau, do you have any advice on this PR? We obviously haven't been using this feature, but do you think it would be helpful to clean it up? Otherwise I'll close it.

talumbau:
Ah, yes I think this has been abandoned and now what it is timing is not representative of what the code actually does. I'm more inclined to handle a timing measurement with a decorator style so that it is forwards compatible with future changes to Tax-Calculator. We can discuss details if desired. Closing for now.

MattHJensen:
Thanks @talumbau

PR #87, opened 2015-01-06
calc/apply beyond taxgains - vanilla

MattHJensen:
This is a vanilla-python implementation of the calc/apply style for the functions beyond taxgains. It still needs to be massaged to work well with jit (i.e., I haven't ensured that variable types are static).


jlyons871:
@MattHJensen, have you tested this on the full puf file? I get different results compared to the master when I run test.py

MattHJensen:
Is it expected that this fails the tests?

talumbau:
No it is not expected. We will need to track this down.

talumbau:
Looks like a Python 3 error according to Travis CI. @jlyons871, can you reproduce?

jlyons871:
It works in both Python 2 and 3 on my mac but I can reproduce on windows. I am looking into it.

talumbau:
I can reproduce with Python 3 on Mac. Weird.

PR #88, opened 2015-01-07
Refactor of taxcalc

talumbau:
Main contributions:

- complete localization of all variables
- year-indexed parameters are available through direct access, underlying arrays through `_` (e.g. `pm.almdep` and `pm._aldmdep`)
- `expand_array` can be used to expand parameters through the desired number of budget years with specified inflation rate

many more commits to come....


talumbau:
Clean up needed, but all variables localized now. Thinking about pushing to its own branch and then merging once the @jit stuff is ready.

talumbau:
the mega merge is complete. All the work happens in a `function.py` The "outer" functions reference parameter variable with `pm.` and the taxpayer data is reference via `pf.` This could also be abstracted out but it is not done in this PR.

MattHJensen:
This looks like a great improvement all around. A few comments and questions. I think the first one is the only that may be relevant pre merge. 

1. `functions.py` contains both numpy and jit versions of taxgains (and taxer). Is there a reason for this? 

2. Do you recommend that we provide IRS-form/line# comments in `parameters.py` (and easier-to-read formatting), or are you thinking that those are for a JSON file? 

3.  Later, inflation rates in `parameters.py` will need to vary year by year. I can create an issue. 
  
4. Let's revisit the structure of `functions.py` after merging this PR.

5. Later, we will want to call the tax record data something other than `puf`; one option would be `records`. I can create an issue.

talumbau:
1. The version of `Taxer` is just a slightly modified version of what is in the `jit` branch. There was a leftover additional definition of `TaxGains`, which I fixed.

2. Yes, this would be a good thing to figure out. I saw Amy's PR and my `parameters.py` has basically messed things up. Ideally we would have something that a tool could read and make nice documentation for. Maybe something with doc strings? Not sure here yet.

3. - 5. All sounds good to me.

MattHJensen:
1. Ok. We should be ready to drop taxer() as well, in favor of taxer_i. 
2. Ok. 

From my perspective, this is ready to be merged.

talumbau:
merged. Congrats @MattHJensen and @jlyons871!

MattHJensen:
You too, @talumbau!

jlyons871:
Looks great, very excited about the improvements!

PR #89, opened 2015-01-12
calc/apply beyond taxgains, vanilla 2

Copper-Head:
I'm terribly sorry for the delay with this. I'll be adding more commits throughout the day.
For some reason my `py.test` would report errors unless I tweaked the code a bit (see comments). I did so in order to have some sort of working test to compare my refactoring to. We should be able to add these jit statements back if we need them to later.

That being said, turns out the Travis CI build seems to have failed, complaining about `xrange`. I'm not sure why the tests on my machine have completed fine though. @talumbau, @jlyons871 thoughts?

### DIFF
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -159,28 +159,36 @@ def CapGains(p):
     header = ['c23650', 'c01000', 'c02700', '_ymod1', '_ymod2', '_ymod3', '_ymod']
     return DataFrame(data=np.column_stack(outputs), columns=header)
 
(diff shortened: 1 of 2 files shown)


MattHJensen:
@Copper-Head check out the progress in #87, which refactors everything before and after taxgains into a vanilla python calc/apply style. @jlyons871 is currently jitting these functions on his fork. The next steps are to 'pythonize' the taxcalc package, largely according to @talumbau's example that he sent around via email, and to combine functions. @Jlyons and @talumbau will coordinate these two steps.

MattHJensen:
Closing this as it covers ground already in master.

PR #91, opened 2015-01-12
fixed accuracy: bug in DataFrames concat with duplicate variables

jlyons871:
There are a few functions which return duplicates of the same variable to be saved in the output `results.csv` file.  These duplicates are causing them to be saved inaccurately when they are [concatenated](https://github.com/jlyons871/Tax-Calculator/blob/a0bc49e1f84b28676440c8ab8b1cc430ce56b6b3/test.py#L45), as is done numerous times in `test.py`.   As the test package uses [append](https://github.com/jlyons871/Tax-Calculator/blob/a0bc49e1f84b28676440c8ab8b1cc430ce56b6b3/taxcalc/tests/test_calculate.py#L20) instead, it does not seem to have the same inaccuracies. 

As for a few specific examples: 
The `_numextra` variable had the values from `c04100` and the `_txpyers` variable had the data from `c15200`, in [`StdDed`](https://github.com/jlyons871/Tax-Calculator/blob/a0bc49e1f84b28676440c8ab8b1cc430ce56b6b3/taxcalc/calculate.py#L417-427). And the `c26700` variable was inaccurately storing the values from `c262600` in [`AMTI`](https://github.com/jlyons871/Tax-Calculator/blob/a0bc49e1f84b28676440c8ab8b1cc430ce56b6b3/taxcalc/calculate.py#L931-947).  

They were being calculated correctly, but were being stored in the `results.csv`with the wrong values. These changes fix that, storing the correct values in the `results.csv` for continued use in accuracy checking.

### DIFF
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -366,6 +366,7 @@ def StdDed(p):
     global _feitax
     global _standard
 
+
     c15100 = np.zeros((dim,))
     StdDed_c15100(_earned, DSI, FLPDYR, p.DEFAULT_YR, p._stded, c15100)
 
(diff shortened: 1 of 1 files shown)


MattHJensen:
Looks good!

Copper-Head:
Oh, this is relevant to my refactoring. If it's ok with everyone I'd merge this and pull the changes into my "jit" branch.

PR #102, opened 2015-01-30
iterate jit and apply jit

talumbau:
### Decorate with `@apply_jit`

The `_apply` functions in `functions.py` are pretty mechanical. They just provide a looping mechanism for the `_calc`-style functions. It seems reasonable to automate their construction and use. `@apply_jit` does this. It programmatically creates functions of this form:

```
def f(x_0, x_1, x_2, x_3):
    for i in range(len(x_0)):
        x_0[i], x_1[i] = calc_function(x_2[i], x_3[i])
    return x_0, x_1
```
Where the number of `x_i`'s and how many inputs/outputs are determined by the arguments to the decorator. As long as each `_apply` function obeys this style, we can replace all of them in the code.


talumbau:
@MattHJensen @Amy-Xu, @jlyons871, I have used the new @iterator_jit in a few functions. Please have a look. I was hoping to merge this before doing all of the translation. That way others can pitch in. It's pretty simple to add the @iterator_jit. One thing is that there is a `parameters` argument so that we know whether to add an `[i]` when creating the `apply`-style function. We have to either be told that information or discover it at jit time. The burden seems low. Let me know your thoughts!

MattHJensen:
This is *very* cool! 

Do you expect that searching for parameters at jit time would be expensive? Would it be less expensive if our variable naming convention distinguished parameters (e.g., the leading `_` or something similar?)

On one hand, I like specifying parameters because it is a bit more explicit. That said, we will likely include this information in the doc string and so there's not a lot of added info(note 1, see below). On the other hand, it does require more work from the user and creates an opportunity for mistakes. 

(1) I'm not suggesting that we should grab the info from the doc string when jitting -- i don't trust our dedication to documentation thaaaat much!

Amy-Xu:
I feel whether we should ask users to specify parameters in the function signature should depend largely on how time consuming it is for the program to discover them at jit time. If it's very time consuming to interpret at jit time, then we should ask users to specify the parameters. If not, we probably should hide them because making the user specify them could be burdensome and confusing in some situations, and I don’t think the user/contributor gains much in terms of information from having them. 

Regarding using the underscore to distinguish variable from either puf or parameters, it might be able to shorten processing time compared to searching through the entire variable name list. We would need to take care of two things if we are going to let the program read where the variable's from using an underscore, though. First, there are many variables in PUF that have underscores at the beginning. We would need to change them. Second, in functions.py, many variables from parameters.py do not have underscores in front of them because we retrieved information for one year only and removed the underscore. We might need to have a slightly different naming convention to handle this — moving to the json lines might help solve this problem.

MattHJensen:
>  If it's very time consuming to interpret at jit time, then we should ask users to specify the parameters. If not, we probably should hide them because making the user specify them could be burdensome and confusing in some situations, and I don’t think the user/contributor gains much in terms of information from having them.

Yes, this makes sense to me as well. 

>  First, there are many variables in PUF that have underscores at the beginning. We would need to change them. 

Doesn't seem too hard, especially since these are all intermediate variables. But it could make test against the results from the SAS code in the future more challenging. 

>Second, in functions.py, many variables from parameters.py do not have underscores in front of them because we retrieved information for one year only and removed the underscore.

Is this going to be a problem anyways when we want webapp users to be able to turn inflation indexing on and off?

> We might need to have a slightly different naming convention to handle this — moving to the json lines might help solve this problem.

talumbau:
The main reason that I created the `parameters` argument for `iterate_jit` is that it wasn't clear to me what is a good algorithm for automatically discovering them. The basic idea is that the function will get arguments with these kinds of shapes passed in:

ndarray([130000, 1]), ndarray([130000, 1]), ndarray([2, 6]), ndarray([130000, 1]), ndarray([3, 1])

We can look at this and obviously know that the first two arguments and the fourth argument are columns of the PUF data (and so need a `[i]` when we call the `calc`-style function in the generated `apply`-style function. The third and fifth arguments should not be called with the `[i]`. Describing how to make this determination seems error prone. Something like "find the longest X by 1 array. anything with that dimension gets called with `[i]`. Anything else doesn't get called that way." That isn't hard but it seems like it might miss certain cases. Is that an appropriate algorithm for making the determination?

MattHJensen:
Ahhh, this is very elegant. I think it would work in 99% of cases and it wouldn't be too hard to build in a little warning system that will tell the user of the package that their tax data and parameters are incompatible (i,e., there is a param as long or longer than the records). 

BUT this solution might prevent some reasonable use cases in which you could envision someone wanting to run the calculator on just one tax record. For instance, the IRS could want to check one "problem'" return, or there might be a little tool where users could input their own tax characteristics and see how different tax reforms would affect themselves (Here is an example of a relatively simple version of something like this http://calculator.taxpolicycenter.org/.)  In these cases each record attribute would be one 1 by 1.

feenberg:
I should explain the purpose of the underscores. In SAS the tax calculator 
is supplied as an include file, and the scope of the identifiers in the 
include file is the same as the user written wrapper. The underscores 
allow a simple way for users to ensure that their variables are not 
overwritten by taxcalc variables, and that they do not overwrite taxcalc 
variables - the user just avoids beginning a variable name with an 
underscore.

dan

MattHJensen:
Would it be crazy/confusing to do something like this? In a way it would be a bit more explicit about what we are reading in and returning, with some magic in the middle... 
    
    @iterate_jit
    def func(p.param, r.record):
        calculated_value =  param*characteristic
        return r.calculated_value

talumbau:
With regard to the above, where does `characteristic` come from? Anyway, that idea is a bit too clever in my view. I believe we can write down some reasonable documentation that explains what `@iterate_jit` does and sufficiently scares people into using it with care. The basic idea is that the wrapped function is itself wrapped in two functions. That seems doable. I wouldn't want to modify the body of the wrapped function as well. Too magical in my view.

I am still unresolved about how to proceed with `parameters`. As you can see with your example of one tax record, we actually get a vital piece of information from a user-specified `apply`-style function (although 99% of it is purely mechanical) so my vote for now is to keep the `parameters` argument. Using this as it is will shrink `functions.py` by about half, which is pretty nice.

MattHJensen:
Meant r.characteristic rather than r.record. updated. But yea, see your point on being a bit too magical.... :sparkles:

Yes, I'm fine keeping the parameters argument in the decorator for now, and we can deal with it later if we come up with something nice upon more reflection. 

Want to merge this and then I can get started with the rest of the translation?

talumbau:
sounds good. merging!

MattHJensen:
Excellent. On Friday, @Amy-Xu found a case in which the function signatures were not consistent across the three layers of StdDed() -- this was the source of many inaccurate variables. In case more situations like this turn up, I am going to hold off on translating the rest of the functions to the @iterate_jit style until @Amy-Xu and @jlyons871  complete accuracy checking.

PR #103, opened 2015-02-02
[WIP] User-defined credits as parameters

talumbau:
- Parameterize UDFs as a single arithmetic formula where each group of parameters can take on different
  values. User specifies sequence of values which are executed for each record. Unlimited number of UDFs are allowed, but the formula for each "instance" of the UDF is the same. Only the values can change.


martinholmer:
Pull request #103, which was submitted almost 22 months ago on 02-Feb-2015, would seem to have been superseded by the the general method incorporated by pull request #1081.
Is there any need to keep #103 open?

@talumbau @MattHJensen @feenberg

martinholmer:
Pull request #103 has been superseded by the the general method in merged pull request #1081.

PR #104, opened 2015-02-02
updated test.py to reflect refactored style changes

jlyons871:
@talumbau can you confirm that the accuracy is correct?


jlyons871:
The output results.csv is the same as before. This should be ready to merge.

talumbau:
LGTM. +1

PR #120, opened 2015-02-10
Data Extrapolation; Marginal Tax Rate Calculator; Behavioral Effects Calculator; Git Setup Guide

MattHJensen:
First of all, sorry for not separating all of these things out. I now understand the value of branching...

The **data extrapolation** involves two stages, both of which are implemented in `puf.py`:

1. STAGE 1: Upon initialization of a puf, `read_blowup()` reads in a csv of "blowup factors" and does some data-prep. When the user increments the year, `blowup()` blows up the records data for the next year according to a pre-specified mapping between puf variables and the blowup factors.
2. STAGE 2: Upon initialization of a puf, `read_weights()` reads in a file of weights for every year in the budget window. Then, every time the year is incremented, `increment_year()` assigns s006 to the appropriate column in the weights dataframe. 

The **marginal tax rates calculator**, mtr() is a method on a Calculator object and is defined in `calculate.py`. When called, it returns an array of marginal tax rates with respect to some pre-specified income variable. An example of the mtr calculator in use can be found in `behavior.py`. 

The **behavioral effects calculator** is a class that we describe in `behavior.py`. `Behavioral example.ipynb` shows how to use the behavioral effects calculator. I am not sure whether a pure function might be better here than a class. 

The **Git setup guide** in `Git setup and workflow.md` is meant to get our contributors to the point where they can ask people for feedback on their work on a branch of their fork. 

Please give me your feedback on the content and implementation of these pieces


Behavioral example.ipynb
```diff
@@ -0,0 +1,152 @@
...
+     "input": [],
+     "language": "python",
+     "metadata": {},
+     "outputs": []
+    }
+   ],
+   "metadata": {}
+  }
+ ]
+}
```
talumbau: Would add more comments before each cell in the notebook, to explain what is happening.

Git setup and workflow.md
```diff
@@ -0,0 +1,61 @@
+**SETUP**
```
talumbau: suggest `Git_setup_and_workflow.md` instead of using spaces in the filename.

taxcalc/behavior.py
```diff
@@ -0,0 +1,67 @@
...
+class Behavior(object): 
+
+	def __init__(self, calcX, calcY, elast_wrt_atr = 0.4, inc_effect = 0.15):
+		self.calcX = calcX
+		self.calcY = calcY
+		self.elast_wrt_atr = elast_wrt_atr
+		self.inc_effect = inc_effect
+
+	@property
+	def calc(self):
```
talumbau: Since `calc` mutates the state of a `Behavior` object (well, actually its constituent calculators), I think it makes more sense to have this as a function, instead of a property.

Git setup and workflow.md
MattHJensen: Gotcha. b489837

Behavioral example.ipynb
MattHJensen: Fleshed this out c0d00aa . Will need to revisit after we move to `Calculator(params="filename", records="filename")`

taxcalc/behavior.py
MattHJensen: Made behavior into a function with 12118db 

Now you can run calc_y_behavioral = behavior(calc_x, calc_y)

This also allows you to compare calc_y without behavior to calc_y_behavioral, which is a useful feature.

MattHJensen:
@TJ, the Travis CI build failure continues to confuse me. In MANIFEST.in, I added

    include taxcalc/StageIFactors.csv
    include taxcalc/WEIGHTS.csv

In taxcalc/records.py, the default arguments are `StageIFactors.csv` and `WEIGHTS.csv`. In `tests/test_calculate.py` and `tests/test_records.py`, the default arguments are always relied upon for creating records.  Everything works fine on my machine, but the CI build fails.

PR #126, opened 2015-02-13
From file

talumbau:
Adds:

`Calculator.from_files`, `Parameters.from_file` (reading from JSON) and `Records.from_file`. `Calculator.from_files` dispatches to the other two.

### DIFF
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -4,6 +4,8 @@ import math
 import numpy as np
 from .utils import *
 from .functions import *
+from .parameters import Parameters
+from .records import Records
 
 
 all_cols = set()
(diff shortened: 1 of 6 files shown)


MattHJensen:
@talumbau, I just want to confirm that we will stop setting defaults in parameters.py once we have a default json parameter file ready. Is that correct?

talumbau:
Oh. that sounds fine to me. In that case, I suppose we would look for a default filename and throw an exception if it is not found, right?

MattHJensen:
Yes, that seems right. I just don't want us to have to maintain the list in parameters.py as well as maintain the default json param file.

talumbau:
Also, are you OK with:
```
calc = Calculator.from_files(parameters="paramfile.json", records="puf2.csv")
```
instead of:
```
calc = Calculator(parameters="paramfile.json", records="puf2.csv")
```

I don't have a strong preference.

MattHJensen:
My preference is the second since the only option we will be giving people is to read from a file and the second is simpler. Is there a downside?

talumbau:
OK. I'll add support for the second. This PR contains support for reading parameters from JSON (which is a blocking issue for @Amy-Xu ), so once I add in that change, if TravisCI is good, I'd like to merge.

MattHJensen:
Great!

PR #129, opened 2015-02-19
add _ospctax and _refund vars

MattHJensen:
Incorporate changes suggested by @feenberg here https://github.com/OpenSourcePolicyCenter/Tax-Calculator/issues/95#issuecomment-71083006. Note that I add _ospctax and _refund to the set of column names in test_calculate.py https://github.com/MattHJensen/taxcalc/compare/ospc_tax?expand=1#diff-03c8c8e4581079a84e9a80e191c841a3R65

### DIFF
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -143,7 +143,7 @@ class Calculator(object):
         F5405(self.parameters, self.records)
         C1040(self.parameters, self.records)
         DEITC(self.parameters, self.records)
-        SOIT(self.parameters, self.records)
+        OSPC_TAX(self.parameters, self.records)
 
     def calc_all_test(self):
         all_dfs = []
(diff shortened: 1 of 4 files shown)


MattHJensen:
Just added the new definitions of _ospctax and _refund from the latest sas taxcalc. This should mean a bit less work for @Amy-Xu in testing. 

    _refund = (c59660 + c11070 + c11600 + c10960 + c10950 + e11580 + e11450 +
               e11500)
    
    _ospctax = c09200 - _refund - e82040

talumbau:
hmm... does that mean this definition:

https://github.com/OpenSourcePolicyCenter/Tax-Calculator/blob/master/taxcalc/utils.py#L156

should be changed?

MattHJensen:
Once this is merged, you won't need to recalculate _refund and _ospctax -- you'll just be able to grab them from the calculator, right? 

I am going to followup with @feenberg about what some of these new variables are, but I wanted to get this up quick so @Amy-Xu can carry on with testing and you could use this instead of c05200 for the tables (although I see now that you didn't necessarily need this).

talumbau:
yep that sounds right. +1 from me.

PR #133, opened 2015-02-20
matching SAS changes

Amy-Xu:
I'm checking taxcalc's results using the data and SAS code from exall.2 folder, and the testing files from Jake's repo. 

So far, I have already moved all the relevant changes in Jake's repo to this latest version of taxcalc package, fixed a few bugs that Jake and I talked through emails, and updated a few functions including F2441 (still working on it). There's a temporary change made on _othadd just for testing purposes -- the value of this variable is missing in exall.csv so it's initialized to zero in our testing routine, but following the tax logic in both SAS and python it's not zero in every single case.

At this point, there're two more variables, c10300 and _eitc, not matching the SAS results. I just finished merging with the ospc master, and I am about to look into details.


feenberg:
_eitc should be the sum of c59680 + c59700 + c59720. If it isn't, then the 
mistake might be in my code. I'll be at the IRS tomorrow and that was 
something I was going to check. There were quite a few changes in the EITC 
calculation recently, so I would expect you to need to make changes in 
that area.

dan

Amy-Xu:
In the latest version of exall.csv from exall.2 folder, there are 6 variables (22250, 23250, 04470, 25470, 08000, 60100) that show up as both e-variables and p-variables. From the testing results I guess the e-values are the ones we should use, but in records.py we aliased the p-variables to the e-variables. The calculator might be confused on which one to use, so just for the sake of testing I deleted the p-varaibles columns (usually the values are missing) for these variables and changed the Es to Ps.  I have uploaded an exall.csv with all these changes to dropbox. https://www.dropbox.com/s/850uwqunupfoyly/exall.csv?dl=0

After review by @talumbau @jlyons871 , this is ready to be merged.

feenberg:
On Fri, 20 Feb 2015, Amy-Xu wrote:

> 
> In the latest version of exall.csv from exall.2 folder, there are 6 variables (22250,
> 23250, 04470, 25470, 08000, 60100) that show up as both e-variables and p-variables.
> From the testing results I guess the e-values are the ones we should use, but in

In the puf SOI renames some variables to P, because they have slightly 
different definitions. The SAS code creates appropriate E values in the 
INIT macro for these variables, so that COMP can be the same for the 
complete file or the PUF.

dan

MattHJensen:
@jlyons871, let us know when you have a chance to review this, and then I will merge.

jlyons871:
@Amy-Xu, good work it looks like this is close to being complete. 

However, I am unable to run this updated version without getting Typing Errors regarding what type variables numba expects to have returned. These are occurring in both the AMTI() and the iterate_jit() functions, when running test.py with puf2.csv and exall.csv as inputs and also when I run py.test. Are you able to replicate these errors?

jlyons871:
changing this [line]( https://github.com/Amy-Xu/Tax-Calculator/blob/testing/taxcalc/functions.py#L609) to 
`_addamt = 0.`
fixes the first error regarding AMTI for me.

Amy-Xu:
hmmmm it works fine on my laptop. I don't know whether it's relevant, but I indeed updated my numba recently. Do you mind show me the typing error message you are getting?

jlyons871:
Updating the numba did the trick. 
Upon initial review it looks good to me, I will dig in a bit deeper before we merge this.

MattHJensen:
@Amy-Xu, I just noticed that the changes in test.py shouldn't be included in this merge. 

@jlyons871, how are things looking to you?

jlyons871:
@MattHJensen  I found a few additional updates from the SAS code and made a PR to include them in this merge. Overall, the accuracy changes look good, nice work @Amy-Xu!

These changes do not affect the accuracy of the "important" variables, as indicated by @feenberg:
`'c00100', 'c04100', 'c04600', 'c04470', 'c04800', 'c05200', 'c09600', 'c05800', 'c09200'`

The accuracy of `c07100` has been fixed in 5707ca38a3bed6a55a87993d629a65c95f388bb5, although with the updates made in ffb6ea6d26d81addfd5236e2647ed18ecab01a92 there is one record ,`recid=115611`, where the results do not match for `c60000`. @feenberg I think this is OK, do you agree?

MattHJensen:
@jlyons871, this looks great! @Amy-Xu, could you look through @jlyons871 updates and then merge them into your branch if everything looks good? @Amy-Xu, could you also reverse the changes in test.py (and look through things to see if there are any similar quirks) and let me know when you are ok for me to merge this?

Amy-Xu:
Thank you @jlyons871, all the changes look good to me. @MattHJensen  Just fixed test.py. It should be ready for merging.

MattHJensen:
Merging. Congratulations everyone!

Issue #98, opened 2015-01-26
Parameterize hard-coded variables


Amy-Xu:
For the hard-coded variables in functions.py, I came up with some naming conventions to parameterize them, which should be universal and flexible enough to apply to other parameters in parameters.py at some point.

The variable names generally have two fixed components, one for tax/tax credit category (upper case) and the other for the function of this number in calculation (lower case). 

--------------------------------
The category names so far include the following:

CTC: Child Tax Credit
ACTC: Additional Child Tax Credit
AMT: Alternative Minimum Tax
II: Individual Income (Including personal exemptions and tax brackets)
CG: Capital Gain
EITC: Earned Income Tax Credit
ETC: Education Tax Credit
FEI: Foreign Earned Income
ID: Itemized Deduction
STD: Standard Deduction
FICA: Federal Income Contributions Act
SS: Social Security
MED: Medicare
AMED: Additional Medicare

--------------------------------
The possible functions of HC numbers:

c: ceiling
f: floor
t: tax
p: phaseout
rt: rate
em: exemption
ec: exclusion
thd: threshold

For example, the FICA tax rate would be _FICA_trt, t for tax and rt for rate. Similarly, prt represents a phaseout rate, crt represents a ceiling rate etc. 

In addition to these two essential components, there might be sub-components to make the var names more informational. For example, _ID_Madical_frt refers to the medical deduction floor rate under Itemized Deduction. All these subcomponents should be very intuitive that people don’t need a reference to understand their meanings. 

All above is what I came up with at this point. Please let me know if you have any suggestions on the format/abbreviation.

I added some hard coded values at the end of parameters.py and revised the corresponding section in functions.py. Here are the links to the modified files:  functions.py (https://github.com/Amy-Xu/Tax-Calculator/blob/master/taxcalc/functions.py) and parameters.py (https://github.com/Amy-Xu/Tax-Calculator/blob/master/taxcalc/parameters.py).

Here's a list of the parameters I created:

![list_params](https://cloud.githubusercontent.com/assets/10549597/5907886/aeb46ab8-a570-11e4-8b01-e3c4716e3f12.png)

MattHJensen:
@amy-xu, since we are adopting this naming convention in #143, could you use whatever information from this issue you think is relevant to make a .md document for the docs folder that describes the convention?

Amy-Xu:
Yes. Will do.

MattHJensen:
[...]

PR #143, opened 2015-02-24
Parameterize hard-coded values and rename all params for consistency

Amy-Xu:
I made several changes on this branch. First, I added the hard-coded values that are going to show up on the webpage to our params.json file. Second, I included hair cut parameters for a few e-codes representing adjustments. These variables are all in function Adj(). Third, I renamed the parameters with the naming convention proposed in #98 and switched all names in both params.json and function.py. To run py.test, I also changed a few lines in test_calculate.py to match with this set of new param names.


MattHJensen:
@Amy-Xu could you compare results.csv from this branch to results.csv from master? If the two are the same then I will merge.

MattHJensen:
If you load them both as dataframes, you can run: assert_frame_equal(df1, df2)

Amy-Xu:
should be good right now

MattHJensen:
Why is AMT exemption phaseout start "_AMT_em_ps" but Child Tax Credit Phaseout Start is "_CTC_pe"? In particular, why are the last letters different?

Amy-Xu:
...........I think that's a typo...............let me fix it........
Is there anything else weird you noticed?

MattHJensen:
Looks good now. Merging.

Issue #100, opened 2015-01-27
Parameters.py should accept year-by-year inflation rates


MattHJensen:
The user of the tax calculator should be able to vary, by year, the inflation rate that indexes parameters. The latest projection from CBO for inflation (cpi-u), for instance, is 1.1% in 2015, 2.2% in 2016, 2.3% in 2017, and 2.4% thereafter. Webapp users do not need this functionality at this point. 
[CBO economic projections](https://www.cbo.gov/publication/45066)

talumbau:
How about:
````
calc.inflation_rates = [0.011, 0.022, 0.023] + [0.024]*7
````

You get an exception if the length of `inflation_rates` is not the same as `budget_years` in the Parameters object. You could also do:

```
rates = [0.011, 0.022, 0.023] + [0.024]*7
calculator(parameters="params.json", records="puf2.csv", inflation_rates=rates)
```

So, in this case, because the default size fo `budget_years` is 10, if you specify the rates, it has to be of length 10. Thoughts?

-T.J.

MattHJensen:
@talumbau The top seems fine for now, although 99% of the time we will just use the defaults values. This is because the inflation rates and the blowup factors are linked (i.e., the records blowup factors are partially determined by the inflation rate, and it doesn't make sense to use different inflation rates for blowing up the tax law parameters from the one that we used to generate the records blowup factors)

A couple of notes:

1. I am now running the calculator locally with default budget_years = 12, so that we can blow up params from 2013 through 2024. We will still only want to report the 10 years 2015-2024. If you don't, I will probably make this update in an upcoming PR. 

2. Just a heads up: we may need to modify how the data blowup, reweighting, and param inflation are implemented when we link the tax calculator to a macroeconomic model. Then we will want to be able to give the calculator a new set of blowup factors, weights, and inflation rates each time we increment the year (not just when we set up the calculator.)

feenberg:
I think we need to make the default inflation rates the same as those 
Ohara assumes for the data. These seem close  to his CPIU numbers, but not 
exact.

dan

MattHJensen:
Dealt with in #155. Merging.

PR #146, opened 2015-02-25
Check 'cpi_inflated' value when expanding array

talumbau:
Checks for the existence of 'cpi_inflated' for each value in a Parameter file. If the value is not present, the default is to not inflate.


MattHJensen:
@talumbau this looks good. It also looks fairly easy to incorporate #100, now, which is also good.

PR #147, opened 2015-02-26
blowup puf.csv to 2013 in calculator init

MattHJensen:
Our puf data are from 2008, but in puf2.csv, we have hardcoded flpdyr = 2013. This was preventing us from blowing up the data properly because records.blowup() needs to start blowing up the data from 2008 -- it multiplies the current year’s data by (1+pct change between the current years’ data and the next years’). 

This PR fixes the issue by 

1. Requiring the use of puf.csv rather than puf2.csv (flpdyr should start at 2008 rather than 2013)
2. When a calculator is initialized, it runs records.increment_year() while records.current_year < parameters.current_year. 

After this is merged, everyone will need to use puf.csv rather than puf2.csv when running the calculator (including when testing with py.test.)


MattHJensen:
@talumbau, if this looks like a suitable solution to you, at least for now, I'd like to merge.

taxcalc/calculate.py
```diff
@@ -58,6 +58,9 @@ def __init__(self, parameters, records, **kwargs):
         self._records = (records if not isinstance(records, str) else
                          Records.from_file(records, **kwargs))
 
+        while self._records.current_year < self._parameters.current_year:
```
talumbau: Can we have a keyword arg that would specify this behavior? I'm fine if the default is true. Something like:

```
calc = Calculator(parameters=p, records=r, sync_years=True)
```

I'm concerned about unintended consequences here.

talumbau: What I mean to say is that a user may unintentionally get incremented records when they accidentally specified the wrong data. In that case, it would be good to give an error message instead of just incrementing the records until the years match.

MattHJensen: @TJ, could you give an example of the situation you envision? 

One option is:

        if sync_years:
            print("You loaded data for "+str(self._records.current_year))
            while self._records.current_year < self._parameters.current_year:
                self._records.increment_year()
            print("Your data have been extrapolated to "+str(self._records.current_year))
        assert self._parameters.current_year == self._records.current_year

But I'm not sure this deals with your concern. 

Also, if `sync_years==False`, we are always going to fail the assertion at the end anyways.

talumbau:
Here is a current test:

```
    puf_basic = Records(tax_dta, start_year=2013)
    calc_basic = Calculator(records=puf_basic)
```

This uses the new capability for not specifying a Parameters object and just getting the default parameters. If you don't specify the start year for a Records object, you get `FLPDYR`. In the case of the test data, this is 1991. Currently, you get an error if you don't specify the start year, because the test data is 1991 and the parameters default is 2013. With the proposed change, the Calculator would automatically update the data from 1991 to 2013. Is this the desired behavior? I would think that the more likely case is that the user specified the wrong data file. In this test scenario, it's a contrived problem because we are just trying to exercise some functionality of the calculator for a test.

MattHJensen:
Since users will be using the 2008 data 98% of the time, I added a condition that the extrapolation should only occur if self._records.current_year==2008. 

Once we start getting feedback from users, we can make a better-informed decision about whether this behavior works or not. 

@talumbau, could you take another look before I merge?

talumbau:
this looks good. the print statements are ripe for a refactor with the `logging` module, but that probably shouldn't hold up this merge.

MattHJensen:
Ok. I'll open an issue on logging. Merging.

PR #155, opened 2015-03-04
single inflation_rate -> array of inflation_rates

talumbau:
- for 1D and 2D arrays
- for arrays of length j that need to be expanded to length m, the
  chosen inflation rates will start at index j and go up to m-1
- other small additions

addresses https://github.com/OpenSourcePolicyCenter/Tax-Calculator/issues/100

(moves to 12 as the default number of budget years)

### DIFF
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -51,9 +51,15 @@ class Calculator(object):
         return cls(params, recs)
 
 
-    def __init__(self, parameters, records, **kwargs):
-        self._parameters = (parameters if not isinstance(parameters, str) else
-                            Parameters.from_file(parameters, **kwargs))
+    def __init__(self, parameters=None, records=None, **kwargs):
+
+        if isinstance(parameters, Parameters):
+            self._parameters = parameters
+        else:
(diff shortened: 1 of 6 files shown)


talumbau:
Need feedback from @MattHJensen @feenberg if the default inflation rates are correct.

feenberg:
We need to use the same inflation rates as Ohara.

dan

MattHJensen:
Our extrapolation routine (from John O'Hare) is currently using the August 2014 CBO baseline.  The CPI-U inflation rates [2013:2024] from that baseline are:

__rates = [0.015, 0.020, 0.022, 0.020, 0.021, 0.022, 0.023, 0.024 0.024, 0.024, 0.024, 0.024]

https://www.cbo.gov/sites/default/files/cbofiles/attachments/45066-2014-08-EconomicProjections.xlsx

Let's use these as our defaults for now. 

John is currently updating the extrapolation to use the Jan 2015 CBO baseline, and when he does I will update these rates and confirm with him that they are correct.

talumbau:
I updated the rates so I think this can go in. Will merge soon if no comments to the contrary.

MattHJensen:
:+1:

PR #166, opened 2015-03-06
fixed the parameter for additional CTC

Amy-Xu:
This parameter wasn't put in the right place. Just fixed it. I also switch '>' to '>=' to fit the actual value of the parameter better.

@MattHJensen Could you review the change and merge it? Thanks


MattHJensen:
Looks good. Merging!

Issue #149, opened 2015-02-26
fica medicare taxes should not be subject to the social security wage limit


MattHJensen:
_fica is currently calculated as 
_fica   = max(0,0.153*min(ssmax, e00200+max(0,_sey)*0.9235)) 
combined

1. The medicare contribution should not be subject to ssmax
2. The medicare contribution should incorporate the new additional medicare tax

MattHJensen:
I am dealing with this in #170.

MattHJensen:
Dealt with in #170. Closing.

PR #170, opened 2015-03-09
[WIP] Accuracy fixes 

MattHJensen:
Dealing with a number of tax bugs. So far:

- [x] #183 
- [x] #180
- [x] #179 
- [x] #178 
- [x] #174 
- [x] #149


taxcalc/functions.py
```diff
@@ -216,19 +216,20 @@ def EI_FICA(   e00900, e02100, SS_Income_c, e00200,
...
+            II_brk1, II_brk2, II_brk3, II_brk4, II_brk5, II_brk6, _fixup, std2008, puf):
 
     if DSI == 1:
         c15100 = max(350 + _earned, STD[6])
     else:
         c15100 = 0.
 
-    if e04470 > 0 and e04470 < STD[MARS-1]:
+    # std2008a = [5450, 10900, 5450, 8000, 10900, 5450, 900]     
+    if FDED==1 and (e04470 < std2008[MARS-1]): #TODO should be 2008 vals, not current law.
```
MattHJensen: @talumbau, there are a few places in the code where we need to use 2008 values of various parameters to impute record attributes (in this case, we are imputing _compitem based on e04470 and the 2008 value of standard deduction.) As an expediency, I added std2008 as a parameter (https://github.com/OpenSourcePolicyCenter/Tax-Calculator/pull/170/files#diff-c13090d5f6d7cc59a8c6e1d6c716e553R1291), but it would be better to have this value in functions.py, but I am not sure how to do that given iterator_jit. Alternatively, we could perform the imputation as part of the initialization of our records object, and that is what I am leaning towards doing. Do you have an opinion about this?

talumbau: I like the idea of initializing the records with this value. While it is possible to put this in functions.py, I think that would lead to a pattern where eventually someone might put mutable state in functions.py, in which case we would be back to the problems we had before. Putting it in records.py seems to work with the idioms we have right now.

MattHJensen:
@Amy-Xu  could you review the changes in this PR? If the documentation in the cited issues is not clear, please let me know or expand on what I have in the comments sections.

taxcalc/functions.py
MattHJensen: Ok. I moved this stuff over to records.py. Two improvements for the future are probably to 

1. Keep imputations in a separate module
2. See if we can move this to vanilla python with a decorator.

Amy-Xu:
look good to me

MattHJensen:
Thanks @Amy-Xu. Merging.

PR #172, opened 2015-03-09
X-year baseline table for testing purposes

Amy-Xu:
In calculate.py, I added a function named baseline_table to create X-year milestone-variable results from the default plan. The number of years is an input argument of the function. It is set to 5 by default and can be any number between 1 and 10.

Currently this function calls calc_all() X times and blows up the data to year (2013+X-1).  So it's not very time-efficient and does not give enough flexibility for users. For now this function is just for testing purposes.

@MattHJensen Could you review the function? Thanks


MattHJensen:
Hi @talumbau do you have any thoughts on the implementation here? This function just creates the "diagnostics table" that we have been using for testing.

talumbau:
Two thoughts:

1. I think there should be a test in `test_calculate.py` that shows calling this function and doing something reasonable with the output. That way we would know if it ever broke.

2. Other mutating methods on a Calculator have names that can help one know that some state is changing (e.g. `calc_all`, `increment_year`). Here, `baseline_table` doesn't just return a table, but does quite a bit of work to produce the table. I just wonder if a different name would be better.

Amy-Xu:
@talumbau 
1. It makes a lot sense to me to add a test. There are two issues in creating this test due to the increment_year() function used in this table generating process. First, our blow-up factors are only available for 2008 and years afterward, which means the blow-up factor table is indexed starting from 2008. So using 91 data to test this function generates an index error. Second, the re-weighting process currently is using a csv file of the same length as 2008 PUF. Because the number of records 91 PUF is way shorter, the length mismatch will throw an error in the re-weighting process.

I fixed the first issue by changing the year variable (FLPDYR) to 2008 before generating a calculator object in the test function. For the second issue, I created a separate weights file (WEIGHTS_testing.csv) of the same length as 91 PUF and save it in the same directory of 91 PUF.

However, it seems all the changes passed python 2.7 but not python 3.4. I'm trying to see what's happening here. Do you have any suggestions?

2. Matt and I decided to make a deep copy of the calculator so this function will not change the state of the original calculator anymore.  But we still changed the name from baseline to diagnostic, thinking it might be a better fit for this function.

talumbau:
I took a look at the code and the error:

- I think the magic I did so the `calc.xxx` looks for `xxx` in the underlying Records and Parameters object does not play nicely with `copy.deepcopy`. We could make a `__deepcopy__` method on Calculator to handle this.

- Would this make more sense as a free standing function? That way you could intentionally decide to only pass in a Calculator object that you are OK with modifying. Either way, you should be able to deep copy a Calculator. I'll add an issue.

MattHJensen:
Ok. We thought it was interesting that the python 2.7 build was passing but not 3.4. 

fyi, this is not an urgent issue since the function is working for both of us locally.

talumbau:
https://github.com/OpenSourcePolicyCenter/Tax-Calculator/pull/189 will likely resolve the issue.

MattHJensen:
Looking at the other table generators in utils.py, it seems that this may make more sense as a freestanding function that we keep with the others. What do you think @Amy-Xu?

talumbau:
https://github.com/OpenSourcePolicyCenter/Tax-Calculator/pull/189 has been merged. Please merge master in to your branch and retry.

MattHJensen:
@talumbau, now that you've bumped version for webapp, are you still ok with us merging this and opening an issue to refactor it as we discussed?

MattHJensen:
Merging. I will open an issue about the refactor.

PR #190, opened 2015-03-23
Distribution table

Amy-Xu:
In this PR:

1. Modified the AGI bins and column variables in the distribution table, to compare with SOI and TPC's tables. Regarding AGI bins, I added one set of bins, named small_agi_bins, to match SOI tables as opposed to large_agi_bins that matches the TPC tables. For the column variables, I added the number of   taxpayers using standard deduction, itemized deduction and AMT. The sequence of column variables was also adjusted slightly to keep in line with the diagnostic table.

2. Changed the simply sum and average to weighted sum and average for column variables in the distribution table. Originally only the create_difference_table() function for two calculators used the weighted results, but not the distribution table for a single calculator.


taxcalc/utils.py
```diff
@@ -231,33 +235,66 @@ def means_and_comparisons(df, col_name, gp, weighted_total):
...
+    res['num_returns_AMT'] = np.where(res['c09600']>0, res['s006'], 0)
+
+    
     if groupby == "weighted_deciles":
-        gp = groupby_weighted_decile(res)
-    elif groupby == "agi_bins":
-        gp = groupby_income_bins(res)
+        df = groupby_weighted_decile(res)
+    elif groupby == "small_agi_bins":       
+        df = groupby_income_bins(res, "soi")
```
MattHJensen: There are a few places that "soi" and "tpc" are still popping up. Is this correct?

Amy-Xu: Yes. I only edited the layer where users need to give explicit inputs

taxcalc/utils.py
```diff
@@ -2,13 +2,12 @@
...
-                 'c09600', 'c07100', 'c09200', '_refund', '_ospctax',
-                 'c10300', 'e00100', 's006']
+STATS_COLUMNS = ['c00100', '_standard', 'c04470', 'c04600','c04800', 'c05200',
+                 'c09600', 'c05800', 'c09200', '_refund', 'c07100','_ospctax',
+                 's006']
 
-TABLE_COLUMNS = ['c00100', 'c04100', 'c04470', 'c04800', 'c05200',
-                 'c09600', 'c07100', 'c09200', '_refund', '_ospctax',
-                 'c10300']
+TABLE_COLUMNS = ['c00100', 'num_returns_StandardDed','_standard', 'num_returns_ItemDed', 'c04470',
```
talumbau: The value of `TABLE_COLUMNS` was based on the understanding I had for what was needed for the web app. Are these the new columns needed for display for the webapp? If so, there is no issue. If not (and therefore the original value of `TABLE_COLUMNS` is needed), a separate variable called `WEB_APP_TABLE_COLUMNS` or similar should be created.

MattHJensen: These new columns are good for the webapp too.

taxcalc/utils.py
```diff
@@ -178,11 +177,11 @@ def groupby_weighted_decile(df):
     labels = [range(1,11)]
     # Groupby weighted deciles
     decile_bins = list(range(1, 11))
-    df['wdecs'] = pd.cut(df['cumsum_weights'], bins, labels)
-    return df.groupby('wdecs')
+    df['bins'] = pd.cut(df['cumsum_weights'], bins, labels)
```
talumbau: `groupby_weighted_decile` and `groupby_income_bins` served the purpose of doing a grouping operation. They returned a Pandas GroupBy object, which could be handled generically. These two functions now just add a column named `bins` to the DataFrame, which makes the names confusing. I think we should either name them differently (e.g. `add_weighted_decile_bins`) or do the grouping in the function (I prefer the latter).

taxcalc/utils.py
```diff
@@ -178,11 +177,11 @@ def groupby_weighted_decile(df):
...
+    return df
 
 
-def groupby_income_bins(df, bins=None, right=True):
+def groupby_income_bins(df, compare_with = "soi", bins=None, right=True):
```
talumbau: Need an explanation of what `compare_with` does in the doc string.

taxcalc/utils.py
```diff
@@ -197,14 +196,19 @@ def groupby_income_bins(df, bins=None, right=True):
             indicate (1,2], (2,3], (3,4].
 
     """
-
-    if not bins:
+    if compare_with == "tpc":
```
talumbau: similar comment about webapp settings. Please indicate if one of these pre-defined bins is the desired one for the webapp.

MattHJensen: 'Large bins' or the pre-existing bins are equally fine.

taxcalc/utils.py
```diff
@@ -231,33 +235,66 @@ def means_and_comparisons(df, col_name, gp, weighted_total):
...
+    
+    for col in cols:
+        if col == "s006" or col =='num_returns_StandardDed' or col == 'num_returns_ItemDed' or col == 'num_returns_AMT':
+            diff[col] = df.groupby('bins')[col].sum()
+        elif col != "c00100":
+            diff[col] = df.groupby('bins').apply(weighted_mean, col)
+            
+    return diff
+
+def create_distribution_table(calc, groupby, result_type):
```
talumbau: This function is definitely an improvement over the previous version, but I think there might be a nicer way to write the math. I'll have a go at it.

taxcalc/utils.py
```diff
@@ -197,14 +196,19 @@ def groupby_income_bins(df, bins=None, right=True):
...
     """
-
-    if not bins:
+    if compare_with == "tpc":
         bins = [-1e14, 0, 9999, 19999, 29999, 39999, 49999, 74999, 99999,
                 200000, 1e14]
+    elif compare_with == "soi":
+        bins = [-1e14, 0, 4999, 9999, 14999, 19999, 24999, 29999, 39999, 49999, 74999, 99999,
+                199999, 499999, 999999, 1499999,1999999, 4999999, 9999999, 1e14]
+    elif not bins:
```
talumbau: For this function, if `compare_with` is not specified, you will get the bins of line 203 regardless of whether you specify `bins` or not. If `compare_with` == `tpc`, you will get the bins of line 205 regardless of whether you specify `bins` or not. The only way to specify your own `bins` is to pass `compare_with` as something other than `soi` or `tpc`. I think this will be error prone. In general, I think that by not specifying `compare_with` and specifying your own `bins`, you should get those bins.

taxcalc/utils.py
MattHJensen: On second thought, the pre-existing bins are slightly preferable: 
"preexisting bins": [-1e14, 0, 9999, 19999, 29999, 39999, 49999, 74999, 99999, 200000, 1e14]

talumbau:
will merge in master and submit a PR to your branch @Amy-Xu

talumbau:
Need the `weighted_sum` here for the distribution tables, as this will now be featured in the web app. Merging.

PR #193, opened 2015-03-24
Prep webapp

MattHJensen:
This includes a few fixes to the tax calculator to make sure that all of the linkages described in this file are working properly. https://github.com/OpenSourcePolicyCenter/webapp/blob/master/Docs/taxbrain_inputs_layout.xlsx

In particular, 

1. There were some parameters being used with a leading 'underscore' in functions.py and included as 1-d arrays in params.json, even though they are indexed by marital status. This wasn't causing any errors since none of these params are indexed by inflation by default, but it would cause errors if the webapp user tried to turn on inflation indexing. I made these params 2-d arrays in params.json and then passed them without the leading underscore in functions.py. 

2. I indexed STD_Aged by filing status rather than by "maried"/"unmarried". 

3. Update long_name attributes in params.json for readability. 

None of these changes impacted the baseline results of the calculator. 

@Amy-Xu, could you please review these changes and let me know if there is anything else that you think we need to do before putting up the next version of the taxcalc for our webapp rollout?


MattHJensen:
This *should* be good to go. @talumbau, if you need this ASAP, feel free to merge. Otherwise @Amy-Xu and I will make a final review pass tomorrow morning.

PR #197, opened 2015-03-26
[WIP]Distribution Table Formatting

Amy-Xu:
Changed Distribution table column names from variable names to labels that are more descriptive.

### DIFF
--- a/taxcalc/utils.py
+++ b/taxcalc/utils.py
@@ -1,16 +1,21 @@
 import numpy as np
 import pandas as pd
 from pandas import DataFrame
+import copy
 
-STATS_COLUMNS = ['c00100', '_standard', 'c04470', 'c04600', 'c04800', 'c05200',
-                 'c09600', 'c05800', 'c09200', '_refund', 'c07100', '_ospctax',
-                 's006']
+STATS_COLUMNS = ['s006', 'c00100', 'c04470', '_standard', 'c04600', 'c04800', 'c05200',
+                 'c09600', 'c05800', 'c09200', '_refund', 'c07100', '_ospctax']
 
-TABLE_COLUMNS = ['c00100', 'num_returns_StandardDed', '_standard',
-                 'num_returns_ItemDed', 'c04470', 'c04600', 'c04800', 'c05200',
+
+TABLE_COLUMNS = ['s006', 'c00100', 'num_returns_ItemDed', 'c04470', 
+                 'num_returns_StandardDed', '_standard', 'c04600', 'c04800', 'c05200',
                  'num_returns_AMT', 'c09600', 'c05800', 'c09200', '_refund',
-                 'c07100', '_ospctax', 's006']
+                 'c07100', '_ospctax']
 
+TABLE_LABELS = ['Returns', 'AGI', 'Itemizers', 'Itemized Deductions',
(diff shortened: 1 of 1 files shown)


MattHJensen:
The column labels look good to me. @talumbau, could you review the implementation before we merge? Not urgent.

taxcalc/utils.py
```diff
@@ -301,10 +310,14 @@ def create_distribution_table(calc, groupby, result_type):
                "or 'large_agi_bins'")
         raise ValueError(err)
 
-    pd.options.display.float_format = '{:8,.0f}'.format
+    pd.options.display.float_format = '{:14,.0f}'.format
+    pd.set_option('precision',4)
     if result_type == "weighted_sum":
         df = weighted(df, STATS_COLUMNS)
-        return df.groupby('bins')[TABLE_COLUMNS].sum()
+        temp = copy.deepcopy(TABLE_LABELS)
```
talumbau: suggest

```
temp = TABLE_LABELS + ['bins']
```

taxcalc/utils.py
```diff
@@ -263,33 +269,36 @@ def results(c):
 
 
 def weighted_avg_allcols(df, cols):
-    diff = DataFrame(df.groupby('bins').apply(weighted_mean, "c00100"),
-                     columns=['c00100'])
+    diff = DataFrame(df.groupby('bins')['s006'].sum(), columns=['s006'])
```
talumbau: I am a bit stuck here. The general idea of this function was to start with with weighted average of the amount of tax paid (`c00100`), and then add columns to that DataFrame by doing the weighted average of the rest of the columns in TABLE_LABELS. But now, `diff` starts out as the weighted average of `s006` which are the weights themselves. Can you help me understand why we want to see the weighted average of `s006` here?

Amy-Xu: You are right that we always want to see the sum of s006 instead of the weighted average. (In this average table, four weights related variables are summed up and all others are taken weighted average.) So here I use the sum() function for the s006 column in the groupby object and I thought this step would return the sum instead of the weighted average. Do I understand this sum() function right?
The reason why I moved s006 to the first column is that 1) it's in line with the diagnostic table format 2) it's easier to compare with SOI tables, which always list the number of returns for each bin before the amount of tax/credit.

taxcalc/utils.py
Amy-Xu: Thanks! Just changed temp to this definition.

MattHJensen:
@Amy-Xu and @talumbau, is getting this PR up to date the best way for me to make labels available to a non-webapp-user, or would a different path be better? 

Adding labels to the tables will make our 10_Minutes_to_Taxcalc notebook much more compelling for new users. (I am making additional updates to that notebook as a part of #201).

Amy-Xu:
My concern is that I'm not sure whether this is the optimal way to add labels. After reading the code for drop q, I think I can make changes so the code won't look this ugly. Should I revise it first?

MattHJensen:
I was wondering whether this angle was outdated as well. I say tackle the problem however you think is best. Should I close this PR?

Amy-Xu:
Yea I think so. I can open a new one if needed.

MattHJensen:
Ok. Closing.

PR #199, opened 2015-03-27
[WIP] Update taxcalc with 2012 and 2011 tax logic

Amy-Xu:
In this PR, 

1) Updated parameters with 2011 and 2012 numbers
2) Added switches for Personal Exemption phaseout, Itemized Deduction limitation, and NIIT. None of them existed before 2013. The switches are realized through if loops and boolean array parameters
3) Parameterized all three cap gain rates


MattHJensen:
2011 and 2012 inflation rates should be prepended to the __rates list in parameters.py. They are 3.2 in 2011 and 2.1 in 2012. 

http://www.usinflationcalculator.com/inflation/consumer-price-index-and-annual-percent-changes-from-1913-to-2008/

talumbau:
If this PR is merged, does it mean that the correct way to do a calculation for budget years 2015-2024 is:

1. Create Parameters object
2. load puf.csv into Records object
3. create calculator from 1 and 2. Calculator will automatically advance records from 2008 to 2011.
4. manually increment year from 2011 to 2015.
5. begin calculations for 2015 and go up to 2024.

(currently we only increment the year twice (2013 -> 2015))

MattHJensen:
Yes, that is correct. Does that seem unwieldy/is there a better way to do things? 

We added 2011 and 2012 tax law because there are a lot of tables from the Statistics of Income division of the IRS that we can compare our results to and assess the effectiveness of our data-extrapolation, but those tables are only available starting in 2012. 

See, e.g., http://www.irs.gov/uac/SOI-Tax-Stats---Individual-Statistical-Tables-by-Size-of-Adjusted-Gross-Income

talumbau:
OK, just checking. It's not a big deal, it's just that the way we do the calculation for the webapp will change slightly (just increment years a couple more times).

talumbau:
One further issue: for the webapp, the user specified parameters will be taken as applying in 2015 and beyond (up to 10 budget years), whereas the initial defaults in params.json will mean "starting in 2011". For example, if a parameters is specified for 6 budget years in params.json, that parameter has values for 2011, 2012, 2013, 2014, 2015, and 2016. When the user supplies a series of values, their values will apply for year 2015, 2016, and so on. Correct?

MattHJensen:
@talumbau, yes, that is correct. That is the behavior that we should be applying currently as well, given the gap between 2013 and 2015.

MattHJensen:
The same calculator is giving different tax computation under a variety of use cases:

Scenario 1: Load one calculator, increment year, and then calculate taxes.

image

Scenario 2: Load one calculator, calculate taxes, increment year, and then calculate taxes again.

image

Scenario 3: Load two calculators properly (using records and records2, params and parms2), calculate taxes for the first, increment year, and then calculate taxes again.

image

Scenario 4: Load two calculators improperly, calculate taxes for the first, increment year, and then calculate taxes again.

image

Notes:

1. It turns out that everything is the same under all of these scenarios up until the calculation of credits. 

2. I believe that the numbers from scenarios 1 and 3 are correct based on proper usage of master. 

3. We can get the "wrong" results on master from a second calculator if we load it without using records2 and params2

talumbau:
scenario 2 is clearly a bug, right?

MattHJensen:
Yes, I thought it was interesting that it produced the exact same result as scenario 4. Scenario 4 happens on master, but scenario 2 does not happen on master.

taxcalc/parameters.py
```diff
@@ -24,7 +24,7 @@ def from_file(cls, file_name, **kwargs):
         return cls(data=params, **kwargs)
 
 
-    def __init__(self, start_year=2013, budget_years=12, inflation_rate=None,
+    def __init__(self, start_year=2013, budget_years=14, inflation_rate=None,
```
talumbau: I think the `start_year` keyword argument default should be 2011.

talumbau:
would you be willing to create a new issue with these results and properly categorize correct/incorrect behavior for these scenarios? The most helpful would only be describing behavior on current master

MattHJensen:
Yes. Will do.

MattHJensen:
Hmm, turns out scenario 2 also happens on master. I will include it in my report.

talumbau:
This work is closely related to https://github.com/OpenSourcePolicyCenter/Tax-Calculator/pull/204. Once Amy has responded to my comment about the start year, I am good with this PR, but the merge might be non-trivial with 204.

MattHJensen:
Ok. It is not urgent to merge this PR since we are using it effectively for testing already. Feel free to go ahead with #204.

taxcalc/parameters.py
Amy-Xu: Yes! Just updated it.

PR #200, opened 2015-03-30
Documentation using docstring and comments for functions.py

Amy-Xu:
This PR includes documentation for three functions: AmOppCr(), LLC() and RefAmOpp(). The docstring section describes what this function does, what parameters get involved and all the key inputs/outputs/intermediate variables. The comments label the important steps in each function.


MattHJensen:
@jlyons871 Could you review prior to merge?

MattHJensen:
@Amy-Xu, is this WIP or are you happy with it? Might make sense for @xiyuw123 to review since she just looked into education credits and for @jlyons871 to review to make sure we are on the right track with our first numpy-style docstring in functions.py

Amy-Xu:
Didn't realize I have already had a PR for this branch. I think I will just keep working on it for the rest of functions.py instead of just education credits. @xiyuw123 I commented the first three education credit functions but not the last nonrefundable one. Let me know your thoughts.

MattHJensen:
Sounds good!

xiyuw123:
Hi Amy, the comments for the education credits look good!

taxcalc/functions.py
```diff
@@ -188,17 +266,16 @@ def ItemDed(_posagi, e17500, e18400, e18425, e18450, e18500, e18800, e18900,
         c21040 = 0.0
 
 
-    if c21060 > _nonlimited and c00100 > _limitratio:
```
Amy-Xu: The if statements here is redundant, because the previous if statement has already handled the issue.

taxcalc/functions.py
```diff
@@ -188,17 +266,16 @@ def ItemDed(_posagi, e17500, e18400, e18425, e18450, e18500, e18800, e18900,
...
-        c04470 = c21060 - c21040
-    else:
-        c04470 = c21060
+    # Itemized deductions amount after limitation if any
+    c04470 = c21060 - c21040
 
-    c20400 = float(c20400)
+
```
Amy-Xu: Moved to section down below.

Amy-Xu:
@MattHJensen I cleaned up this branch and update it with ospc master. All docstrings added so far should be good to merge. I commented the two places where the codes were changed slightly for the sake of documentation. @jlyons871 Could you review the format of docstring? thanks!

MattHJensen:
@Amy-Xu, you should be able to build the readthedocs html locally to see if sphinx is interpreting your docstrings correctly. 

First, pip install the following dependencies: 

sphinx
sphinxcontrib-napoleon==0.2.11


Then, from ~/docs directory type `make html` from the terminal and this will build a local version into the ~/docs/build/html directory.

Amy-Xu:
@MattHJensen Could you review this PR? I wrapped up the docstring section for several functions I worked on.

taxcalc/functions.py
```diff
@@ -41,10 +92,17 @@ def CapGains(  e23250, e22250, e23660, _sep, _feided, FEI_ec_c, ALD_StudentLoan_
...
+    # Net capital gain (long term + short term) before exclusion
     c23650 = e23250 + e22250 + e23660
+    
+    # Limitation for capital loss
     c01000 = max(-3000 / _sep, c23650)
+    
+    # Foreign earned income exclusion
     c02700 = min(_feided, FEI_ec_c * f2555)
+    
+    #
```
MattHJensen: Did you mean to leave a comment here?

taxcalc/functions.py
```diff
@@ -120,23 +179,82 @@ def ItemDed(_posagi, e17500, e18400, e18425, e18450, e18500, e18800, e18900,
                  ID_Charity_frt, puf):
 
     """
-    WARNING: Any additional keyword args, such as 'puf=True' here, must be 
+    Itemized Deduction; Form 1040, Schedule A
+    
+    Notes
+    -----
+    Tax Law Parameters:
```
MattHJensen: We should be consistent about whether we leave taxpayer characteristics or tax law parameters first. Do you have a preference? I am leaning towards taxpayer characteristics first as in Adjustments().

taxcalc/functions.py
```diff
@@ -120,23 +179,82 @@ def ItemDed(_posagi, e17500, e18400, e18425, e18450, e18500, e18800, e18900,
...
+        
+        ID_Miscellaneous_frt : Deduction for miscellaneous expenses; floor as a percent of AGI
+        
+        ID_Charity_crt_Cash : Deduction for charitable cash contributions; ceiling as a percent of AGI
+        
+        ID_Charity_crt_Asset : Deduction for charitable asset contributions; ceiling as a percent of AGI
+        
+        ID_Charity_frt : Deduction for charitable contributions; floor as a percent of AGI
+    
+    Taxpayer Charateristics:
```
MattHJensen: Spell check on characteristics (and intermediate variables, below), and we should be consistent w capitalization (Taxpayer Characteristics vs. Taxpayer characteristics).

taxcalc/functions.py
```diff
@@ -238,6 +358,34 @@ def EI_FICA(   e00900, e02100, SS_Earnings_c, e00200,
 @iterate_jit(parameters=["AMED_thd", "AMED_trt", "FICA_ss_trt", 
                          "FICA_mc_trt"], nopython=True)
 def AMED(e00200, MARS, AMED_thd, _sey, AMED_trt, FICA_mc_trt, FICA_ss_trt):
+    """
+    Additional Medicare Tax as a part of FICA
+    
+    
+    Notes
+    -----
+    Tax Low Parameters:
```
MattHJensen: sp

MattHJensen:
I think the comments are a big improvement. 

The docstrings are an improvement in terms of readability. Having them there keeps you from having to flip to params.json and the taxpayer characteristics spreadsheet all the time for definitions, and the fact that you only include characteristics that are used in the calculations (not no-puf vars) is an additional piece of useful information. 

I still have mixed feelings about the docstrings, though, because they have some major downsides:

1. Writing them is work. 
2. We will have to keep parameter names and taxpayer characteristics synced across functions.py, params.json, and our taxpayer characteristics spreadsheet. 
3. functions.py is considerably longer. 

cc @rkuchan, @Amy-Xu, @talumbau, @jlyons871 for your views on this.

Amy-Xu:
I have mix feelings as well, for slightly different reasons. 

I always think it's important to give users multiple access to one piece of information, to make the calculator easier to use. Docstring would be one handy source of key parameters and variables, so users do not need to jump between files to look for just one variable name for example. And of course, being able to retrieve docstrings to the documentation page is an obvious advantage.

The downside, from my perspective, comes from the options provided by Sphinx. As for section titles, Sphinx only have a limited amount http://sphinx-doc.org/ext/napoleon.html?highlight=parameters of choices that do not fit the need of most functions for tax logic. For most functions, we just need to distinguish between e-variables and parameters. These two do not fall exactly into any of the existing categories. Thus we have to kind of compromise by putting everything in 'Notes', and using sub-section titles. I'm not sure whether there's a better way to achieve this.

Definitely I agree -- maintenance would be a pain. And for sure, it's a lot of work to write. But we need to do sort of similar amount of work somewhere, if not in the format of docstring. Just my opinion though.

MattHJensen:
@Amy-Xu, I'm definitely leaning towards agreement that it is an unalloyed good to give the user access to variable names in functions.py. Seems from your reservations that the important followup question is, 'is this the best way to give users that access.' @talumbau and @jlyons871, any thoughts on this?

talumbau:
From my perspective, the doc strings are a huge value add. It is true that the effort level is high to create them. It appears that functions in functions.py are fairly static though, relative to other parts of the package. I think doc strings like these would help a new person get up to speed more quickly on contributing to the calculator. Just my two cents.

MattHJensen:
Thanks @talumbau. @Amy-Xu, once you resolve conflicts, I'll go ahead and merge.

Amy-Xu:
Should be good to merge.

talumbau:
looks good!

PR #206, opened 2015-04-01
[WIP] A script to show what calculations we do

talumbau:
- this script shows what calculations are performed to
  compare two tax policies

### DIFF
--- a/docs/table_results.py
+++ b/docs/table_results.py
@@ -1,0 +1,116 @@
+from taxcalc import *
+import pandas as pd
+import taxcalc
+import json
+import numpy.testing as npt
+
+STATS_COLUMNS = ['c00100', '_standard', 'c04470', 'c04600', 'c04800', 'c05200',
+                 'c09600', 'c05800', 'c09200', '_refund', 'c07100', '_ospctax',
+                 's006']
+
+TABLE_COLUMNS = ['c00100', 'num_returns_StandardDed', '_standard',
+                 'num_returns_ItemDed', 'c04470', 'c04600', 'c04800', 'c05200',
+                 'num_returns_AMT', 'c09600', 'c05800', 'c09200', '_refund',
+                 'c07100', '_ospctax', 's006']
+
+def results(c):
+    outputs = [getattr(c, col) for col in STATS_COLUMNS]
+    return DataFrame(data=np.column_stack(outputs), columns=STATS_COLUMNS)
+
+myvars = {2015:{}}
+myvars[2015]['_II_rt4'] = [0.39]
+
+#Create a Public Use File object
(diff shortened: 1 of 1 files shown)


talumbau:
@MattHJensen, @Amy-Xu , the purpose of this PR is to make explicit the calculations for table creation in the webapp. Here we do everything except the particulars of the dropq algorithm. The idea is that this script will be the exact calculation the webapp produces (except for what must be done for dropq). In order to release the webapp, I'll need someone to say "yep, this script does the correct calculations" or let me know how to change it to make it right. One thing that occurs right away: should the webapp do `large_agi_bins`? I wasn't clear on that point. Thanks!

MattHJensen:
Ok, we will look through this. 

As for agi bins, the bins we are already using for the webapp are somewhat preferable to either of the built in options that Amy added for taxcalc package user. If it would be easier to use one of the built-in bins, though (small_agi_bins or large_agi_bins), large_agi_bins are the preferred of the two.

docs/table_results.py
```diff
@@ -0,0 +1,116 @@
...
+
+    #Select out the columns we want
+    df1 = results(calc1)
+    df2 = results(calc2)
+
+    # Difference in plans
+    # Positive values are the magnitude of the tax increase
+    # Negative values are the magnitude of the tax decrease
+    df2['tax_diff_dec'] = df2['_ospctax'] - df1['_ospctax']
+    df2['tax_diff_bin'] = df2['_ospctax'] - df1['_ospctax']
```
talumbau: this looks strange (because `tax_diff_dec` and `tax_diff_bin` are the same expression). In the real dropq though, these columns are different because records are dropped within each bin, and there are two binning strategies. So, in this idealized case (no dropping of records), the columns `tax_diff_dec` and `tax_diff_bin` are the same.

MattHJensen:
@talumbau, this seems like a useful script to keep around as documentation. Is that your view? If so, do you think there is a more appropriate name/docstring for it before we add it to the docs?

MattHJensen:
@talumbau, what do you think we should do with this PR?

MattHJensen:
@talumbau, could you check out the last couple of comments here when you have a chance?

talumbau:
@MattHJensen now that we have https://github.com/open-source-economics/drop-q/, I would be inclined to close this PR, since we could just point folks at the code there to see how the TaxBrain tables are created. Thoughts?

MattHJensen:
>@MattHJensen now that we have https://github.com/open-source-economics/drop-q/, I would be inclined to close this PR, since we could just point folks at the code there to see how the TaxBrain tables are created.  

Makes sense to me too. Thanks @talumbau. Closing.

PR #208, opened 2015-04-01
utils.expand_array accepts arrays with None

talumbau:
This is needed to handle variation in how a user may specify multi-valued parameters in the webapp. In particular, they may provide different numbers of values (per year) for each component of a multi-valued parameter.


talumbau:
merging soon in preparation for new deployment.

Issue #181, opened 2015-03-11
Add a results row to the distribution and difference tables


jlyons871:
In both DataFrames returned by `utils.create_distribution_table()` and `utils.create_difference_table()`, a "results" row (at the bottom) needs to be added. This will contain a sum of all the corresponding data for that column, from all of the rows.

talumbau:
it's probably more accurate to call it a "sum" or "all" row, rather than "results", but yes it is basically just a sum.

MattHJensen:
Resolved. Closing.

PR #209, opened 2015-04-02
added sum row for diff and dist table

jlyons871:
This adds an additional row to the bottom of the distribution_table and difference_table, containing the summation of the column above, where applicable. #181

### DIFF
--- a/taxcalc/tests/test_utils.py
+++ b/taxcalc/tests/test_utils.py
@@ -274,6 +274,10 @@ def test_dist_table_sum_row():
     t2 = create_distribution_table(calc1, groupby="large_agi_bins", result_type="weighted_sum")
     assert(np.allclose(t1[-1:], t2[-1:]))
 
+    t3 = create_distribution_table(calc1, groupby="small_agi_bins", result_type="weighted_avg")
+    for col in t3:
+        assert(t3.loc['sums', col] == 'n/a')
+
 
 def test_diff_table_sum_row():
     cur_path = os.path.abspath(os.path.dirname(__file__))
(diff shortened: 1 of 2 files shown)


jlyons871:
currently if `create_distribution_table ` is called with `result_type == "weighted_avg"`, it does not append the row to the bottom. Since the numbers are averages it does not seem to make sense to have a sum row.

feenberg:
On Thu, 2 Apr 2015, Jake Lyons wrote:

> 
> currently if create_distribution_table is called with result_type == "weighted_avg", it does not append the row to
> the bottom. Since the numbers are averages it does not seem to make sense to have a sum row.

Can we put in a weighted average instead?

dan

jlyons871:
@feenberg I'm not sure I understand your question. Do you mean to have the table only display weighted averages and not weighted sums?  

Currently, the distribution table currently has rows corresponding to the [binning methods](https://github.com/OpenSourcePolicyCenter/Tax-Calculator/blob/master/taxcalc/utils.py#L270-278) (either large agi/tpc, small agi/soi, or by weighted deciles) of incomes and the columns corresponding to these [tax variables](https://github.com/OpenSourcePolicyCenter/Tax-Calculator/blob/master/taxcalc/utils.py#L9-12) with the output of each cell representing either the weighted sum or the weighted average. 

And this new "summation" row will be displayed in the webapp. For example, one cell in the table, when `result_type == "weighted_sum"` and `groupby == "large_agi_bins"`,  would represent the weighted sum of those with incomes between 49,999 and 74,999 for `_ospctax`. This new sums row at the bottom of the table, sums up all of the cells for each column. So there would be a cell with the summation of the `_ospctax`  for all of the incomes/rows. 

When `result_type == "weighted_avg"`, do you agree that it makes sense for this summation row to have values of 'n/a' for each column? To me this makes the most sense in the current format. Is there another format that would make more sense?

talumbau:
I'm interested in merging this, but I'm unsure if Dan's question has been addressed. I need the "sum" row for the next iteration of dropq/taxcalc/webapp.

MattHJensen:
@talumbau, let's go ahead with the sum row for now to get everything up and running, and then I want to go over everything in much more depth with Dan before the following iteration.

PR #215, opened 2015-04-06
impute extra AMTI from 2008 puf

MattHJensen:
There are several AMT preference items that we do not observe on the public use file. We do, however, observe total AMTI. Therefore, we can impute the value of the unobserved AMT preferences by using the difference between the AMTI that we calculate using observed AMT preferences and total AMTI in 2008. 

This PR turns on that imputation. 

Without this imputation, our AMTI (and AMT) is lower than where we would expect it to be from TPC and SOI tables. With this imputation, our AMTI is a little too high. Later we will want to impute the AMTI preference items directly using a 2000 Treasury paper on ["Who Pays the Individual AMT?"](http://www.treasury.gov/resource-center/tax-policy/tax-analysis/Documents/ota87.pdf)


MattHJensen:
@Amy-Xu could you review before I merge? 

This PR changes taxes:

Master
image

PR
image

Amy-Xu:
:+1:

PR #217, opened 2015-04-06
Education credit changes

xiyuw123:
Added 3 lines of code to NonEdCr that should calculate the amount of nonrefundable education credits

### DIFF
--- a/taxcalc/functions.py
+++ b/taxcalc/functions.py
@@ -1133,7 +1133,7 @@ def RefAmOpp(_cmp, c87521, _num, c00100, EDCRAGE, c87668):
 @iterate_jit(parameters=["ETC_pe_Married", "ETC_pe_Single"], nopython=True)
 def NonEdCr(c87550, MARS, ETC_pe_Married, c00100, _num,
     c07180, e07200, c07230, e07240, e07960, e07260, e07300,
-    e07700, e07250, t07950, c05800, _precrd, ETC_pe_Single):
+    e07700, e07250, t07950, c05800, _precrd, ETC_pe_Single, _xlin3, _xlin6, c87668, c87620):
 
     # Nonrefundable Education Credits
     # Form 8863 Tentative Education Credits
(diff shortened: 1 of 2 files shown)


MattHJensen:
Could you open a corresponding issue, with the "tax" and "bug" labels where you describe the problem in the current code? Then you can link to that issue from this PR.

xiyuw123:
Hi Matt, I just opened an issue for the education credits. I don't have permission to add labels though. Here it is: https://github.com/OpenSourcePolicyCenter/Tax-Calculator/issues/218

MattHJensen:
@xiyuw123, could show the diagnostic table before and after this change?

xiyuw123:
Sure. This is the original master output:
ospc master output

This is the output after the education credits change:
ospc education credits

MattHJensen:
Did you expect both refundable and nonrefundable credits to change?

xiyuw123:
Only nonrefundable credits should have changed. I'm not sure at the moment why the nonrefundable credits change also.

xiyuw123:
So the education credits also seem to change for the 2011 calculations, the one that was originally included in the SAS code.

2011 with the code for education credits commented out:
ospc 2011 output no educr changes

2011 with the code left in:
ospc education credits

MattHJensen:
Looks like the second table is for 2013 instead of 2011.

xiyuw123:
Sorry about that; I posted the wrong picture! It should be fixed now.

MattHJensen:
Thanks! It seems like you should keep digging into the link in our code between the education credit and other refundable credits.

xiyuw123:
So the reason the nonrefundable credits change is because the education credits are used in the calculations for the child tax credit. A higher education credit decreases the amount of the child tax credit. The child tax credit is then used in the (refundable) additional child tax credit, which will increase if the child tax credit decreases. As a result, the refundable credits also increase. The additional child tax credit is pretty much the only refundable credit affected by the education credit.

MattHJensen:
Opps, looks like this got merged in as a part of #220. @xiyuw123, could you open an issue that describes the fact that you're still figuring out the interaction between the nonrefundable education credit and the refundable ACTC?

xiyuw123:
Yep, sure. I've been looking through the forms for the child tax credit and the ACTC and it seems like that's what supposed to happen, but here is the issue: https://github.com/OpenSourcePolicyCenter/Tax-Calculator/issues/239

PR #220, opened 2015-04-06
Account for float arithmetic in utils.py 'count' functions

xiyuw123:
Weighted_count_gt_zero and weighted_count_lt_zero now properly account for arithmetic between floats.

### DIFF
--- a/taxcalc/utils.py
+++ b/taxcalc/utils.py
@@ -207,19 +207,19 @@ def expand_array(x, inflate, inflation_rates, num_years):
 
 
 def count_gt_zero(agg):
-    return sum([1 for a in agg if a > 0])
+    return sum([1 for a in agg if a > 0.0001])
 
 
 def count_lt_zero(agg):
-    return sum([1 for a in agg if a < 0])
+    return sum([1 for a in agg if a < -0.0001])
 
 
 def weighted_count_lt_zero(agg, col_name):
-    return agg[agg[col_name] < 0]['s006'].sum()
+    return agg[agg[col_name] < -0.0001]['s006'].sum()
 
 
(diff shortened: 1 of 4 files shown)


MattHJensen:
@talumbau, could you review before we merge? The problem is that when we run policy reforms like increasing the top tax rate, some taxpayers will have a fractional reduction in total taxes and show up in the tax cut column -- instead their taxes should be considered unchanged. .

taxcalc/utils.py
```diff
@@ -215,11 +215,11 @@ def count_lt_zero(agg):
 
 
 def weighted_count_lt_zero(agg, col_name):
-    return agg[agg[col_name] < 0]['s006'].sum()
+    return agg[agg[col_name] < -0.0001]['s006'].sum()
 
 
 def weighted_count_gt_zero(agg, col_name):
-    return agg[agg[col_name] > 0]['s006'].sum()
+    return agg[agg[col_name] > 0.0001]['s006'].sum()
```
talumbau: I strongly prefer a `tolerance` keyword arg with a default of `0.0001` or `-0.0001` for both functions. Also, a test function including with something like `weighted_count_gt_zero(agg, col, tolerance=0.004)` or similar would prove it was doing the right thing.

MattHJensen:
@xiyuw123, could you add a test for this as well? @Amy-Xu can help you get started and then T.J. can review.

MattHJensen:
@talumbau, could you review again before we merge?

talumbau:
looks good to me!

MattHJensen:
Thanks Xiyu, T.J., and Amy! Merging

PR #231, opened 2015-04-07
Interest paid deduction blow up factor

Amy-Xu:
In this PR, I added one new column of blow up factor in StageIFactors.csv file for interest paid deduction and apply this blow up factor (AIPD) to the corresponding variable in PUF (e19200) in the blowup function of taxcalc.

Originally we used personal income first and then interest income blow up factors for the extrapolation of interest paid deduction, but neither of them produces results close to SOI statistics. To hit the actual deduction numbers perfectly, at least for years from 2008 to 2012,  we can use SOI statistics to generate a blow up factor just for interest paid deduction. 

First, I collected the interest paid deduction total amount from SOI tax stats, table 2.1 for each year from 08 to 12, and calculated blow up factors using 08 value as base. Then I applied the growth rate of personal income blow up factor (ATXPY) to this new blow up factor to generate values from 2013 onwards.


MattHJensen:
@Amy-Xu, could you verify that if you change the blowup factor for e19200 back to ATXPY on this branch that you get the same ospctax as before? If that works fine, I'll go ahead and merge.

taxcalc/utils.py
```diff
@@ -4,10 +4,10 @@
 from collections import defaultdict
 
 STATS_COLUMNS = ['c00100', '_standard', 'c04470', 'c04600', 'c04800', 'c05200',
-                 'c62100','c09600', 'c05800', 'c09200', '_refund', 'c07100',
-                 '_ospctax','s006']
+                 'c62100','c09600', 'c05800', 'c09200', '_refund', 'c07100', '_ospctax',
+                 's006']
 
-TABLE_COLUMNS = ['s006','c00100', 'num_returns_StandardDed', '_standard',
```
MattHJensen: What is the reason for these changes in utils.py?

Amy-Xu:
screen shot 2015-04-07 at 3 36 51 pm
So here's the results if I switch it back to ATXPY. I think they are the same right?

taxcalc/utils.py
Amy-Xu: This a merge conflict where I deleted my version and keep the ospc master version. Not sure why it shows the diff like this.

MattHJensen: Seems like something might have gotten thrown off somehow. TABLE_COLUMNS on OSPC master right now are 

TABLE_COLUMNS = ['s006','c00100', 'num_returns_StandardDed', '_standard',
                 'num_returns_ItemDed', 'c04470', 'c04600', 'c04800', 'c05200',
                 'c62100','num_returns_AMT', 'c09600', 'c05800',  'c07100','c09200',
                 '_refund','_ospctax']

MattHJensen:
Great. They are the same. just waiting on the utils.py issue now.

Amy-Xu:
Yep should be good right now

MattHJensen:
Sweet. merging.

PR #240, opened 2015-04-09
10_Min notebook fixed and improved

MattHJensen:
This PR updates the 10_Minutes_to_Taxcalc notebook so that it now reads the calcY parameters from an example parameter JSON file. 

A  concern that I have is that the new docs/example_param_files/repeal_state_local_tax_deduction.json will now need to be kept in sync with changes to params.json. The way that I can see doing this right now seems a bit onerous: 

1. copy the updated params.json
2. paste params.json into the example_param_files folder
3. delete repeal_state_local_tax_deduction.json
4. rename example_param_files/params.json as example_param_files/repeal_state_local_tax_deduction.json. 

@talumbau, @jlyons871, @Amy-Xu, any other ideas?


MattHJensen:
In the last two commits, I have changed the way in which a user specifies plan Y parameters in our example notebook. 

Before, the user loaded a plan y json file. Now, the user specifies user_mods.  

I made this change for two reasons. 

1. it would be difficult for us to keep an example plan Y json file up to date, as described in the comment above.  
2. For a user who conda installs the calculator, loading a plan Y json file would have been their first exposure to our parameter files and that would have been both confusing and logistically complicated (we would have had to give them an up-to-date- parameter file to use).

Does anyone oppose using user_mods, at least in the near term? I imagine we will also want docs telling 'power' users that they can load a plan Y param file, but it would be better if that is not an ipython notebook so we don't have to keep the an example param file up to date.

Amy-Xu:
I agree - user_mods would be a better choice. It doesn't require users to know much about params.json so it's easier for them to change just one or two parameters through user_mods.

MattHJensen:
Ok. Thanks @Amy-Xu. I'm going to merge for now so that we have something that is working and we can revisit later.

PR #234, opened 2015-04-08
Add a haircut to the state and local tax deduction

MattHJensen:
This PR adds the ability to apply a haircut to the state and local tax deduction. 

See issue #144.

This is [CBO budget options](https://www.cbo.gov/sites/default/files/cbofiles/attachments/49638-BudgetOptions.pdf) 51.

I am breaking PR #201 into pieces, and this is one of those.

@Amy-Xu could you please review before I merge?

PR #243, opened 2015-04-10
Haircut parameters for student loan deduction and self employment tax deduction

Amy-Xu:
Haircut parameters were added in function Adj() for several adjustments. Two of these adjustment variables (e03210: student loan adjustment, e03260: self employment tax adjustment) are used for other calculations later on in functions.py. To make sure the haircut parameters function properly, we need to have them applied to the corresponding adjustment variables each time they are used in functions.py. In this PR, I added the haircut parameters in the needed spots.

### DIFF
--- a/taxcalc/functions.py
+++ b/taxcalc/functions.py
@@ -34,8 +34,8 @@ def Adj(   e35300_0, e35600_0, e35910_0, e03150, e03210, e03600, e03260,
     return (_feided, c02900)
 
 
-@iterate_jit(parameters=['FEI_ec_c'], nopython=True)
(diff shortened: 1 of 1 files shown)


Amy-Xu:
screen shot 2015-04-10 at 5 14 24 pm
The diagnostic table won't be changed as a result of this PR. @MattHJensen Could you review the changes?

MattHJensen:
@Amy-Xu, looks good. Could you add a similar diagnostics table for master before we merge, though? See, #234 for an example.

Amy-Xu:
Yep here's the table from master

screen shot 2015-04-13 at 9 13 58 am

MattHJensen:
Thanks. Looks good and merging.

PR #259, opened 2015-04-29
edits contributor_guide.rst

rkuchan:
Thanks to @martinholmer for his feedback on `contributor_guide.rst.` This PR makes the following changes to the Setup Git section:
* In step 4, prompt contributors to sign in to GitHub before creating a fork.
* Change all `>>>` to `$` in the code-blocks - and add a footnote about prompts.
* In step 8, remove the `git add origin` step.
* Add step 9 to make sure that py.test works after the initial setup.

This PR also fixes a typo in step 8 of the Workflow section and removes the Behavior Example from the Example Code section

### DIFF
--- a/docs/source/contributor_guide.rst
+++ b/docs/source/contributor_guide.rst
@@ -21,15 +21,15 @@ Setup Git
 
 3. Tell Git to remember your GitHub password by following steps 1-4 on `password setup`_. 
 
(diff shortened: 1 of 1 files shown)


MattHJensen:
Looks good. Merging! If anything comes up later we can tinker then.

PR #261, opened 2015-05-06
remove Calculator attribute access to params/records 

MattHJensen:
I am wondering what you all think about removing the Calculator's attribute access to params and records. This PR demonstrates what that would look like. 

My sense from observing new users is that having the API rely on the Calculator's attribute access to params and records makes it more difficult for people to grasp the structure of the calculator and has been a point of confusion for our new users/contributors. 

For example, rather than writing  `calc.II_em` to access the personal exemption amount, it would be more expressive if the user were to write `calc.params.II_em`. 

This would also simplify our codebase considerably. 

A penalty is that the API will be more verbose, and the user will have to remember which data are tax law parameters (params) and which data are tax unit characteristics (records) when using the API. That said, nothing will change in functions.py, where the bulk of the tax logic is -- we won't have to use this notation there. Furthermore, our encouragement for the user to think of the data as belonging to either params or records may be educational.  

In recognition that there would be more to type under this scheme, I renamed `parameters` to `params`.

### DIFF
--- a/taxcalc/parameters.py
+++ b/taxcalc/parameters.py
@@ -6,7 +6,7 @@ from pkg_resources import resource_stream, Requirement
 
 DEFAULT_START_YEAR = 2013
 
-class Params(object):
+class Parameters(object):
 
 
     CUR_PATH = os.path.abspath(os.path.dirname(__file__))
@@ -159,7 +159,7 @@ class Params(object):
 
 def default_data(metadata=False, start_year=None):
     """ Retreive of default parameters """
-    parampath = Params.params_path
+    parampath = Parameters.params_path
     if not os.path.exists(parampath):
         path_in_egg = os.path.join("taxcalc", Params.PARAM_FILENAME)
         buf = resource_stream(Requirement.parse("taxcalc"), path_in_egg)
@@ -167,7 +167,7 @@ def default_data(metadata=False, start_year=None):
         as_string = _bytes.decode("utf-8")
         params = json.loads(as_string)
     else:
-        with open(Params.params_path) as f:
+        with open(Parameters.params_path) as f:
             params = json.load(f)
 
     if start_year:
@@ -194,10 +194,10 @@ def default_data(metadata=False, start_year=None):
                 if v['cpi_inflated'] is True:
                     if isinstance(new_val, list):
                         for y in range(last_year_for_data, start_year):
-                            new_val = [x * (1.0 + Params._Params__rates[y]) for x in new_val]
+                            new_val = [x * (1.0 + Parameters._Parameters__rates[y]) for x in new_val]
                     else:
                         for y in range(last_year_for_data, start_year):
-                            new_val *= 1.0 + Params._Params__rates[y]
+                            new_val *= 1.0 + Parameters._Parameters__rates[y]
                 #Set the new values
                 v['value'] = [new_val]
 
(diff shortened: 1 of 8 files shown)


MattHJensen:
Here are the diagnostics tables demonstrating no change in output, as expected 

###### Before (master branch)
image

###### After (api branch)
image

taxcalc/parameters.py
```diff
@@ -6,7 +6,7 @@
 
 DEFAULT_START_YEAR = 2013
 
-class Parameters(object):
+class Params(object):
```
talumbau: I'm fine with calling the attribute on the `Calculator` object `params` instead of `parameters`, but I think it's OK to leave the class name as `Parameters` as one is rarely typing out the full class name. Just a thought.

MattHJensen: I thought that it might be confusing to API users if Calculator attribute and class name match for Records and records, but not for Parameters and params.

The question is whether this is confusing: 

    tax_dta = pd.read_csv("../../puf.csv")
    # Create a default Parameters object
    paramsX = Parameters(start_year=2013)
    recordsX = Records(tax_dta)
    # Create a Calculator
    calcX = Calculator(records=recordsX, params=paramsX)

 I do agree that Parameters is a better name for the class, though. What do you think?

zrisher: I think it's generally expected that object names are never abbreviated, but parameter names (including those that refer to objects) can be depending on context. Even in Ruby, `parameters` is one of those words used so commonly that we're ok abbreviating it to `params`. So I would doubly expect python coders to understand:
```
paramsX = Parameters(start_year=2013)
calcX = Calculator(records=recordsX, params=paramsX)
```
pretty easily.

MattHJensen:
Per @talumbau's and @zrisher's suggestion, I just reverted the class name to `Parameters`, leaving the Calculator attribute with the new name `params`. 

As a more general matter, what do people think about forcing API users to type out calc.params.II_em rather than allowing them to type calc.II_em? I believe I was one of the original advocates for the calc.II_em approach, but lately it feels like the shortcut makes it harder for newcomers to conceptualize how the calculator is structured. 

Looking for feedback from @talumbau and @zrisher, but also from @rkuchan and @SherwinLott as relatively new contributors. 

I spoke with @Amy-Xu, and she was a bit concerned by the extra typing that would be required when playing around with the calculator in a notebook, which I totally get.

talumbau:
About the extra typing in the notebook, with this proposed style, tab-completion in the IPython notebook will work fine, so typing `calc.p<tab>` will get completed to `calc.params`. I suspect people would quickly get used to that.

I am fine with these changes, as it simplifies the implementation (e.g. `__deepcopy__` is no longer needed since we are not mucking with attribute access anymore). If user of taxcalc are happy with it too, I think it's a win-win.

MattHJensen:
Ahh, yes just removed __deepcopy__. 

Also, here is the latest diagnostic_table (still no change from master) 

image

Amy-Xu:
Almost forget tab-completion. Then it's almost the same in terms of typing variable names. I'm totally good with this change.

rkuchan:
I wouldn't mind the extra typing. I think the longer version would help me keep the structure of the tax calculator in mind when working - I like that.

zrisher:
I'm definitely a big :+1:

MattHJensen:
Sounds like it's a go -- I'll remove the question label and WIP tag. @Amy-Xu, could you review the diffs before I merge?

Amy-Xu:
look good to me!

MattHJensen:
Thanks Amy. Merging.

Issue #203, opened 2015-03-30
The same calculator gives different tax computations under a variety of use cases


MattHJensen:
Scenario 2 and 4, below, are problematic. 

Scenario 1: Load one calculator, increment year, and then calculate taxes.
![image](https://cloud.githubusercontent.com/assets/8301092/6907202/2271fc50-d702-11e4-9ced-5959444d20c1.png)

Scenario 2: Load one calculator, calculate taxes, increment year, and then calculate taxes again.
![image](https://cloud.githubusercontent.com/assets/8301092/6907214/4125350e-d702-11e4-99f7-fa3445be5025.png)

Scenario 3: Load two calculators 'properly' (using records and records2, params and parms2), calculate taxes for the first, increment year, and then calculate taxes again.
![image](https://cloud.githubusercontent.com/assets/8301092/6907241/8b537f50-d702-11e4-9d59-863354f44d40.png)

Scenario 4: Load two calculators improperly, increment year for the first, calculate taxes for the first, increment year for the second, and then calculate taxes for the second.
![image](https://cloud.githubusercontent.com/assets/8301092/6907339/5289f5f4-d703-11e4-9d24-a5e1747a62be.png)

While scenario 2 is clearly a bug, scenario 4 is suboptimal as well because it requires creating records1, records 2, params1, and params 2 to load two calculators, while penalizing you with an unreported error if you do not do so.

MattHJensen:
Fixed with 0b24eba. Continuing discussion of how to avoid bugs like this in the future with #272. Closing.

PR #262, opened 2015-05-06
Fix bug in NIIT

MattHJensen:
We currently have a bug that causes us to get different results under the two scenarios below:

1. load a calculator, increment year, and calculate taxes 
2. load a calculator, calculate taxes, increment year, and then calculate taxes again. 

This corresponds to Scenario 1 and Scenario 2 of issue #203. 

I have identified the net investment income tax (NIIT) as the source of this bug. 

NIIT is currently reassigned if AGI is greater than the NIIT threshold: 

    if c00100 > NIIT_thd[MARS - 1]:
        NIIT  = NIIT_trt * min(e00300 + e00600 + max(0, c01000)
                + max(0, e02000), c00100 - NIIT_thd[MARS - 1])

This means that if we calculate taxes, increment year, and calculate taxes again, something like this could happen to a record: In the first year, AGI is greater than the NIIT threshold, so NIIT is assigned to a non-zero value. In the second year, AGI is less than the NIIT threshold, so NIIT is neither reassigned nor is it zero-- rather it would be the value that it was assigned in the first year. This is a bug.

I fixed this problem by adding an else statement that explicitly assigns NIIT to zero if it does not pass the if condition:

    if c00100 > NIIT_thd[MARS - 1]:
        NIIT  = NIIT_trt * min(e00300 + e00600 + max(0, c01000)
                + max(0, e02000), c00100 - NIIT_thd[MARS - 1])
    else: NIIT = 0
    return NIIT

Alternatively, I could have replaced all of that with: 
        
    NIIT  = max(0, NIIT_trt * min(e00300 + e00600 + max(0, c01000)
                + max(0, e02000), c00100 - NIIT_thd[MARS - 1]))

One question is whether we should be worried about this type of problem occurring again in the future. We could, for instance, re-zero all of the `zeroed_names` from records.py as a part of increment_year().


feenberg:
I have found it helpful to zero everything in both the fortran and SAS versions of my calculator.

MattHJensen:
FYI, I just reworded the description of the bug in the first comment. It should be clearer now.

MattHJensen:
Note that this bugfix only slightly affects our diagnostics table since the bug is so small and we round everything off. 

###### BEFORE (branch master)
image



###### AFTER (branch issue203)
image

MattHJensen:
Fixed with 0b24eba. Continuing discussion of how to avoid bugs like this in the future with #272. Closing.

PR #269, opened 2015-05-18
fix for weighting function

talumbau:
- any column that starts with 's006' is not weighted.

This is necessary to fix a problem in dropq.


MattHJensen:
Looks good to me.

PR #273, opened 2015-05-22
[WIP] Ascii output

xiyuw123:
Created a file that converts the csv results from test.py into ascii with uniform columns. The ascii output generator records certain lines in the csv file corresponding to the row you would see in excel. It creates a text file with this data and then transposes it to a final results file for easier viewing.


Amy-Xu:
This looks good to me. It will be very handy for users who want to retrieve results in a nice format. There's one thing I want to confirm, at [line 32](https://github.com/OpenSourcePolicyCenter/Tax-Calculator/pull/273/files#diff-4fa4ed70d0d9bc886f9a52733adb9a36R32), the list variable holds the row numbers of records rather than the actual record IDs (recid) right?

xiyuw123:
Yep, this was mostly because it was easier to reference and check the data by row numbers in the csv file. If it's easier to just input the recid's it should also be possible to match based on the recid's.

Amy-Xu:
Row numbers definitely work for me. I have not thought about using record IDs directly, but to me it seems easier to use recid directly because I usually get recid first from ipython notebook and then open puf to get row numbers based on recid. But the downside would be a fair amount of work to revise this script. I'm not sure it worth the effort or not. What do you think? @MattHJensen @xiyuw123  

If we keep using row numbers, do you think it might a bit more straightforward if we use something like row_number_list to name this variable?

MattHJensen:
As a general matter, I'm a proponent of merging in useful features and then opening an issue for enhancements and then adding them through an additional PR. That philosophy might apply in this case after Amy's var-name suggestions are discussed.

MattHJensen:
@xiyuw123  It looks like the workflow is to run test.py to generate a csv and then read in the csv created by test.py to create an ascii. Is that correct? If so, is there a reason for doing this rather than directly creating an ascii?

xiyuw123:
Yes, that's correct. This is mainly because at the time I already had some result csv files saved on my computer from previous tests I did that I had to convert to ascii files for Dan. I don't see any reason why the feature couldn't be part of test.py, and I'd be happy to implement that if you want.

MattHJensen:
Ahh, that makes sense. I was thinking that it would be better as a standalone script actually, that doesn't use test.py at all (it may end up calling similar functions, though). I suppose an alternative might be to give test.py users an option of outputting either csv or ascii, which seems to be what you are suggesting. @talumbau, do you have an opinion on this?

talumbau:
I'm not sure I'm 100% dialed in to what the final result of this script produces. I'll add a few comments though that have occurred to me while looking at this PR.

ascii_output_v3.py
```diff
@@ -0,0 +1,100 @@
+"""
```
talumbau: Why is this file `v3`? I don't think there's a `v1` or `v2`

ascii_output_v3.py
```diff
@@ -0,0 +1,100 @@
...
+            col_elem = col_elem[:cut]
+        else:
+        #adds spaces to fill out data string if length is smaller than col_size
+            col_elem = col_elem.ljust(col_size)
+        temp.write(col_elem)
+        nrows += 1
+    return nrows
+
+def ascii_output():
+    csv_results = "results_puf.csv" #input csv file
```
talumbau: It seems odd that we have a hard coded filename here that reads a particular output file that has a hard coded filename in test.py. Why not have some kind of flag that indicates, when one does the work that produces `results_puf.csv` that you also want what happens in this file? That way you wouldn't have to execute `test.py` and then this script.

ascii_output_v3.py
```diff
@@ -0,0 +1,100 @@
+"""
+Turns csv output of tax calculator to ascii output with uniform columns
+and transposes data so columns are rows, rows are columns
+
```
talumbau: This description of the work makes me think that this should be done with `pandas` commands. Is there a strong reason to not consider DataFrame manipulations and then the `to_csv` method on DataFrames to do this work?

ascii_output_v3.py
```diff
@@ -0,0 +1,100 @@
...
+    nrows = 0
+    while (data != []):
+        col_elem = data.pop()
+        #cuts a data string from the end if length is larger than col_size
+        if (len(col_elem) > col_size):
+            cut = col_size - len(col_elem)
+            col_elem = col_elem[:cut]
+        else:
+        #adds spaces to fill out data string if length is smaller than col_size
+            col_elem = col_elem.ljust(col_size)
```
talumbau: For this kind of work (formatting a column to match up some size), I would suggest "formatters" in DataFrames:
http://stackoverflow.com/questions/20937538/how-to-display-pandas-dataframe-using-a-format-string-for-columns

talumbau:
I think a good idea here would be to create a (gist)[https://gist.github.com/] that shows a trivial file (~10-15 lines, 6-10 columns) of a toy "results_puf.csv" and show what this script produces. Then we could decided if this should be handled with DataFrame operations and formatters, or if this more special case code is called for.

xiyuw123:
@talumbau Sorry, I'll give a bit more background on the script. I had been running a few tests with the tax calculator and I produced a few results csv files from it. We later learned that, to compare these results with Dan's, they needed to be converted into a ascii file with fixed column widths. So the file on hand right now I originally designed mostly just to convert the result csv files I had on hand.

So it had to change something like this:

<script src="https://gist.github.com/xiyuw123/636d48015c77d0297082.js"></script>

To this:

<script src="https://gist.github.com/xiyuw123/2093157ec17c5dde9b3e.js"></script>

So in the first one, the csv file, each different row is a different tax payer, and the columns are the information about them. In the second one, this data is pretty much transposed.

Also, for the v3 part, that's because there's actually a few different files on my computer, I was just too lazy to change the name before the PR.

Hope this helps!

talumbau:
Ok. This makes sense to me now. Is there any reason not to use the tools in `pandas` to do this for us? I prefer to do that as much as possible because then I don't have to worry about finding mistakes in my own code. For example, If I have a small text file that looks like this:
```
A,B,C,D
1,2,3,4
5,6,7,8
9,10,11,12
```
The following code will do (as far as I can tell) the kind of operation you want to do:
```
import pandas as pd
df = pd.read_csv("test.txt")
out = df.T.reset_index()
out[2] = out[2].map('{:5.2f}'.format)
out.to_csv("testout.txt", header=False, index=False)
```
The file `testout.txt` now contains:
```
A,1,5, 9.00
B,2,6,10.00
C,3,7,11.00
D,4,8,12.00
```
Notice that the last column uses Python string formatting to line up the numbers. You can apply such formatting to any columns you want. If you start with what I have written here, I think you will get a much more compact script - also faster too because `pandas` has fast CSV reading and writing.

xiyuw123:
There's really no reason why not to use pandas, I just didn't use it because I wasn't too familiar with it.

talumbau:
OK, I think the best long term asset here would be delivered by doing as follows:

- [ ] renaming the script to a better name
- [ ] modifying the script to use a pandas-style modification and output of the data
- [ ] creating a small test that shows how the script modifies a table

I would keep all of the comments (or as many as makes sense) as they are quite helpful

xiyuw123:
ok, I will start working on it

xiyuw123:
Here is the new script using pandas:
https://gist.github.com/xiyuw123/d819f728de1859df5a98

Here is a test file that I ran it on:
https://gist.github.com/xiyuw123/57d5edd6c2cfdf3bac8a

For the output file, I set recids = [1,4,5]. This is the result:
https://gist.github.com/xiyuw123/57d5edd6c2cfdf3bac8a

Tell me how it looks!

talumbau:
Ok, those gists look great. Seems like a very useful helper script to add. The next step is to either take that code and make it part of this pull request, or close out this pull request and create a new one with the new additions. The helper script should live at the top of the repo and the small test case should live in taxcalc/tests (and be callable by pytest, just like our other tests)

xiyuw123:
I've been looking through some of the other test files, but I'm still not exactly sure how pytest works - should I edit any other file to have it include a new test or just add a file in the tests folder with assert statements?

talumbau:
I would either add to `taxcalc/tests/test_utils.py` or create a new file called `taxcalc/tests/test_helpers.py`.  A good description of how pytest finds tests to run can be found in [this part of the docs](https://pytest.org/latest/goodpractises.html#test-discovery)

Issue #213, opened 2015-04-04
diagnostic_table() should be a standalone function


MattHJensen:
[diganostic_table() is currently a Calculator method](https://github.com/OpenSourcePolicyCenter/Tax-Calculator/blob/master/taxcalc/calculate.py#L277-L360). Instead, it should be broken into two standalone functions in utils.py. 

1. diagnostic_table() should be a non-mutating function that accepts a calculator and produces a table for a single year, similar to the other table generators in utils.py
2. There should also be a "multi year diagnostics table generator" that accepts a calculator and a number of years, creates a hardcopy of that calculator, and then does something like

```  
df = []
for i in num_years:
    calc.calc_all()
    df1 = diagnostic_table(calc)
    df = df.append(df1)
    calc.iterate_year()
return df
```

talumbau:
Agree. @MattHJensen, is this a task that you see as falling in my bucket or someone else's?

MattHJensen:
I'm on this one.

martinholmer:
Issue #213 has been resolved by pull requests #848 and #852.

PR #274, opened 2015-05-24
Move diagnostic_table to utils

MattHJensen:
Deals with #213.  

1. Moves the modification of `c04470` and creation of `num_returns_ItemDed`, `num_returns_StandardDed`, and `num_returns_AMT` from `create_distribution_table()` to `results()` . 
2. Adds an "everyone" groupby option to `create_distribution_table()` 
    - Yes, this is bit odd since a table of aggregates is not technically a "distribution" table. Still, it didn't seem worth a new function.
3. Adds a new `create_diagnostic_table` function to `utils.py`. 
    - The functionality is a bit different than `diagnostic_table()` in ways that I hope make it better for diagnosing problems with a calculator and comparing the results of a topic branch against master.
        1. The default number of years takes you through the budget window (2013 - 2024)
        2. None of the numbers are rounded. (We may want to add the option of applying formatting later)
    - One downside of these changes is that a full table requires a scrollbar in ipython notebook, which will make it hard to take a screenshot of the entire table and paste it into a PR. I am going to try to turn off the word wrapping of the index, which will actually aggravate the horizontal problem. I'm not sure what to do about this yet.
4. Deletes the `diagnostic_table()` Calculator module
5. Rewrites the relevant test to use create_diagnostic_table() rather than the old diagnostic_table() and make the calculator used in the test start at 2013 instead of 2008. 
    - I could not think of a good reason to start at 2008 instead of 2013 -- if there is one, please let me know. 

cc for review @talumbau and @Amy-Xu 

NEW TABLE

image

OLD TABLE

image


taxcalc/utils.py
```diff
@@ -428,8 +445,48 @@ def create_distribution_table(calc, groupby, result_type):
...
+
+    num_years : Number of years to include in the diagnostic table. 
+                Default is set to 12 to include 2013 to 2024.
+
+    Returns
+    -------
+    DataFrame where columns are years and rows are aggregated tax variables. 
+    """
 
+    calc = copy.deepcopy(calc)
```
talumbau: Could be reasonable to allow the caller to specify whether or not to mutate the `calc` object. So something like a keyword argument `make_copy=True`, so that way if you don't mind messing up the `calc` object to make the table you don't have to pay for a copy. Just a suggestion though. I don't feel strongly.

talumbau:
Looks like a definite improvement in the table. My one comment aside, I am +1 on this.

MattHJensen:
Thanks T.J. I'll clean this up, incorporate the `make_copy=True` kwarg, and ask for a merge.

MattHJensen:
@talumbau, is 62a8180 as you expected?

talumbau:
Yes, looks good!

MattHJensen:
PR Branch
image

Master
image

MattHJensen:
The content here is good now and ready to be merged. One thing that is bugging me is that I can't figure out how to prevent the index column from wrapping. Our old diagnostic table didn't wrap the row labels, but the new one does, and it means I often have to zoom out to get all of the rows in a screenshot. 

If anyone has a suggestion before I merge, it would definitely be appreciated!

MattHJensen:
@talumbau, if everything looks good to you, could you merge? I will then open an issue on the index column wrapping thing from my previous comment.

taxcalc/utils.py
```diff
@@ -406,7 +411,26 @@ def results(c):
                 outputs.append(getattr(c.records, col))
         else:
             outputs.append(getattr(c, col))
-    return DataFrame(data=np.column_stack(outputs), columns=STATS_COLUMNS)
+
+    res = DataFrame(data=np.column_stack(outputs), columns=STATS_COLUMNS)
+
+    # Keep itemized deductions (c04470) where itemized deductions reduce AGI
```
talumbau: Hmm.. this function had previously served the purpose of "collect up the columns we care about and return them." Now it will do some useful work. Is there a better place for this logic? Totally possible that the answer is 'no', or 'maybe, but we can figure it out after merging'. This PR has been up for a while, apologies for just raising this now.

feenberg: On Fri, 10 Jul 2015, T.J. Alumbaugh wrote:

> 
> In taxcalc/utils.py:
> 
> > @@ -406,7 +411,26 @@ def results(c):
> >                  outputs.append(getattr(c.records, col))
> >          else:
> >              outputs.append(getattr(c, col))
> > -    return DataFrame(data=np.column_stack(outputs), columns=STATS_COLUMNS
> )
> > +
> > +    res = DataFrame(data=np.column_stack(outputs), columns=STATS_COLUMNS)
> > +
> > +    # Keep itemized deductions (c04470) where itemized deductions reduce 
> AGI

The comment should read "keep itemized deductions where they reduce 
taxable income" (not AGI).

dan

MattHJensen:
Closing this in favor of #848. Thanks @martinholmer.

Issue #271, opened 2015-05-19
The 'Imputations' in records.py should be refactored


talumbau:
This code:

```
        """Imputations"""  # TODO Move these into a separate script with a decorator
        self._cmbtp_itemizer = (-1 * np.minimum(np.maximum(0., self.e17500 - np.maximum(0., self.e00100) * 0.075), 0.025 * np.maximum(0., self.e00100))
                        + self.e62100 + self.e00700 + self.e04470 + self.e21040
                        - np.maximum(0, np.maximum( self.e18400, self.e18425)) - self.e00100 - self.e18500
                        - self.e20800)
        self._cmbtp_standard = self.e62100 - self.e00100 + self.e00700

```
Should be a numba-ized function or something more readable.

MattHJensen:
Resolved with #279. Closing.

PR #279, opened 2015-05-28
Cleaning Up Files

mmessick:
This branch contains small clean-ups using pep 8 style for writing Python.

I also have the change for #271, although I was not sure which decorator to use for the function.
Or if you would like this function to be a decorator itself, I wasn't sure how you wanted it to look.
Right now I just made temporary variables to make it easier to read, but these can easily be condensed
to get the code back to the original version.


talumbau:
I haven't looked at this too much, but I'm wondering if the regular `jit` or `vectorize` decorators from numba would be a better choice? http://numba.pydata.org/numba-doc/0.18.2/index.html

mmessick:
@talumbau, I have been playing around with this for most of the morning, but I just saw your comment now. I think that's a great suggestion and one I was planning to move to next. I will let you know how it goes and if that fixes the problems. 

Earlier I had been trying to use the iterate_jit decorator from the decorators.py, but that is tailored for a Calculator object because it requires the input of both a Parameter object and a Record object. I was trying to do some workarounds on that, but it doesn't seem to quite be working.

After I try using jit or vectorize, if those do not work, I am going to try writing a decorator similar to iterate_jit, but just for a record object. I was working on doing this earlier, however, I was getting a little confused on some of the code's logic

mmessick:
@MattHJensen, will you take a look at my latest push? I was looking into writing another decorator based off of iterate_jit (which I made a copy of in decorators.py so that I could add a bunch of comments to get familiar with its structure), however, after not having the most success understanding how all the pieces fit together, I decided to go with TJ's suggestion to just try using the vectorize decorator. This seems to work as of now. But please let me know if it is not working the way you want it to.

talumbau:
`iterate_jit` is a special hammer to handle the particular case of the kinds of functions found in `functions.py`. @mmessick we could schedule a chat to discuss `iterate_jit` and answer any questions you might have. For this case, I think `vectorize` is probably a good choice. Also, the latest numba should allow you to use `@vectorize` without specifying types. I would give that a go with `conda update numba`.

I would hesitate in creating a new version of `iterate_jit` just to handle the case of one function.

Would also be good to see a test in `test_records.py` that tests this function showing some input, an desired output, and then how the function produces the desired output.

mmessick:
Numba is apparently up to date on my machine, but it still won't allow me to forgo specifying types.

I even used an example from their documentation:
>>> @vectorize
... def g(a, b): return a / b
...
>>> g(2.,3.)
0.66666666666666663

and when I ran it, even this simple function required specifying types for vectorize

taxcalc/records.py
```diff
@@ -7,7 +7,11 @@
 import numpy as np
 import os.path
 import os
+from numba import vectorize, float64
+from .parameters import Parameters
```
talumbau: This import is not needed, correct?

taxcalc/records.py
```diff
@@ -7,7 +7,11 @@
...
 from pkg_resources import resource_stream, Requirement
+from .decorators import *
```
talumbau: Now that you are not using decorators from `decorators.py`, this import is also not needed, correct?

taxcalc/records.py
```diff
@@ -661,3 +656,30 @@ def read(self, data):
...
+
+    def mutate_imputations(self):
+        self._cmbtp_itemizer = imputation(self.e17500, self.e00100,
+            self.e18400, self.e18425, self.e62100, self.e00700,
+            self.e04470, self.e21040, self.e18500, self.e20800)
+
+
+@vectorize([float64(float64, float64, float64, float64, float64, float64,
+            float64, float64, float64, float64)])
+def imputation(e17500, e00100, e18400, e18425, e62100, e00700, e04470,
```
talumbau: this looks good to me!

taxcalc/records.py
mmessick: correct, line 11 is not needed

taxcalc/records.py
mmessick: also correct

taxcalc/tests/test_records.py
```diff
@@ -18,6 +18,46 @@ def test_create_records():
     r = Records(tax_dta_path)
     assert r
 
+
 def test_create_records_from_file():
     r = Records.from_file(tax_dta_path)
     assert r
+
+
+def test_imputation():
```
talumbau: I think a small vector test would be preferred to a scalar, but I don't think that should hold up the commit.

taxcalc/records.py
mmessick: ok great! @MattHJensen was hoping to get rid of specifying what type of function vectorize was using ie her preferred not to have: [float64(float64, float64, float64, float64, float64, float64, float64, float64, float64, float64)]. However, I could not get the code to run without it (despite updating numba)

talumbau:
@MattHJensen, for changes like this, you often run a more comprehensive test against the full data set to see the effect on overall results? Do you judge that this is necessary for this change?

talumbau:
@mmessick since you are cleaning up the PEP-8 stuff, you can fix the print statements in records.py around line 286 and line 309? Sort of non-standard to use line break character `\` unless there is a really good reason.

taxcalc/records.py
talumbau: Yes I believe I misspoke. This (very cool) feature of Numba is coming in the new version, which will drop next week (or maybe this week? I think). So, for now, `@vectorize` does need the type info.

taxcalc/records.py
```diff
@@ -661,3 +656,30 @@ def read(self, data):
...
+
+    """
+    Calculates _cmbtp_itemizer
+    Uses vectorize decorator to speed-up calculation process
+    with NumPy data structures
+    """
+
+    # temp variables to make it easier to read
+    x = max(0., e17500 - max(0., e00100) * 0.075)
+    y = -1 * min(x, 0.025 * max(0., e00100))
```
MattHJensen: let's rename "y", "medical_adjustment" or similar and let's rename "z", "state_adjustment" or similar.

taxcalc/records.py
```diff
@@ -661,3 +656,30 @@ def read(self, data):
...
+    z = max(0, max(e18400, e18425))
+
+    _cmbtp_itemizer = (y + e62100 + e00700 + e04470 + e21040 - z - e00100
```
MattHJensen: Also, it probably makes more sense to get rid of the -1 coefficient on y and then subtract y right before we subtract z rather than add y (with the -1 coef) at the beginning of this expression.

MattHJensen:
> @MattHJensen, for changes like this, you often run a more comprehensive test against the full data set to see the effect on overall results? Do you judge that this is necessary for this change?

Yes, @mmessick, could you paste into the comments the results from calc.diagnostic_table() from master branch? See the last comment from https://github.com/OpenSourcePolicyCenter/Tax-Calculator/pull/262 as an example. (I don't think your changes should result in any difference between master and branch diagnostic tables)

mmessick:
Here is the diagnostic table for my branch (I wasn't sure how to get it in the same format):
image

This is for the master branch:
image

mmessick:
Should be ready to be merged

talumbau:
This looks good to me, and as long as the results above are ok with Matt, I am +1 on merging. On a style note, I think the best way to do PEP-8 changes is a PR that's only for those changes. This PR does more than just "clean up files", so I think it helps for purposes of review to separate out 'real' changes from formatting changes.

MattHJensen:
Merging!

PR #285, opened 2015-06-15
Edit tax-calculator docs/source/*.rst files to eliminate all Sphinx warnings.

martinholmer:
This first round of Tax Calculator documentation edits contains
changes to eliminate Sphinx warnings and to standardize page
organization.  It also makes some (relatively minor) changes to the
documentation text.  The next round of documentation edits needs to
focus on the intended audience for the documentation: people who just
want to use the Tax Calculator versus people who want to contribute to
the development of the Tax Calculator.  The next round should also
consider making the guide a single .rst file because now it is awkward
to direct users back to a certain place in another .rst file (as
within page anchors would allow in HTML).


MattHJensen:
Thanks Martin, this is great! Flagging this for @rkuchan to take a look.

taxcalc/decorators.py
```diff
@@ -283,7 +285,7 @@ def make_wrapper(func):
         try:
             jit_args = inspect.getargspec(jit).args + ['nopython']
         except TypeError:
-            print ("This should only be seen in RTD, if not install numba!")
+            #print ("This should only be seen in RTD, if not install numba!")
```
MattHJensen: @talumbau, this was printing repeatedly when building html (presumably when sphinx looked inside of each function in functions.py), which is a bit disconcerting/confusing. If this is ever seen outside of RTD, I presume Numba's nonexistence would have been noticed already. If we are worried about confusing error messages when packages are missing, should we check for the existence of dependencies explicitly?

talumbau: I would definitely advocate for something other than just commenting out the print statement. One way to handle this is to attempt the import of `jit` at the top and if `numba` is not found, produce a dummy function. Something like this:

```
# Numba is not available when building on ReadTheDocs
try:
    from numba import jit
except ImportError:
    jit = lambda x: return x
```
then this line inside the function would simply be:

```
    jit_args = inspect.getargspec(jit).args + ['nopython']
```
because `jit` is defined in either case. How does that sound?

martinholmer: On Mon, Jun 22, 2015 at 9:42 AM, T.J. Alumbaugh <notifications@github.com>
wrote:

> In taxcalc/decorators.py
> <https://github.com/OpenSourcePolicyCenter/Tax-Calculator/pull/285#discussion_r32934642>
> :
>
> > @@ -283,7 +285,7 @@ def make_wrapper(func):
> >          try:
> >              jit_args = inspect.getargspec(jit).args + ['nopython']
> >          except TypeError:
> > -            print ("This should only be seen in RTD, if not install numba!")
> > +            #print ("This should only be seen in RTD, if not install numba!")
>
> I would definitely advocate for something other than just commenting out
> the print statement. One way to handle this is to attempt the import of
> jit at the top and if numba is not found, produce a dummy function.
> Something like this:
>
> # Numba is not available when building on ReadTheDocs
> try:
>     from numba import jit
> except ImportError:
>     jit = lambda x: return x
>
> then this line inside the function would simply be:
>
>     jit_args = inspect.getargspec(jit).args + ['nopython']
>
> because jit is defined in either case. How does that sound?
>

Does the above suggested code revision mean that a user does not need numba
as part of the Python distribution to execute the Tax Calculator?  Because
with
your suggested code revision jit is just a dummy function that does nothing
when
numba in not installed?  Right?  Or am I missing something?

 If that is your intention, then I think it is an improvement in the design
of the Tax Calculator.  I have always thought that the way things should
work is that if numba is installed tax calculations will execute fast, but
if
not installed the calculator should run and get the same answers but
take longer to execute.  Is that what you are suggesting?  If so, then
I like your suggested revision.  If not, see my (actually Matt's) question
below.

What Matt asked you was something different: is there (or should there be)
at the beginning of the Tax Calculator logic code that checks for numba
installation and stops Tax Calculator execution with an error message when
numba is not part of the installed Python distribution.  I'm interested in
your
answer to that question.

talumbau:
The answer to the above is:

No I do not think there should be logic in the Tax-Calculator that checks for the existence of numba and stops execution with an error message when numba is not part of the installed Python distribution. The reasons for this are:

 - Python will already do this (print an error message if `numba` is not found). The error message will be an `ImportError` exception
 - it's un-pythonic. The Python style is typically known as ["Easier to Ask Forgiveness Than Permission"](https://docs.python.org/2/glossary.html#term-eafp). The suggested change (add a more explicit check for numba) is more of a ["Look Before You Leap"](https://docs.python.org/2/glossary.html#term-lbyl)

Yes, you are correct that the suggested change makes `jit` a dummy function. My assumption here was that we are trying to solve the problem "RTD needs to build the docs, but to do so it ends up calling code that needs numba, and numba isn't easily callable in the RTD build system."

We could certainly add a feature to make the Tax-Calculator run if `numba` isn't installed. I don't see any reason why such a thing is impossible. In fact, what I have suggested above might actually be enough to make it work. Basically we would just add a different test environment to TravisCI. @MattHJensen do you want me to create an issue? Is the intended use case something beyond the issue with RTD?

MattHJensen:
@talumbau, your response helps me to understand that the primary issue with the RTD build system. Being able to run taxcalc without `numba` would be useful, although it doesn't seem urgent -- I anticipate that it could be generally helpful with debugging.

talumbau:
Is this good to go? I think the answer is 'yes'

MattHJensen:
@Tj, this is good to go assuming that you are ok leaving the commented out line (discussed above) while #292 is pending.

talumbau:
A merge is necessary now since I updated the contributor guide in https://github.com/OpenSourcePolicyCenter/Tax-Calculator/pull/297

martinholmer:
I am resolving merge comments and making a few other changes.
I will post a message here when this PR is ready to be merged.

martinholmer:
Sorry about the jet-lag, I meant to say that I am resolving the merge **conflicts**.

martinholmer:
As of about 7:10PM EDT on Friday, this pull request has passed all Travis checks and has no merge conflicts.

MattHJensen:
Excellent. Thanks Martin. Merging now.

PR #288, opened 2015-06-19
Revise tax-calculator/README.md file along the lines suggested by Mat…

martinholmer:
…t Jensen.


martinholmer:
Revise README.md so that no lines longer than 80 characters,
so that text more closely corresponds to docs/source/*.rst content,
and so that the confusing Installation instructions are replaced by
two ways of getting started with the Tax Calculator.

PR #293, opened 2015-06-23
Delete example.py

MattHJensen:
This appears to have been committed accidentally and is not a working example. It was originally created by T.J., I believe as an example of his vision for the taxcalc package. 

@talumbau, if I've got the history of thought right, could you merge? 

This isn't to say that we shouldn't have an example.py -- perhaps one that mirrors our 10_min notebook --  as part of the docs sometime in the future, but I think it would be different enough from this one to warrant starting over.


talumbau:
yep. out with the old, in with the new!

Issue #296, opened 2015-06-25
conda package and source install conflict


MattHJensen:
@talumbau, this is mostly a question for you. 

Conda installing the taxcalc package locally causes errors for a contributor who wants to interact with the taxcalc source code. Since many of our contributors will be new to conda, this might trip them up if they ever 'accidentally' execute `conda install taxcalc`, even just to play around. This raises a couple of questions:

1. Is there a way to keep the two from conflicting when they are in the same environment? 

2. If the answer to the first question is  "no," then it seems we need to decide whether it should be users or contributors who need to create and activate a new conda environment. 

Am I thinking about this the right way? 

cc @martinholmer

talumbau:
Ah, maybe I should start sort of one step 'higher'. I thought there would be a class of users where they could do productive things with just the following: a JSON file of the parameters that they wanted to tweak, a PUF (or equivalent) dataset, and some basic Python skills. For this use case, a conda package is a great way to go. You don't have to mess with the "guts", you can just make changes to your "Plan Y",  run the calculator as shown in the notebooks, and then you are good to go.

If you are interacting with the source, this is more what I would consider 'development'. In that case, you don't want the conda package installed because you are probably in some kind of iteration of "make changes to source, see how they impact 'the answer', make more changes to source, etc." In this case, I typically set up some kind of 'run script' with a terminal or IPython notebook at the top of the repo. Then, when I make changes, I will do something like `python check.py` from the top of the repo. In that script, the `import taxcalc` will import the package found in the `taxcalc/` directory which has my latest changes.

So, the overall answer is that development work should be done in an environment where the package is not installed. If one wants to uninstall a package, you can just do `conda remove taxcalc`.

The developer case would still want a conda environment because, presumably, you would want to run the tests (and likely add tests) and also have the proper version of pandas, numpy, numba, etc. That's actually my fault for not creating an `environment.yaml` file. I'll be adding one soon. That way collaborators who want to develop the package could do something like:

```
git clone git@github.com:OpenSourcePolicyCenter/Tax-Calculator
cd Tax-Calculator
conda env create
py.test
```

The `conda env create` would install everything needed to make the package run, but it wouldn't install the package itself.

MattHJensen:
[...]

PR #297, opened 2015-06-26
Add environment.yml

talumbau:
- This file permits one to execute the "conda env" command to create an
  environment that is appropriate for development and testing of the
  taxcalc package

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,0 +1,12 @@
+name: tax
+dependencies:
+- pytest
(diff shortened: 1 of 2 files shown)


environment.yml
```diff
@@ -0,0 +1,12 @@
+name: tax
```
MattHJensen: Just added a comment pertaining to this name at #296.

PR #301, opened 2015-07-09
New Weights based on updated factors

Amy-Xu:
This PR has the new weights created using the updated Stage I factors. In this version, we remain the assumption of constant wage distribution. Revenue projection is generally lower than the current master version. Here's a comparison.

Current Master:
screen shot 2015-07-09 at 2 29 43 pm

This PR:
screen shot 2015-07-09 at 2 30 46 pm


MattHJensen:
Looks good to me.

MattHJensen:
Merging. Thanks @Amy-Xu!

PR #306, opened 2015-07-12
Pandas ascii output

xiyuw123:
Ascii output file renamed, edited to use Pandas. I also added a test for it in test_utils.py


MattHJensen:
@xiyuw123, when you open a PR, Travis CI runs the test suite in both a python 3 and a python 2 environment. If you click on 'details' to the right of "some checks have failed" it will actually show you the log for setting up the environment and running the tests. In this case it turns out that you need to use the print function print("hello") to be compatible. 

As an aside, does this replace and can we close #273?

xiyuw123:
Ok, thanks for letting me know that!

And yes, this replaces #273. I can close that.

talumbau:
this looks good to me. Compact and likely will be extended to other uses case. :+1:

MattHJensen:
@Amy-Xu, could you review this one and merge when you think it is ready?

Amy-Xu:
Looks good to me. Thanks @xiyuw123  @talumbau. Merging.

Issue #305, opened 2015-07-11
RTD build is failing 


MattHJensen:
Here is the trace. I have not had a chance to do much investigation yet. 

cc @talumbau, @jlyons871, @martinholmer 

Build for taxcalc

Built: July 11, 2015. 10:19 a.m.

State: Finished

Outcome: Failed (Status Code: 1)

Version: latest

Type: html

Commit: 7be74b02e1e4d59d5673195f30db3dd5abea7c99

Length: 22 seconds

Build Standard Output

html
-----

Running Sphinx v1.3.1
making output directory...
Build Standard Error

html
-----

Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/taxcalc/envs/latest/local/lib/python2.7/site-packages/sphinx/cmdline.py", line 244, in main
    opts.warningiserror, opts.tags, opts.verbosity, opts.jobs)
  File "/home/docs/checkouts/readthedocs.org/user_builds/taxcalc/envs/latest/local/lib/python2.7/site-packages/sphinx/application.py", line 143, in __init__
    self.setup_extension(extension)
  File "/home/docs/checkouts/readthedocs.org/user_builds/taxcalc/envs/latest/local/lib/python2.7/site-packages/sphinx/application.py", line 433, in setup_extension
    err)
ExtensionError: Could not import extension sphinxcontrib.napoleon (exception: No module named sphinxcontrib.napoleon)

Extension error:
Could not import extension sphinxcontrib.napoleon (exception: No module named sphinxcontrib.napoleon)
Setup Output

checkout
-----

venv
-----

Using real prefix '/usr'
New python executable in /home/docs/checkouts/readthedocs.org/user_builds/taxcalc/envs/latest/bin/python
Installing setuptools, pip...done.

doc_builder
-----

Downloading/unpacking sphinx==1.3.1
Downloading/unpacking Pygments==2.0.2
Downloading/unpacking virtualenv==1.10.1
Downloading/unpacking setuptools==1.1
Downloading/unpacking docutils==0.11
Downloading/unpacking mkdocs==0.14.0
Downloading/unpacking mock==1.0.1
Downloading/unpacking pillow==2.6.1
Downloading/unpacking readthedocs-sphinx-ext==0.5.4
  Downloading readthedocs_sphinx_ext-0.5.4-py2.py3-none-any.whl
Downloading/unpacking sphinx-rtd-theme==0.1.8
[...]

PR #310, opened 2015-07-13
latest version of sphinxcontrib-napoleon on rtd

MattHJensen:
@talumbau suggested "Seems worth trying the latest version and seeing if that fixes the issue." to see if that solves #305. I believe this is the way to implement that. After this is merged, I will trigger another build on RTD.

### DIFF
--- a/docs/requirements-docs.txt
+++ b/docs/requirements-docs.txt
@@ -10,4 +10,4 @@ toolz
 
 # for ReadTheDocs
 sphinx
-sphinxcontrib-napoleon==0.2.11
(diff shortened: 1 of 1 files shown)


jlyons871:
@MattHJensen adding the requirement of version `0.2.11` to the `sphinxcontrib-napoleon` was the fix to the previous rtd build issue https://github.com/rtfd/readthedocs.org/issues/1164. 

By updating the Sphinx version we can bypass this, and is the best way moving forward. Please see #311.

MattHJensen:
Cool. Thanks! Closing this.

PR #311, opened 2015-07-14
updated min sphinx version pip-requirement and pkg import

jlyons871:
As of Sphinx version 1.3, the napoleon package is included by default in its extensions. (no need to pip install sphinxcontrib-napoleon anymore for local development)

RTD was pulling the latest sphinx, which has a slightly different import statement. Setting the requirement to use at least Sphinx 1.3 will keep this fixed moving forward.

### DIFF
--- a/docs/requirements-docs.txt
+++ b/docs/requirements-docs.txt
@@ -9,5 +9,4 @@ six
 toolz
 
 # for ReadTheDocs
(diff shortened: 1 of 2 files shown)


MattHJensen:
Merging. Thanks @jlyons871

PR #312, opened 2015-07-15
Add four new tests to taxcalc/tests/test_parameters.py

martinholmer:
The four additions test non-default (constant and variable) inflation rates
with and without a policy reform.  There are failures in two of the new tests.
The following patch to the update method of the Parameters class fixes all
the failures and all the old parameter tests continue to pass:

```
--- a/taxcalc/parameters.py
+++ b/taxcalc/parameters.py
@@ -110,7 +110,8 @@ class Parameters(object):
                     # advance until the parameter is in line with the current
                     # year
                     num_years_to_skip=self.current_year - year
-                    inf_rates = [self.__rates[year + i]
+                    offset_year = year - self.start_year
+                    inf_rates = [self._inflation_rates[offset_year + i]
                                  for i in range(0, num_years_to_expand)]
 
                     nval = expand_array(values,
```

After this pull request is accepted, I will prepare another pull request
that includes the above patch unless the discussion here reveals 
problems with the proposed patch.

### DIFF
--- a/taxcalc/parameters.py
+++ b/taxcalc/parameters.py
@@ -110,7 +110,8 @@ class Parameters(object):
                     # advance until the parameter is in line with the current
                     # year
                     num_years_to_skip=self.current_year - year
-                    inf_rates = [self.__rates[year + i]
+                    offset_year = year - self.start_year
+                    inf_rates = [self._inflation_rates[offset_year + i]
                                  for i in range(0, num_years_to_expand)]
 
                     nval = expand_array(values,
--- a/taxcalc/tests/test_parameters.py
+++ b/taxcalc/tests/test_parameters.py
@@ -41,6 +41,66 @@ def test_create_parameters():
     assert p
 
 
+def test_constant_inflation_rate_without_reform():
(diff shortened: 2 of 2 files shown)


talumbau:
@martinholmer nice tests! The typical workflow is to add new code + new tests at the same time in a pull request. We can't merge the new tests first because then we would be checking in failed tests, and Travis CI can't differentiate between this case (new tests for not-yet-added features) from the case of 'added new code and broke a test'. So we typically put new code plus the tests that test the new capabilities of the new code in the same PR.

martinholmer:
OK, I'll add the patch to this pull request, which will eliminate the test failures.

martinholmer:
@talumbau thanks for the help on the OSPC workflow.
The pull request now passes the Travis CI build process.

talumbau:
Ah yes I see. The `update` method on Parameters didn't take into account user-supplied inflation rates. This looks good to me. +1

MattHJensen:
Looks good to me as well. Merging.

PR #314, opened 2015-07-16
New weights from CLP solver

Amy-Xu:
The new weights file in this PR is generated by the CLP solver through its python interface, instead of MINOS (fortran). There are some difference in weights I think mainly because of two reasons. First, CLP and MINOS might implement the Simplex algorithm in slightly different ways so we are getting slightly different results from the LP process. Second, previously there was a copy mistake I made in excel, which led to weird results of 2018. This mistake is corrected by automating Stage II in python.

Wage distribution assumption remains the same.


Amy-Xu:
Here are the diagnostic tables.

Current master:



This PR:
screen shot 2015-07-16 at 1 26 29 pm

@MattHJensen Could you review? Thanks!

MattHJensen:
This looks good to me! Merging.

PR #320, opened 2015-07-19
Correct Calculator class constructor call in test.py

martinholmer:
Also, make other non-substantive changes to eliminate pylint warnings.
All these changes leave the results_puf.csv output file produced by the
test.py program unchanged.


talumbau:
LGTM! It's not clear to me what the long term disposition of this file should be, but it certainly wasn't good in its current state. Open to ideas on that...

Issue #325, opened 2015-07-27
BUG: expand_1D behaves differently from expand_2D when providing inflation rates


talumbau:
PR #321  demonstrates a difference in behavior between the two version of `expand_array`. In `expand_1D`, one provides a list of interest rates equal to the desired number of resulting values. Thus, the current behavior is master is as follows:

```
    x = np.array([1, 2], dtype='f4')
    irates = [0.0, 0.1, 0.25]
    exp = np.array([1, 2, 2.5])
    res = expand_1D(x, inflate=True, inflation_rates=irates, num_years=3)
    # res == exp is True
```

The idea is that we desire three scalar values but we only have two. `expand_1D` is implemented such that to get the nth value in the new array, we apply the inflation rate at `irates[n]` to the `n-1` value. This is odd because we never make use of the inflation rate at index 0 (i.e. `irates[0]`).

`expand_2D` takes a different approach: the nth value in the expanded array is based on applying the inflation rate at n-1 to the value at n-1. Thus, to produce an output with k values, we only need k-1 interest rates. 

I think this is more sensical, so I propose that we fix this difference in behavior by modifying expand_1D to work as follows:

```
   x = np.array([1, 2], dtype='f4')
    irates = [0.1, 0.25]
    exp = np.array([1, 2, 2.5])
    res = expand_1D(x, inflate=True, inflation_rates=irates, num_years=3)
    # res == exp is True
```
cc @martinholmer @Amy-Xu

martinholmer:
Pull request #321 resolves the difference in how inflation rates are applied between the
expand_1D() and expand_2D() utility functions in favor of how the expand_1D() function
works.  That decision was based on the acceptance by @MattHJensen after review by
@talumbau of pull request #312, which included the following text in test_parameters.py:
```
# Note: above assert shows that inflation indexing of policy parameters
# works like this:  param(t) = param(t-1) * ( 1 + inflation_rate(t) ).
```
That is, it works like in the expand_1D() function.  However, the current version of the 
expand_2D() function works this way:
```
param(t) = param(t-1) * ( 1 + inflation_rate(t-1) ).
```
So, why does it now seem "more sensical" to apply the indexing in a different way than
was accepted in pull request #312?

Surely the choice should be determined by how CBO applies inflation rates because
the default inflation rates are taken from CBO.  @MattHJensen, what did you find out
about CBO's method after we discussed this issue last month?

talumbau:
[...]

PR #326, opened 2015-07-28
fix expand_1D to match convention of expand_2D

talumbau:
- producing new array of n values requires only n-1 inflation rates

### DIFF
--- a/taxcalc/tests/test_parameters.py
+++ b/taxcalc/tests/test_parameters.py
@@ -82,9 +82,7 @@ def test_variable_inflation_rate_without_reform():
     assert round(grate, 3) == round(0.04, 3)
     # check implied inflation rate between 2021 and 2022
     grate = float(p._II_em[2022 - syr]) / float(p._II_em[2021 - syr]) - 1.0
-    assert round(grate, 3) == round(0.08, 3)
-    # Note: above assert shows that inflation indexing of policy parameters
-    # works like this:  param(t) = param(t-1) * ( 1 + inflation_rate(t) ).
+    assert round(grate, 3) == round(0.04, 3)
 
 
 def test_variable_inflation_rate_with_reform():
@@ -107,7 +105,7 @@ def test_variable_inflation_rate_with_reform():
     assert round(grate, 3) == round(0.04, 3)
     # check implied inflation rate between 2021 and 2022 (after then reform)
     grate = float(p._II_em[2022 - syr]) / float(p._II_em[2021 - syr]) - 1.0
-    assert round(grate, 3) == round(0.08, 3)
+    assert round(grate, 3) == round(0.04, 3)
 
 
 def test_create_parameters_from_file(paramsfile):
--- a/taxcalc/tests/test_utils.py
+++ b/taxcalc/tests/test_utils.py
@@ -46,18 +46,10 @@ def test_expand_1D_short_array():
 
 def test_expand_1D_variable_rates():
     x = np.array([4, 5, 9], dtype='f4')
-    irates = [0.02, 0.02, 0.02, 0.03, 0.035]
+    irates = [0.02, 0.02, 0.03, 0.035]
     exp2 = []
     cur = 9.0
-    for i in range(1, 3):
-        idx = i + len(x) - 1
-        cur *= (1.0 + irates[idx])
-        exp2.append(cur)
-
-    exp1 = np.array([4, 5, 9])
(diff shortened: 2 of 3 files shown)


martinholmer:
**Don't you dare commit this!**  This will break substantial parts of pull request #321.
A better approach is to commit pull request #321 and then figure out how CBO
applies inflation rates, which is something @MattHJensen is going to do.  Once,
we know what CBO does with its inflation rates we will be able to make a sensible
decision about how to do this in the OSPC Tax-Calculator.

MattHJensen:
@martinholmer, it seems like T.J. gave you a scare! But don’t worry about anything ever getting merged before a relevant issue (in this case #325) has been fully discussed — we have done a very good job as a community at upholding this principle. (And I should probably add something on this to our contributor docs). That said, it is certainly possible to make mistakes or for someone not to notice that an issue is relevant, so it is very helpful for you to note that this PR depends on the discussion in #325, especially since it wasn’t marked as such already. 

I’ll follow up over at #325 on how CBO reports inflation numbers shortly

talumbau:
I opened #326 as part of a discussion of #325 . If we decide the fix should be different after Matt’s followup, then I would still advocate for having a separate PR that only implements the fix. My intent was never to merge my fix with no discussion.

The reason for having a fix-specific PR, more generally, is that it fits into a workflow for handling a bug that tends to be very successful on github:

- Create an issue that describes the bug as succinctly as possible, preferably with some small code example
- If possible, create a PR that resolves that bug (and only that bug) in as few lines as is reasonably possible. Then invite feedback on your solution.

There are many reasons why this is a good way to resolve a bug. Some of them are:

- It allows others to minimize the time they need to spend to come up to speed on the particular issue. If they can read a few lines of code/text that describe the problem, and then a few lines of code that fix the problem, it’s much more efficient than reading a large PR that has other items inside it.
- It invites the maximum possible scrutiny to the community to discuss the issue. The issue is labeled as a bug, and interested parties can now comment in one place on this particular issue. The discussion can range from possible solutions, to addressing whether or not the behavior is actually a bug (not relevant in this case, but often that does come up!)
- It provides a clean history of the issue. If a PR is labeled as ‘Fix for #325’ or something similar, it’s easy to look through the history of the repository and see how the bug got fixed.

#321 on the other hand is a very large PR and needs additional review. There are many diffs of the form
```
-           with open(file_name) as f:
-               params = json.loads(f.read())
+          with open(file_name) as pfile:
+              params = json.loads(pfile.read())
```
which one has to look through, which are more formatting/style changes rather than substantive changes. Typically, one would have a separate style/formatting PR. For the parts that I have looked at, I haven’t had a chance to put my comments/questions in. There are likely good reasons for the choices made, but I haven’t had time to consider them yet.

martinholmer:
If the project team decides to change it prior inflation-indexing scheme, then it would be less
work to commit pull request #321 before making the change to a new inflation-indexing scheme.
Remember pull request #321 found another bug as well, which has not been discussed in its
own issue as has the inflation-indexing bug.

martinholmer:
I suggest that if everybody is OK with pull request #326, then we should proceed
as follows: (1) commit #326 to master, (2) I will fix the test results in pull request #321
to agree with the new method of applying inflation rates, and (3) commit #321 to master.

Does this seem like a reasonable approach? If not, what would be a better approach?

talumbau:
I think that is the right flow, with one or two small additional details. I will post a PR with an additional test that reveals a particular issue with how `dropq` uses `taxcalc`. If I am correct, once you merge in that change with the `parameters-edit2` branch, you will see a test failure. In my testing, this actually reveals a bug in a different place and not in the code in `parameters-edit2` branch. Anyway, details to come. I think a good first step though is the merge of this PR, and then merging master in to your branch.

martinholmer:
TJ @talumbau, Just go ahead and commit pull request #326 and then
I'll fix up pull request #321.  Isn't that the easiest way to proceed?
@MattHJensen, do you see any problems with that workflow?

MattHJensen:
Sounds like we're all in agreement on merging #326 -- It might make sense to wait on fixing up #321 until TJ has had a chance to open the test PR in case that will affect how you should fix #321, but we'll know better whether that is the case once T.J. has a chance to post the forthcoming details on that. @talumbau, just because I'm am only 98% certain that your forthcoming test PR doesn't have anything to do with merging #326, I'll let you go ahead and make this merge.

Issue #328, opened 2015-07-30
Enforce PEP8 Compliance 


MattHJensen:
- [x] Open and then merge a PR that brings current code base to PEP8 compliance. This is forthcoming from me.
- [x] Enforce PEP8 compliance in the test suite. 

This is a followup from the discussion in #30.

martinholmer:
@MattHJensen said on 30-Jul-2015:
> Open and then merge a PR that brings current code base to PEP8 compliance. 
> This is forthcoming from me.
>
>  Enforce PEP8 compliance in the test suite.

Where do things stand on this effort?

MattHJensen:
Latest discussion is in #329. In particular, see the second to last comment.

martinholmer:
So, @MattHJensen, the discussion at the end of pull request #329 is pretty
complicated.  What is the situation?  Are TJ and Jake saying the Travis CI
tool cannot accurately check PEP8-compliant code (and thus, Travis CI is
returning a false negative)?  Or, are they saying they don't know how to 
make the Tax-Calculator code fully PEP8 compliant?

MattHJensen:
The only method that I have discovered to resolve the last PEP8 error (E402) and thereby make the code fully PEP8 compliant is incompatible with our current method of testing with Travis CI. @talumbau has asked @jlyons871 to look into whether we can modify our current method of testing with Travis CI to make it compatible with my E402 fix or whether some other solution to E402 is available. 

Alternatively, I could roll back my fix for E402, we could merge every other PEP8 fix, and then we could start enforcing PEP8 excluding E402 in the test suite. This would buy us time to look for an E402 fix on a more leisurely basis.

martinholmer:
@MattHJensen said:
> Alternatively, I could roll back my fix for E402, we could merge every other PEP8 fix,
> and then we could start enforcing PEP8 excluding E402 in the test suite.
> This would buy us time to look for an E402 fix on a more leisurely basis.

This sounds like an excellent plan to me.

MattHJensen:
Agreed. I will do this tomorrow.

MattHJensen:
Ok, opened #338 in accordance with this plan: 

> Alternatively, I could roll back my fix for E402, we could merge every other PEP8 fix,
> and then we could start enforcing PEP8 excluding E402 in the test suite.
> This would buy us time to look for an E402 fix on a more leisurely basis.

MattHJensen:
Resolved with #338. Closing.

Issue #30, opened 2014-09-18
coding style


MattHJensen:
At some point we need to go through our code and make it conform to a set of style standards. PEP 8 seems like a good starting point. Is it possible to make pylint run automatically as part of our commit suite (along with Travis-CI?)

talumbau:
We have not strictly enforced PEP-8, but have eliminated tabs in the code. If there is a desire to enforce PEP-8, we can run `flake8` on the source. Closing.

MattHJensen:
Based on discussion with @martinholmer, I believe it would be useful to begin enforcing some style rules more systematically than we do now. From early discussion, the three rules that have been requested are

- max of 80 characters
- 4 space indents (no tabs)
- no lines with blank spaces at the end

I'm hoping to discuss:

1. Is this the right set of rules. 
2. How should we enforce rules.

zrisher:
In my experience, those are standard to every major open source project on github (besides indent, where 2 spaces is the norm in other languages, but of course in Python 4 spaces is expected).

While automated linting tools can be a pain in the bum sometimes, I believe most projects with large contributor bases have found them to be a significant boon to productivity. While some things will likely be more elegant if the rules are broken slightly, its hard to imagine style enforcement really stopping a useful PR.

martinholmer:
Based on my experience developing the OSPC ACA-refundable-tax-credit calculator in Python,
I think these three style rules are way too limited.  As you may know, Python is the official scripting 
language at  Google, and they have a small team who has worked for years developing the
[Google Python Style Guide](https://google-styleguide.googlecode.com/svn/trunk/pyguide.html).
I'm not advocating that this project try to follow that Style Guide, but it is important to note that
the very first item in the Style Guide is an admonition to use ```pylint``` to check your code.

Here is what the Google Python Style Guide says about this:
```
Lint

Run pylint over your code.

Definition:
pylint is a tool for finding bugs and style problems in Python source code.  
It finds problems that are typically caught by a compiler for less dynamic 
languages like C and C++.  Because of the dynamic nature of Python, 
some warnings may be incorrect; however, spurious warnings should 
be fairly infrequent.

Pros:
Catches easy-to-miss errors like typos, using-vars-before-assignment, etc.

Cons:
pylint isn't perfect.  To take advantage of it, we'll need to sometimes: 
a) Write around it,  b) Suppress its warnings, or c) Improve it.

Decision:
Make sure you run pylint on your code.  Suppress warnings if they are
[...]

PR #329, opened 2015-07-30
[WIP] PEP8 compliance

MattHJensen:
Dealing with #328.


MattHJensen:
@martinholmer, note that some of the changes in this PR may cause conflicts with the PEP8 resolutions you apply in #321. I included the parameter-related scripts here so that we have the flexibility to add a pep8 compliance checker to the test suite before #321 gets merged in. If it looks like #321 is going to get merged beforehand, then I can remove the conflicting commits (files) from this PR.

MattHJensen:
Two outstanding issues. I will do some digging into these tomorrow morning if no one has a ready answer. 

1. We have a common pattern in the tests: 

```
    import stuff
    cur_path = os.path.abspath(os.path.dirname(__file__))
    sys.path.append(os.path.join(cur_path, "../../"))
    sys.path.append(os.path.join(cur_path, "../"))
    import more stuff
    from taxcalc import *
```

PEP8 is does not like having the path-related, non-import lines before imports, but we need these to find taxcalc. 

2. In `taxcalc/tests/test_utils.py`  we have a function `asciifile()` with a block of text that is 81 characters long (two chars too long for PEP8). I could break this over the lines, but that would be really ugly. 

```
    txt = ("A              \t1              \t100            \t123.45         \n"
           "B              \t2              \t200            \t678.912        \n"
           "C              \t3              \t300            \t000.000        \n"
           "D              \t4              \t400            \t87             \n"
           "EFGH           \t0              \t500            \t92             "
           )
```

cc @talumbau @martinholmer

martinholmer:
I haven't tested this, but won't something like this work for the long lines?
(Note, if you don't like X, then pick any other character not used elsewhere in the string.
```
EXISTING CODE IN TEST_UTILS:

    txt = ("A              \t1              \t100            \t123.45         \n"
           "B              \t2              \t200            \t678.912        \n"
           "C              \t3              \t300            \t000.000        \n"
           "D              \t4              \t400            \t87             \n"
           "EFGH           \t0              \t500            \t92             "
        )



POSSIBLE REPLACEMENT CODE:

    txt = ("A              \t1              \t100            \t123.45  X"
           "B              \t2              \t200            \t678.912 X"
           "C              \t3              \t300            \t000.000 X"
           "D              \t4              \t400            \t87      X"
           "EFGH           \t0              \t500            \t92      X"
        ).replace("X", "       \n")
```

martinholmer:
@MattHJensen said:
> PEP8 does not like having the path-related, non-import lines before imports,
> but we need these to find taxcalc.

My (perhaps incorrect) understanding is that the Python library is written in
Python code that is PEP8 compliant.  If that is so, why not look there to see
how they have solved this problem.  I'm sure they face similar issues.

talumbau:
regarding the existing code in test_utils.py, the white space should not be significant in the string `txt`. I would do something like this:

```
    txt = ("A           \t1           \t100         \t123.45         \n"
           "B           \t2           \t200         \t678.912        \n"
           "C           \t3           \t300         \t000.000        \n"
           "D           \t4           \t400         \t87             \n"
           "EFGH        \t0           \t500         \t92             "
        )
```

martinholmer:
TJ @talumbau,
From a quick glance at the documentation for the filecmp.cmp() function,
it is not apparent that your statement about white-space being irrelevant
is true.   Here is what the Python 2.7 documentation says:
```
10.5. filecmp — File and Directory Comparisons
Source code: Lib/filecmp.py

The filecmp module defines functions to compare files and directories, with various optional time/correctness trade-offs. For comparing files, see also the difflib module.

The filecmp module defines the following functions:

filecmp.cmp(f1, f2[, shallow])
Compare the files named f1 and f2, returning True if they seem equal, False otherwise.

Unless shallow is given and is false, files with identical os.stat() signatures are taken to be equal.

Files that were compared using this function will not be compared again unless their os.stat() signature changes.

Note that no external programs are called from this function, giving it portability and efficiency.
```

Am I missing something?

MattHJensen:
Thank you both very much. I had trouble with the ascii methods, so I just wrapped the text. I will make a note for myself to come back to this later, when it's not holding other things up, and see if I can do something more elegant. 

Moving on to imports now. Thanks for the suggestion to look at the python library @martinholmer.

talumbau:
No you are not missing something. It was my mistake. I glanced at the test and actually assumed that this was the input for this test that someone added. The `asciifile` test fixture has to be exactly like that, as you noted. hmm.. I would be inclined to then do each file line as two separate lines. or call the variable `x` instead of `txt`.

MattHJensen:
Ahha, thanks T.J. Going with `x` instead of `txt`.

martinholmer:
@MattHJensen said:
> Ahha, thanks T.J.  Going with x instead of txt.

That will eliminate the PEP8 warning, but at the expense of adding another pylint warning.
Both pep8 and pylint are happy with the following "txt =" statement:
```
def asciifile():
    txt = (
        "A              \t1              \t100            \t123.45         \n"
        "B              \t2              \t200            \t678.912        \n"
        "C              \t3              \t300            \t000.000        \n"
        "D              \t4              \t400            \t87             \n"
        "EFGH           \t0              \t500            \t92             "
    )
```

MattHJensen:
Thanks Martin. I'm going to leave that out for now so that when we move on to the discussion of beyond-pep8 style checking (as prescribed by #30), it can serve as one of many nice examples of the additional suggestions we would get from pylint.

martinholmer:
@MattHJensen said:
> We have a common pattern in the tests:
> ```
>     import stuff
>     cur_path = os.path.abspath(os.path.dirname(__file__))
>     sys.path.append(os.path.join(cur_path, "../../"))
>     sys.path.append(os.path.join(cur_path, "../"))
>     import more stuff
>     from taxcalc import *
> ```
> PEP8 does not like having the path-related, non-import lines before imports, but we need these to find taxcalc.

In the taxcalc/tests directory, for example, the following use of relative imports works for me:
```
from .. taxcalc import Parameters
```
Have you tried that approach?

MattHJensen:
@martinholmer I am actually trying to get the relative imports to work right now. Here is the bug I am running into when I apply them in test_calculate.py:

````
tests/test_calculate.py:9: in <module>
    from .. taxcalc import *
E   ValueError: Attempted relative import in non-package
```

martinholmer:
I guess I know even less than the meager amount that I thought I knew
about Python importing.  It is all very easy in C, but I find the Python
documentation very confusing.  The documenation on these matters
seems to be [6.4 Packages](https://docs.python.org/2/tutorial/modules.html#packages).
What does ```ValueError: Attempted relative import in non-package``` mean?
What is not a package?  And how does one make it into a package?
Sorry, I can't be of more help on this matter.  But we must be doing something
that is pretty basic wrong for it to be caught by PEP8.

MattHJensen:
Ok. I resolved this by adding an __init__.py file with `from taxcalc import *` to the `tests` directory, which turns `tests` into a package, I believe. Will be interseted to hear @talumbau's feedback on this. Pushing the commits now.

MattHJensen:
Passing locally but failing on Travis CI. Investigating.

MattHJensen:
Ok. Calling in the calvary. @talumbau, do you have any suggestions on this? In order to import taxcalc in the tests without adding the file to the path first (which is not a pep8 compliant solution), I made tests into a package by adding an __init__.py that imports everything from taxcalc. This works fine on my local machine, but travis CI is having a hard time finding the tests now.

MattHJensen:
Other than the (albeit major) problem of this PR not passing the test suite, this is ready for review. It eliminates all pep8 warnings in the tax-calculator file tree. I tried to change *as little as possible* on the way to reaching pep8 compliance.

talumbau:
I actually did not know that messing with sys.path and then importing was a violation. The sys.path manipulation was just a convenience because I like to be in the same directory as the tests and do:

```
vi test_foo..py
<edit, exit>
py.test -x test_foo.py
<rinse and repeat>
```
Without the path manipulation, you have to be at the top of the repo and do:
```
vi taxcalc/tests/test_foo.py
<edit, exit>
py.test -x taxcalc/tests/test_food.py
```
which is really not that big of a deal. It's just a workflow that I was comfortable with. I don't think it should hold up PEP8 compliance.

talumbau:
I haven't thought very much about whether the `tests/` directory should actually be considered a subpackage of `taxcalc`. This is not my usual set up. Usually, `tests/` lives inside the package, but is not a subpackage. I should probably look around and see if there are other precedents here. If `tests` is a subpackage of `taxcalc` then we could do the relative imports, but I don't know yet if that is a standard thing to do.

talumbau:
for the purposes of getting this to pass, you could add this to the `setup.py` at the top of the repo:

```
   'packages': ['taxcalc', 'taxcalc.tests'],
```

instead of the current value for `packages`

MattHJensen:
Thanks @talumbau 
>for the purposes of getting this to pass, you could add this to the setup.py at the top of the repo:
   'packages': ['taxcalc', 'taxcalc.tests'],
instead of the current value for packages

Implemented this with bfa9981. Chasing down add on bugs.

talumbau:
Ah @MattHJensen I just looked at the Travis CI output. This might be a reason to decide not to add `tests/` as a subpackage. The tests are now running through the installed subpackage `tests` as opposed to the directory "\<starting directory on Travis CI machine\>/taxcalc/tests". The 1991 fake 'PUF' for example, lives at the top of the repo, which for travis CI would be "\<starting directory on Travis CI machine\>". Since we are running the tests on the `tests` subpackage, it attempts to find this file (and other files) at this path:
"\<starting directory for installed packages on Travis CI machine\>/site-packages/taxcalc/tests/../../"
does that make sense?

MattHJensen:
Thanks @talumbau, I'll unpackage `tests`, but that brings us back to the original problem.  

There has been fairly extensive discussion of this PEP warning (E304) online, but I still haven't found a satisfactory solution for packages that are not always installed, as in our development case.

A few highlights of the discussions:
https://bugs.python.org/issue23061
https://github.com/PyCQA/pep8/issues/264
http://stackoverflow.com/questions/7505988/importing-from-a-relative-path-in-python

I'm not quite ready to say that we turn off the E304 warning...yet

talumbau:
but it would be OK if you removed the sys.path modifications right? I am fine with that if everyone else is. I'm the culprit for putting those in there.

MattHJensen:
The issue is if I move the sys.path modification then taxcalc isn't found. (I can actually remove it from all but the first test used: test_calculate.py)

MattHJensen:
@talumbau, is there a way around this? 

>Ah @MattHJensen I just looked at the Travis CI output. This might be a reason to decide not to add tests/ as a subpackage. The tests are now running through the installed subpackage tests as opposed to the directory "<starting directory on Travis CI machine>/taxcalc/tests". The 1991 fake 'PUF' for example, lives at the top of the repo, which for travis CI would be "<starting directory on Travis CI machine>". Since we are running the tests on the tests subpackage, it attempts to find this file (and other files) at this path:
"<starting directory for installed packages on Travis CI machine>/site-packages/taxcalc/tests/../../"
does that make sense?

If not, I suggest that we move forward by ignoring PEP8 E402 ("module level import not at top of file").               
Making `tests` into a subpackage and then using relative imports to find `taxcalc` from a test file is still the only solution I have found for:

>We have a common pattern in the tests:
    import stuff
    cur_path = os.path.abspath(os.path.dirname(__file__))
    sys.path.append(os.path.join(cur_path, "../../"))
    sys.path.append(os.path.join(cur_path, "../"))
    import more stuff
    from taxcalc import *
PEP8 does not like having the path-related, non-import lines before imports, but we need these to find taxcalc.

talumbau:
@jlyons871 can you sync with @MattHJensen's branch here and dig in to this failure?

MattHJensen:
Closing in favor of #338.

PR #332, opened 2015-08-03
Reference updated for parameters in params.json

GoFroggyRun:
IRS reference has been added to three parameters: _ID_Charity_crt_Cash, _ID_Charity_crt_Asset, and _ID_Charity_frt.


MattHJensen:
@Amy-Xu, could you review and merge when ready? The one thing I noticed that might require some more thought is that there is a stylistic difference between the additions here and some existing references in the way they cite the schedule. 

Example of addition:
`SCHEDULE A, Form 1040, Line 16, Pub. 526 (Limits on Deductions: 50% Limit Organizations.` 

Example of existing reference: 
`Form 1040, line 44, instruction (Schedule X)`

Amy-Xu:
I'll include this part in my PR. Merging.

PR #340, opened 2015-08-11
[WIP]New Weights and Factors

Amy-Xu:
I moved Stage I from excel to python and updated the data projection on population (CENSUS) and number of returns (IRS). In this PR, the weights and factors files are generated through python scripts, using the updated projection.

This PR:
screen shot 2015-08-11 at 10 09 51 am

Master:
screen shot 2015-08-11 at 10 11 29 am


Amy-Xu:
This issue will be dealt with later on when we refactor taxcalc to accommodate both 08 and 09 PUF. Close for now.

Issue #336, opened 2015-08-05
Proposed redesign of the Parameters class API


martinholmer:
**Background**

The classes in taxcalc have been developed initially to support the
kind of tax policy analysis provided by the web application TaxBrain.
The considerable power of these classes have been packaged in an API
(that is, a set of public class methods) that supports the kind of
ten-budget-year revenue estimation and distributional analysis
produced by the TaxBrain app.  This is a considerable accomplishment
that has attracted interest in using the taxcalc capabilities in other
ways. One of the new ways of using the taxcalc capabilities is
described in issue #291, which calls for the development of a
TAXSIM-like capability.  See issue #319 for a broader discussion of
other new ways of using the taxcalc classes.

The development of these other ways of using taxcalc classes is likely
to require additions, and possibly revisions, in the current taxcalc
API.  This is part of what any successful project goes through as it
gains popularity: the original API that was focused on the original
use of the classes needs to be generalized to support a wider variety
of uses.  This process often causes tensions between the newcomers who
want a broader API to support their new applications, and the veterans
who designed the original API and have a responsibility to ensure that
any changes to the API do not undermine its original use.

**Some API requirements for a TAXSIM-like capability**

Users of this new taxcalc application will specify a text file of
income tax filing units (one unit per row that includes a unit id and
the unit's tax year) and get in return a text output file containing a
row for each input unit that includes the unit's tax liability and
intermediate results such as adjusted gross income and taxable income.
Users of this new application will have the option to specify a policy
reform file containing JSON (if no policy reform file is specified,
then the application uses current-law policy parameters from the
params.json file) and the option to specify an inflation rate file
containing JSON (if no inflation rate file is specified, then the
application uses the default inflation rates specified in the taxcalc
Parameters class.

To clarify the simple input and output style of the initial version of
this TAXSIM-like capability, consider an example of the 22-variable
input file that looks like this:

```
11 2013 0 1 0 0 58000 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
37 2015 0 3 1 1 46000 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
21 2014 0 1 0 0 18000 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
19 2013 0 2 0 1 49000 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
[...]

PR #341, opened 2015-08-11
Add multi-year reform test to test_parameters.py file

martinholmer:
Having this test in place is essential for the Parameters class API redesign work described in issue #336.


MattHJensen:
:+1:   @talumbau?

talumbau:
Yes, this is the explicit test that we needed for how inflation works for both scalar and vector quantity tax parameters. In particular, we encode that, for both scalar and vector, the year N rate is equal to the year N-1 * (1 + inflation[year N -1]).

+1 from me.

MattHJensen:
Great. Merging.

Issue #342, opened 2015-08-14
Questions about default_data() function in parameters.py


martinholmer:
There is no documentation on this function, so I'm confused about what it is supposed to return.

Read the following test (which is on the master branch) and look for my questions below.

```
def test_parameters_get_default_start_year():
    paramdata = taxcalc.parameters.default_data(start_year=2015, metadata=True)
    # 1D data, has 2015 values
    meta_II_em = paramdata['_II_em']
    assert meta_II_em['start_year'] == 2015
    assert meta_II_em['row_label'] == ["2015"]
    assert meta_II_em['value'] == [4000]

    # 2D data, has 2015 values
    meta_std_aged = paramdata['_STD_Aged']
    assert meta_std_aged['start_year'] == 2015
    assert meta_std_aged['row_label'] == ["2015"]
    assert meta_std_aged['value'] == [[1550, 1250, 1250, 1550, 1550, 1250]]

    # 1D data, doesn't have 2015 values, is CPI inflated
    meta_amt_thd_marrieds = paramdata['_AMT_thd_MarriedS']
    assert meta_amt_thd_marrieds['start_year'] == 2015
    assert meta_amt_thd_marrieds['row_label'] == ["2015"]
    # Take the 2014 rate and multiply by inflation for that year
    assert meta_amt_thd_marrieds['value'] == (
        [41050 * (1.0 + Parameters._Parameters__rates[2014])])

    # 1D data, doesn't have 2015 values, is not CPI inflated
    meta_kt_c_age = paramdata['_KT_c_Age']
    assert meta_kt_c_age['start_year'] == 2015
    assert meta_kt_c_age['row_label'] == ""
    assert meta_kt_c_age['value'] == [24]

    # 1D data, does have 2015 values, goes up to 2018
    meta_ctc_c = paramdata['_CTC_c']
    assert meta_ctc_c['start_year'] == 2015
    assert meta_ctc_c['row_label'] == ["2015", "2016", "2017", "2018"]
    assert meta_ctc_c['value'] == [1000, 1000, 1000, 500]
```

I have two sets of questions about the above test.

First, why does the _AMT_thd_MarriedS parameter return ["2015"] for
its 'row_label' attribute when the _KT_c_Age parameter returns ""?
Why the difference?  Neither have a 2015 row label value in params.json,
so why are they treated differently?

Second, why does the _CTC_c parameter have row labels and values
beyond 2015 when almost all the other parameters stop at 2015?
[...]

Issue #284, opened 2015-06-10
"write_params_to_json"


MattHJensen:
I can see two major uses for a "write_params_to_json" method that would create a json params file from the param attributes on the calculator. 
1. We could allow webapp users to download the json for use on their local version of taxcalc. This could be useful in many situations, including if we wanted to be able to run a tax plan that was created through the webapp on a local machine, without drop-q. 
2. One of the reasons why we use user_modes to create a plan y calculator in our 10_min notebook, rather than loading a planY.json params file, is that it would be a pain to maintain an example planY.json params file. Everytime we add or change a parameter to the taxcalc package, we would have to remember to duplicate that modification in both params.json and the new planY.json. With  `write_params_to_json`, we could have a little script that loads a calculator with params.json and a user_mod (params repeal state and local tax) and then calls `write_params_to_json` in order to create planY.json. 

A note on implementation: we would probably want "write_params_to_json" to build off of the default params.json rather than writing the notebook from scratch. 

@talumbau, what do you think about this?

talumbau:
very much :+1: on this. I don't know how useful the downloading of inputs as CSV is write now on the webapp. It is sufficient for reproducibility, but not particularly user friendly. This would be more user friendly.

zrisher:
And much easier to maintain!

MattHJensen:
I think this would excuse us from needing to display and maintain "inputs used" at the bottom of the results page. Rather, we can offer for download both `planX.json` and `planY.json`, and then only display the parameters changed (user_mods). 

As an aside, we may also want a `write_user_mods' function.

1. this would help with displaying "parameters changed" on the webapp (see above). 
2. I am not sure about this yet, but I think it might be easier to maintain forward compatibility with user_mods rather than with planY_params.json. (My thinking is that a user_mods structure generated from calc v1 would not work with calc v2 if the user_mod variables had been renamed or otherwise modified between v1 and v2 or if the user_mod api itself is changed. A planY.json generated from v1 would not work with v2 if *any* variable had been renamed or otherwise modified between v1 and v2 or if the json structure itself had been changed.)

feenberg:
On Fri, 12 Jun 2015, Matt Jensen wrote:

> 
> I think this would excuse us from needing to display and maintain "inputs
> used" at the bottom of the results page. Rather, we can offer for download
> both planX.json and planY.json, and then only display the parameters changed
> (user_mods).
>

My thought was PlanX would be existing law, and we would show in the 
[...]

PR #343, opened 2015-08-14
Parameters class API redesign

martinholmer:
This pull request includes a redesign of the Parameters class API plus Sphinx-aware documentation.  The new API is as described in issue #336.  The old Parameters class had no documentation, but a combination of the source code, the tests in test_parameters.py and test_calculate.py, and the helpful discussion in issue #342, provided a basis for the redesign.  The redesign includes a replacement for the old global default_data function, which passes all the tests and works as described in the issue #342 discussion.  Several changes have been made on Sep 1 (see details below in commit history) in response to helpful suggestions made by T.J. Alumbaugh (@talumbau) and @MattHJensen.


taxcalc/parameters.py
```diff
@@ -1,170 +1,496 @@
...
+
+    Returns
+    -------
+    class instance: Parameters
+    """
+
+    PARAMS_FILENAME = 'params.json'
+    IRATES_FILENAME = 'irates.json'  # TODO: move __rates there & add wages
+    JSON_START_YEAR = 2013  # remains the same unless earlier data added
+    FIRST_BUDGET_YEAR = 2015  # increases by one every calendar year
```
talumbau: This is surprising. We have a hard coded start year for web app users, but is it correct that the first 'usable' year for a Parameters instance is 2015? In other words, I thought the web app use case was special to the web app and not a restriction on usage of the class in general. Here, the code seems to imply that the first budget year for which one can use a Parameters instance in a valid way is 2015. Is that correct?

taxcalc/calculate.py
```diff
@@ -59,41 +59,28 @@ def calculator(params, records, mods="", **kwargs):
...
-        params = Parameters.from_file(pfname, **kwargs)
-        recs = Records.from_file(rfname, **kwargs)
-        return cls(params, recs)
-
     def __init__(self, params=None, records=None, sync_years=True, **kwargs):
 
         if isinstance(params, Parameters):
             self._params = params
         else:
-            self._params = Parameters.from_file(params, **kwargs)
```
talumbau: It's asymmetric that it is permissable to create a Records object from a string which is a path to a file, but not a Parameters object from a string which is a path to a file. We don't have to use a `from_file` class factory method, but we should probably permit one to do so from the __init__. More detail in a later comment.

taxcalc/parameters.py
```diff
@@ -1,170 +1,496 @@
...
+            msg = 'year passed to set_year() must be in [{},{}] range.'
+            raise ValueError(msg.format(self.start_year, self.end_year))
+        self._current_year = year
+        year_zero_indexed = year - self._start_year
+        for name in self._vals:
             arr = getattr(self, name)
-            setattr(self, name[1:], arr[yr - self._start_year])
+            setattr(self, name[1:], arr[year_zero_indexed])
+
+    @classmethod
```
talumbau: In prior discussions, I had advocated that, if there is an objection to leaving the `default_data` as a free-standing function, one could move it inside the Parameters class as a `@staticmethod` and not a `@classmethod`. Do you advocate for this to be a `@classmethod` or was this unintentional?

taxcalc/parameters.py
```diff
@@ -1,170 +1,496 @@
...
+                                inflate=cpi_inflated,
+                                inflation_rates=inf_rates,
+                                num_years=num_years_to_expand)
+            cval = getattr(self, name)
+            cval[(self.current_year - self.start_year):] = nval
+            setattr(self, name, cval)
+        self.set_year(self._current_year)
+
+    # TODO: eventually remove the following variable that is used
+    #       only in the depreciated global default_data() function
```
talumbau: *deprecated

taxcalc/parameters.py
```diff
@@ -1,170 +1,496 @@
...
+                default_cpi = False
+            cpi_inflated = year_mods[year].get(name + '_cpi', default_cpi)
+            # set post-reform values of parameter with name
+            nval = expand_array(values,
+                                inflate=cpi_inflated,
+                                inflation_rates=inf_rates,
+                                num_years=num_years_to_expand)
+            cval = getattr(self, name)
+            cval[(self.current_year - self.start_year):] = nval
+            setattr(self, name, cval)
```
talumbau: This line is not necessary. `cval` is just another name for the attribute whose name is the contents of `name`. In line 475, the assignment to some of its indices happens and after that, there is no more work to do.

taxcalc/parameters.py
```diff
@@ -1,170 +1,496 @@
...
+        if num_years is less than one.
+        if len(inflation_rates) is not equal to num_years.
+        if min(inflation_rates.keys()) is not equal to start_year.
+
+    Returns
+    -------
+    class instance: Parameters
+    """
+
+    PARAMS_FILENAME = 'params.json'
```
talumbau: Is there a reason to not leave these variables at module scope and instead scope them inside the class?

taxcalc/parameters.py
```diff
@@ -1,170 +1,496 @@
...
+            raise ValueError(msg.format(self.start_year, self.end_year))
+        self._current_year = year
+        year_zero_indexed = year - self._start_year
+        for name in self._vals:
             arr = getattr(self, name)
-            setattr(self, name[1:], arr[yr - self._start_year])
+            setattr(self, name[1:], arr[year_zero_indexed])
+
+    @classmethod
+    def default_data(cls, metadata=False, start_year=None):
```
talumbau: It seems like a bad idea to include this function but not remove the `default_data` free-standing function. Although we would all agree that this should not be the long term solution, things have a way of 'lingering' in code if not dealt with right away. I would suggest either verifying that this new `default_data` works in 

https://github.com/OpenSourcePolicyCenter/webapp-public/blob/master/webapp/

and in particular the test here:

https://github.com/OpenSourcePolicyCenter/webapp-public/blob/master/webapp/apps/taxbrain/tests.py

and then removing the `default_data` free-standing function, or not including the new `default_data` function in this PR.

taxcalc/tests/test_calculate.py
```diff
@@ -117,14 +118,21 @@ def test_make_Calculator_deepcopy():
 
 
 def test_make_Calculator_from_files(paramsfile):
-    calc = Calculator.from_files(paramsfile.name, tax_dta_path,
-                                 start_year=1991,
-                                 inflation_rates=irates)
+    with open(paramsfile.name) as pfile:
+        params = json.load(pfile)
+    ppo = Parameters(parameter_dict=params, start_year=1991,
```
talumbau: Here is a case where it seems that we have "handicapped" the Parameters class. Why not permit:

```
ppo = Parameters("myparams.json", start_year=1991, num_years=10)
```
or something similar to
```
ppo = Parameters(path="myparams.json", start_year=1991, num_years=10)
```

It seems particularly useful to have such a feature in light of the upcoming feature https://github.com/OpenSourcePolicyCenter/Tax-Calculator/issues/284, which we hope to integrate to the webapp. This would allow a non-coder to create a policy reform, download a JSON file, give that to someone who could use `taxcalc` in a notebook, and then the Python user could create the individual's reform with something like:

```
ppo = Parameters(path="bobs_reform.json", start_year=1991, num_years=10)
```

which seems nice.

talumbau:
I made some comments/asked some questions inline. There are significant changes to test_calculate.py and test_parameters.py, and in this case I don't think Github diff-ing is sufficient to see what behaviors are changing, so I'll do some offline diff-ing. The motivation for the refactor is described in https://github.com/OpenSourcePolicyCenter/Tax-Calculator/issues/336, so I was looking for new tests that show increased abilities and some path toward the TAXSIM capabilities. Is `test_multi_year_reform` the right place to see a "sneak peek" of the new capabilities?

taxcalc/tests/test_parameters.py
```diff
@@ -41,71 +42,96 @@ def test_create_parameters():
...
+            expected = [str(start_year + i) for i in range(len(value))]
+            assert row_label == expected
+        else:
+            num_missing_row_labels += 1
+    if num_missing_row_labels != 0:
+        msg = 'num_parameters_in_total={}\n'
+        sys.stdout.write(msg.format(num_parameters))
+        msg = 'num_without_row_label_list={}\n'
+        sys.stdout.write(msg.format(num_missing_row_labels))
+        # assert num_missing_row_labels == 0
```
talumbau: I think it's fine to `assert` here or raise some kind of exception. I think it's also fine to `print` here, or if there is a desire to really direct this output somewhere, I would suggest `stderr`.

taxcalc/tests/test_parameters.py
```diff
@@ -330,86 +345,96 @@ def check_ss_earnings_c(ppo, reform, ifactor):
...
+        paramdata = Parameters.default_data()
+    else:  # use legacy global default_data function
+        paramdata = taxcalc.parameters.default_data()
     assert paramdata['_CDCC_ps'] == [15000]
 
 
-def test_update_Parameters_raises_on_no_year(paramsfile):
-    p = Parameters.from_file(paramsfile.name)
-    user_mods = {"_STD_Aged": [[1400, 1200]]}
+def test_reform_with_no_year():
```
talumbau: Do you object to the style of 'test that expects an exception to be raised has "raises" in the name'? I find that it is a helpful idiom to keep it straight in my brain that this is a test wherein, if an exception is *not* raised, the test fails. I have tried to be consistent with such a convention throughout the tests. Now that I've explained the idiom, what are your thoughts on it?

taxcalc/tests/test_parameters.py
```diff
@@ -330,86 +345,96 @@ def check_ss_earnings_c(ppo, reform, ifactor):
...
+    ppo = Parameters(start_year=2013)
+    reform = {2013: {"_STD_Aged": [[1400, 1100, 1100, 1400, 1400, 1199]]}}
+    ppo.implement_reform(reform)
+    assert_array_equal(ppo.STD_Aged,
                        np.array([1400, 1100, 1100, 1400, 1400, 1199]))
 
 
-def test_update_Parameters_raises_on_future_year():
-    p = Parameters(start_year=2013)
+def test_Parameters_reform_before_start_year():
```
talumbau: same comment as above here, regarding 'raises' in the test name.

taxcalc/tests/test_parameters.py
```diff
@@ -330,86 +345,96 @@ def check_ss_earnings_c(ppo, reform, ifactor):
...
+
+
+def test_Parameters_reform_with_default_cpi_flags():
+    ppo = Parameters(start_year=2013)
+    reform = {2015: {"_II_em": [4300]}}
+    ppo.implement_reform(reform)
+    # '_II_em' has a default cpi_flag of True, so
+    # in 2016 its value should be greater than 4300
+    ppo.set_year(2016)
+    assert ppo.II_em > 4300
```
talumbau: yes this is a better test of what should happen.

taxcalc/tests/test_calculate.py
```diff
@@ -117,14 +118,21 @@ def test_make_Calculator_deepcopy():
 
 
 def test_make_Calculator_from_files(paramsfile):
```
talumbau: I still advocate for a way to make a Parameters object from a file name, but I would actually kill this test since there is no longer a `from_files` class method on `Calculator`

taxcalc/tests/test_calculate.py
```diff
@@ -170,54 +178,62 @@ def test_make_Calculator_user_mods_as_dict():
...
+                             [1532, 1227, 1533, 1228, 1535, 1229]])
     assert_array_equal(calc.params._STD_Aged, exp_STD_Aged)
+    exp_II_em = np.array([3900, 3950, 5000, 6000])
     assert_array_equal(calc.params._II_em, exp_II_em)
-    assert_array_equal(calc.params.STD_Aged, exp_cur_STD_Aged)
-    assert_array_equal(calc.params.II_em, exp_cur_II_em)
 
+    # compare actual and expected values for 2010
+    calc.increment_year()
+    calc.increment_year()
```
talumbau: This is the main heart of the change here. When using the `calculator` function to build the desired Parameters and Calculator object, we automatically increment the year of the Calculator instance. This way is fine with me, as long as we make similar changes in `dropq`. So, we should attach some kind of milestone name and use it consistently across the two repos to make sure that the removal of testing of the `calculator` function goes along with the removal of its actual usage in the `dropq` repo.

taxcalc/tests/test_calculate.py
```diff
@@ -170,54 +178,62 @@ def test_make_Calculator_user_mods_as_dict():
...
+    exp_2010_II_em = 5000
+    assert_array_equal(calc.params.II_em, exp_2010_II_em)
+    exp_2010_STD_Aged = np.array([1501, 1202, 1502, 1203, 1504, 1204])
+    assert_array_equal(calc.params.STD_Aged, exp_2010_STD_Aged)
 
-def test_make_Calculator_user_mods_with_cpi_flags(paramsfile):
 
-    calc = Calculator(params=paramsfile.name,
-                      records=tax_dta_path, start_year=1991,
+def test_make_Calculator_user_mods_with_cpi_flags(paramsfile):
```
talumbau: I would say this test is no longer necessary in its modified form. The original purpose of the test was to make sure that a `Calculator` instance made with the `calculator` function would properly handle a `user_mods` dictionary and expand the array in the right way. Here, the `calculator` function is no longer used, and so we directly apply the reform the `Parameters` instance, for which we already have tests in `test_parameters.py`

taxcalc/tests/test_calculate.py
```diff
@@ -226,27 +242,24 @@ def test_make_Calculator_user_mods_with_cpi_flags(paramsfile):
...
     exp_almsep_values = [40400] + [41050] * 11
     exp_almsep = np.array(exp_almsep_values)
-
-    assert_array_equal(calc2.params._almdep, exp_almdep)
-    assert_array_equal(calc2.params._almsep, exp_almsep)
+    act_almsep = getattr(calc.params, '_almsep')
+    assert_array_equal(act_almsep, exp_almsep)
 
 
 def test_make_Calculator_empty_params_is_default_params():
```
talumbau: For this one, the purpose of the test is to assert that the creation of a `Calculator` object with no passed `Parameters` instance is allowed. But in the modified test, we create one and pass it in. So, if it is no longer permitted to create a `Calculator` instance in such a way, I would suggest modifying this test to `test_make_Calculator_no_params_raises` or a similar name to explicitly assert that this is no longer allowed.

taxcalc/parameters.py
MattHJensen: Agreed with T.J. that it would be better to figure out whether it is ok to remove the earlier `default_data` global function before adding the new class method. @talumbau, could you do the testing you propose with the webapp so that @martinholmer doesn't need to get up and running with that code base yet?

taxcalc/parameters.py
MattHJensen: > cval is just another name for the attribute whose name is the contents of name. 

@talumbau, to verify my understanding, this is because `cval` isn't a deepcopy of `name`, right?

taxcalc/tests/test_calculate.py
MattHJensen: @talumbau, I'm glad you bring this up, and @martinholmer, it would probably be useful for you to review the entire discussion in #284 as background, rather than just the top post. Apologies for not suggesting this earlier.

It's actually a bit more complicated than what the discussion in #284 might imply, which I get to two paragraphs down. 

First, the core takeaway is that it has long been @feenberg and my view that some users might prefer to specify a reform by way of a description of the reformed law in its entirety (a JSON file) rather than by way of a description of the changes in law from baseline (a parameter_dict). The way that this would work in practice is that a user would create a planX calculator from a PPO using the default `params.json` and a planY calculator from a PPO using some `planY_params.json` and then the user would compare the planX and planY calculators.  

Here is the additional complexity: a smaller subset of users might not even want to use `params.json` at all  -- there have, after all, been serious disagreements in the past over what the appropriate baseline is. These users might want to create the *planX* calculator from a PPO with `params_mybaseline.json` rather than `params.json` and then either use a JSON file *or* a parameter_dict to create the PPO for the planY calculator. Following this train of thought for the first time, I see that if the user wants to specify a different baseline with a JSON file to create the PPO for the planX calculator and then use a dictionary of reforms to specify changes from that baseline for the PPO for the planY calculator, then it means that a user will want to specify *both* a JSON file and a parameter_dict in order to create the PPO for the planY calculator. I don't see why this complexity would be a problem but wanted to point it out since it isn't documented elsewhere.

taxcalc/calculate.py
MattHJensen: See relevant discussion on line 122 of `test_calculate.py`

taxcalc/parameters.py
MattHJensen: @talumbau, what are the pros and cons of having a `@staticmethod` vs. `@classmethod` here?

taxcalc/parameters.py
talumbau: yep. I'll give it a whirl and report back here.

taxcalc/parameters.py
talumbau: It's correct to note that `cval` is not a copy (deep or shallow) of the attribute (whose name is the contents of `name`). Sometimes it's helpful to compare variables in Python to "references" in other languages, say C++, although that analogy falls apart when you start to push it too hard. I've found that the best way to think about variables in Python is just to think of them as names, or even just "name tags" that we stick on things, the "things" being values.

taxcalc/parameters.py
talumbau: I would say the main ’pro’ of going with @staticmethod is that it most closely reflects programmer intent for the function. All of the existing use cases for the defaultdata function are essentially about grabbing out the default data that “comes with” taxcalc. It’s just one step above the raw opening of a file and grabbing out JSON data. The ‘con’ of the three new @classmethod functions is that it communicates that there is something stateful about the behavior, or that this behavior is dependent in some way on the Parameters class itself. For example, the style of these three @classmethod functions is desirable if you have a situation where there is a subclass of your base class, and wish to customize behavior of one of your class methods. The advantage of OO kicks in at that point, because the `cls` argument would be the proper class object (either Parameters or its subclass) so you get code reuse. In the existing use cases, we don’t have any need for behavior like that. In the refactor, there are a number of mechanisms in play (several bound methods of the Parameters class, two of which use the convention for “private to this class”) which don’t help communicate the stateless-ness of the function. With a "transplant" of the module-scope function to @staticmethod, you are communicating 'this function is logically grouped with what's happening in this class but it doesn't need the binding mechanism of class methods and I'm just putting it here because if you're reading the code of this class you should probably know what this function does too'. I think that's an accurate picture of the intent of the function.

taxcalc/parameters.py
MattHJensen: @talumbau, what are the tradeoffs between module scope and class scope for these variables?

taxcalc/parameters.py
MattHJensen: The calculator should be kept 'usable' for any year for which we have tax logic (currently the first year is 2013, we hope to add 2011 and 2012 in the future). That said, it is almost certainly true that the vast majority of users will want to start with the first year in the federal budget window, which is currently 2015.

taxcalc/parameters.py
talumbau: The tradeoff that I see is that if we keep these constants at module level scope we don't have to type `Parameters.PARAMS_FILENAME` inside this module, so the code is just more concise. Also, the name of the file in which we store the data seems to me to be more of a characteristic of our package itself, and not particular to the Parameters class. The filename is about the physical layout of our package data; the class is an abstraction. Typically, variables like this are scoped inside classes to prevent pollution of a global namespace. This is more of an issue when you have languages with a true preprocessor, like C++ or Java. In Python, the only way that these variables would "leak" is if you did this in some other module:

```
from parameters import *
```

which is an atypical thing to do anyway. Scoping inside the class would keep these variables "inside" in that case, but that doesn't seem like a compelling argument to me. So I was wondering if there was another reason.

taxcalc/parameters.py
talumbau: @MattHJensen I tried to test this in dropq and the web application. Here is what I found:

- the tests in webapp-public pass in the one case where we call default_data using the new default_data

- I was unable to test this branch in dropq because of a behavior change between the existing `update` method and the new `implement_reform` method. I don't know if this is an intentional change or not. If it is intentional, @martinholmer and I will need to walk through how one should create a Calculator instance and add reforms. Here is the new behavior vs. the old:

```
from taxcalc import Calculator, Records, Parameters

# User specified plan, start at 2013
params = Parameters(start_year=2013)
# setting up for reform for 2015
params.set_year(2014)
params.set_year(2015)
assert params.current_year == 2015
mods = {2015: {'_II_rt3': [0.31], '_II_rt4': [0.39]}}
params.implement_reform(mods)
# I would assume params.current_year is still 2015
#this fails because params.current_year goes back to 2013
assert params.current_year == 2015
```

`dropq` builds Parameter instances assuming that after implementing a reform (or "updating" in the old way) the current_year remains the same. So, I could impose a solution in my branch and try to get dropq to run. Should I do that?

martinholmer:
> @MattHJensen I tried to test this in dropq and the web application. Here is what I found:

> the tests in webapp-public pass in the one case where we call default_data using the new default_data

> I was unable to test this branch in dropq because of a behavior change between the existing update method and the new implement_reform method. I don't know if this is an intentional change or not. If it is intentional, @martinholmer and I will need to walk through how one should create a Calculator instance and add reforms. Here is the new behavior vs. the old:

```
from taxcalc import Calculator, Records, Parameters

# User specified plan, start at 2013
params = Parameters(start_year=2013)
# setting up for reform for 2015
params.set_year(2014)
params.set_year(2015)
assert params.current_year == 2015
mods = {2015: {'_II_rt3': [0.31], '_II_rt4': [0.39]}}
params.implement_reform(mods)
# I would assume params.current_year is still 2015
#this fails because params.current_year goes back to 2013
assert params.current_year == 2015
```

> dropq builds Parameter instances assuming that after implementing a reform (or "updating" in the old way) the current_year remains the same. So, I could impose a solution in my branch and try to get dropq to run. Should I do that?

T.J. @talumbau (and @MattHJensen) would something like the following work?

```
params = Parameters()
mods = {2015: {'_II_rt3': [0.31], '_II_rt4': [0.39]}}
params.implement_reform(mods)
params.set_year(2015)
```

talumbau:
@martinholmer for inline conversations, would you mind keeping the conversation inline using the Add a Line Note button?

screen shot 2015-08-31 at 12 53 29 pm

We have many inline conversations on this PR and if we use the main Conversation thread I think it will get too complicated.

taxcalc/parameters.py
talumbau: From @martinholmer's comment, below the suggestion is to do as follows:

```
params = Parameters()
mods = {2015: {'_II_rt3': [0.31], '_II_rt4': [0.39]}}
params.implement_reform(mods)
params.set_year(2015)
```
That would definitely work for the code snippet I used above. The questions I then have would be:

 - What is benefit of changing the current year to the start year via `implement_reform` and is it intuitive to do such a thing? My answer to the second question is 'no'. 

- In particular, it does not seem to affect the overall algorithm if line 186 is changed to:
```
   self.set_year(temp_yr)
```

where `temp_yr` could just be the current year saved off in a temp variable at the beginning of the function.

Also, the functionality described above is inside the `calculator` function, which dropq uses, so the suggested changes (setting the year back to the desired after calling `implement_reform` would need to happen inside the `calculator` function before creating the `Calculator` instance. This use case for `dropq` is handled in the test `test_make_Calculator_increment_years_first` in test_calculator.py. 

The PR would be much easier to merge and, as a preliminary setp, test against dropq if `test_make_Calculator_increment_years_first` was preserved.

martinholmer: Why is it not "intuitive" to leave the current_year after implementing a reform at
the value it was when the Parameters object was instantiated?

What is "line 186"?

Why all the talk about the calculator factory function?  I thought we were in agreement that it would be eliminated.

talumbau: It would seem to me that one should argue from the perspective that the mutation of some state is intuitive, rather than force someone to argue that *not* mutating state is intuitive. In other words, all I'm saying is 'why should this value change?' and it seems like a bad way to go to say "why shouldn't it change?" If that is the case, then perhaps any member data should be allowed to change, but then how could the user have an expectation on what the state should be after calling a method?

"line 186" refers to line 186 of `calculate.py` in this PR.

It seems reasonable to deprecate the `calculator` function. To eliminate it, or as in this PR, to make it trip an assertion, and then merge that to main would instantly break the dropq package. That seems like a bad idea because then we are faced with either merging a PR to master in taxcalc that instantly breaks master on dropq, or doing a simultaneous refactor of dropq at the same time as keeping a PR open on taxcalc, which seems like an unnecessary roadblock. If we deprecate it, then provide a similar idiom as to what the `calculator` function does, then we can independently merge PRs on taxcalc and dropq.

martinholmer: T.J. @talumbau, I'm totally confused by your comments.

Last week you reported that the redesigned Parameters class had worked fiine with one part of the  webapp, but you couldn't see how to use the redesigned Parameters class in the other part of the webapp (and posted the code you couldn't make work).

Then today, I responded to you suggesting a different way to use the redesigned Parameters class in that second test.  Now in your latest comment in this thread you seem to be saying that the second webapp test fails when using the redesigned Parameters class.  Is that true?  If so, can you post the webapp code that fails when trying to using the redesigned Parameters class and also post the error message you get?  It would be helpful to the OSPC team if you could post a complete report of what testing you have done and what the test results have been.  If there is not a test failure when using the redesigned Parameters class in the webapp, then I don't understand what all the conversation is about.

Can we both argee to focus on the facts of the matter and put off an exchange of opionions until everybody understands the facts?

talumbau: Maybe we should schedule a code review to increase clarity. There are three repositories in play here. There is this one (Tax-Calculator), dropq, and webapp-public. Last week, I reported that the new class method `default_data` passes a test in `webapp-public`. `webapp-public` is a user of both the `taxcalc` package and the `dropq` package. `dropq` is a user of the `taxcalc` package. In order to fully run the web application (and not just the unit tests in `webapp-public`, I attempted to run the `dropq` package in a way similar to how it is done in `webapp-public`. I was unable to do this for the above stated reasons. I did not post code that I was unable to get work; I posted code that illustrated the problem with the code in this branch. The code in `dropq` that fails when run against this branch is best illustrated by looking at `test_make_Calculator_increment_years_first`. If this test were to pass as it exists in master, then the `dropq` package would work, because the test is designed to show how `dropq` uses the package. It seems more helpful to focus on a test or show a small snippet of code indicating the failure in assumptions than to say something like: 

Run this file:
https://github.com/OpenSourcePolicyCenter/dropQ/blob/master/example.py

And you'll get an assertion violation at this line in `calculate.py`:

```
 assert self._params.current_year == self._records.current_year
```

That would send someone down a path of digging in to a bunch of code attempting to discover why the assertion was violated. My goal was to save time and get to the heart of the issue, which for this method, is a mutation of `current_year` where there previously had not been one.

martinholmer: T.J. @talumbau said on Monday:
> Run this file:
> https://github.com/OpenSourcePolicyCenter/dropQ/blob/master/example.py
>
> And you'll get an assertion violation at this line in calculate.py:
> ```
> assert self._params.current_year == self._records.current_year
> ```

Can you post the **full** assert error message including the calling trace,
so everybody can see what code was executing when this error occurred?
That would be very helpful in our efforts to resolve the problems you are
having.

talumbau: Can you explain what you mean by "the problems you are having"? In this thread, I'm attempting to explain why this branch won't run the functionality in dropq. Is that what you mean by "the problems you are having?" Here are some ways to reproduce the issue:

- run `test_make_Calculator_increment_years_first` as it exists in master on this branch, changing only the `budget_years` keyword arg from master to `num_years`. Then run pytest as follows:

```
py.test taxcalc/tests/test_calculate.py -k test_make_Calculator_increment_years_first
```

A second way to reproduce is as I mentioned above:

- install via `python setup.py install` taxcalc synced with this branch, checkout the latest master of dropq, then run `example.py` at the top of the repo.

Then you will see the assertion violation. Are you saying that you cannot reproduce this behavior? 

```
  File "example.py", line 14, in <module>
    mY_dec, mX_dec, df_dec, mY_bin, mX_bin, df_bin, fiscal_tots = dropq.run_models(tax_dta, num_years=10, user_mods=user_mods)
  File "/Users/talumbau/dev/dropq/dropQ/dropq/dropq.py", line 320, in run_models
    calc3 = calculator(params=params3, records=records3, mods=user_mods)
  File "/Users/talumbau/miniconda3/envs/heroku_aei/lib/python2.7/site-packages/taxcalc-0.2_644_g13a470e-py2.7.egg/taxcalc/calculate.py", line 63, in calculator
  File "/Users/talumbau/miniconda3/envs/heroku_aei/lib/python2.7/site-packages/taxcalc-0.2_644_g13a470e-py2.7.egg/taxcalc/calculate.py", line 95, in __init__
AssertionError
```

talumbau: I posted a PR against your branch that resolves the issues:

https://github.com/martinholmer/Tax-Calculator/pull/1

martinholmer: T.J. @talumbau reported this error when using the redesigned Parameters class in the dropQ repository:
```
  File "example.py",
       line 14, in <module>
       mY_dec... = dropq.run_models(tax_dta, num_years=10, user_mods=user_mods)
  File "/Users/talumbau/dev/dropq/dropQ/dropq/dropq.py",
       line 320, in run_models:
       calc3 = calculator(params=params3, records=records3, mods=user_mods)
  File "egg/taxcalc/calculate.py",
       line 63, in calculator [factory function]
  File "egg/taxcalc/calculate.py",
       line 95, in __init__:
       AssertionError
       assert self._params.current_year == self._records.current_year
```
The relevant dropq.run_models code is as follows:
```
def run_models(tax_dta, user_mods="", num_years=NUM_YEARS_DEFAULT,
               start_year=2015):

    # 'fake' advance to handle type changes for jitting
    ...

    #########################################################################
    #   Create Calculators and Masks
    #########################################################################
    records = Records(tax_dta)
    records2 = Records(tax_dta.copy(deep=True))
    records3 = Records(tax_dta.copy(deep=True))
    # add 1 dollar to gross income
    records2.e00200 += 1
    records2.e02000 += 1
    records2.e02615 += 1
    # Default Plans
    # Create a default Parameters object
    params = Parameters(start_year=2013)
    # Create a Calculator
    calc1 = Calculator(params=params, records=records)
    calc1.increment_year()
    calc1.increment_year()
    calc1.calc_all()
    assert calc1.current_year == start_year

    params2 = Parameters(start_year=2013)
    # Create a Calculator with one extra dollar of income
    calc2 = Calculator(params=params2, records=records2)
    calc2.increment_year()
    calc2.increment_year()
    calc2.calc_all()
    assert calc2.current_year == start_year

    # where do the results differ..
    soit1 = results(calc1)
    soit2 = results(calc2)
    mask = (soit1._ospctax != soit2._ospctax)

    # User specified Plans
    params3 = Parameters(start_year=2013)
    # Create a Calculator for the user specified plan
    calc3 = calculator(params=params3, records=records3, mods=user_mods)
    assert calc3.current_year == start_year
    calc3.calc_all()
    soit3 = results(calc3)
    ...
```
Notice that calc1 and calc2, both of which are created using the Calculator constructor, work without error.  The problem comes in using the calculator factory function, which we agreed (at least I thought we agreed) in July to eliminate soon.  One way to think about the error T.J. is getting is that it is caused by an error of mine in the calculator factory function (which is largely mysterious to me).  One strategy is to fix that error, but I'm reluctant to put in effort on something that is slated for extinction.  The other strategy is to simply use the new Parameters class the way it is supposed to be used and rewrite the calc3 part of the dropq.run_models code as follows:
```
    # User specified Plans
    params3 = Parameters(start_year=2013)
    params3.implement_reform(user_mods)
    # Create a Calculator for the user specified plan
    calc3 = Calculator(params=params3, records=records3)
    calc3.increment_year()
    calc3.increment_year()
    calc3.calc_all()
    assert calc3.current_year == start_year
```
Does this code produce the same error as T.J. reported above?  I'm asking because I do not have the knowledge or capabilities to run the webapp on my computer.

talumbau: Hi Martin,

It seems good to deprecate the `calculator` function, but causing to raise an assertion (as is done in this PR) seems not good, because it breaks existing code in two other repositories. A solution that involves zero additional effort is to merge this PR martinholmer#1 in to your branch. The reasons this is good are as follows:

 - it means dropq and webapp-public will run, instead of raising asserts
 - it doesn't break any tests in this branch
 - it explicitly asserts that the `calculator` function should still work, which it should because it is used in master in dropq.

I don't know is meant by the question: "Does this code produce the same error as T.J. reported above? "

There's no need to run the webapp to demonstrate this issue. One question I asked above was "Are you unable to reproduce this problem?" Let me know if the instructions on reproducing above are unclear.

martinholmer: T.J. @talumbau,  @MattHJensen asked me to download the dropQ repo and try out my ideas.  Here is my report on the run_models code changes (to avoid the calculator factory function) that eliminate the errors you were experiencing:
```
NOTES ON DROPQ-MASTER/EXAMPLE.PY USE WITH REDESIGNED PARAMETERS API
===================================================================

(A) out of the box example.py (with only path changes to find files):

iMac2:dropQ-master mrh$ python example.py
You loaded data for 2008.
Your data have beeen extrapolated to 2013.
You loaded data for 2008.
Your data have beeen extrapolated to 2013.
Modifications are for year 2015 and Parameters are for year 2013.
  Parameters will be advanced to year 2015
Modifications are for year 2015 and Records are for year 2008.
  Records will be advanced to year 2015
  File "example.py", line 13, in <module>
    mY_dec, mX_dec, df_dec, mY_bin, mX_bin, df_bin, fiscal_tots = dropq.run_models(tax_dta, num_years=10, user_mods=user_mods)
  File "/Users/mrh/work/OSPC/dropQ-master/dropq/dropq.py", line 324, in run_models
    calc3 = calculator(params=params3, records=records3, mods=user_mods)
  File "/Users/mrh/work/OSPC/dropQ-master/dropq/../../tax-calculator/taxcalc/calculate.py", line 63, in calculator
    calc = Calculator(params, records)
  File "/Users/mrh/work/OSPC/dropQ-master/dropq/../../tax-calculator/taxcalc/calculate.py", line 95, in __init__
    assert self._params.current_year == self._records.current_year
AssertionError

This is the error T.J. reported; the relevant dropq.run_models calc3 code is:
    # User specified Plans
    params3 = Parameters(start_year=2013)
    # Create a Calculator for the user specified plan
    calc3 = calculator(params=params3, records=records3, mods=user_mods)
    assert calc3.current_year == start_year
    calc3.calc_all()

(B) revise calc3 code to avoid the calculator factory function as follows:

    # User specified Plans
    params3 = Parameters(start_year=2013)
    params3.implement_reform(user_mods)
    params3.set_year(start_year)
    # Create a Calculator for the user specified plan
    calc3 = Calculator(params=params3, records=records3)
    calc3.calc_all()
    assert calc3.current_year == start_year

Also, make the following change to handle default_data now being a static
method, not a global function:

    #all_params = sorted([str(k) for k in taxcalc.parameters.default_data()])
    all_params = sorted([str(k) for k in Parameters.default_data()])

With these two sets of run_models code changes, we have the following results:

iMac2:dropQ-master mrh$ python example.py
You loaded data for 2008.
Your data have beeen extrapolated to 2013.
You loaded data for 2008.
Your data have beeen extrapolated to 2013.
You loaded data for 2008.
Your data have beeen extrapolated to 2015.
('elapsed time for second run: ', 3.204669952392578)
('elapsed time for second run: ', 6.002561807632446)
('elapsed time for second run: ', 5.773066997528076)
('elapsed time for second run: ', 5.514545917510986)
('elapsed time for second run: ', 5.405025959014893)
('elapsed time for second run: ', 5.396327018737793)
('elapsed time for second run: ', 5.584659099578857)
('elapsed time for second run: ', 5.554149866104126)
('elapsed time for second run: ', 5.574904918670654)
('elapsed time for second run: ', 5.687175035476685)

Is this the expected output from example.py?
```

talumbau: Yes, this is how the file should run. To be explicit about our options, there seem to be two paths available:

- merge the PR I suggested above and restore the functionality of the test that it contains. Then continue on with this PR and hopefully merge it to master. Leave dropq as it is. Then open a follow-on PR to dropq and hopefully merge it to master. Then remove the calculator function. 

-  change dropq to conform to this PR and then create a PR on dropq that conforms to this PR. Hopefully merge both of those to master.

Although you are not saying so explicitly, you are proceeding down the path of the second option. Would you say that your preference is the second option? You would not have this information available, but the option that saves the most time is the first option. The reason has to do with the deployment of this code to our production servers. The heart of the issue is this: it is much, much harder to deploy two pull requests (this PR and the dropq PR) to a test server than it is to deploy the two repos at master. I can't see a reason not to do option 1.

MattHJensen:
I believe this PR is ready to be merged based on the following analysis:

There are two outstanding lines of discussion that have not yet been fully addressed, but after conversing with @martinholmer on the phone this morning, I believe both of them ought to be dealt with in follow-on work. These two areas, and the reason for separating them from this PR, are as follows:

1. @talumbau raised several points about test names and test coverage (or overcoverage) which have not yet been addressed. The first and foremost reason for dealing with them separately is that they are not urgent and do not affect the existing functionality of the code. Second, I believe our test naming conventions relate to our ongoing conversation over coding style (https://github.com/OpenSourcePolicyCenter/Tax-Calculator/issues/30#issuecomment-126374926); in particular, test naming conventions will come up when we move on to the second step of our coding style conversation:
> "After that, @martinholmer will open a separate issue that lists all of the classes of warnings that pylint generates and opens a discussion on which to keep or whether to use some alternative to pylint." 

2. @talumbau and I raised a point about allowing the user to specify a JSON file of tax parameters other than `params.json`. This too is not urgent for our users, and can therefore -- from a timing perspective -- be dealt with later. Moreover, @martinholmer raised the point that we should be doing some checking when such a file is read to ensure that the user supplied JSON is valid, which we do not currently do on master. When we reintroduce this feature, we ought to include some of this checking. 
As @zrisher has said,
>  users will make eventually make every usage error you can think of. : )

An additional reason for separating both these issues into follow on work is that they might both require quite a bit of discussion, and this PR is already quite cluttered. 

@talumbau, if this analysis seems reasonable to you, could you go ahead and merge this PR?

talumbau:
@MattHJensen OK, I'll merge this and create follow-on issues.

PR #347, opened 2015-09-08
Various fixes on tests

talumbau:
- test that assert the raise of an Exception contain 'raise' in name
- remove 'test_make_Calculator_from_files' since it is identical to
  following test. There is no longer a 'from_files' class method
  so the test should have been removed.

### DIFF
--- a/taxcalc/tests/test_calculate.py
+++ b/taxcalc/tests/test_calculate.py
@@ -117,16 +117,6 @@ def test_make_Calculator_deepcopy():
     calc2 = copy.deepcopy(calc)
 
 
-def test_make_Calculator_from_files(paramsfile):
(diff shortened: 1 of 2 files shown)


talumbau:
looking to merge this soon... Any objections?

MattHJensen:
On naming: it makes sense to me to keep the naming scheme consistent across tests for now (by making this change) and then to review the naming scheme more systematically when we proceed with the second step of our code style review https://github.com/OpenSourcePolicyCenter/Tax-Calculator/issues/30#issuecomment-126374926. 

On removing the "from_file" test, that makes sense as well. 

:+1: 

@martinholmer, could you look this over and merge it if all seems good?

Issue #291, opened 2015-06-22
Add TAXSIM-like input and output capabilities


martinholmer:
Currently users following the Contributor Guide cannot do much with
the Tax Calculator source code beyond running `py.test`.  The
experience of users just getting started with the Tax Calculator could
be improved considerably if there was a capability of reading in one
or a few simple tax filing units from a text file and getting back
from the Tax Calculator a text file with the results.  I think the
best way to start adding this capability would be to adopt the
Internet TAXSIM file input and output format.  I am imagining that
this model enhancement can be accomplished by leaving the taxcalc code
unchanged and adding two functions.  One would read a TAXSIM-like
input file and write a PUF-like csv file for taxcalc input.  And the
other would write a TAXSIM-like output file by extracting information
from the Tax Calculator results records.  These two functions could
be combined into a single Python script that reads the TAXSIM-like
input file and the specified policy parameters, uses taxcalc to do
the calculations, and then writes the TAXSIM-like output file.

MattHJensen:
For reference, here is internet taxsim: http://users.nber.org/~taxsim/taxsim9/

mmessick:
@martinholmer, I just opened a PR ( #309 ) to facilitate the incorporation of Luca into the code base. I was facing a similar issue with trying to run individual files on our calculator and I thought that refactoring the initialization method for a Records object might help with this. That way, you would not have to write output to a csv file (which might end up making the code run faster). Let me know what you think and if this change would be helpful to you

martinholmer:
The enhancement described in this issue #291 --- addin a capabiltiy like Internet TAXSIM --- was originally implemented in pull reequest #376.

PR #348, opened 2015-09-09
Update taxcalc with 09 PUF

Amy-Xu:
In this PR, several changes have been made to accommodate 09 PUF.
1. removed variables (mainly e18425 and e18450) that don't exist in 09 anymore, and added new variables (mainly e18400).
2. excluded the aggregated record in PUF and added a function to blow up 09 PUF. This new function right now hard coded all the blow-up factors. 
3. modified the tests to adjust for all the changes. Mainly regenerated two gzipped files for testing purposes.


Amy-Xu:
@MattHJensen I'm not really sure why AGI was increased by about 60b in 2013. Will look into that tomorrow. But generally this is how I modified the code to read 09 PUF directly. Could you take a look?

This PR:
screen shot 2015-09-09 at 5 22 10 pm


OSPC master:
screen shot 2015-07-16 at 1 26 29 pm

Amy-Xu:
It seems all the failures have something to do with PEP8. Will fix that soon.

MattHJensen:
@Amy-Xu, this looks good to me as a short-run solution. Could you leave a comment with a list of issues that you think will need to be dealt with later on to allow for flexible changing between base data years? (The comment you leave here will be useful for when @talumbau or others review this PR, and then the comment can be turned into an open issue when this is merged.).

Amy-Xu:
Sure. @talumbau Here's a bit background. The new PUF has a slightly different set of variables (so some functions in functions.py have been adjusted), and more importantly, has a giant record which aggregates all the records with extreme values. As this giant record would potentially distort the results, we remove this record when importing puf, and then distribute the information in this record to all other records using our extrapolation results (implemented in records.py, adding extrapolate_09_puf() function and removing the giant record). 

As you can tell, two gzipped files for testing have also been replaced because of the different base file.

The goal to make our calculator to be able to accommodate both 08 PUF and 09 PUF. Originally @MattHJensen (pls correct me if I misspeak about anything) and I thought our calculator could achieve this simply by reading the variables names, and checking whether certain variables exist or not, without adding a global variable indicating what year of PUF is used. 

But right now we have an extra blow up for 09 PUF, and a new set of WEIGHTS, Stage I factors, and testing files. It seems to be pretty difficult not having a standalone variable for the year of PUF. No matter how we're going to implement this, here are several issues to keep in mind.

1. 09 PUF should only be turned on when running with 09 PUF.
2. 08 and 09 PUF have their own versions of WEIGHTS.csv and StageIFactors.csv.
3. Some variables only exist in either 08 PUF or 09 PUF. This would influence the initialization of Record, some functions in functions.py and test_functions.py. test_all1991_puf.gz should have two versions for the same reason.
4. As 08 and 09 PUF contain essentially different data, exp_results.csv.gz should have two versions.

Amy-Xu:
@talumbau Aside from this refactoring, I also have a question on aliasing from testing the results with 09 PUF. In records.py, variable p23250 was aliased as e23250. But in the blow up function, the blow up factor for this variable is still applied to p23250 for all the projection years. Can e23250 actually get the blow-up as well through aliasing at initialization?

feenberg:
On Thu, 10 Sep 2015, Amy Xu wrote:

> 
> Sure. @talumbau Here's a bit background. The new PUF has a slightly
> different set of variables (so some functions in functions.py have been
> adjusted), and more importantly, has a giant record which aggregates all the
> records with extreme values. As this giant record would potentially distort
> the results, we remove this record when importing puf, and then distribute
> the information in this record to all other records using our extrapolation

I don't know what "distribute" means in this context. I would have 
expected you to divide the aggregate values by the weight, and include the 
resulting single record.

> results (implemented in records.py, adding extrapolate_09_puf() function and
> removing the giant record).
> 
> As you can tell, two gzipped files for testing have also been replaced
> because of the different base file.
> 
> The goal to make our calculator to be able to accommodate both 08 PUF and 09
> PUF. Originally @MattHJensen (pls correct me if I misspeak about anything)
> and I thought our calculator could achieve this simply by reading the
> variables names, and checking whether certain variables exist or not,
> without adding a global variable indicating what year of PUF is used.
>

I do think it is very important to include a year variable in the data, so 
that different years can be concatenated when necessary. I know that we 
don't need this for the current project, but there will be times the data 
is read by stat packages or other programs and it will be convenient to 
have the year as part of the record for each taxpayer.

That said, I don't think our calculator needs to know which year the data 
come from provided it has been inflated to the proper level. If a variable 
is not present in every year, the calculator should be written so that 
having a zero value does not invalidate the result. Take for example the 
"foo" credit, which is present in the law and in the data 2008 and not in 
either in 2009. The dataset for both years should have a "foo" item, but 
equal to zero in 2009. Then when the calculator does 2008 tax it will 
include foo if present in the data and not if not. Of course that means 
that calculating 2009 tax from 2008 data will be inaccurate by the amount 
of omitted foo, but that is the best we can do.

> But right now we have an extra blow up for 09 PUF, and a new set of WEIGHTS,
> Stage I factors, and testing files. It seems to be pretty difficult not
> having a standalone variable for the year of PUF. No matter how we're going
> to implement this, here are several issues to keep in mind.
>
>  1. 09 PUF should only be turned on when running with 09 PUF.
>  2. 08 and 09 PUF have their own versions of WEIGHTS.csv and
>     StageIFactors.csv.

I wonder if we wouldn't be better off creating and storing the inflated 
datasets and then reading them for each run, rather than doing the 
inflation for each run. That is, have an earlier binding for inflation.

>  3. Some variables only exist in either 08 PUF or 09 PUF. This would
>     influence the initialization of Record, some functions in functions.py
>     and test_functions.py. test_all1991_puf.gz should have two versions for
>     the same reason.

All variables should exist in all years, but may be uniformly zero 
sometimes.

>  4. As 08 and 09 PUF contain essentially different data, exp_results.csv.gz
>     should have two versions.

Not sure what this is.

dan

Amy-Xu:
From Dan's comment:
`All variables should exist in all years, but may be uniformly zero
sometimes.`

@feenberg 09 PUF has e18400 which I believe doesn't exist in 08 PUF. e18425 and e18450 are also removed from 09 PUF. All three variables are for state and local taxes.

feenberg:
On Thu, 10 Sep 2015, Amy Xu wrote:

> 
> From Dan's comment:
> All variables should exist in all years, but may be uniformly zero
> sometimes.
> 
> @feenberg 09 PUF has e18400 which I believe doesn't exist in 08 PUF. e18425
> and e18450 are also removed from 09 PUF. All three variables are for state
> and local taxes.

How about:

    _sltax = max(e18400,e18450,e18425)

I think this works for either year of data and either year of law since 
the taxpayer is allowed to deduct the larger of income or sales tax but 
recently we are not provided with both. (I don't have the relevant docs 
available to me here, but I think this is right).

dan

Amy-Xu:
It seems there was a typo in our blow-up function that no one noticed before. Schedule E income (e02000) should be blown up based on whether this income variable is positive or not, instead of whether wage (e00200) was positive or not. After fixing this typo, the diagnostic table looks like this:

screen shot 2015-09-11 at 5 05 08 pm

Even though AGI is higher, but two versions (09 extrapolation prior taxcalc vs inside taxcalc) of diagnostic table become the same. I believe this is the right results under our current assumptions. @MattHJensen  @talumbau This PR is ready for review now.

MattHJensen:
Excellent fine, @amy-xu!

talumbau:
> @talumbau Aside from this refactoring, I also had a question on aliasing when testing the results with 09 PUF. In records.py, variable p23250 is aliased as e23250. But in the blow up function, the blow up factor for this variable is still applied to p23250 for all the projection years. Can e23250 actually get the blow-up as well through aliasing at initialization?

@Amy-Xu yes the attribute `e23250` on the records objects is a name for the same data with the attribute name `p23250` on the records objects. So, from this line of code onward:

https://github.com/Amy-Xu/Tax-Calculator/blob/09_PUF/taxcalc/records.py#L676

a modification of either attribute is equivalent, because it refers to the same data.

talumbau:
> 09 PUF extrapolation should only be turned on when running with 09 PUF.
> 08 and 09 PUF should have their own versions of WEIGHTS.csv and StageIFactors.csv.
> Some variables only exist in either 08 PUF or 09 PUF. This would influence the initialization of Record, some functions in functions.py and test_functions.py. test_all1991_puf.gz should have two versions for the same reason.
> As 08 and 09 PUF contain essentially different data, exp_results.csv.gz should have two versions.

So @Amy-Xu you are saying the goal is to support all of these features, right? What occurs to me is that we might want to create a separate class each for the 2008 PUF and the 2009 PUF. There are many, many ways to do this. I think it would definitely be easier using two classes than it would be to have a bunch of `if-else` checks inside the single Records class. Each class would have its own `blowup` method for example. One question about the two `PUF` files: since the two files contain a difference set of variables, how should this be handled in `functions.py`? If there is a reference to a variable that occurs in one but not the other, it would seem we would have to have a bunch of `if/else` checks OR we would have to guarantee that at the conclusion of initialization, any `Records` instance (or subclass) would have to have some set of attributes on it.  Is this the path you would like to go down? @MattHJensen is this along the lines of what you were thinking?

MattHJensen:
@talumbau, I'm not sure on the two classes concept...I'm thinking we should proceed as is for the sake of expediency -- which will break compatibility with the 08 puf in favor of the 09 puf -- and then open an issue to discuss how to implement that compatibility at a more leisurely pace. Delaying this discussion will also serve to give @martinholmer a chance to get better aquatinted with the records class and think about how it might need to be refactored for his ongoing project #291.

talumbau:
Ah OK now this PR makes more sense to me. So we are fine with breaking backwards compatibility and then adding it back in afterwards.

MattHJensen:
> So we are fine with breaking backwards compatibility and then adding it back in afterwards

Exactly

Amy-Xu:
@talumbau Thanks! That's what I thought as well for e23250. But then I realized it might not be what actually happened. As you can see here,

screen shot 2015-09-14 at 10 07 44 am

This is running the weighted sum of e23250 and p23250 right after initializing the calculator. Theoretically the data extrapolation has already been done inside `calc1 = Calculator(records = records1, params = params1)`, but it turns out the weighted sum of these two variables are quite different. What do you think is happening here?

talumbau:
@Amy-Xu , I took a closer look at what is happening with `e23250` and `p23250`. The problem occurs in line 209 of `records.py`. The line is in the `blowup` method:

```
    self.p23250 = self.p23250 * self.BF.ACGNS[self._current_year]
```

In this line, the Records object `self` gets a new value for the attribute named `p23250`. So, this "breaks" the aliasing of `p23250` and `e23250`. The idea is that you can modify the underlying numpy array that `e23250` and `p23250` both refer to, but if you do an assignment, you are saying "take this name and associate with the value on the right side of the equal sign". So, to preserve the aliasing, you could do this:

```
    self.p23250 *= self.BF.ACGNS[self._current_year]
```

I tried this in master and got as follows:

```
In [9]: calc1.records.e23250.sum()
Out[9]: 25119224529

In [10]: calc1.records.p23250.sum()
Out[10]: 25119224529

```

talumbau:
Just to give a complete answer here, my suggestion would be to refactor all of blowup to use the `*=` syntax for applicable lines. For the `np.where` statements, I would do as follows:

```
        self.e00900[:] = np.where(self.e00900 >= 0,
                               self.e00900 *
                               self.BF.ASCHCI[self._current_year],
                               self.e00900 *
                               self.BF.ASCHCL[self._current_year])
```

The `[:]` operator at the end basically says "all indices of this array". So we assign the values on the right into those indices, instead of assigning an attribute of name "eXXXX" to the value on the right. This will preserve any and all aliasing.

cc @MattHJensen let me know if you want me to do this in a PR.

Amy-Xu:
@talumbau So you mean by using just the equal sign, we actually assign a new address to p23250 while e23250 is still pointing to the old one? By using `*=`, we are modifying the value stored at the old location which p23250 and e23250 both point to? 

I can do this in this PR as well. Let me know what you prefer @MattHJensen

talumbau:
> @talumbau So you mean by using just the equal sign, we actually assign a new address to p23250 while e23250 is still pointing to the old one? By using *=, we are modifying the value stored at the old location which p23250 and e23250 both point to?

That is essentially correct. The only modification I would make is to say that by using the equal sign we are assigning a new *value* to `p23250` while `e23250` is still pointing to the old one. Everything else is exactly correct.

Amy-Xu:
@talumbau I just asked @MattHJensen, it seems he's fine with either way. I also have several other changes to make relevant to this aliases & blow up issue. How about I go ahead and modify this section as you suggested, and then you review it in the end?

talumbau:
sounds great to me.

Amy-Xu:
@talumbau I finished modifying this PR for the blow up section. Could you review the changes I made? The diagnostic table remain the same as the last one I post. This PR will be ready to merge in after review by you and @MattHJensen.

@talumbau There's one more thing to think about before this PR gets merged in. I'm playing with TaxBrain to identity weird behaviors in the webapp. To help you guys later debugging, I'm also generating 10-year revenue tables so you'll have a sort of "answer key" to compare the webapp results with. But as TaxBrain is currently using 08 PUF, I'll use 08 PUF for these tables as well. So the problem turns out that if this PR's merged in OSPC master, then OSPC master can not use 08 PUF anymore. Matt suggested to make a new branch in OSPC Tax-Calculator repo for 08 PUF and merge this 09_PUF into master. What do you think?

talumbau:
yes I think a branch on this repo is a good idea. Will review changes shortly.

talumbau:
this looks good but I would put in a test like this, because we have fixed this behavior and want it to stay fixed:

```
In [1]: import taxcalc

In [2]: import pandas as pd

In [3]: import numpy as np

In [4]: td = pd.read_csv("puf.csv")

In [5]: params1 = taxcalc.Parameters(start_year=2013)

In [6]: records1 = taxcalc.Records(td)

In [7]: calc1 = taxcalc.Calculator(records=records1, params=params1)
You loaded data for 2008.
Your data have beeen extrapolated to 2013.

In [9]: assert calc1.records.e23250.sum()==calc1.records.p23250.sum()

```

Thoughts?

Amy-Xu:
Yes! We definitely need a test for this. Do you mean something like [this](https://github.com/OpenSourcePolicyCenter/Tax-Calculator/commit/ba7474d251a01224b91ff5f3b7be0eb84604047a) in the test suite? I was wondering how to use 2008 puf directly for the tests? Or it's fine to use 1991 puf here?

Amy-Xu:
@talumbau just wondering whether you've had a chance to take a look at this? No rush tho

talumbau:
Yes, I reviewed the changes to `blowup` and the added test. Looks good! :+1:

MattHJensen:
Looks good to me. Thank you @Amy-Xu! :bow: Merging.

PR #350, opened 2015-09-11
fix and document the marginal tax rate function

MattHJensen:
The `mtr()` Calculator method was broken. This PR fixes it and adds a simple test and docstring. 

cc @martinholmer, @feenberg, and @talumbau. Feedback on the documentation and test is appreciated. 

I will be following this with PRs to clean up `behavior.py`. The goal is to get to the point in the next week to week and a half where we can incorporate some micro-feedback behavior in the webapp calculations.


MattHJensen:
Note that I also included 57b6984 in #350, the commit that gets this code working. I expect #350 to be merged before this one, and so this PR will focus primarily on documentation and tests.

martinholmer:
@MattHJensen, These revisions seem to compute a marginal federal income tax rate.  Wouldn't it be useful to also compute a marginal FICA tax rate?  Both of these marginal rates are in the Internet TAXSIM output as ovar[7] and ovar[9].  Given what you have done so far, it seems trivial to compute the marginal FICA tax rate for each filing unit.  And the marginal FICA tax rate is substantial for most people.

feenberg:
On Fri, 18 Sep 2015, Martin Holmer wrote:

> 
> @MattHJensen, These revisions seem to compute a marginal federal income tax rate.
> Wouldn't it be useful to also compute a marginal FICA tax rate? Both of these marginal
> rates are in the Internet TAXSIM output as ovar[7] and ovar[9]. Given what you have
> done so far, it seems trivial to compute the marginal FICA tax rate for each filing
> unit. And the marginal FICA tax rate is substantial for most people.

I assume we will be calculating marginal rates with a finite difference. 
That is, change the income a bit and find the difference in tax liability. 
Just using the bracket rates omits the effect of clawbacks, ceilings and 
floors.

In my experience there are some notches in the tax law where a small 
change in income generates a large change in tax. To avoid these I did at 
one time try to increase the span of the finite difference, but found that 
this was not sufficiently effective unless the span was made very large. 
So instead I reduced the span to a penny. Now I rarely run in to notches, 
and if I do the resulting rate is so ridiculous it is easily noticed and 
tossed out. (Actually, I calculate plus and minus a penny, and use the 
lessor rate, so I never see a wild rate). All my calculations are done in 
double precision, so the penny is sufficient to get accurate results.

Note that the combined rate is not the sum of the individual rates, but is 
calculated as a finite difference in tax over the finite difference in 
income.

This is an opportune time to bring up the calculation of the marginal tax 
rate. The rate we want is the "after-tax share of tax in compensation". 
Since only half of the social security tax is including in wages for 
income tax purposes, we need to increase the denominator by the excluded 
portion of FICA. That is, if t0 is the tax, and t1 is the tax we $1 more 
of w-2 earnings, then the marginal rate is:

     ((t1-t0)/(1.+.0765)

if the taxpayer is below the FICA maximum.

See page 55 here:

   http://www.nber.org/chapters/c6236.pdf

Dan

MattHJensen:
@martinholmer, thanks for the suggestion. 

@feenberg:
>(Actually, I calculate plus and minus a penny, and use the
lessor rate, so I never see a wild rate) 

Yes, I am following your example on this (using $1 as default rather than 1 penny for now -- this difference is a kwarg, though). 

In other words, we calculate IIT (individual income tax), add a dollar to income, recalculate IIT, and take the difference between the two IIT liabilities and then divide that difference by $1 (the default difference). We also do the same minus a dollar and take the lessor. 

To get the FICA MTR, I thought I would just replace IIT with FICA in the above sentence, where FICA includes both employer and employee shares.  

To get the combined IIT and FICA MTR, I thought I would just replace IIT with the sum IIT + FICA (again, with FICA including both employer and employee shares).  

Then to get the right after tax rate, I just take 1-MTR.

feenberg:
On Tue, 22 Sep 2015, Matt Jensen wrote:

> 
> @martinholmer, thanks for the suggestion.
> 
> @feenberg:
>
>       (Actually, I calculate plus and minus a penny, and use the
>       lessor rate, so I never see a wild rate)
> 
> Yes, I am following your example on this (using $1 as default rather than 1
> penny for now -- this difference is a kwarg, though).
> 
> In other words, we calculate IIT (individual income tax), add a dollar to
> income, recalculate IIT, and take the difference between the two IIT
> liabilities and then divide that difference by $1 (the default difference).
> We also do the same minus a dollar and take the lessor.

Did you see what I wrote about the divisor including the employer share of 
FICA?

dan

MattHJensen:
>Did you see what I wrote about the divisor including the employer share of
FICA? 

I did but I misinterpreted it. Now I understand and will add this.

MattHJensen:
@martinholmer and @feenberg,

I added the capability to calculate FICA mtr, IIT mtr, or FICA+IIT mtr. I also added the difference in employer-side payroll tax to the finite_difference for the denominator of the mtr calculation. It is now:
taxes_diff/(finite_diff+ 0.5 * FICA_diff)

I will calculate and post some summary statistics on the MTRs on the 09 puf tomorrow to make sure they are in the right ballpark.

MattHJensen:
Average combined MTR is 28%, unweighted after a few bugfixes. I'll give the weighted values and post a csv by wages tomorrow.

MattHJensen:
- mtr() now produces IIT, FICA, and combined MTRs with a single call. 

- I realized that the employer payroll tax adjustment to the denominator in the MTR calculation was incorrect. 

    It is now: 

        employer_fica_adjustment = np.where(self.records.e00200 <
                                            self.params.SS_Earnings_c,
                                            0.5 * self.params.FICA_ss_trt +
                                            0.5 * self.params.FICA_mc_trt,
                                            0.5 * self.params.FICA_mc_trt)

        mtr_combined = delta_combined_taxes / (finite_diff + employer_fica_adjustment)

- I also realized that the "down" version of the calculation, where we subtract the finite_difference from wages and then calculate taxes, doesn't make sense for records with wages <= finite_difference; so I fixed this as well.

MattHJensen:
Here are medians by _expanded income bin. 

Note that 0.153/1.0765  = .142127
and
0.038/1.0145 = 0.037457

Also note that the medians are taken independently across columns. 

```
                            mtr_FICA   mtr_IIT  mtr_combined
bins                                                        
(-100000000000000, 0]       0.142127  0.000000      0.142127
(0, 4999]                   0.142127 -0.071064      0.071064
(4999, 9999]                0.142127 -0.071064      0.071064
(9999, 14999]               0.142127  0.000000      0.142127
(14999, 19999]              0.142127  0.092894      0.235021
(19999, 24999]              0.142127  0.139340      0.281468
(24999, 29999]              0.142127  0.139340      0.281468
(29999, 39999]              0.142127  0.139340      0.281468
(39999, 49999]              0.142127  0.139340      0.281468
(49999, 74999]              0.142127  0.142127      0.284255
(74999, 99999]              0.142127  0.139340      0.281468
(99999, 199999]             0.142127  0.246427      0.324298
(199999, 499999]            0.037457  0.335042      0.382454
(499999, 999999]            0.037457  0.390340      0.439507
(999999, 1499999]           0.037457  0.402050      0.439507
(1499999, 1999999]          0.037457  0.402050      0.439507
(1999999, 4999999]          0.037457  0.402050      0.439507
(4999999, 9999999]          0.037457  0.402050      0.439507
(9999999, 100000000000000]  0.037457  0.402050      0.439507
```

@martinholmer, could you review this PR; I'd like to include these mtrs in testing? Currently this method does not mutate the calculator in any way, but one thing that we should store the mtrs somehow rather than just returning them.

martinholmer:
@MattHJensen,  This looks good to me.  The only problem I noticed is that the documenation of the mtr method is out of sync with the method definition, as can be seen at
https://github.com/MattHJensen/taxcalc/blob/mtr-update/taxcalc/calculate.py#L150-L175.

MattHJensen:
Thanks @martiholmer, I just updated the docstring. 

One thing that worries me is that I don't have an intuition for why the median IIT mtr jumps here: 

```
                            mtr_FICA   mtr_IIT  mtr_combined
bins                                                        
(499999, 999999]            0.037457  0.390340      0.439507
(999999, 1499999]           0.037457  0.402050      0.439507

```
This might be something best sorted out in testing rather than PR-review since it isn't clear whether the MTR method is at fault, the tax calculation is at fault, or my intuition is at fault.

martinholmer:
@MattHJensen said:
> Note that 0.153/1.0765 = .142127 and 0.038/1.0145 = 0.037457

If you are trying to compute the employee-share marginal tax rate, then shouldn't you be subtracting instead of dividing?

martinholmer:
@MattHJensen said:
> One thing that worries me is that I don't have an intuition for why the median IIT mtr jumps here:
> ```
>                             mtr_FICA   mtr_IIT  mtr_combined
>bins                                                        
>(499999, 999999]            0.037457  0.390340      0.439507
>(999999, 1499999]           0.037457  0.402050      0.439507
> ```
>
>This might be something best sorted out in testing rather than PR-review since it isn't clear whether the MTR method is at fault, the tax calculation is at fault, or my intuition is at fault.

I agree.  Wait until later to investigate this.

feenberg:
On Sat, 10 Oct 2015, Matt Jensen wrote:

> 
> Thanks @martiholmer, I just updated the docstring.
> 
> One thing that worries me is that I don't have an intuition for why the median IIT mtr
> jumps here:
>
>                             mtr_FICA   mtr_IIT  mtr_combined
> bins 
> (499999, 999999]            0.037457  0.390340      0.439507
> (999999, 1499999]           0.037457  0.402050      0.439507
> 
> This might be something best sorted out in testing rather than PR-review since it
> isn't clear whether the MTR method is at fault, the tax calculation is at fault, or my
> intuition is at fault.

It isn't always easy to determine why the mtr is changing. Does the mean 
change in a similar way?  Can you tab by taxable income?  By alternative 
minimum income? I wonder if there are some low taxable income folk in 
these income categories. If one of those makes the jump go away, that is 
an explanation.

You can calculate the mtr on regular tax separately, but that is probably 
not the cause. The next place I would look is the AMT and the AMT 
exclusion.

dan

feenberg:
On Sat, 10 Oct 2015, Martin Holmer wrote:

> 
> @MattHJensen said:
>
>       Note that 0.153/1.0765 = .142127 and 0.038/1.0145 = 0.037457
> 
> If you are trying to compute the employee-share marginal tax rate, then shouldn't you
> be subtracting instead of dividing?

The divison is there to make the mtr the marginal rate on the employer 
wage cost, which includes the other half of FICA, and that other half is 
not subject itself to FICA. That is, total compensation is

    wage_paid + employer_tax_on_wage_paid = wage + wage*.153/2

So if you add a dollar to wage, that adds 1.0765 to the employer wage bill 
that the mtr needs to include.

dan

martinholmer:
@MattHJensen,  The mtr method has a parameter called ```income_type_string``` that is used to set the variable ```income_type```, but then ```income_type``` is never used as far as I can see.  Using a change in ```e00200``` seems to be hardwired into the logic.  So, I guess, the logic should be generalized to use ```income_type``` or that parameter and its documentation should be dropped.  The code in question is at
https://github.com/MattHJensen/taxcalc/blob/mtr-update/taxcalc/calculate.py#L150-L172

By the way, the default option for calculating mtr in Internet TAXSIM is "Taxpayer Earnings", but "Spouse Earnings" is an option and so is "Wage Income" (which may be equivalent to e00200, but I don't really know).  And then there are other options for using changes in unearned income to compute mtr.

MattHJensen:
>The mtr method has a parameter called income_type_string that is used to set the variable income_type, but then income_type is never used as far as I can see. 

Thanks. I just added a commit that re-generalizes. Results are unchanged from what I reported before when e00200 is passed as income_type_string. 

Note that I also added a condition to the employer_fica_adjustment that income_type_string must be e00200 (or e00200s or e00200p). 

> By the way, the default option for calculating mtr in Internet TAXSIM is "Taxpayer Earnings", but "Spouse Earnings" is an option and so is "Wage Income" (which may be equivalent to e00200, but I don't really know). And then there are other options for using changes in unearned income to compute mtr.

We will have to get some guidance from @feenberg on the difference between "Taxpayer Earnings" and "Wage Income".

 I expect we will also want to use this function for some capital income items.

feenberg:
On Sat, 10 Oct 2015, Martin Holmer wrote:

> 
> @MattHJensen, The mtr method has a parameter called income_type_string that is used to
> set the variable income_type, but then income_type is never used as far as I can see.
> Using a change in e00200 seems to be hardwired into the logic. So, I guess, the logic
> should be generalized to use income_type or that parameter and its documentation
> should be dropped. The code in question is at
> https://github.com/MattHJensen/taxcalc/blob/mtr-update/taxcalc/calculate.py#L150-L172
> By the way, the default option for calculating mtr in Internet TAXSIM is "Taxpayer
> Earnings", but "Spouse Earnings" is an option and so is "Wage Income" (which may be
> equivalent to e00200, but I don't really know). And then there are other options for
> using changes in unearned income to compute mtr.

e00200 is the sum of taxpayer and spouse earnings, and those are treated 
the same in current law, but not in some proposals. There is also interest 
in the rate of tax on long term capital gains, and on itemized deductions.

dan

taxcalc/behavior.py
```diff
@@ -54,9 +54,9 @@ def behavior(calc_x, calc_y, update_income=update_income):
     """
 
     # Calculate marginal tax rates for plan x and plan y.
-    mtrX = calc_x.mtr('e00200')
+    mtrX = calc_x.mtr('e00200')[2]
```
talumbau: this seems a bit confusing. Based on the function signature of `mtr`, I would suggest:

```
fica, iit, combined = calc_x.mtr('e00200')
mtrX = combined
```

taxcalc/behavior.py
```diff
@@ -54,9 +54,9 @@ def behavior(calc_x, calc_y, update_income=update_income):
...
 
-    mtrY = calc_y.mtr('e00200')
+    mtrY = calc_y.mtr('e00200')[2]
```
talumbau: same comment as above.

MattHJensen:
Just made call to mtr() clearer based on @talumbau's suggestion.

MattHJensen:
@martinholmer, could you merge this if it looks good now?

PR #351, opened 2015-09-11
Simple bug fixes to get the behavior code working

MattHJensen:
There are really three issues with our behavioral calculations:

1. The code is not currently working. 
2.  The documentation and tests are poor. 
3. There is a chance that the calculation that is being done in `behavior.py` is not correct from an economic perspective -- i.e., that what claims to calculate the income effect and substitution effect does not actually calculate those concepts.

This PR  deals only with (1). 

Follow on PRs will deal with (2) and (3). 

This sequencing will allow our webapp backend folks to begin thinking about how to incorporate behavioral calculations into the webapp while I work on documentation and tests and while the taxcalc folk review those documentation and tests. The downside is that we will have poorly tested, poorly documented, and likely inaccurate code in master. Well, currently we have poorly tested, poorly documented, likely inaccurate, *and* definitely broken code in master. (All of this is my fault for not writing better code to begin with months ago). 

Another downside is that we won't be able to have a lengthy discussion over implementation of this behavioral feature in taxcalc before we start experimenting with use of its API in the webapp. This means that the webapp backend folks should give suggestions on the behavior API as they begin to use it, and it means that they should expect that it might change later on as the taxcalc people have more time to look over it. 

Usage of the behavior feature is demonstrated in `docs/notebooks/Behavioral_example.py`

cc @talumbau @feenberg @martinholmer


feenberg:
The thing to remember is that the published elasticities, typically in the 
range -.2 to -1.0 are the estimated "percentage change in taxable income 
for a percentage change in the after-tax share". So, for instance if the 
planX tax rate is .25, the planY rate is .15 then the after-tax share is 
.75. Wiith an elasticity of .5 you would want to impose a change in 
taxable income of:

                elasticity* ((t1-t0)/(1-t0))*taxable_income

or

                -.5*(.15-.25)/(1-.25)*taxable_income

This gives us a 13% change in after-tax share, or a 6.6% change in taxable 
income. I just add that to "other income" and "other deductions" in my 
work.

Now that has to be apportioned between AGI and itemized deductions. In the 
past I have allocated to deductions in proportion to the share of itemized 
deductions in AGI, but that is arbitrary and Matt may want to consult 
others for opinions.

The capital gains elasticity is calculated in the same way. If we have a 
separate cg elasticity, probably "taxable income" in the above calculation 
should be "non-gain taxable income".

Hope this clarifies things.

dan

MattHJensen:
@feenberg, thank you. We had discussed allowing the user to include an income effect as well. Do you still think that is best? If so, I have a particular question about that.

The income effect is calculated as:

inc_effect_param * (calcY._ospctax - calcX.ospctax). 

My question is, when we calculate the income effect should we use calcY._ospctax *before* or *after* updating income from the substitution effect. 

In other words should the workflow be: 

Scenario A. 

1. Calculate taxes for plan X and plan Y
2. Calculate substitution effect
3. Calculate income effect
4. Update the taxpayers' income for the combined income and substitution effects 
5. Recalculate taxes under plan Y. 

Or 

Scenario B.

1. Calculate taxes for plan X
2. Calculate substitution effect
3. Update taxpayers' income for the substitution effect
4. Calculate taxes for plan Y 
5. Calculate income effect
6. Update taxpayers' income for the income effect
7. Recalculate taxes under plan Y. 


A separate issue is how we are updating the taxpayers' income. You suggest adding it to other income and other deductions. Currently we are spreading it between *wages*, other income, and other deductions (other deductions only if the taxpayer is an itemizer):

```    
    delta_wages = (delta_inc * calcY.records.e00200 /
                   (calcY.records.c00100 + _itemized))

    other_inc = calcY.records.c00100 - calcY.records.e00200

    delta_other_inc = (delta_inc * other_inc /
                       (calcY.records.c00100 + _itemized))

    delta_itemized = (delta_inc * _itemized /
                      (calcY.records.c00100 + _itemized))

    calcY.records.e00200 = calcY.records.e00200 + delta_wages

    calcY.records.e00300 = calcY.records.e00300 + delta_other_inc

    calcY.records.e19570 = np.where(_itemized > 0,
                                    calcY.records.e19570 + delta_itemized, 
                                    0)

```

feenberg:
On Tue, 15 Sep 2015, Matt Jensen wrote:

> 
> @feenberg, thank you. We had discussed allowing the user to include an
> income effect as well. Do you still think that is best? If so, I have a
> particular question about that.
> 
> The income effect is calculated as:
> 
> inc_effect_param * (calcY._ospctax - calcX.ospctax).
> 
> My question is, when we calculate the income effect should we use
> calcY._ospctax before or after updating income from the substitution
> effect.

Use the "before" income, that is income at the default tax rates.

> 
> A separate issue is how we are updating the taxpayers' income. You suggest
> adding it to other income and other deductions. Currently we are spreading
> it between wages, other income, and other deductions (other deductions only
> if the taxpayer is an itemizer):
>

You are right. The change in income can be split between wages and other 
income. This is arbitrary and we should be alert for suggestions of a 
better method.

dan

MattHJensen:
@talumbau, the notebook example is now functioning properly. And thanks to @feenberg, I am much more confident that we are calculating the right thing now. Docstrings and tests coming in a separate PR.

talumbau:
Is it possible to put a basic call to the behavior function in the tests? We would just need one call to behavior similar to this:

```
# Call the behavioral effects calculator and create new Plan Y Calculator obects. 
calcY_behavior1 = behavior(calcX, calcY, elast_wrt_atr=0.4, inc_effect=0.15)
calcY_behavior2 = behavior(calcX, calcY, elast_wrt_atr=0.5, inc_effect=0.15)
calcY_behavior3 = behavior(calcX, calcY, elast_wrt_atr=0.4, inc_effect=0)          
```
from the notebook. The test data set has proven inadequate in certain ways but if we can at least call the code in a test, that would keep us from breaking the code in the future.

MattHJensen:
> if we can at least call the code in a test, that would keep us from breaking the code in the future.

@talumbau, thank you for the good advice. I just added a test to this PR for review.

MattHJensen:
@PeterDSteinberg, could you look over the architecture here with the goal in mind of incorporating a behavioral option into the webapp? Webapp users would choose the income and substitution effect parameter values on the input page and then we would display results from the behavior_plany_calculator (if we keep the current architecture).

PeterDSteinberg:
Sounds good @MattHJensen .  @talumbau  and I discussed yesterday the steps involved there.

MattHJensen:
@PeterDSteinberg, I feel remiss for not mentioning that there will likely be several other behavioral assumptions in the future (in addition to `inc_effect` and `elast_wrt_atr`), the values of which we will want both taxcalc and webapp users to be able to choose for themselves. Some of these "economic parameters" may vary by year or taxpayer characteristic. In other words, in the long run, it might make sense to have something similar to the `Parameters` class and `params.json`, except for economic parameters rather than for tax law parameters. I assume it is faster and more sensible for you to proceed in an ad hoc manner for now, without designing and building all of this machinery, but I thought this might be useful context nonetheless. 

cc @talumbau, @martinholmer

talumbau:
this looks good to me. +1

PR #352, opened 2015-09-14
Move names list out of read method into Records class scope.

martinholmer:
The Records class invites users to specify data as a Pandas DataFrame (as well as a csv file) in the class constructor, but does not provided users with a list of column names that the DataFrame is expected to have.  There is such a list; it is called names and is located in the read method where it is inaccessible to the public.  This is sort of a Catch-22 situation: users who want to construct a DataFrame don't have any access to what the Records class expects of the DataFrame.

This pull request makes no substantive change to the records.py code; it simply moves the names list out of the read() method, renames the list from names to NAMES, and changes one line in the read method so that that method can access the non-local NAMES list.

It would be highly desirable to implement this pull request before making all the PUF2008-to-PUF2009 changes to the Records class.


talumbau:
LGTM. This seems like a step down the path of writing a check in the form of "does this CSV file contain the necessary and sufficient columns for a Records instance?", which I think would be a good thing. +1.

MattHJensen:
Looks Good To Me as well. 

>This seems like a step down the path of writing a check in the form of "does this CSV file contain the necessary and sufficient columns for a Records instance?", which I think would be a good thing.

In the long run, it would be desirable for any column not present to be treated as zeroes. 

@Amy-Xu, could you take a look at this and merge if it looks good to you and your work on PUF2009?

Amy-Xu:
This looks good to me!

PR #353, opened 2015-09-15
added calc to ascii script

xiyuw123:
Takes calc object and converts suitable records into ascii format. Currently I'm not exactly sure how to write a test file for the script. It works when you use it in iPython though.


MattHJensen:
We have been sharing tax record information effectively lately. This PR was very helpful for thinking through those issues. Closing.

PR #355, opened 2015-09-18
Validation nodes

Amy-Xu:
In this PR, I added validation nodes for the parameters that we most likely need to validate. Other parameters do not have that node at this moment. @MattHJensen @zrisher Please let me know if this string format works or not.

### DIFF
--- a/taxcalc/params.json
+++ b/taxcalc/params.json
@@ -93,7 +93,7 @@
         "value":      [[1500, 1200, 1200, 1500, 1500, 1200 ],
 		               [1550, 1200, 1200, 1550, 1550, 1200],
 		               [1550, 1250, 1250, 1550, 1550, 1250]],
-	"validation": {"min": "default", "max": "null"}
+	"validations": {"min": "default"}
 
     },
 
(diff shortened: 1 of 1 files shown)


zrisher:
Hey @Amy-Xu, this is looking good, just a few changes I would make:

**"null" -> null**
`null` in json is represented simply as `null`, so we would use 
```
{ "max": null }
```
instead of 
```
`{ "max": "null" }` 
```
The latter is assigning to key `"max"` a string with the word "null" in it.

**Skip defining null validations**

I actually think we don't need to write out `min` or `max` if the value is `null`. We're already assuming that 

> if a parameter doesn't have a `min` and doesn't have a `max`, we don't need to validate it.

Generalizing that to 

> if a parameter has a validation key defined and assigned to a non-empty value, apply it

 I think will save us all some time down the road.

So instead of 

```
"validation": {"min": "default", "max": "null"}
```

we would write

```
"validation": {"min": "default"}
```

**Maybe: call it "validations" instead of "validation"?**
I believe standard practice for JSON is to pluralize the name of any key that maps to a collection of multiple objects of the same type. I could see the argument being made that `min`, `max`, and any other validations placed in here are really more properties than a collection of similar objects, since they're simple data types. I see them as connected enough that plural syntax makes the most sense to me, but that might be personal. Plus, if we ever adapt this to take one "validation" per column, it might be nice to have the keyword "validations" available so consumption of the old format can be gracefully deprecated.

I wouldn't worry about this last point for now. Easy to change later. Just the first two please, and I'll start working to use this.

Amy-Xu:
@zrisher Thanks for all the suggestions! I just removed the null validations and changed the name of 'validation' to 'validations'. Let me know if there's anything else needed to be changed after you start working.

zrisher:
Looks perfect to me.

MattHJensen:
LGTM too. @talumbau, feel free to merge this before you bump versions if it looks good to you. 

We can have a more extended discussion about which validations are needed later, and revise this list if necessary through follow-on PRs. I'm guessing it would be helpful for @zrisher's work this week to have some examples in master.

PR #356, opened 2015-09-18
params_to_json method on parameters

PeterDSteinberg:
There may be some subtleties I'm missing being new to the code.  Let me know if this is what we are thinking of for issue #284.

### DIFF
--- a/taxcalc/tests/test_parameters.py
+++ b/taxcalc/tests/test_parameters.py
@@ -455,3 +455,14 @@ def test_parameters_get_default_start_year():
     assert meta_ctc_c['start_year'] == 2015
     assert meta_ctc_c['row_label'] == ["2015", "2016", "2017", "2018"]
     assert meta_ctc_c['value'] == [1000, 1000, 1000, 500]
+
+
+def test_params_to_json():
+    p = Parameters()
+    vals_as_json = p.params_to_json()
+    temporary_file = tempfile.NamedTemporaryFile(delete=False)
+    with open(temporary_file.name,'w') as t:
+        t.write(vals_as_json)
+    p2 = Parameters(parameter_dict=json.load(open(temporary_file.name)))
+    assert json.loads(p2.params_to_json()) == json.loads(vals_as_json)
+    os.remove(temporary_file.name)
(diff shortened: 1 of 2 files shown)


taxcalc/tests/test_parameters.py
```diff
@@ -455,3 +455,14 @@ def test_parameters_get_default_start_year():
...
+
+def test_params_to_json():
+    p = Parameters()
+    vals_as_json = p.params_to_json()
+    temporary_file = tempfile.NamedTemporaryFile(delete=False)
+    with open(temporary_file.name, 'w') as t:
+        t.write(vals_as_json)
+    p2 = Parameters(parameter_dict=json.load(open(temporary_file.name)))
+    assert json.loads(p2.params_to_json()) == json.loads(vals_as_json)
+    os.remove(temporary_file.name)
```
talumbau: Hmm.. if we throw an exception somewhere in here before reaching the `os.remove`, we will end up polluting the file system. What do you think about using temporary file test fixtures?

talumbau: Also, I think we should have a test with `metadata=True` as an argument to default_data.

talumbau:
Here is a test fixture I am referring to:

https://github.com/OpenSourcePolicyCenter/Tax-Calculator/blob/master/taxcalc/tests/test_parameters.py#L17

MattHJensen:
@talumbau, do you remember the status of this one? I still think this would be helpful as described in issue #284.

talumbau:
Yes, this would need updates to the implementation now that we have Behavior and Growth parameters. The implementation would not be too difficult and would add utility to TaxBrain. We could provide a download of the raw JSON reform, or use such functionality to create a "download Jupyter notebook" which would have the JSON reform inside. Perhaps it would be nice to tie such a feature to the availability of the public data set. Once that is available, then someone could run TaxBrain, get a Jupyter notebook for their run, a link to the public data set, and then be able to run the exact reform on their machine using the notebook they download from the TaxBrain results page. That's sort of beyond the scope of this feature, but this feature will be needed to support that use case. I suppose one thing I'm saying is that I think it makes the most sense to tie this feature in with that bigger use case, rather than deliver this feature by itself. Thoughts?

martinholmer:
Pull request #356, which was submitted on September 18, 2015, has major conflicts with the master branch.  Given that work on Tax-Calculator has not been hindered over the past 17 months by the absence of this enhancement, it seems as if this pull request should be closed.  If a need for something like this arises in the future, it will be easy to add such a capability that fits with the current Tax-Calculator code structure.

PR #357, opened 2015-09-19
Revise test_blow_up so that blowup method is called

martinholmer:
Unless I am mistaken, the original test never executed the blowup method because the Records object passed to the Calculator __init__ method did not have a current_year value of 2009, which is required for the extrapolate_09_puf and increment_year Records methods (both of which call blowup) to be called.


MattHJensen:
@amy_xu, could you review this and merge if everything looks good?

Amy-Xu:
Ugh my bad. Thank you very much for pointing this out @martinholmer. LGTM, merging.

PR #361, opened 2015-09-23
Implement itemized deduction benefit surtax

Amy-Xu:
In this PR, I added some haircut parameters and an additional table-generator function to calculate and present the results of this specific type of reform. 

@talumbau @MattHJensen Haven't add any tests yet, but need your feedback on whether it's ok to implement it this way and whether we need to make it more generalized or not.

@talumbau This kind of reform is different from other reforms in that this one needs two calculators to generate results. Specifically, one calculator carries the default plan results, and the other calculates results with certain parameters set up in a certain way. Then, get the difference of tax from these two calculators, truncate the difference in a certain way and add the truncated diff to the second one to get the final results.

As you can see, this certain type of reform, involving two calculators, is a bit difficult for our current taxcalc and drop q. What I came up with is to use a parameter as a switch to add functionality to our table generator and drop q. This switch is the cap rate of this reform. If this cap rate is not zero, then we can add the truncated diff. 

To play with this type of reform in notebooks, users can use the `diagnostic_table_special_reform()` to view results. This function takes the default plan calculator as an argument.
screen shot 2015-09-23 at 5 36 57 pm

In drop q, similarly I added a few lines both after initializing calc3 and during the process calculating later year results:
```
    # User specified Plans
    params3 = Parameters(start_year=2013)
    params3.implement_reform(reform)
    # Create a Calculator for the user specified plan
    calc3 = Calculator(params=params3, records=records3)
    calc3.increment_year()
    calc3.increment_year()
    assert calc3.current_year == start_year
    calc3.calc_all()
    
    if calc3.params.ID_DeductionBenefit_crt != 0:
        tax_diff = np.where(calc3.records._ospctax - calc1.records._ospctax > 0,
                            calc3.records._ospctax - calc1.records._ospctax, 0)
        cap = calc3.params.ID_DeductionBenefit_crt * calc3.records.c00100
        calc3.records._ospctax = np.where(tax_diff < cap,
                                         tax_diff + calc3.records._ospctax,
                                         cap + calc3.records._ospctax)
```

Both pure taxcalc and taxcalc with drop q are giving me consistent results under this set up. @MattHJensen The results seem fine to me, but not quite sure how to double check. Could you take a look at it?

Default Plan (calc1):
screen shot 2015-09-23 at 5 40 23 pm

Certain Itemized Deduction eliminated starting in 2015 (calc2):
screen shot 2015-09-23 at 5 41 16 pm

Added the diff (calc2):
screen shot 2015-09-23 at 5 42 16 pm

### DIFF
--- a/taxcalc/functions.py
+++ b/taxcalc/functions.py
@@ -1744,5 +1744,5 @@ def BenefitSurtax(calc):
         calc.records._surtax = np.where(tax_diff < surtax_cap,
                                         tax_diff,
                                         surtax_cap)
-        
+
         calc.records._ospctax += calc.records._surtax
--- a/taxcalc/functions.py
+++ b/taxcalc/functions.py
@@ -168,9 +168,11 @@ def AGI(_ymod1, c02500, c02700, e02615, c02900, e00100, e02500, XTOT,
 def ItemDed(_posagi, e17500, e18400, e18500, e18800, e18900,
             e20500, e20400, e19200, e20550, e20600, e20950, e19500, e19570,
             e19400, e19550, e19800, e20100, e20200, e20900, e21000, e21010,
-            MARS, _sep, c00100, ID_ps, ID_Medical_frt, ID_Casualty_frt,
-            ID_Miscellaneous_frt, ID_Charity_crt_Cash, ID_Charity_crt_Asset,
-            ID_prt, ID_crt, ID_StateLocalTax_HC, ID_Charity_frt, puf):
+            MARS, _sep, c00100, ID_ps, ID_Medical_frt, ID_Medical_HC,
+            ID_Casualty_frt, ID_Casualty_HC, ID_Miscellaneous_frt,
+            ID_Miscellaneous_HC, ID_Charity_crt_Cash, ID_Charity_crt_Asset,
+            ID_prt, ID_crt, ID_StateLocalTax_HC, ID_Charity_frt,
+            ID_Charity_HC, ID_Mortgage_HC, puf):
 
     """
     Itemized Deduction; Form 1040, Schedule A
@@ -253,7 +255,7 @@ def ItemDed(_posagi, e17500, e18400, e18500, e18800, e18900,
     # Other Taxes (including state and local)
     c18300 = _statax + e18500 + e18800 + e18900
 
-    # Casulty
+    # Casualty
     if e20500 > 0:
         c37703 = e20500 + ID_Casualty_frt * _posagi
         c20500 = c37703 - ID_Casualty_frt * _posagi
@@ -289,14 +291,14 @@ def ItemDed(_posagi, e17500, e18400, e18500, e18800, e18900,
 
     # Gross Itemized Deductions
 
-    c21060 = (e20900 + c17000 + (1 - ID_StateLocalTax_HC) * c18300 + c19200 +
-              c19700 + c20500 + c20800 + e21000 + e21010)
+    c21060 = (e20900 + (1- ID_Medical_HC) * c17000 +
+             (1 - ID_StateLocalTax_HC) * c18300 + (1 - ID_Mortgage_HC) * c19200 +
+             (1 - ID_Casualty_HC) * c19700 + (1 - ID_Casualty_HC) * c20500 +
+             (1 - ID_Miscellaneous_HC) * c20800 + e21000 + e21010)
 
     # Limitations on deductions excluding medical, charity etc
     _phase2_i = ID_ps[MARS - 1]
-
     _nonlimited = c17000 + c20500 + e19570 + e21010 + e20900
-    _phase2_i = ID_ps[MARS - 1]
     _limitratio = _phase2_i / _sep
 
     if c21060 > _nonlimited and c00100 > _limitratio:
--- a/taxcalc/functions.py
+++ b/taxcalc/functions.py
@@ -1713,23 +1713,23 @@ def ExpandIncome(FICA_ss_trt, SS_Earnings_c, e00200, FICA_mc_trt, e02400,
     return (_expanded_income)
 
 
-def BenefitCap(calc):
-    if calc.params.ID_DeductionBenefit_crt != 1:
+def BenefitSurtax(calc):
+    if calc.params.ID_BenefitSurtax_trt != 1:
         nobenefits_calc = copy.deepcopy(calc)
 
         # hard code the reform
         nobenefits_calc.params.ID_Medical_HC = \
-            int(nobenefits_calc.params.ID_BenefitCap_Switch[0])
+            int(nobenefits_calc.params.ID_BenefitSurtax_Switch[0])
         nobenefits_calc.params.ID_StateLocalTax_HC = \
-            int(nobenefits_calc.params.ID_BenefitCap_Switch[1])
+            int(nobenefits_calc.params.ID_BenefitSurtax_Switch[1])
         nobenefits_calc.params.ID_casualty_HC = \
-            int(nobenefits_calc.params.ID_BenefitCap_Switch[2])
+            int(nobenefits_calc.params.ID_BenefitSurtax_Switch[2])
         nobenefits_calc.params.ID_Miscellaneous_HC = \
-            int(nobenefits_calc.params.ID_BenefitCap_Switch[3])
+            int(nobenefits_calc.params.ID_BenefitSurtax_Switch[3])
         nobenefits_calc.params.ID_Mortgage_HC = \
-            int(nobenefits_calc.params.ID_BenefitCap_Switch[4])
+            int(nobenefits_calc.params.ID_BenefitSurtax_Switch[4])
         nobenefits_calc.params.ID_Charity_HC = \
-            int(nobenefits_calc.params.ID_BenefitCap_Switch[5])
+            int(nobenefits_calc.params.ID_BenefitSurtax_Switch[5])
 
         nobenefits_calc.calc_one_year()
 
@@ -1738,9 +1738,11 @@ def BenefitCap(calc):
                             nobenefits_calc.records._ospctax -
                             calc.records._ospctax, 0)
 
-        cap = nobenefits_calc.params.ID_DeductionBenefit_crt *\
+        surtax_cap = nobenefits_calc.params.ID_BenefitSurtax_trt *\
             nobenefits_calc.records.c00100
 
-        calc.records._ospctax = np.where(tax_diff < cap,
-                                         tax_diff + calc.records._ospctax,
-                                         cap + calc.records._ospctax)
+        calc.records._surtax = np.where(tax_diff < surtax_cap,
+                                        tax_diff,
+                                        surtax_cap)
+        
+        calc.records._ospctax += calc.records._surtax
--- a/taxcalc/functions.py
+++ b/taxcalc/functions.py
@@ -291,10 +291,13 @@ def ItemDed(_posagi, e17500, e18400, e18500, e18800, e18900,
 
     # Gross Itemized Deductions
 
-    c21060 = (e20900 + (1- ID_Medical_HC) * c17000 +
-             (1 - ID_StateLocalTax_HC) * c18300 + (1 - ID_Mortgage_HC) * c19200 +
-             (1 - ID_Casualty_HC) * c19700 + (1 - ID_Casualty_HC) * c20500 +
-             (1 - ID_Miscellaneous_HC) * c20800 + e21000 + e21010)
+    c21060 = (e20900 + (1 - ID_Medical_HC) * c17000 +
+              (1 - ID_StateLocalTax_HC) * c18300 +
+              (1 - ID_Mortgage_HC) * c19200 +
+              (1 - ID_Casualty_HC) * c19700 +
+              (1 - ID_Casualty_HC) * c20500 +
+              (1 - ID_Miscellaneous_HC) * c20800 +
+              e21000 + e21010)
 
     # Limitations on deductions excluding medical, charity etc
     _phase2_i = ID_ps[MARS - 1]
--- a/taxcalc/functions.py
+++ b/taxcalc/functions.py
@@ -4,7 +4,7 @@ from pandas import DataFrame
 import math
 import numpy as np
 from .decorators import *
-
+import copy
 
 @iterate_jit(nopython=True)
 def FilingStatus(MARS):
@@ -1710,3 +1710,25 @@ def ExpandIncome(FICA_ss_trt, SS_Earnings_c, e00200, FICA_mc_trt, e02400,
                         employer_share_fica)
 
     return (_expanded_income)
+
+
+def BenefitCap(calc):
+    if calc.params.ID_DeductionBenefit_crt != 0:
+        nobenefits_calc = copy.deepcopy(calc)
+        
+        # hard code the reform
+        nobenefits_calc.params.ID_Medical_HC = nobenefits_calc.params.ID_BenefitCap_Switch[0]
+        nobenefits_calc.params.ID_StateLocalTax_HC = nobenefits_calc.params.ID_BenefitCap_Switch[1]
+        nobenefits_calc.params.ID_casualty_HC = nobenefits_calc.params.ID_BenefitCap_Switch[2]
+        nobenefits_calc.params.ID_Miscellaneous_HC = nobenefits_calc.params.ID_BenefitCap_Switch[3]
+        nobenefits_calc.params.ID_Mortgage_HC = nobenefits_calc.params.ID_BenefitCap_Switch[4]
+        nobenefits_calc.params.ID_Charity_HC = nobenefits_calc.params.ID_BenefitCap_Switch[5]
+        
+        nobenefits_calc.calc_one_year()
+        
+        tax_diff = np.where(nobenefits_calc.records._ospctax - calc.records._ospctax > 0, 
+                            nobenefits_calc.records._ospctax - calc.records._ospctax, 0)
+        cap = nobenefits_calc.params.ID_DeductionBenefit_crt * nobenefits_calc.records.c00100
+        calc.records._ospctax =  np.where(tax_diff < cap, 
+                                          tax_diff + calc.records._ospctax, 
+                                          cap + calc.records._ospctax)
--- a/taxcalc/functions.py
+++ b/taxcalc/functions.py
@@ -1719,18 +1719,19 @@ def BenefitCap(calc):
 
         # hard code the reform
         nobenefits_calc.params.ID_Medical_HC = \
-            nobenefits_calc.params.ID_BenefitCap_Switch[0]
+            int(nobenefits_calc.params.ID_BenefitCap_Switch[0])
         nobenefits_calc.params.ID_StateLocalTax_HC = \
-            nobenefits_calc.params.ID_BenefitCap_Switch[1]
+            int(nobenefits_calc.params.ID_BenefitCap_Switch[1])
         nobenefits_calc.params.ID_casualty_HC = \
-            nobenefits_calc.params.ID_BenefitCap_Switch[2]
+            int(nobenefits_calc.params.ID_BenefitCap_Switch[2])
         nobenefits_calc.params.ID_Miscellaneous_HC = \
-            nobenefits_calc.params.ID_BenefitCap_Switch[3]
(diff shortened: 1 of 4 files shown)


MattHJensen:
@amy-xu what do you think about renaming `calc_all` to `calc_all_once()`. Then a new `calc_all()` function would call `calc_all_once()` from `calculate.py` and it would call a new function `BenefitCap()` from `functions.py`. In `functions.py`, `BenefitCap()` would, among other things, call `calc_all_once()` if the first value of `params.ID_DeductionBenefit_crt == 1`. The second through Nth values of params.ID_DeductionBenefit_crt would be switches and each would correspond to an itemized deduction to indicate whether that itemized deduction is included in the cap or not. These would be used to conduct the logic of the cap inside the `BenefitCap()` function of `functions.py`. 

cc @talumbau

MattHJensen:
I suppose if we did something similar to my comment above we'd probably want a separate params attribute than ID_DeductionBenefit_crt for the switches, since that would be used for the cap rate.

Amy-Xu:
Have a few questions for this:
1. Where does the default plan calculation come in? And how do you want to save the results for later diff calculations?
2. What exactly does BenefitCap() do?
3. What's the advantage of doing it this way?

talumbau:
@MattHJensen I had a similar thought to this when you told me about the behavioral calculations. I had thought of changing the method name to `calc_one_year` or perhaps something even more generic like `_execute`. My idea was slightly different, but overall the idea was that the new `calc_all` could check for the condition to do the extra processing and call `calc_one_year` as necessary. To answer @Amy-Xu 's questions, my understanding is that the answers would be as follows:

1. The default plan calculation remains unchanged and would happen in its own Calculator instance

2. Matt would give more detail here, but it sounds like the answer is "test for the conditions that would enable this new calculation and then perform the calculation by copying self, doing `calc_one_year` on both, and then performing the operations you describe in this PR."

3. the benefit would be to have a consistent interface in how to calculate a reform even in the case of this itemized deduction benefit cap.

although I am essentially answering a question that I was not asked, but I am interested to see if I understand Matt's idea correctly.

MattHJensen:
@amy-xu I haven't tried any of this yet, so view it as buggy pseudocode.

I am thinking `calc_all()` in `calculate.py` would look like this:

```
def calc_all(self):
    calc_one_year(self)
    BenefitCap(self)
```

I am thinking `BenefitCap()` in `functions.py` would look something like this. 

```
def BenefitCap(calc):
   if calc.params.ID_DeductionBenefit_crt != 0: 
       nobenefits_calc = copy.hardcopy(calc)
       # might need getattrs here...
       reform = {calc.current_year: 
                      {"ID_Medical_HC": calc.parameters.ID_DeductionBenefit_Switches[0],
                      "ID_Casualty_HC": calc.parameters.ID_DeductionBenefit_Switches[1],
                       etc. }} 
       nobenefits_calc.implement_reform(reform)
       nobenefits_calc.calc_one_year()
       tax_diff = np.where(nonbenefits_calc.records._ospctax - calc.records._ospctax > 0, 
                                       nobenefits_calc.records._ospctax - calc.records._ospctax, 
                                       0)
       cap = calc.params.ID_DeductionBenefit_crt * calc.records.c00100
       calc.records._ospctax =  np.where(tax_diff < cap, 
                                                               tax_diff + calc.records._ospctax, 
                                                               cap + calc.records._ospctax)
       return calc
       

```

As @talumbau mentioned, I think the benefit of this is that we could apply the cap with the same interface as we apply all other reforms, without needing all of the built in machinery in dropq or the diagnostic_table. 

One concern here is that if you view a diagnostic table, we don't actually change the value of c04470. I wonder whether we should add this 'surtax' value someone other than _ospctax. I'll take a look at the SAS code.

Amy-Xu:
Ok, this is much clearer to me. I guess my first question is, for the the calc argument in BenefitCap(), do you mean it's the `self` or an exogenous calculator as I set up in the notebook? If it's the first case, after the reform implemented, the self would be changed. Then for all the calculations for later years, deep copy doesn't work anymore. But we still need this default plan calculator for every year afterward. If it's the second case, do we really want to pass a calculator to a function in functions.py?

talumbau:
@Amy-Xu, yes this is where my idea was different from Matt's. I don't think it would be best to pass the instance of `calc` to a function in `functions.py` for two reasons:

- it would be an outlier from the other functions in that file that are jitted and compiled in `nopython` mode so we get native performance. If we pass a Python object to a function, we can't compile it in `nopython` mode, so we would not get speedup from it

- the fact that we are passing in just a single `Calculator` instance as an argument to the function indicates that it might be best to formulate the function as a method on the `Calculator` class. It would not be jitted, so we would have to understand the performance implications of such a decision.

MattHJensen:
Quick reply from my phone:

Amy, yes, would be passing self. I don't see why this would be a problem in later years though. You would just increment your calc object like always.

On Sep 24, 2015, at 12:08 PM, T.J. Alumbaugh <notifications@github.com<mailto:notifications@github.com>> wrote:


@Amy-Xu<https://github.com/Amy-Xu>, yes this is where my idea was different from Matt's. I don't think it would be best to pass the instance of calc to a function in functions.py for two reasons:

@talumbau, could you give a longer description of what you are thinking of?

  *   it would be an outlier from the other functions in that file that are jitted and compiled in nopython mode so we get native performance. If we pass a Python object to a function, we can't compile it in nopython mode, so we would not get speedup from it

The main action though is the calls to calc_one_year, which would be jitted. I think the bigger issue than performance would be needing to have the numpy logic, which is not easy to read.

  *   the fact that we are passing in just a single Calculator instance as an argument to the function indicates that it might be best to formulate the function as a method on the Calculator class. It would not be jitted, so we would have to understand the performance implications of such a decision.

I put it in functions.py rather than as a calculator method since it is tax logic, all of which resides in functions.py right now.

Amy-Xu:
Regarding this comment from @MattHJensen:
>Amy, yes, would be passing self. I don't see why this would be a problem in later years though. You would just increment your calc object like always.

I see where the confusion is coming from. In your pseudocode, you're adding it to the `self` calc, but I think it's probably a typo. You want to add the diff to the `nobenefits_calc` and return the `nobenefits_calc` instead the original calc, which from my perspective will change the default calc to a calc with reforms parameters. If this BenefitCap() function is a attribute of the `self` calc, I don't think this is the optimal way to do it.

MattHJensen:
fyi, @talumbau, Amy and I sorted out the above on flowdock. `tax_diff` should be added to the original calculator as in my example above, and we should be returning that same calculator (mutated now).....assuming we are to implement things this way. Looking forward to hearing your ideas, or any other ideas from @Amy-Xu.

Amy-Xu:
I take back the previous comment. Sorry about the confusion. I think @MattHJensen is right on the pseudocode. And that's probably the reason why our itemized deduction was much lower when it shouldn't be. 

For @talumbau arguments, my understanding is
> it would be an outlier from the other functions in that file that are jitted and compiled in nopython mode so we get native performance. If we pass a Python object to a function, we can't compile it in nopython mode, so we would not get speedup from it

BenefitCap(), comparing to other functions in functions.py, takes in a whole calculator as an argument instead of parameters or PUF variables. In that sense it's a outlier.

>the fact that we are passing in just a single Calculator instance as an argument to the function indicates that it might be best to formulate the function as a method on the Calculator class. It would not be jitted, so we would have to understand the performance implications of such a decision.

I also want to ask, does this mean to add this as attribute function, or a standalone function like the diagnostic table?

@MattHJensen I don't think this is a pure tax logic function tho. It looks more like a reform.

talumbau:
it seems to me that this could be either a freestanding function or a method on `Calculator`. I don't have a strong opinion either way. As a method, it would probably look like:

```
def benefitCap(self):
   if self.params.ID_DeductionBenefit_crt != 1: 
       nobenefits_calc = copy.deepcopy(self)
       # might need getattrs here...
       reform = {self.current_year: 
                      {"ID_Medical_HC": self.parameters.ID_DeductionBenefit_Switches[0],
                      "ID_Casualty_HC": self.parameters.ID_DeductionBenefit_Switches[1],
                       etc. }} 
       nobenefits_calc.implement_reform(reform)
       nobenefits_calc.calc_one_year()
       tax_diff = np.where(nonbenefits_calc.records._ospctax - self.records._ospctax > 0, 
                                       nobenefits_calc.records._ospctax - self.records._ospctax, 
                                       0)
       cap = self.params.ID_DeductionBenefit_crt * self.records.c00100
       self.records._ospctax =  np.where(tax_diff < cap, 
                                                               tax_diff + self.records._ospctax, 
                                                               cap + self.records._ospctax)
```

I don't return `self`, since it would not be necessary, as one would likely call this as:

```
calc.benefitsCap()
```

or perhaps `self.benefitsCap()` if we already in a method of Calculator. In either case, there is not much benefit to be gained from using `numba` here, since we already are using `np.where` and numpy arithmetic for the computation here.

MattHJensen:
I'm partial to keeping it as a function in functions.py for now and then revisiting that decision later. I'd also like to think about whether there are any ways to refactor away from using numpy arithmetic and np.where...

The reason this should be in functions.py is that it can be viewed as tax logic with self.params.ID_DeductionBenefit_crt == 1 under current law. 

@amy-xu, could you take a shot at implementing something along these lines?

MattHJensen:
We should probably call this something like 'BenefitCapSurtax' and calculate the surtax amount as a separate variable before adding it to _ospctax so that we can report it separately. 

I am thinking about how to handle the following concern I raised earlier. We can discuss this with @feenberg next Monday in DC as well. 

>One concern here is that if you view a diagnostic table you will still see itemized deductions at their pre-cap amount since we don't actually change the value of c04470. I wonder whether we should add this 'surtax' value someone other than _ospctax. I'll take a look at the SAS code.

Amy-Xu:
@MattHJensen Here's a preliminary version of your proposal. Didn't fix the tests, but the calculator works. The default reform inside ID_BenefitCap_Switch is set up as I demonstrated in notebook in is PR. 

No cap rate setup:
screen shot 2015-09-24 at 7 23 05 pm

Cap Rate == 0.02
screen shot 2015-09-24 at 7 23 46 pm

Could you review? 

@talumbau We need to come back and make a decision later. The code looks very ugly right now.

taxcalc/params.json
```diff
@@ -393,6 +393,19 @@
...
+	"description": "This parameter is specified by items and year. ",
+        "irs_ref": "",
+        "notes": "",
+        "start_year": 2013,
+        "col_var": "MARS",
+        "row_var": "FLPDYR",
+        "row_label": ["2013"],
+        "cpi_inflated": true,
+        "col_label": ["Medical", "StateLocal", "Casualty", "Miscellaneous", "Interest", "Charity"],        
+        "value":    [[1, 	   0,             1,             1,             1,       0]]
```
MattHJensen: Let's make all of these 1 by default.

taxcalc/params.json
```diff
@@ -930,6 +995,32 @@
...
+	"description": "This percentage is used to cap the tax benefit from deductions. Users can specify what deductions they want to apply to. ",
+        "irs_ref": "",
+        "notes": "",
+        "start_year": 2013,
+        "col_var": "",
+        "row_var": "",
+        "row_label": ["2013"],
+        "cpi_inflated": false,
+        "col_label": "",
+        "value":     [0.0]
```
MattHJensen: This should be 1 in current law. Well, actually I'm interested to know whether there is any difference if this is 1 or 1e309.

taxcalc/params.json
```diff
@@ -393,6 +393,19 @@
...
+	"description": "This parameter is specified by items and year. ",
+        "irs_ref": "",
+        "notes": "",
+        "start_year": 2013,
+        "col_var": "MARS",
+        "row_var": "FLPDYR",
+        "row_label": ["2013"],
+        "cpi_inflated": true,
+        "col_label": ["Medical", "StateLocal", "Casualty", "Miscellaneous", "Interest", "Charity"],        
+        "value":    [[1, 	   1,             1,             1,             1,         1]]
```
MattHJensen: @talumbau mentioned on the phone that maybe these should be booleans. That makes sense to me too.

MattHJensen: Oh, I see why that might be hard. It would require more logic here:

```
nobenefits_calc.params.ID_casualty_HC = \
nobenefits_calc.params.ID_BenefitCap_Switch[2]
```

MattHJensen: pretty easy to solve, though:  
```
nobenefits_calc.params.ID_casualty_HC = \
int(nobenefits_calc.params.ID_BenefitCap_Switch[2])
```

Amy-Xu:
@talumbau I finished revising the everything but tests. I'm not quite sure whether I need to add separate test for `calc_one_year()` as we already have tests for calc_all(). Could you review the code and give me some suggestions on tests if we need any?

MattHJensen:
After discussing with @feenberg today, let's change everything from BenefitsCap to BenefitsSurtax and then add a Misc. Surtax column (with this included) to our diagnostic tables on the webapp.

Amy-Xu:
@MattHJensen So how about our diagnostic table in taxcalc? Are we leaving it as now or add that column?

MattHJensen:
>@MattHJensen So how about our diagnostic table in taxcalc? Are we leaving it as now or add that column?

let's add it.

Amy-Xu:
@MattHJensen Changed cap to surtax and added this surtax column to the diagnostic table. With surtax rate of 0.02, and the default switch, the table looks like:
screen shot 2015-09-30 at 4 59 22 pm

taxcalc/params.json
```diff
@@ -930,6 +995,32 @@
...
+        "start_year": 2013,
+        "col_var": "",
+        "row_var": "",
+        "row_label": ["2013"],
+        "cpi_inflated": false,
+        "col_label": "",
+        "value":     [0.0] 
+    },
+    "_ID_BenefitSurtax_trt":{
+        "long_name": "Deduction tax benefit surtax rate, ceiling as a percent of AGI",
```
MattHJensen: Let's have one 
_BenefitSurtax_crt:{
"long_name": Credit on itemized deduction benefit surtax (percent of AGI)
"description": The surtax on specified itemized deductions applies to benefits in excess of this percent of AGI. 

and one 
_BenefitSurtax_trt:{
"long_name": Surtax rate on the benefits from specified itemized deductions
"description": The benefit from specified itemized deductions exceeding the credit is taxed at this rate. A surtax rate of 1 strictly limits the benefit from specified itemized deductions to the specified credit.

taxcalc/functions.py
```diff
@@ -1705,3 +1711,38 @@ def ExpandIncome(FICA_ss_trt, SS_Earnings_c, e00200, FICA_mc_trt, e02400,
...
+        nobenefits_calc.calc_one_year()
+
+        tax_diff = np.where(nobenefits_calc.records._ospctax -
+                            calc.records._ospctax > 0,
+                            nobenefits_calc.records._ospctax -
+                            calc.records._ospctax, 0)
+
+        surtax_cap = nobenefits_calc.params.ID_BenefitSurtax_trt *\
+            nobenefits_calc.records.c00100
+
```
MattHJensen: Just leaving a note for us, @amy-xu, this should be:

calc.records._surtax = np.where(tax_diff  >  surtax_cap,
                                                     tax_diff - surtax_cap,
                                                      0)

Amy-Xu:
@MattHJensen could you review?

MattHJensen:
Will do. Could you paste in another diagnostic table?

Amy-Xu:
Diagnostic table after the tax logic correction:
screen shot 2015-10-05 at 5 27 35 pm

MattHJensen:
Also, could you add this wording for the new parameter descriptive and name attributes, (or anything better) to params.json? 

_BenefitSurtax_crt:{
"long_name": Credit on itemized deduction benefit surtax (percent of AGI)
"description": The surtax on specified itemized deductions applies to benefits in excess of this percent of AGI. In http://www.nber.org/papers/w16921, Feldstein, Feenberg, and MacGuineas propose a credit of 2% of AGI against a 100% tax rate; in their proposal, however, a broader set of tax benefits, including the employer provided health exclusion, would be taxed.


_BenefitSurtax_trt:{
"long_name": Surtax rate on the benefits from specified itemized deductions
"description": The benefit from specified itemized deductions exceeding the credit is taxed at this rate. A surtax rate of 1 strictly limits the benefit from specified itemized deductions to the specified credit. In http://www.nber.org/papers/w16921, Feldstein, Feenberg, and MacGuineas propose a credit of 2% of AGI against a 100% tax rate; in their proposal, however, a broader set of tax benefits, including the employer provided health exclusion, would be taxed.

cc @feenberg

Amy-Xu:
Just added @MattHJensen.

taxcalc/functions.py
```diff
@@ -289,14 +292,17 @@ def ItemDed(_posagi, e17500, e18400, e18500, e18800, e18900,
 
     # Gross Itemized Deductions
 
-    c21060 = (e20900 + c17000 + (1 - ID_StateLocalTax_HC) * c18300 + c19200 +
-              c19700 + c20500 + c20800 + e21000 + e21010)
+    c21060 = (e20900 + (1 - ID_Medical_HC) * c17000 +
+              (1 - ID_StateLocalTax_HC) * c18300 +
+              (1 - ID_Mortgage_HC) * c19200 +
+              (1 - ID_Casualty_HC) * c19700 +
```
MattHJensen: Should this one be ID_Charity_HC rather than ID_Casualty_HC?  Applying ID_Casualty_HC to c20500 seems right.

taxcalc/functions.py
```diff
@@ -289,14 +292,17 @@ def ItemDed(_posagi, e17500, e18400, e18500, e18800, e18900,
...
+              (1 - ID_Casualty_HC) * c20500 +
+              (1 - ID_Miscellaneous_HC) * c20800 +
```
MattHJensen: Should ID_Miscellaneous_HC also be applied to c20800 in the calculation of AMTI (like what we do with state and local?) Any others? 

cc @GoFroggyRun.

taxcalc/params.json
```diff
@@ -930,6 +995,45 @@
         "col_label": "",
         "value": [0]
     },
+     "_ID_Mortgage_HC":{
```
MattHJensen: Should this be _ID_InterestPaid_HC rather than _ID_Mortgage_HC?

taxcalc/functions.py
MattHJensen: ^^ Looks like this might be true for medical as well

taxcalc/functions.py
GoFroggyRun: Yeah, `c19700` correspond to Line 19, Schedule A, Gifts to Charity section. So I guess here it's a typo.

taxcalc/functions.py
GoFroggyRun: It's a good idea to treat `c20800` and `c17000` indifferently, like we did for `_statax`, in AMTI. A quick question come across that should we use `c60200` in stead of `min(c17000, 0.025 * _posagi) = min(c17000, 0.025 * max(0., c00100)) = min((1 - ID_Medical_HC) * c17000, 0.025 * max(0., c00100))` in the calculation of `c62100`?

GoFroggyRun: Never mind, we should still keep the original way of calculating `c62100` while applying ID to `c20800` and `c17000`.

Amy-Xu:
screen shot 2015-10-06 at 9 15 50 am
Table after the changes. @MattHJensen

MattHJensen:
Thanks @amy-xu!

@mistakevin, we are hoping you could help us by making a few new parameters from this PR available from the TaxBrain input page. Below the Itemized Deduction Limitation subsection, we are hoping to add a new subsection:

image

There are a few new things going on here. 

1. There is a new UI element, the checkbox for which itemized deductions are subject to the limitation.
2. The checkboxes are all coming from one parameter, _ID_BenefitSurtax_Switch. 
3. A flatblock description for a subsection (currently we only have these between sections.)

Before I merge this PR, would you mind taking a look to make sure that it will play nicely with the webapp as currently structured? 

cc @talumbau

mistakevin:
I don't think there would be any issue. Is there any way I can get a separate issue created with the exact specs and scenarios for the feature?

cc @talumbau

MattHJensen:
@mistakevin, sorry, just saw this. I'm merging this and opening a new issue on the webapp-public repo.

PR #362, opened 2015-09-24
Drop calculator function and guard against Parameter class misuse

martinholmer:
This pull request does two things: (1) it removes the calculator factory function, and (2) it strengthens the Parameters class against misuse of the class constructor.  These two simple changes have been made in the first two commits of this pull request, and those commits provide more explanation about these two changes. All the remaining code revisions in this pull request are in the test_utils.py and test_calculate.py files, and follow from the two substantive changes.

Recent discussion in the dropQ repo suggests there is consensus for removing the calculator factory function.  And recent master branch commits have eliminated the calculator factory function from use in the dropQ, webapp, and webapp-public repos, as best as I can tell.  I'm not aware of its use in any other part of the OSPC code base.

### DIFF
--- a/taxcalc/tests/test_calculate.py
+++ b/taxcalc/tests/test_calculate.py
@@ -8,9 +8,8 @@ from numpy.testing import assert_array_equal
 import pandas as pd
 import tempfile
 import pytest
-from numba import jit, vectorize, guvectorize
-from taxcalc import *
-import copy
+from taxcalc import Parameters, Records, Calculator, expand_array
+from taxcalc import create_distribution_table, create_difference_table
 
 
 # use simulated 1991 PUF to emulate private 2009 PUF
@@ -83,6 +82,7 @@ def test_make_Calculator_deepcopy():
     recs = Records(data=TAX_DTA, weights=WEIGHTS, start_year=2009)
     calc1 = Calculator(params=parm, records=recs)
     calc2 = copy.deepcopy(calc1)
+    assert isinstance(calc2, Calculator)
 
 
 def test_make_Calculator_files_to_ctor(paramsfile):
@@ -118,11 +118,10 @@ def test_make_Calculator_with_policy_reform():
 def test_make_Calculator_with_multiyear_reform():
     # create a Parameters object and apply a policy reform
     params3 = Parameters()
-    reform3 = {2015: {}, 2016: {}}
+    reform3 = {2015: {}}
     reform3[2015]['_STD_Aged'] = [[1600, 1300, 1600, 1300, 1600, 1300]]
-    reform3[2015]['_II_em'] = [5000]
-    reform3[2016]['_II_em'] = [6000]
-    reform3[2016]['_II_em_cpi'] = False
+    reform3[2015]['_II_em'] = [5000, 6000]  # reform values for 2015 and 2016
+    reform3[2015]['_II_em_cpi'] = False
     params3.implement_reform(reform3)
     # create a Calculator object using this policy-reform
     puf = Records(data=TAX_DTA, weights=WEIGHTS, start_year=2009)
@@ -220,20 +219,20 @@ def test_Calculator_create_distribution_table():
     puf = Records(data=TAX_DTA, weights=WEIGHTS, start_year=2009)
     calc = Calculator(params=params, records=puf)
     calc.calc_all()
-    DIST_LABELS = ['Returns', 'AGI', 'Standard Deduction Filers',
+    dist_labels = ['Returns', 'AGI', 'Standard Deduction Filers',
                    'Standard Deduction', 'Itemizers',
                    'Itemized Deduction', 'Personal Exemption',
                    'Taxable Income', 'Regular Tax', 'AMTI', 'AMT Filers',
                    'AMT', 'Tax before Credits', 'Non-refundable Credits',
                    'Tax before Refundable Credits', 'Refundable Credits',
                    'Revenue']
-    t1 = create_distribution_table(calc, groupby="weighted_deciles",
-                                   result_type="weighted_sum")
-    t1.columns = DIST_LABELS
-    t2 = create_distribution_table(calc, groupby="small_income_bins",
-                                   result_type="weighted_avg")
-    assert type(t1) == DataFrame
-    assert type(t2) == DataFrame
+    dt1 = create_distribution_table(calc, groupby="weighted_deciles",
+                                    result_type="weighted_sum")
+    dt1.columns = dist_labels
+    dt2 = create_distribution_table(calc, groupby="small_income_bins",
+                                    result_type="weighted_avg")
+    assert isinstance(dt1, pd.DataFrame)
+    assert isinstance(dt2, pd.DataFrame)
 
 
 def test_Calculator_create_difference_table():
@@ -250,7 +249,7 @@ def test_Calculator_create_difference_table():
     calc2 = Calculator(params=params2, records=puf2)
     # create difference table and check that it is a Pandas DataFrame
     dtable = create_difference_table(calc1, calc2, groupby="weighted_deciles")
-    assert type(dtable) == DataFrame
+    assert isinstance(dtable, pd.DataFrame)
 
 
 def test_diagnostic_table():
--- a/taxcalc/tests/test_calculate.py
+++ b/taxcalc/tests/test_calculate.py
@@ -12,19 +12,18 @@ from numba import jit, vectorize, guvectorize
 from taxcalc import *
 import copy
 
-WEIGHTS_FILENAME = "../../WEIGHTS_testing.csv"
+
+# use 1991 PUF to emulate private 2009 PUF
+TAX_DTA_PATH = os.path.join(CUR_PATH, '../../tax_all1991_puf.gz')
+TAX_DTA = pd.read_csv(TAX_DTA_PATH, compression='gzip')
+# PUF-fix-up: MIdR needs to be type int64 to match PUF
+TAX_DTA['midr'] = TAX_DTA['midr'].astype('int64')
+# specify WEIGHTS appropriate for 1991 PUF
+WEIGHTS_FILENAME = '../../WEIGHTS_testing.csv'
 WEIGHTS_PATH = os.path.join(CUR_PATH, WEIGHTS_FILENAME)
 WEIGHTS = pd.read_csv(WEIGHTS_PATH)
 
-all_cols = set()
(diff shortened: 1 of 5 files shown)


MattHJensen:
@talumbau, could you review this PR? It seems like it might get out of date pretty quickly and need quite a bit of maintenance if we don't get it merged soon.

taxcalc/tests/test_calculate.py
```diff
@@ -12,19 +12,18 @@
...
-all_cols = set()
-tax_dta_path = os.path.join(CUR_PATH, "../../tax_all1991_puf.gz")
-tax_dta = pd.read_csv(tax_dta_path, compression='gzip')
-
-# Fix-up. MIdR needs to be type int64 to match PUF
-tax_dta['midr'] = tax_dta['midr'].astype('int64')
-tax_dta['s006'] = np.arange(0, len(tax_dta['s006']))
-
-irates = {1991: 0.015, 1992: 0.020, 1993: 0.022, 1994: 0.020, 1995: 0.021,
+# use 1991 PUF to emulate private 2009 PUF
```
talumbau: to avoid confusion, I would say "use simulated 1991 PUF", since the dataset is really some modified version of 1991 data that was coerced into something resembling the PUF format.

taxcalc/tests/test_calculate.py
```diff
@@ -45,38 +44,19 @@ def paramsfile():
...
-
-    # Create a Public Use File object
-    puf = Records(tax_dta)
-
-    # Create a Calculator
-    calc = Calculator(params=params, records=puf)
+def run():
+    parm = Parameters()
+    assert parm.current_year == 2013
+    recs = Records(data=TAX_DTA, weights=WEIGHTS, start_year=2009)
```
talumbau: So implicit in this PR, we are saying that it is OK to override what is written in the underlying data when creating a Records instance by passing a `start_year` keyword arg, but it is not OK to do that for the Parameters instance. For example, the tax record data is from 1991, but we create the Records instance for 2009. However, we are not allowed to create a Parameters instance earlier in time, even if we wanted to. I will be using this interface far less often than others, so I don't have particularly strong feelings, it just feels a bit arbitrary and should probably be written down somewhere.

taxcalc/tests/test_calculate.py
```diff
@@ -89,133 +69,130 @@ def test_sequence():
...
-
-    # Create a Public Use File object
-    puf = Records(tax_dta)
-
-    user_mods = """{"1991": { "_STD_Aged": [[1500, 1250, 1200, 1500, 1500, 1200 ]],
-                     "_STD_Aged_cpi": false}}"""
-
-    calc2 = calculator(params, puf, mods=user_mods, _II_em_cpi=False,
-                       _II_em=np.array([4000]))
+def test_make_Calculator_with_policy_reform2():
```
talumbau: This shows the removal of the test of a capability to create a reform from a a JSON string. It turns out these reform dictionaries are serialized and deserialized to JSON in other parts of the web application, and so we haven't used this feature directly. @MattHJensen should sign off on the removal of this feature.

taxcalc/tests/test_calculate.py
```diff
@@ -89,133 +69,130 @@ def test_sequence():
...
     assert_array_equal(calc2.params.STD_Aged,
-                       np.array([1400, 1200, 1000, 1000, 1000, 1000]))
-
-
-def test_make_Calculator_with_reform_after_first_year():
-    fyr = 2008
-
-    # specify reform in in 2010
-    user_mods = {2010: {"_STD_Aged": [[1501, 1202, 1502, 1203, 1504, 1204]]}}
-    user_mods[2010]['_II_em'] = [5000, 6000]
```
talumbau: we need to maintain tests that test reforms specified in this way. Here, a value for parameter `_II_em` is specified for 2010 and 2011. We need to maintain tests like this because this is how we process data from the web application.

martinholmer:
@talumbau said:
> to avoid confusion, I would say "use simulated 1991 PUF", since the dataset is really some modified version of 1991 data that was coerced into something resembling the PUF format.
I didn't know that; thanks for the information.  I've made your suggested change.

taxcalc/tests/test_calculate.py
```diff
@@ -89,133 +69,130 @@ def test_sequence():
...
-def test_make_Calculator_mods():
-
-    # Create a Params object
-    params = Parameters(start_year=1991, inflation_rates=irates)
-
-    # Create a Public Use File object
-    puf = Records(tax_dta)
-
-    calc2 = calculator(params, puf, _II_em=np.array([4000]), _II_em_cpi=False)
+def test_make_Calculator_with_policy_reform1():
```
talumbau: This replacement test for `test_make_Calculator_mods` is redundant because the purpose of the original test was to test the ability to pass in particular reforms as keyword arguments. Again, I think @MattHJensen should chime in here to see if we should note this feature and create an issue for it to be included as a use case in creating a `Calculator` instance.

This new test exercises the same features as `test_make_Calculator_with_policy_reform2`, so I would suggest killing off `test_make_Calculator_with_policy_reform1` and leave  `test_make_Calculator_with_policy_reform2` and just call it  `test_make_Calculator_with_policy_reform`

martinholmer:
@talumbau said:
> So implicit in this PR, we are saying that it is OK to override what is written in the underlying data when creating a Records instance by passing a start_year keyword arg, but it is not OK to do that for the Parameters instance. For example, the tax record data is from 1991, but we create the Records instance for 2009. However, we are not allowed to create a Parameters instance earlier in time, even if we wanted to. I will be using this interface far less often than others, so I don't have particularly strong feelings, it just feels a bit arbitrary and should probably be written down somewhere.

There is a big difference between the params.json file (which has data only for 2013 on) and a data file being used by Records.  How can we make any sense out of saying the Parameters start_year is before 2013 when we are reading data from params.json?  What exactly are the values of policy parameters before 2013?  A user who is willing to specify a custom params dictionary can specify a Parameters start_year before 2013, and this is actually done some in the tests.  On the other hand, I can understand how, for the purpose of testing only, we can say that the 1991 data in the data file is to be considered data for some other year.

taxcalc/tests/test_calculate.py
```diff
@@ -89,133 +69,130 @@ def test_sequence():
...
-    tax_dta2 = pd.read_csv(tax_dta_path, compression='gzip')
-    rec = Records(tax_dta2, start_year=fyr)
-
-    # create Calculator object with params and rec objects
-    calc = Calculator(params, rec)
-
+                       np.array([1600, 1300, 1300, 1600, 1600, 1300]))
+
+
+def test_make_Calculator_with_policy_reform3():
```
talumbau: this test exercises a multi-year reform, so I suggest renaming to `test_make_Calculator_with_mulityear_reform`. This will make explicit how it differs from `test_make_Calculator_with_policy_reform`, so as a developer works on the repo, he/she will immediately see what is broken if one or both of the tests fail.

martinholmer:
@talumbau said:
> This shows the removal of the test of a capability to create a reform from a a JSON string. It turns out these reform dictionaries are serialized and deserialized to JSON in other parts of the web application, and so we haven't used this feature directly. @MattHJensen should sign off on the removal of this feature.

A reform can be created from a JSON string simply be converting it to a dictionary and passing that dictionary to the Parameters implement_reform method.  The conversation of the JSON string to a valid dictionary has nothing to do with the Tax-Calculator Parameters class, so it doesn't seem sensible to be testing that in the test_parameters.py file.

martinholmer:
@talumbau said:
> This new test exercises the same features as test_make_Calculator_with_policy_reform2, so I would suggest killing off test_make_Calculator_with_policy_reform1 and leave test_make_Calculator_with_policy_reform2 and just call it test_make_Calculator_with_policy_reform

and

> this test exercises a multi-year reform, so I suggest renaming to test_make_Calculator_with_mulityear_reform. This will make explicit how it differs from test_make_Calculator_with_policy_reform, so as a developer works on the repo, he/she will immediately see what is broken if one or both of the tests fail.

All of the above quoted suggestions have been implemented in the pull request.

talumbau:
OK, I think the last thing is just to re-add the original `test_make_Calculator_with_reform_after_first_year` that tests a policy reform where multiple years of the reform are described as a sequence of values. To emphasize that the purpose of the test is just to exercise this capability, perhaps a better name would be `test_make_Calculator_reform_multi_value_per_year` or something like that.

martinholmer:
@talumbau said:
> OK, I think the last thing is just to re-add the original test_make_Calculator_with_reform_after_first_year that tests a policy reform where multiple years of the reform are described as a sequence of values. To emphasize that the purpose of the test is just to exercise this capability, perhaps a better name would be test_make_Calculator_reform_multi_value_per_year or something like that.

I'm not sure I understand what you are suggesting above.  Do you mean by "where multiple years of the reform are described as a sequence of values" this kind of specification:
```
reform[2015]['_II_em'] = 5000
reform[2016]['_II_em'] = 6000
```
instead of the all-at-once style you told me to use:
```
reform[2015]['_II_em'] = [5000, 6000]
```
If this is what you mean, then there is already another test that uses the "sequence-of-values" approach to specifying a reform: test_make_Calculator_with_reform_after_start_year.  So, unless I'm misunderstanding you, the pull request already shows both ways of specifying a reform with provisions in more than one year.

taxcalc/tests/test_calculate.py
```diff
@@ -89,133 +68,118 @@ def test_sequence():
...
-
+                       np.array([1600, 1300, 1300, 1600, 1600, 1300]))
+
+
+def test_make_Calculator_with_multiyear_reform():
+    # create a Parameters object and apply a policy reform
+    params3 = Parameters()
+    reform3 = {2015: {}}
+    reform3[2015]['_STD_Aged'] = [[1600, 1300, 1600, 1300, 1600, 1300]]
+    reform3[2015]['_II_em'] = [5000, 6000]  # reform values for 2015 and 2016
```
talumbau: yes I was looking for this!

talumbau:
Yep you are right. I somehow missed what was happening in `test_make_Calculator_with_multiyear_reform`. That's the format we need to test. Thanks!

martinholmer:
@talumbau said:
> Yep you are right. I somehow missed what was happening in test_make_Calculator_with_multiyear_reform. That's the format we need to test. Thanks!

Sure.  Is there anything else that needs to be fixed before merging into master?

talumbau:
no I think this is good to go. +1

MattHJensen:
Merging.

PR #363, opened 2015-09-25
[WIP] Lift code compile out of iterate_jit wrapper

talumbau:
- Instead of doing an inspection of arguments at function run time,
  look at the allowed names in Records and Parameters
  instances at decoration time, and then use those to generate and
  compile the high-level function. Then, at function call time,
  we simply called the jitted function.


talumbau:
I had an idea that this would improve speed, but I don't think it actually does. I want to do testing with dropq to be sure. This needs clean up for it be of any use anyway, so even if it showed good speedup, it wouldn't be mergable as is. If there's no speed-up in my testing, I will close this.

talumbau:
Results with current master:
```
You loaded data for 2009.
You loaded data for 2009.
You loaded data for 2009.
('elapsed time for second run: ', 3.777651071548462)
('elapsed time for second run: ', 4.730506896972656)
('elapsed time for second run: ', 4.6791510581970215)
('elapsed time for second run: ', 4.424412965774536)
('elapsed time for second run: ', 4.327259063720703)
('elapsed time for second run: ', 4.398780107498169)
('elapsed time for second run: ', 4.339246034622192)
('elapsed time for second run: ', 4.352167844772339)
('elapsed time for second run: ', 4.115127801895142)
('elapsed time for second run: ', 4.247420787811279)
```
Results with this branch:
```
You loaded data for 2009.
You loaded data for 2009.
You loaded data for 2009.
('elapsed time for second run: ', 2.8246099948883057)
('elapsed time for second run: ', 4.406286001205444)
('elapsed time for second run: ', 4.271919012069702)
('elapsed time for second run: ', 4.336216926574707)
('elapsed time for second run: ', 4.189327001571655)
('elapsed time for second run: ', 4.21439003944397)
('elapsed time for second run: ', 4.181212902069092)
('elapsed time for second run: ', 4.126456022262573)
('elapsed time for second run: ', 4.195266962051392)
('elapsed time for second run: ', 4.177548885345459)
```

So it is slightly faster with this technique, but not much. It would actually be interesting to run `cProfile` on the code and see where most of the time is spent. I think I looked at this once and found about 50% of the time was spent in `calc_all` and about 50% in all of the pandas operations for `dropq`. Anyway, this seems like a bust.

PR #360, opened 2015-09-23
[WIP] Validation:Updates in SAS

GoFroggyRun:
This PR deals with the changes that have been made recently in the SAS calculator.

PR #364, opened 2015-09-26
Do Records class imputations only if current_year is 2009

martinholmer:
This pull request adds an "if self._current_year == 2009:" statement before the imputation code in the Records class constructor.  This is done because the imputations make no sense if the Records data are for any other year (as is the case, for example, in the new TAXSIM-like capability described in issue #291).  When making this change, the imputation code was put into its own private method to avoid PEP8 warnings about lines being too long.  Also, several other non-substantive code changes were made to make the imputation logic easier to follow or to eliminate pylint warnings.  After all these changes, the test.py script produces exactly the same results (using the 2009 puf.csv as Records data) as test.py did before these changes.

The last three commits were added to revise test code to use the revisions in the records.py file.

### DIFF
--- a/taxcalc/records.py
+++ b/taxcalc/records.py
@@ -50,7 +50,7 @@ class Records(object):
     parameters of the constructor, and therefore, imputed variables
     are generated to augment the data and initial-year blowup factors
     are applied to the data. Explicitly setting start_year to some
-    value other than CURRENT_PUF_YEAR will cause this variable-imputation
+    value other than Records.PUF_YEAR will cause this variable-imputation
     and initial-year-blowup logic to be skipped.  There are situations in
     which this is exactly what is desired, but more often than not,
     skipping the imputation and blowup logic would be a mistake.  In
@@ -58,7 +58,7 @@ class Records(object):
     class constructor unless you know exactly what you are doing.
     """
 
-    CURRENT_PUF_YEAR = 2009
+    PUF_YEAR = 2009
 
     CUR_PATH = os.path.abspath(os.path.dirname(__file__))
     WEIGHTS_FILENAME = "WEIGHTS.csv"
@@ -289,7 +289,7 @@ class Records(object):
             msg = ('Records.constructor start_year is neither None nor '
                    'an integer')
             raise ValueError(msg)
-        if self._current_year == Records.CURRENT_PUF_YEAR:
+        if self._current_year == Records.PUF_YEAR:
             self._impute_variables()
 
     @property
@@ -304,7 +304,7 @@ class Records(object):
         # Implement Stage 2 Extrapolation reweighting.
         self.s006 = self.WT["WT" + str(self.current_year)] / 100
 
-    def extrapolate_09_puf(self):
+    def extrapolate_2009_puf(self):
         year = 2009
         self.BF.AGDPN[year] = 1
         self.BF.ATXPY[year] = 1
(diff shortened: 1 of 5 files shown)


MattHJensen:
> This pull request adds an "if self._current_year == 2009:" statement before the imputation code in the Records class constructor.

This seems like a great idea (and I see the discussion over at #360). The other changes look like improvements as well. 

@Amy-Xu, could you also review and merge if everything looks good?

taxcalc/records.py
```diff
@@ -240,45 +237,18 @@ def __init__(self,
...
-        self.read_weights(weights)
-        if (start_year):
+        """
+        Records class constructor
+        """
+        self._read_data(data)
+        self._read_blowup(blowup_factors)
+        self._read_weights(weights)
+        if start_year:
             self._current_year = start_year
```
Amy-Xu: Just as this PR mentioned, it doesn't really make any sense if we run any data other than 2009. Then it means if users random put some other start_year here, for example 2013, the calculator will just automatically skipped the blow-up process without any warning message. So my concerns are 1) what are the cases where users might put in random years here? Are those cases necessary? 2) if we want to keep this start_year attribute, should we add a message letting users know no blow-up is involved due to the start_year set up?

taxcalc/records.py
```diff
@@ -240,45 +237,18 @@ def __init__(self,
...
-                                  np.logical_and(self.MARS != 2,
-                                                 self.MARS != 3),
-                                  (self.e04470 - std2009[self.MARS - 1]) /
-                                  STD_Aged_2009[0],
-                                  (self.e04470 - std2009[self.MARS - 1]) /
-                                  STD_Aged_2009[1]),
-                                  np.where(self.e02400 > 0, self._txpyers, 0))
-        self._puf_year = 2009
+        if self._current_year == 2009:
+            self._impute_variables()
```
Amy-Xu: Similarly here, if the _current_year was set to some other year other 2009 due to the start_year attribute, we're not doing the imputations anymore. Should we add a warning message in an else loop?

Amy-Xu:
@martinholmer This looks good to me in general. But I have a few concerns and put them in the notes for corresponding lines. Could you take a look?

martinholmer:
@Amy-Xu said:
> Just as this PR mentioned, it doesn't really make any sense if we run any data other than 2009. Then it means if users randomly put some other start_year [in the Parameters constructor call], for example 2013, the calculator will just automatically skip the blow-up process without any warning message. So my concerns are 1) what are the cases where users might put in random years here? Are those cases necessary? 2) if we want to keep this start_year attribute, should we add a message letting users know no blow-up is involved due to the start_year set up?

Amy, these are legitimate concerns.  Here are my reactions to the issues you raise.  First of all, almost all the time in real use (as opposed to the "fake" use of 1991 data to emulate the 2009 PUF in various tests) the Parameters class constructor will be called like this:
```
parms = Parameters()
```
In this case, the situations you are concerned about ("random years") do not occur.  And second, we need to keep the start_year in the Parameters constructor because otherwise we will not be able to use the 1991 date to emulate the 2009 PUF in the tests and I will not be able to develop the TAXSIM-like capability described in issue #291.  I don't like your idea of writing a message because it will complicate the output generated by the TAXSIM-like capability, which is __intentionally__ skipping the blow-up process by setting start_year to something other than 2009.  But perhaps we can add a strong warning about "random years" to the documentation of the Parameters constructor.  I'm am going to go ahead and do that, and then we can discuss these issues further.

martinholmer:
@Amy-Xu,  Sorry about my confusion between the Records and Parameters class in my prior post.  I have edited that post to make more sense.

martinholmer:
@Amy-Xu,
In yesterday's discussion, I forgot to add the basic point that a mistaken value of start_year will raise an error when Calculator.calc_all() is called.  This means that if somebody doesn't read the Records class documentation and explicitly sets the Records class constructor parameter start_year to some value other than Records.PUF_YEAR (which is currently 2009), they won't get far because an error will be raised since the imputed _compitem variable does not exist.  I have added a new test to test_records.py that checks for this condition.  I think that between the enhanced documentation of the Records class constructor and the fact that a user mistake in setting start_year will cause a fatal error, we have done enough to guard against mistakes in setting start_year.  Is there anything else I should change in this pull request before you merge it into the master branch?

martinholmer:
@Amy-Xu,  I'm interested in your view on the status of this pull request.

Amy-Xu:
LGTM! Thanks @martinholmer. Merging.

PR #375, opened 2015-10-08
Proxy for Foreign Earned Income Exclusion Indicator

Amy-Xu:
@feenberg We just realized that all the variables currently used right now (_feided<-(e35300, e35600, e35910), f2555) to calculate Foreign Earned Income (FEI) exclusion are non-puf variables. I was wondering whether we could use the variable TXST as a proxy for f2555, and use e00200 to replace _feided for now in order to get a rough estimate of FEI exclusion (c02700).

I want to give this TXST a try because if this variable has the value of 10, it means this taxpayer has FEI tax liability. 
screen shot 2015-10-08 at 5 14 28 pm

Then I would set `f2555` to 1 if this TXST is 10. Then use what you had in SAS `c02700 = min(_feided,_feimax{flpdyr}*F2555)` but replace `_feided` with `e00200` to approximate FEI exclusion. Do you think this is a reasonable way to do it?

For 2013, FEI exclusion sums up to 27 billion. We could get 11 billion through the way I described above, but zero in the current version of code.

cc @MattHJensen


feenberg:
TXST==10 is a good proxy for F2555.

dan

MattHJensen:
@feenberg, could you also verify that replacing _feided with e00200 seems roughly reasonable?

feenberg:
On Thu, 8 Oct 2015, Matt Jensen wrote:

> 
> @feenberg, could you also verify that replacing _feided with e00200 seems
> roughly reasonable?

No, that isn't right. e00200 is wages, _feided is e02900 but not more than 
the allowed amount.

dan

Amy-Xu:
@feenberg I don't see e02900 in 09 PUF. Only e62900, alternative tax foreign tax credit, is available. Is there any other ways we might be able to approximate _feided? 

Originally we thought `e00200` might work, the reason being taxpayers' foreign earned income is included in wages, although this might over estimate their foreign earned income. We thought `e00200` includes both domestic and foreign earned income. Is that right?

feenberg:
On Fri, 9 Oct 2015, Amy Xu wrote:

> 
> @feenberg I don't see e02900 in 09 PUF. Only e62900, alternative tax foreign tax
> credit, is available. Is there any other ways we might be able to approximate _feided?
> 
> Originally we thought e00200 might work, the reason being taxpayers' foreign earned
> income is included in wages, although this might over estimate their foreign earned
> income. We thought e00200 includes both domestic and foreign earned income. Is that
> right?

It does, but many F2555 returns have only a fraction from abroad.  You can 
try assuming all the income is eligible - see if that improves the overall 
accuracy.

dan

MattHJensen:
@Amy-Xu we should probably add an `if puf:` condition to switching out e02900 with e00200.

Amy-Xu:
screen shot 2015-10-09 at 12 25 07 pm
Diagnostic table with this modification.

MattHJensen:
Could you also post a a) diagnostic table from master so we can see the revenue change, and b) the amount of FEI tax in the 2013 SOI tables?

feenberg:
On Fri, 9 Oct 2015, Matt Jensen wrote:

> 
> Could you also post a a) diagnostic table from master so we can see the revenue
> change, and b) the amount of FEI tax in the 2013 SOI tables?

Recall that FEI is more in the nature of a credit than a tax. The credit 
allows the FEI to be taxed as if it were the taxpayers only income, and 
the taxpayer had no deductions or exemptions.

dan

Amy-Xu:
FEI exclusion in 2013 using this method: 10.73 billions
SOI estimates: 29.0 billions

Master:
screen shot 2015-10-09 at 2 41 43 pm

This PR:
screen shot 2015-10-09 at 2 40 41 pm

MattHJensen:
I'm still a little puzzled by why our estimate of the FEI exclusion using e00200 would be lower than what is reported in the SOI table. @feenberg, is e00200 strictly higher than e02900?

feenberg:
On Fri, 9 Oct 2015, Matt Jensen wrote:

> 
> I'm still a little puzzled by why our estimate of the FEI exclusion using e00200 would
> be lower than what is reported in the SOI table. @feenberg, is e00200 strictly higher
> than e02900?

My first thought is that the if you take everyone with TXST of 10 and give 
them the FEI maximum exclusion, that must be more than the actual amount 
in the SOI table, unless something is very wrong. In the table they give 
the number of taxpayers taking the exclusion - is that comparable to our 
number?

dan

PR #376, opened 2015-10-08
Add Internet-TAXSIM-like capability to taxcalc package

martinholmer:
This pull request is the initial implementation of the capability described in issue #291.


taxcalc/simpletaxio.py
```diff
@@ -0,0 +1,648 @@
...
+"""
+# CODING-STYLE CHECKS:
+# pep8 --ignore=E402 simpletaxio.py
+# pylint --disable=locally-disabled simpletaxio.py
+
+import os
+import sys
+import json
+import re
+import numpy.core as np
```
talumbau: the standard idiom here would be
```
import numpy as np
```
There is only one usage of the `np` alias below, referring to `numpy.core.array`, which is the same as `numpy.array`, but additional uses of `np` by other developers would likely produce unexpected results.

taxcalc/simpletaxio.py
```diff
@@ -0,0 +1,648 @@
...
+from numpy.core import int32 as np_int32
```
talumbau: suggest removal of this and specifying dtype below as the string `int32` which is a common pattern and prevents the need for an import.

taxcalc/simpletaxio.py
```diff
@@ -0,0 +1,648 @@
...
+    class instance: SimpleTaxIO
+    """
+
+    def __init__(self, input_filename, reform_filename):
+        """
+        SimpleTaxIO class constructor.
+        """
+        # construct output_filename and delete old output file if it exists
+        self._output_filename = '{}.out-simtax'.format(input_filename)
+        if os.access(self._output_filename, os.F_OK):
```
talumbau: suggest usage of `os.path.exists` for this check.

taxcalc/simpletaxio.py
```diff
@@ -0,0 +1,648 @@
...
+            os.remove(self._output_filename)
+        # check for existence of file named input_filename
+        if not os.access(input_filename, os.F_OK):
```
talumbau: suggest usage of `os.path.exists` for this check.

taxcalc/simpletaxio.py
```diff
@@ -0,0 +1,648 @@
...
+        Returns
+        -------
+        nothing: void
+        """
+        # loop through self._year_set doing tax calculations and saving output
+        for calcyear in self._year_set:
+            if calcyear != self._calc.params.current_year:
+                self._calc.params.set_year(calcyear)
+            self._calc.calc_all()
+            calc_records_tax_year = getattr(self._calc.records, 'FLPDYR')
```
talumbau: since `FLPDYR` is a string literal and also a valid Python identifier, suggest:
```
calc_records_tax_year = self._calc.records.FLPDYR
```

taxcalc/simpletaxio.py
```diff
@@ -0,0 +1,648 @@
...
+        Read reform file, strip //-comments, and return dict based on JSON.
+        The reform file is JSON with string policy-parameter primary keys and
+           string years as secondary keys.
+        Returned dictionary has integer years as primary keys and
+           string policy-parameters as secondary keys.
+        The returned dictionary is suitable as the argument to the Parameters
+           class implement_reform(reform_dict) method.
+        """
+        # check existence of specified reform file
+        if not os.access(reform_filename, os.F_OK):
```
talumbau: suggest usage of os.path.exists for this check.

taxcalc/simpletaxio.py
```diff
@@ -0,0 +1,648 @@
...
+                    raise ValueError(msg.format(num_aged, lnum))
+            else:  # if filing_status is 1=single or 3=head_of_household
+                if num_aged > 1:
+                    msg = ('var[6]={} on line {} of simtax INPUT is not '
+                           'less than or equal to one')
+                    raise ValueError(msg.format(num_aged, lnum))
+            tran = var[13]
+            if tran != 0:
+                msg = ('var[13]={} on line {} of simtax INPUT is not zero '
+                       'to indicate no state income tax calculations')
```
talumbau: I can't be sure, but it looks to me that, based on the pattern from the other error checking in this function, the intention is to raise an exception if `tran !=0`. But after the `msg` variable is created, we leave this `for` block. Are you sure you did not want to raise a ValueError here?

taxcalc/simpletaxio.py
```diff
@@ -0,0 +1,648 @@
...
+        -------
+        calc: Calculator
+        """
+        # create all-zeros dictionary and then list of all-zero dictionaries
+        zero_dict = {}
+        for names in Records.NAMES:
+            zero_dict[names[1]] = 0
+        dict_list = [zero_dict for _ in range(0, len(self._input))]
+        # use dict_list to create a Pandas DataFrame and Records object
+        recsdf = pd.DataFrame(dict_list, dtype=np_int32)
```
talumbau: suggest:

```
 recsdf = pd.DataFrame(dict_list, dtype="int32")
```

taxcalc/tests/test_simpletaxio.py
```diff
@@ -0,0 +1,141 @@
...
+    Temporary reform file for SimpleTaxIO constructor.
+    """
+    rfile = tempfile.NamedTemporaryFile(mode='a', delete=False)
+    rfile.write(REFORM_CONTENTS)
+    rfile.close()
+    # must close and then yield for Windows platform
+    yield rfile
+
+
+def test_1(input_file):  # pylint: disable=redefined-outer-name
```
talumbau: I think the best idea is to give tests a name that describes what the test is testing. The idea is to provide context to the person who runs the test and gets failures.

taxcalc/tests/test_simpletaxio.py
```diff
@@ -0,0 +1,141 @@
...
+    """
+    Test SimpleTaxIO contructor with no policy reform.
+    """
+    reform_file_name = None
+    simtax = SimpleTaxIO(input_file.name, reform_file_name)
+    assert simtax.number_input_lines() == NUM_INPUT_LINES
+
+
+def test_2(input_file,  # pylint: disable=redefined-outer-name
```
talumbau: see comment above.

taxcalc/tests/test_simpletaxio.py
```diff
@@ -0,0 +1,141 @@
...
+           reform_file):  # pylint: disable=redefined-outer-name
+    """
+    Test SimpleTaxIO contructor with a policy reform.
+    """
+    simtax = SimpleTaxIO(input_file.name, reform_file.name)
+    assert simtax.number_input_lines() == NUM_INPUT_LINES
+
+
+def test_cleanup(input_file,  # pylint: disable=redefined-outer-name
```
talumbau: We can use the test fixture itself to clean up these files. I'll show an example chunk of code in the `reform_file` fixture.

taxcalc/tests/test_simpletaxio.py
```diff
@@ -0,0 +1,141 @@
...
+@pytest.yield_fixture
+def reform_file():
+    """
+    Temporary reform file for SimpleTaxIO constructor.
+    """
+    rfile = tempfile.NamedTemporaryFile(mode='a', delete=False)
+    rfile.write(REFORM_CONTENTS)
+    rfile.close()
+    # must close and then yield for Windows platform
+    yield rfile
```
talumbau: The clean up activity has been neglected in our tests so far. We can do this on all of our temp file test fixtures. After the `yield` we just need:

```
    if os.path.exists(filename):
        try:
            os.remove(filename)
        except OSError:  # sometimes we can't remove a generated temp file
            pass
```

martinholmer: @talumbau, OK, I followed your suggestion but used os.path.isfile(fname) instead of os.path.exists(fname) because the later returns true if fname is a directory or a file, and I don't want to remove dirctories.

taxcalc/tests/test_simpletaxio.py
martinholmer: @talumbau, Yes all the other test file use very long test_ function name and have no docstring.  This style, whatever its merits, drives pylint crazy.  Pylint assigns two warnings to each test_ function: one for too long a function name and one for the missing docstring.  The style here eliminates both warnings and provides unlimited space (not just an arcane, long function name) in the docstring to explain what is being tested.  If a test fails the programmer is going to have to look at the full code for the test anyway, so I don't see that a long function name is much of an advantage.  So, thanks for the suggestion, but I think I'll stick to the style in the test_simpletaxio.py file.

taxcalc/simpletaxio.py
martinholmer: @talumbau,  YES, that is my mistake.  Thanks for catching this.

taxcalc/simpletaxio.py
martinholmer: @talumbau,  Good suggestion.  This has been changed as above.

taxcalc/simpletaxio.py
martinholmer: @talumbau,  Good suggestion to move away from os.access(...), but I've decided to use os.path.isfile instead of os.path.exists, which returns true if the supplied path is a directory (not a file).  Same changes are made in several places, so I'll thank you for your suggestion just once here.

taxcalc/simpletaxio.py
martinholmer: @talumbau,  Again, thanks for the useful suggestion, which has been incorporated in the pull request.

taxcalc/simpletaxio.py
martinholmer: @talumbau,  You are right about this.  I've change to "import numpy as np" and "np.core.array(...)".  Need the .core to avoid pylint warning (seems that numpy has an 'unusual' importing philosophy that pylint has trouble with).  But working around this is now better than before, so thanks for the suggestion.

taxcalc/simpletaxio.py
martinholmer: See comment below on how os.path.isfile is used instead.

taxcalc/simpletaxio.py
martinholmer: See comment below on how os.path.isfile is used instead.

taxcalc/simpletaxio.py
martinholmer: @talumbau,  I had this kind of code to start with, but it produces the following pylint warning:
```
pylint --disable=locally-disabled simpletaxio.py
************* Module taxcalc.simpletaxio
simpletaxio.py:87: [E1101(no-member), SimpleTaxIO.calculate] Instance of 'Records' has no 'FLPDYR' member
```
This is probably because of the indirect way Records members are specified.  So, everything in your comment is true, except you are not using pylint, so you don't see the same picture I'm looking at.  What I'm doing is correct, right?  It just seems indirect to you, right?  Or am I missing something else that is important?  In other words, is there anything here other the style preferences?

martinholmer:
@MattHJensen and @talumbau,
This pull request is ready to be merged as far as I can see.  I've incorporated almost all of T.J.'s many suggestions and provided him reasons for doing something slightly different in one case and doing nothing at all in two cases.  If either of you think more needs to be done on this pull request, we need to move quickly because there is another pull request (with validation tools) and a subsequent pull request (with bug fixes found with the validation tools) that Matt wants incorporated quickly into the master branch.

taxcalc/simpletaxio.py
```diff
@@ -0,0 +1,649 @@
...
+        Returns
+        -------
+        nothing: void
+        """
+        # loop through self._year_set doing tax calculations and saving output
+        for calcyear in self._year_set:
+            if calcyear != self._calc.policy.current_year:
+                self._calc.policy.set_year(calcyear)
+            self._calc.calc_all()
+            calc_records_tax_year = getattr(self._calc.records, 'FLPDYR')
```
talumbau: Comment from @martinholmer :
   "This is probably because of the indirect way Records members are specified. So, everything in your comment is true, except you are not using pylint, so you don't see the same picture I'm looking at. What I'm doing is correct, right? It just seems indirect to you, right? Or am I missing something else that is important? In other words, is there anything here other the style preferences?"

The code produces the correct result, but it would cause anyone looking at it to say, 'wait, why do it that way?' and then they would wonder if they were missing something. There's no default value added to the `getattr` call, and `FLPDYR` is a string literal and valid Python identifier. The only way this makes sense to me is because you have provided the answer - to silence a warning from pylint. I don't think the best approach is 'get no output from pylint' and I think this is a great example of why it is not always best to do whatever is necessary to get rid of all pylint warnings. pylint is a helpful tool, but it's not flawless.

martinholmer: @talumbau,  I've changed the code related to ```FLPDYR``` to be more in line with your preferences.

taxcalc/simpletaxio.py
```diff
@@ -0,0 +1,649 @@
...
+            self._calc.calc_all()
+            cr_taxyr = self._calc.records.FLPDYR  # pylint: disable=no-member
+            for idx in range(0, self._calc.records.dim):
+                indyr = cr_taxyr[idx]
+                if indyr == calcyr:
+                    lnum = idx + 1
+                    ovar = SimpleTaxIO._extract_output(self._calc.records, idx,
+                                                       self._input[lnum])
+                    self._output[lnum] = ovar
+            if calc_marginal_tax_rates:
```
MattHJensen: Happy to see this!

talumbau:
this looks good. merging.

PR #377, opened 2015-10-08
Update Weights and Stage I factors for CPS matched file

Amy-Xu:
In this PR:

1) SOI statistics has been updated to 2013 release, and therefore corresponding Stage I factors and Stage II targets are updated as well.

2) To accommodate this CPS matched file, a new WEIGHTS.csv is generated to match the number of records. This new CPS matched file has about 250k records. Therefore if this PR got merged, 09 PUF is not compatible with taxcalc anymore. Everyone need to switch to this CPS matched PUF.

Here's the diagnostic table:

screen shot 2015-10-08 at 7 18 31 pm


martinholmer:
@Amy-Xu, If you are removing the call to extrapolate_2009_puf() in calculate.py, is there any reason to have that method in the Records class?

Amy-Xu:
@martinholmer You're right. If we are going to remove this extra step then there no need to keep that method in Records.py anymore. But I just email John yesterday. It seems like we still need to keep this 09 extrapolation. I'll update this PR soon with new weights.

feenberg:
On Thu, 8 Oct 2015, Amy Xu wrote:

> 
> In this PR:
> 
> 1) SOI statistics has been updated to 2013 release, and therefore corresponding Stage
> I factors and Stage II targets are updated as well.
> 
> 2) To accommodate this CPS matched file, a new WEIGHTS.csv is generated to match the
> number of records. This new CPS matched file has about 250k records. Therefore if this
> PR got merged, 09 PUF is not compatible with taxcalc anymore. Everyone need to switch
> to this CPS matched PUF.

The calculator needs to be able to use different datasets. Can't the 
weights data be incorporated into the main datafile?

dan

Amy-Xu:
@MattHJensen I added the 09 extrapolation back and re-did the extrapolation of all other projection years. Here's the table.
screen shot 2015-10-09 at 12 07 06 pm

cc @martinholmer I think we need to keep the 09 extrapolation for now because the CPS matched file needs it, which I didn't know yesterday.

martinholmer:
@Amy-Xu said:
> I think we need to keep the 09 extrapolation for now because the CPS matched file needs it, which I didn't know yesterday.

OK, that's fine.

MattHJensen:
Merging. Everyone note that you will need to begin using the newest data file from @amy-xu now. Please contact her or me via a private channel if you didn't get it.

PR #379, opened 2015-10-09
Use CPS earning split ratio to calculate FICA social security tax

Amy-Xu:
CPS provides the wage of household head and spouse separately. In this PR, earning split ratios are calculated based on the two CPS variables, for married filing jointly couples only. Then FICA social security ceiling can be applied to the head and spouse separately for a more accurate calculation of payroll tax. But it seems the current payroll total is about 100 billions lower than CBO estimates. @martinholmer @feenberg @MattHJensen Any suggestions/feedback on this? Thanks!


Amy-Xu:
Diagnostic table for this PR:
screen shot 2015-10-09 at 2 03 02 pm

feenberg:
On Fri, 9 Oct 2015, Amy Xu wrote:

> 
> CPS has provided the wage of household head and spouse separately. In this PR, earning
> split ratios are calculated based on the two CPS variables, for married filing jointly
> couples only. Then FICA social security ceiling can be applied to the head and spouse
> separately for a more accurate calculation of payroll tax. But it seems the current
> payroll total is about 100 billions lower than CBO estimates. @martinholmer @feenberg
> @MattHJensen Any suggestions/feedback on this? Thanks!

CPS isn't very good at recording high incomes, and those are necessary for 
the part of FICA that isn't capped. I don't suppose there is an estimate 
for the 15.3% portion alone. There might be from SSA, though.

dan

MattHJensen:
>CPS isn't very good at recording high incomes, and those are necessary for
the part of FICA that isn't capped. 

@feenberg, I don't think that would explain it. @amy-xu is still calculating _fica using e00200:

    _sey = e00900 + e02100
    _fica_ss_head = max(0, FICA_ss_trt * min(SS_Earnings_c,
                        e00200 * _earning_split + max(0, _sey) *
                        (1 - 0.5 * (FICA_mc_trt + FICA_ss_trt))))
    _fica_ss_spouse = max(0, FICA_ss_trt * min(SS_Earnings_c,
                          e00200 * (1 - _earning_split) + max(0, _sey) *
                          (1 - 0.5 * (FICA_mc_trt + FICA_ss_trt))))

    _fica_ss = _fica_ss_head + _fica_ss_spouse
    _fica_mc = max(0, FICA_mc_trt * (e00200 + max(0, _sey) *
                   (1 - 0.5 * (FICA_mc_trt + FICA_ss_trt))))

    _fica = _fica_mc + _fica_ss

She is just getting the _earning_split variable from CPS (`wage_head/total` where `total!=0`)

MattHJensen:
This might have something to do with the fact that we are indexing the ceiling at prices instead of wages right now as @martinholmer describes in  https://github.com/OpenSourcePolicyCenter/Tax-Calculator/issues/369

MattHJensen:
@feenberg, note that we are also adding _amed appropriately to _fica already (I accidentally left this out in my previous comment):

    _amed = AMED_trt * max(0, e00200 +
                           max(0, _sey) * (1 - 0.5 *
                                           (FICA_mc_trt + FICA_ss_trt)) -
                           AMED_thd[MARS - 1])
    _fica = _fica + _amed

feenberg:
On Fri, 9 Oct 2015, Matt Jensen wrote:

>
>       CPS isn't very good at recording high incomes, and those are necessary for
>       the part of FICA that isn't capped.
> 
> @feenberg, I don't think that would explain it. @amy-xu is still calculating _fica
> using e00200:
> 
> _sey = e00900 + e02100
> _fica_ss_head = max(0, FICA_ss_trt * min(SS_Earnings_c,
>                     e00200 * _earning_split + max(0, _sey) *
>                     (1 - 0.5 * (FICA_mc_trt + FICA_ss_trt))))
> _fica_ss_spouse = max(0, FICA_ss_trt * min(SS_Earnings_c,
>                       e00200 * (1 - _earning_split) + max(0, _sey) *
>                       (1 - 0.5 * (FICA_mc_trt + FICA_ss_trt))))
> 
> _fica_ss = _fica_ss_head + _fica_ss_spouse
> _fica_mc = max(0, FICA_mc_trt * (e00200 + max(0, _sey) *
>                (1 - 0.5 * (FICA_mc_trt + FICA_ss_trt))))
> 
> _fica = _fica_mc + _fica_ss
> 
> She is just getting the _earning_split variable from CPS (wage_head/total where
> total!=0)

That makes it a bigger mystery, since I would expect CPS splits to be less 
extreme than SOI splits. Can Amy go back to 2009 data and compare to SSA 
rather than CBO, without aging it might be easier to isolate the problem.

dan

martinholmer:
@Amy-Xu, @feenberg, @MattHJensen,
I don't have any idea at this stage about what factor(s) are causing the underestimate.
However, it seems to me that the way to proceed is to get our FICA tax calculations
logically correct, and then investigate the underestimation problem (if it is still there).

MattHJensen:
> it seems to me that the way to proceed is to get our FICA tax calculations
logically correct

@martinholmer, are you referring to the indexing issue, or are there other errors that you have identified?

martinholmer:
@Amy-Xu, @feenberg, @MattHJensen,
Where is e00200 set as the sum of wage_head and wage_spouse?  If you still have e00200 in puf.csv, then why do you include both wage_head and wage_spouse?  Why not just add one new variable, e00200s, which is zero for MARS != 2 and otherwise is the spouse dollar earnings (which range from zero to e00200)?  That would save space and eliminate redundancy, and get rid of the ambiguous earning_split variable, whose name tells you nothing about whose share is being discussed.  Plus, you have extra multiplications to do with the split variable.

martinholmer:
@Amy-Xu, @feenberg, @MattHJensen,
Matt said:
> > it seems to me that the way to proceed is to get our FICA tax calculations logically correct

> @martinholmer, are you referring to the indexing issue, or are there other errors that you have identified?

I'm referring to the topic of this pull request: adding spouse earnings, say e00200s, as an extra puf.csv variable.

Amy-Xu:
@martinholmer Let me clarify this a bit more. Currently we are not using wage_head and wage_spouse from CPS directly. We use the the ratio calculated from those two variables, and e00200 to approximate earning split ratios for married filing jointly couples. So there was no extra e00200s added to our code base.

One reason why we didn't use the CPS variables directly is that those two variable may or may not sum up to e00200. And theoretically we want to base our calculation more on PUF variables than CPS variables if I assume it right? @MattHJensen

MattHJensen:
@amy-xu, very clear comment. Here is what I was in the middle of writing that gets at basically the same thing:

@martinholmer,  the advantage of using earning_split (perhaps this should be renamed `primary_earner_wage_share` or similar) rather than `e00200s` taken from the CPS as you propose is that `e00200s` and `e00200p` from the CPS might not add up to `e00200` from the puf; this is just a statistical match. It is not even clear to me that e00200s would necessarily be smaller than e00200 from the PUF. So relying on e00200 as much as possible seems safer.

martinholmer:
@Amy-Xu, @feenberg, @MattHJensen,
Matt said:
> @martinholmer, the advantage of using earning_split (perhaps this should be renamed primary_earner_wage_share or similar) rather than e00200s taken from the CPS as you propose is that e00200s and e00200p from the CPS might not add up to e00200 from the puf; this is just a statistical match. It is not even clear to me that e00200s would necessarily be smaller than e00200 from the PUF. So relying on e00200 as much as possible seems safer.

I didn't mean to suggest using the CPS values directly.  It just seems to me that there is a whole lot of calculation that could be done __once__ when the puf.csv file is prepared (instead of being done every time the FICA calculations are done).  By all means use the SOI e00200, but impute to the puf.csv file just one extra variable, which is the dollar amount the e00200 total that is attributable to the spouse.  And yes, you are right, e00200s could be more than e00200 if the head has negative self-employment earnings.  We're concerned about speed of execution, right?  So, why not take an easy speed-up opportunity when it arises?

MattHJensen:
>I didn't mean to suggest using the CPS values directly. It just seems to me that there is a whole lot of calculation that could be done once when the puf.csv file is prepared (instead of being done every time the FICA calculations are done). By all means use the SOI e00200, but impute to the puf.csv file just one extra variable, which is the dollar amount the e00200 total that is attributable to the spouse.

@martinholmer, sorry for the misunderstanding. This makes a lot of sense. So when preparing the puf-cps match file we'd use the CPS ratios to create imputed e00200s and e00200p variables @amy-xu, what do you think?

GoFroggyRun:
While dealing with the weighted averaged of `_fica` under 09 logic, it turns out that there are quite a few observations that have `MARS == 2` but `(wage_head + wage_spouse) == 0`, for some reason. So I'd suggest to modify the way `total` is calculated a little bit to take care of this special case:
`total = np.where(np.logical_and(self.MARS == 2, (self.wage_head + self.wage_spouse) != 0), self.wage_head + self.wage_spouse, 0)`. 

cc @MattHJensen

feenberg:
On Sun, 11 Oct 2015, Sean.Wang wrote:

> 
> While dealing with the weighted averaged of _fica under 09 logic, it turns out that there are quite a few
> observations that have MARS == 2 but (wage_head + wage_spouse) == 0, for some reason. So I'd suggest to
> modify the way total is calculated a little bit to take care of this special case:
> total = np.where(np.logical_and(self.MARS == 2, (self.wage_head + self.wage_spouse) != 0), self.wage_head +
> self.wage_spouse, 0).

Total wages zero is perfectly legitimate. I can't tell what the proposed 
code does - does it just make total wage zero? Odd that that would be a 
special case. Or does it make the total missing. That shouldn't be 
necessary, the wage isn't missing, it is zero.

Is the problem not having a way to allocate the finite increment to wage 
between husband and wife? In that case there needs to be an arbitrary 
allocation, or perhaps one informed by the CPS average split for low wage 
taxpayers.

dan

GoFroggyRun:
Yes. You are right. But I was thinking if we could skip the intermediate
`total` value while get the `_earning_split` within one function. And
ruling out such case would be helpful. I guess at this moment we are happy
with the current code.

martinholmer:
@MattHJensen and @Amy-Xu,  I'm curious about the timing of work on this pull request #379.  If it is likely to be merged into the master branch sometime on Tuesday (Oct 13), then I would wait on my next pull request, otherwise I'll go ahead with mine and fix it up (to include spouse earnings) later when work on this pull request is finished.

Amy-Xu:
@martinholmer Just moved the imputations of e00200p and e00200s out of `functions.py` to `records.py`. Could you review? If everything looks good, then @MattHJensen will have this PR merged today.

martinholmer:
@Amy-Xu, @feenberg, @MattHJensen,
When splitting e00200 between taxpayer and spouse (if present) in the FICA tax calculations (at https://github.com/Amy-Xu/Tax-Calculator/blob/payroll/taxcalc/functions.py#L342-L360), how is ```_sey``` being handled?  Isn't ```_sey``` a combined taxpayer+spouse variable?  If so, should we be adding the full couple amount into both the taxpayers and the spouses OASDI payroll tax calculation?

martinholmer:
@Amy-Xu,
Is the ```_earning_split``` variable (at https://github.com/Amy-Xu/Tax-Calculator/blob/payroll/taxcalc/functions.py#L342-L344) still need?

MattHJensen:
>@Amy-Xu, @feenberg, @MattHJensen,
When splitting e00200 between taxpayer and spouse (if present) in the FICA tax calculations (at https://github.com/Amy-Xu/Tax-Calculator/blob/payroll/taxcalc/functions.py#L342-L360), how is _sey being handled? Isn't _sey a combined taxpayer+spouse variable? If so, should we be adding the full couple amount into both the taxpayers and the spouses OASDI payroll tax calculation?

@feenberg, @Amy-Xu and @martinholmer, do you think it would also make sense to impute primary and secondary _sey using the ratio of primary to secondary wages from the CPS?

martinholmer:
@MattHJensen said:
> do you think it would also make sense to impute primary [taxpayer/head] and secondary [spouse] `_sey` using the ratio of primary to secondary wages from the CPS? The assumption, of course, would be that the split between primary and secondary self-employment income follows the split of primary and secondary wage income -- this doesn't seem well founded, but I'm not sure of a better option.

I agree because it seems to me, whatever the shortcomings of your proposal, that your proposal is better than the double counting of self-employment income that seems to be going on in the current version of this pull request.    cc @Amy-Xu and @feenberg

Amy-Xu:
@MattHJensen Is this what you meant? Could you review?

Here's the table after split the self-employed income:
screen shot 2015-10-13 at 3 55 07 pm

cc @martinholmer

Amy-Xu:
@martinholmer I feel `_earning_split` is not quiet necessary in terms of algebra, but it might help users to understand the code better by separating out this variable from the rest of imputation logic?

martinholmer:
@Amy-Xu said:
> I feel _earning_split is not quiet necessary in terms of algebra, but it might help users to understand the code better by separating out this variable from the rest of imputation logic?

I probably didn't make myself clear.  My point is much more simple:  In this pull request the `EI_FICA` function in functions.py has the `_earning_split` variable as a function argument, but it is never used in that function, so why is it being passed to that function?  Here is the function definition:
```
def EI_FICA(SS_Earnings_c, e00200, e00200p, e00200s,
            e11055, e00250, e30100, FICA_ss_trt, FICA_mc_trt,
            e00900p, e00900s, e02100p, e02100s, _earning_split):
```
But the variable ```_earning_split``` is never used in the logic of the function.  Seems as if it is unnecessary to include it in the list of function arguments.  Or am I missing something?

cc @MattHJensen

feenberg:
On Tue, 13 Oct 2015, Martin Holmer wrote:

> 
> @MattHJensen said:
>
>       do you think it would also make sense to impute primary [taxpayer/head]
>       and secondary [spouse] _sey using the ratio of primary to secondary wages
>       from the CPS? The assumption, of course, would be that the split between
>       primary and secondary self-employment income follows the split of primary
>       and secondary wage income -- this doesn't seem well founded, but I'm not
>       sure of a better option.

It is ok for now.

dan

MattHJensen:
@Amy-Xu, this looks like it is ready to merge except for: https://github.com/OpenSourcePolicyCenter/Tax-Calculator/pull/379#issuecomment-147844247. 

Is that right?

Amy-Xu:
you're right. Just removed that earning_split @MattHJensen

MattHJensen:
Thanks @Amy-Xu, merging. 
cc @martinholmer

martinholmer:
@MattHJensen said:
> Thanks @Amy-Xu, merging.   cc @martinholmer

@MattHJensen and @Amy-Xu,
Pull request #379 is still open (30 minutes later) as far as I can see.  Am I missing something?

MattHJensen:
Sorry @martinholmer, seems I forgot to click merge before. It is in now.

PR #382, opened 2015-10-10
keep s006 as an an array rather than data series 

MattHJensen:
In increment year we replace s006 with a dataseries from our weights dataframe. This is a quick fix to keep s006 as an array and maintain consistency with the other records attributes. 

cc @martinholmer and @Amy-Xu

### DIFF
--- a/taxcalc/records.py
+++ b/taxcalc/records.py
@@ -396,7 +396,7 @@ class Records(object):
         # Implement Stage 1 Extrapolation blowup factors
         self._blowup(self._current_year)
         # Implement Stage 2 Extrapolation reweighting.
-        self.s006 = (self.WT["WT" + str(self.current_year)] / 100).as_matrix()
+        self.s006 = (self.WT["WT" + str(self.current_year)] / 100).values
 
     def extrapolate_2009_puf(self):
         year = 2009
--- a/taxcalc/records.py
+++ b/taxcalc/records.py
@@ -396,7 +396,7 @@ class Records(object):
         # Implement Stage 1 Extrapolation blowup factors
         self._blowup(self._current_year)
         # Implement Stage 2 Extrapolation reweighting.
-        self.s006 = self.WT["WT" + str(self.current_year)] / 100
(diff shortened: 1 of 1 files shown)


martinholmer:
@MattHJensen,  Seems fine to me, but @Amy-Xu is the one whose opinion really counts.

Amy-Xu:
So what is the data type of the resulting s006? I thought all of other puf variables are numpy arrays instead of simple arrays. In addition to maintaining consistency,  is there any other reason you want to make this change?

MattHJensen:
> So what is the data type of the resulting s006? I thought all of other puf variables are numpy arrays instead of simple arrays.

The resulting data type is an numpy array, (as_matrix, somewhat confusingly, produces a numpy array rather than a matrix).

>In addition to maintaining consistency, is there any other reason you want to make this change?

The purpose is merely to maintain consistency. I was working on a separate script to test the mtr calculations and ended up getting a little confused because e00200 and s006 had different types. Just trying to prevent other people from encountering similar confusion in the future.

Amy-Xu:
Looks good to me!

taxcalc/records.py
```diff
@@ -396,7 +396,7 @@ def increment_year(self):
         # Implement Stage 1 Extrapolation blowup factors
         self._blowup(self._current_year)
         # Implement Stage 2 Extrapolation reweighting.
-        self.s006 = self.WT["WT" + str(self.current_year)] / 100
+        self.s006 = (self.WT["WT" + str(self.current_year)] / 100).as_matrix()
```
talumbau: I think the call you want is `values`. I just ran this and it seems we have `WT` is a DataFrame, so accessing one of its columns, e.g. `WT['WT2016']` produces a Series, and to get a numpy array from that series you can access it via the `values` attribute.

MattHJensen:
>I think the call you want is values. I just ran this and it seems we have WT is a DataFrame, so accessing one of its columns, e.g. WT['WT2016'] produces a Series, and to get a numpy array from that series you can access it via the values attribute.

Just updated to use .values. @talumbau could you merge if it looks good now?

talumbau:
+1

PR #383, opened 2015-10-11
New growth rate handles

Amy-Xu:
In this PR, all the growth rate handles are refactored as the new class style in #367. In terms of functionality, this PR is equivalent to #368, which can be closed after this PR's open. Similar to the Behavior class, this Growth class would copy both the baseline and reform calculators and apply user-specified rate on both calculators. A quick demo in a notebook is as follows:

screen shot 2015-10-11 at 6 47 14 pm

This is an example changing the general economic growth rate to 4%, in real terms, for both baseline and reform (add surtax on Itemized Deduction Benefits).

**Without changing the growth rates:**
Baseline:
screen shot 2015-10-11 at 6 44 47 pm

Reform:
screen shot 2015-10-11 at 6 44 59 pm

**Increase general economic growth to 4% a year**
Baseline:
screen shot 2015-10-11 at 6 45 19 pm

Reform:
screen shot 2015-10-11 at 6 45 30 pm

Haven't been able to change the algebra of growth rates for webapp yet.


MattHJensen:
@amy-xu, sounds like calling these things during increment_year() are fine. Just on the phone with @talumbau. He might provide some more detail soon.

Amy-Xu:
Got you. Sounds good. I was wondering whether you have seen that json file/path not found error? @MattHJensen I got confused because I think I have that growth.json on this PR.

taxcalc/calculate.py
```diff
@@ -26,18 +27,24 @@ def add_df(alldfs, df):
...
+
         if isinstance(behavior, Behavior):
             self.behavior = behavior
         else:
             self.behavior = Behavior(start_year=policy.start_year)
 
+        if isinstance(growth, Growth):
+            self.growth = growth
+        else:
+            self.growth = Growth(start_year=policy.start_year)
```
talumbau: Hmm.. this idiom seems bad to me (used for both Growth and Behavior). If someone passes a Growth object for `growth`, then we set it to `self.growth`. If they pass anything else (a string, some other kind of object, whatever), we just silently make a new Growth object and assign it to `self.growth`. Basically the user could assume he/she is sending in an object for `growth` but actually ending up with something else (with no error/warning message). That seems bad. How about something like this?
```
if growth:
    if isinstance(growth, Growth):
        self.growth = growth
    else:
        raise ValueError("Must supply growth as a Growth object")
else:
    self.growth = Growth(start_year=policy.start_year
```

taxcalc/calculate.py
```diff
@@ -8,6 +8,7 @@
 from .policy import Policy
 from .records import Records
 from .behavior import Behavior
+from .growth import *
```
talumbau: suggest:

```
from .growth import adjustment, target
```
(and any other functions needed here)

Amy-Xu:
@talumbau Thanks! Will fix those very soon. I got very confused that Travis CI keeps telling me it cannot find that `growth.json` even thought all my local tests are passing. Could you take a glance of that error?

MattHJensen:
>@talumbau Thanks! Will fix those very soon. I got very confused that Travis CI keeps telling me it cannot find that growth.json even thought all my local tests are passing. Could you take a glance of that error?

Try adding it to MANIFEST.in

talumbau:
yes, I agree with @MattHJensen, I believe that adding it to the MANIFEST.in will put it in the egg file so that it can be read when running via Travis CI.

taxcalc/growth.json
```diff
@@ -0,0 +1,26 @@
+{
```
MattHJensen: @Amy-Xu could you take a crack at filling these in? Here is a suggestion, although please feel free to modify these. 

long_name: "Deviation from CBO forecast of baseline economic growth (percentage point)" 

description: "The data underlying this model are extrapolated to roughly match the CBO's projection of the economy's development over the 10-year federal budget window, with each type of economic data extrapolated at a different growth rate. This parameter allows a factor to be subtracted or added to those growth rates for the construction of the economic baseline. For example if you supply .02 (or 2%), then 0.02 will be added to the wage and salary growth rate, interest income growth rate, dividend growth rate, schedule E income growth rate, and all other growth rates used to extrapolate the underlying dataset."

long_name: "Replacement for CBO real GDP growth in economic baseline (percent)". 

description: "The data underlying this model are extrapolated to roughly match the CBO's projection of the economy's development over the 10-year federal budget window, with each type of economic data extrapolated at a different growth rate. One of the growth rates taken from the CBO is GDP growth. This parameter allows you to specify a real GDP growth rate, and all other rates will be modified to maintain the distance between them and GDP growth in the CBO baseline. For example, if the CBO growth rate for one year is 0.02 and the user enters 0.018 for this parameter, then 0.002 will be subtracted from every growth rate in the construction of the economic baseline, including wage and salary growth, interest income growth, dividend growth, and many others. 

cc @feenberg @martinholmer

taxcalc/growth.json
```diff
@@ -0,0 +1,26 @@
...
+    "_factor_adjustment":{
+        "long_name": "",
+	"description": "",
+	"irs_ref": "",
+        "start_year": 2013,
+        "col_var": "",
+        "row_var": "",
+        "row_label": ["2013"],
+        "cpi_inflated": false,
```
MattHJensen: we don't need the cpi_inflated attribute for this param.

taxcalc/growth.json
```diff
@@ -0,0 +1,26 @@
...
+    },
+     "_factor_target":{
+        "long_name": "",
+	"description": "",
+	"irs_ref": "",
+        "start_year": 2013,
+        "col_var": "",
+        "row_var": "",
+        "row_label": ["2013"],
+        "cpi_inflated": false,
```
MattHJensen: we don't need the cpi_inflated attribute for this param.

Amy-Xu:
I changed the application of adjustment and targets. More detailed description can be found in growth.json description node. Here are the tables after applying adjustments or targets. Default results remain the same.

GDP growth (real terms) at 4%:
Baseline:
screen shot 2015-10-13 at 4 48 34 pm

User-specified Plan (add 0.02 Itemized Deduction Benefit Surtax):
screen shot 2015-10-13 at 4 50 47 pm

General economic growth rate increase by 1%:
Baseline:
screen shot 2015-10-13 at 4 52 43 pm

User specified plan:
screen shot 2015-10-13 at 4 52 53 pm

@MattHJensen Could you take a look at the tables? I will add tests for this branch soon.

Amy-Xu:
@MattHJensen @talumbau just added tests for this PR. Should have finished everything I have in mind. Could you review? Thanks!

talumbau:
Here are the questions that I have:

1. Does it make sense to specify both a factor target and a factor adjustment? In my conversation with Matt, my understanding is that it does not make sense. Should we check if both are specified and raise an exception if that occurs?

2. Is it the case that the following user_mods dictionaries should produce all the same results:
   - user_mods = {}
   - user_mods = {2015: {'_factor_target': [[0.0]]}}
   - user_mods = {2015: {'_factor_adjustment': [[0.0]]}}
(this is relevant for certain ways a user could interact with the webapp)

3. following the maxim of "make the common case fast", would it be OK to implement `increment_year` as follows:

```
    def increment_year(self):
        if self.growth.factor_adjustment != 0.0:
            adjustment(self, self.growth.factor_adjustment,
                       self.policy.current_year + 1)
        if np.any(self.growth._factor_target != 0.0):
            target(self, self.growth._factor_target,
                   self.policy._inflation_rates,
                   self.policy.current_year + 1)
        self.records.increment_year()
        self.policy.set_year(self.policy.current_year + 1)
        self.behavior.set_year(self.policy.current_year)
```

MattHJensen:
@amy-xu, quick question/comment on issue #1 from @talumbau. Leaving replies on everything to you.

>Does it make sense to specify both a factor target and a factor adjustment? In my conversation with Matt, my understanding is that it does not make sense. Should we check if both are specified and raise an exception if that occurs?

Would it make sense to do this through the validations attribute? As to raising an exception in taxcalc, we might want to do this something more general and raise an exception in taxcalc when any validation rule is violated.

Amy-Xu:
@talumbau for the first question, yes we should check whether both are specified at the same time. But as @MattHJensen just pointed out here, I'm not quite sure where's the best place to implement this check. I can definitely do it in this PR if necessary.

2. Yes, if in each case that's the only thing users specified in user_mods.

3. Yes, that would make the common case fastest. And if I move those two condition from where they are right now to this new increment_year, those growth-rate cases wouldn't be slowed down either. I guess I can just go ahead and make the changes in this PR?

talumbau:
> Q3. Yes, that would make the common case fastest. And if I move those two condition from where they are right now to this new increment_year, those growth-rate cases wouldn't be slowed down either. I guess I can just go ahead and make the changes in this PR?

Sure I think it would make sense to make the changes in this PR. On my end, I am working on getting the webapp ready for these additional inputs.

taxcalc/growth.json
```diff
@@ -0,0 +1,24 @@
+{
+    "_factor_adjustment":{
+        "long_name": "Deviation from CBO forecast of baseline economic growth (percentage point",
```
talumbau: edit: "(percentage point" -> "(percentage point)"

talumbau: actually why not just "(percent)" like in `_factor_target`

Amy-Xu: Good point. I'm not sure :palm_tree:  @MattHJensen?

MattHJensen: the _factor_target allows the user to specify a GDP growth rate as a percent. The _factor_adjustment allows the user to specify a percentage point deviation from existing rates (If the existing growth rate is 10%, a 10 percent increase would make it 11%, a 10 percentage point increase would make it 20%.)

talumbau: got it. thanks

Amy-Xu:
@talumbau @MattHJensen  It doesn't seem the validation for these two parameters fits the min/max format I proposed a while ago. So I guess the easiest way to add validation would just be throwing a value error if the two parameters are non-zero at the same time. We can probably think of more general solutions later on. What do you think?

talumbau:
I would be a proponent of "ugly exception raising now, clean generalized validation behavior later" but that is just me.

Amy-Xu:
@MattHJensen Then I guess this PR should be good now.

MattHJensen:
Thanks Amy. Merging.

PR #385, opened 2015-10-13
Add taxcalc/validation tools, results, and documentation

martinholmer:
Still hoping pull request #379 will be merged soon and this pull request can be modified to generate tax filing units that have non-zero spouse earnings.


martinholmer:
@MattHJensen,  If we are going to keep anywhere close to the schedule, this pull request needs to be reviewed very quickly.  As soon as this pull request is merged into the master branch, I'll be able to prepare a tax-calculation-bug pull request.

MattHJensen:
@martinholmer, this LGTM for now. I will use and review more closely after it is merged. @talumbau, could you review as well; if there is any essential problem could you point it out here, for non-essential improvements could you open issues or followon PRs?

talumbau:
Based on a quick review, the only thing I would say is that I think it would be best to put the `validation` directory at the root level instead of inside `taxcalc`. Putting it inside `taxcalc` makes it look like a sub-package of the Python package `taxcalc`, which of course it is not. Other than that, I am fine with merging it.

martinholmer:
@talumbau said:
> Based on a quick review, the only thing I would say is that I think it would be best to put the validation directory at the root level instead of inside taxcalc. Putting it inside taxcalc makes it look like a sub-package of the Python package taxcalc, which of course it is not. Other than that, I am fine with merging it.

Well, why then, is the tests directory a subdirectory of the taxcalc directory?  Doesn't it "look like a sub-package of the Python package taxcalc, which of course it is not"?

The placement of the validation directory below taxcalc makes it clear that it is part of our suite of tests of the taxcalc package: we have unit tests in the tests subdirectory and we have cross-model validation tests in the validation subdirectory.  I don't see what the problem is.

cc @MattHJensen

talumbau:
> Well, why then, is the tests directory a subdirectory of the taxcalc directory? Doesn't it "look like a sub-package of the Python package taxcalc, which of course it is not"?

Actually, maybe it should be a subpackage. `pandas` puts a `__init__.py` in its `tests` subdirectory, making it a subpackage. `numpy` does not do this. I could go either way. The problem I am pointing out is this: putting tcl and awk code inside a subdirectory of a Python package which calls a Python script up two directories (outside of the package) would read as a pretty odd thing for someone trying to understand the layout of the code.

MattHJensen:
@martinholmer, why don't you go ahead and merge this when you are ready, and then we can follow up with an issue if needed to discuss anything outstanding.

martinholmer:
Merged into master branch per @MattHJensen instructions.

PR #388, opened 2015-10-14
Revise test.py to work with Policy class

martinholmer:
Nothing else depends on test.py, so I'll merge this myself.


MattHJensen:
Sounds good. Thanks @martinholmer

PR #389, opened 2015-10-14
Add eNNNNNp and eNNNNNs variables to Records.ZEROED_NAMES

martinholmer:
Without this change the SimpleTaxIO class, which does not call the
Records class impute_variables method, crashes because the six new
p/s variables are undefined.


martinholmer:
@MattHJensen and @Amy-Xu,
Shall I go ahead and merge this?  I can't do any work with the SimpleTaxIO class until this is fixed.

MattHJensen:
Looks good to me. Merging. 

@martinholmer, in the future you can always checkout another branch on top of a branch that is part of an open PR. I do this rarely, but it is useful in situations like this where you are highly confident that the first PR will get merged before you move to the next one.

PR #396, opened 2015-10-15
Initial fix of Calculator class mtr method

martinholmer:
This pull request fixes the mtr method to work with the new FICA tax calculation logic introduced in pull request #379.  It does that by restricting the mtr income_type_string to being only one value, 'e00200p' (taxpayer earnings).  Future development will expand the number of valid income types.  Given this income type restriction, a number of mtr calls (including two in the behavior function) had to be changed from mtr('e00200') to mtr('e00200p') in order to get all the unit tests to pass.

The revisions to the mtr method also include the addition of a new parameter called wrt_adjusted_income, which has a default value of True.  When it is True, the marginal tax rates are computed with respect to (wrt) an adjusted income concept that includes the employer share of OASDI+HI payroll taxes.

The only mtr call that uses wrt_adjusted_income=False is in SimpleTaxIO and that is done to facilitate comparison of marginal tax rates with Internet TAXSIM, the current version of which is not yet computing marginal tax rates with respect to expanded income.  When used in this way, the OSPC vs TAXSIM comparisons for a random sample of 100,000 tax filing units with no itemized-deduction expenses and no child-care expenses in 2013 are as follows:
```
TAXDIFF:ovar,#diffs,#1cdiffs,maxdiff[id]=  7     58     32    -15.00 [305]
TAXDIFF:ovar,#diffs,#1cdiffs,maxdiff[id]=  9     13      0      0.90 [2158]
TAXDIFF:ovar,#diffs,#1cdiffs,maxdiff[id]= 19    209    209     -0.01 [36]
TAXDIFF:ovar,#diffs,#1cdiffs,maxdiff[id]= 25    115    115      0.01 [2574]
TAXDIFF:ovar,#diffs,#1cdiffs,maxdiff[id]= 26      2      0  -2900.00 [95197]
TAXDIFF:ovar,#diffs,#1cdiffs,maxdiff[id]= 27    150    150     -0.01 [1761]
TAXDIFF:ovar,#diffs,#1cdiffs,maxdiff[id]= 28    132    132     -0.01 [36]
TAXDIFF:ovar,#diffs,#1cdiffs,maxdiff[id]=  4    328    328     -0.01 [2339]
```
These results show that the income tax liability (ovar=4) differences are no more than one cent, but there are a few marginal tax rate differences among the 100,000 filing units: 26 income tax rate differences greater than one basis point and 13 FICA tax rate differences greater than one basis point.  These mtr differences will be investigated in the future.


martinholmer:
@MattHJensen,  Here are the changes to get the mtr method working reasonably well for the 'e00200p' income type.  We can add other income types in the future.

MattHJensen:
Thank you @martinholmer! This LGTM. Could you merge when you are ready? 

There are some things I'd change on a more leisurely basis: 

`wrt_expanded_income` is a bit confusing as a name because we use a different kind of "_expanded_income" as the tab variable for distribution tables.

@feenberg has advocated using 0.01 as the finite_diff, so I think we should make sure the allowable range includes this. (In fact, he suggested exploring we can avoid all the kinks by first adding a penny to income and then calculating mtr using a 1 cent finite diff, since kinks tend to dollar levels. If this worked, we could avoid doing the 'up' and 'down' calculation and potentially save ourselves two calculator runs.)

martinholmer:
@MattHJensen,  Thanks for the suggestions.
I'll drop the low end of the valid finite_diff range right away.
Do you have a suggestion for an alternative to wrt_expanded_income?

feenberg:
On Thu, 15 Oct 2015, Matt Jensen wrote:

> 
> Thank you @martinholmer! This LGTM. Could you merge when you are ready?
> 
> There are some things I'd change on a more leisurely basis:
> 
> wrt_expanded_income is a bit confusing as a name because we use a different kind of
> "_expanded_income" as the tab variable for distribution tables.
> 
> @feenberg has advocated using 0.01 as the finite_diff, so I think we should make sure
> the allowable range includes this. (In fact, he suggested exploring we can avoid all
> the kinks by first adding a penny to income and then calculating mtr using a 1 cent
> finite diff, since kinks tend to dollar levels. If this worked, we could avoid doing
> the 'up' and 'down' calculation and potentially save ourselves two calculator runs.)
>

After Ohara is through with our data - is it still intergers?

dan

MattHJensen:
@martinholmer 
>Do you have a suggestion for an alternative to wrt_expanded_income?

What about wrt_employer_adjusted_income? 

or, more terse, wrt_adjusted_income?

martinholmer:
@MattHJensen said:
> What about wrt_employer_adjusted_income?  Or, more terse, wrt_adjusted_income?

Both good suggestions; like the terse one better for that reason.  Then started thinking it is not just any adjustment, it is a larger income.  So, what about `wrt_extended_income`?  Is that too close to expanded income?

MattHJensen:
I prefer adjusted to extended bc of distance from expanded, but don't feel strongly either way. Does seem valuable to know that income is bigger....

MattHJensen:
Thanks again @martinholmer. The simtax testing facility was a big help with my work this morning as well.

PR #398, opened 2015-10-15
Added blowup for e87521, American Opportunity Credit

GoFroggyRun:
I noticed that we don't have `e87521` blowed-up in records.py and thus added, please review and see if this is a good way to do that. Thanks!
@Amy-Xu


Amy-Xu:
LGTM. Thanks Sean. Merging.

PR #394, opened 2015-10-15
fully optimize itemization

MattHJensen:
@martinholmer, I believe this solves the itemization optimization issue you identified. 

There are still some differences with internet taxsim that I am trying to deal with now, but we might want to wait for another PR:

```
TAXDIFF:ovar,#diffs,#1cdiffs,maxdiff[id]= 17   2179      0  76000.00 [86567]
      #big_vardiffs_with_big_inctax_diff=                3
TAXDIFF:ovar,#diffs,#1cdiffs,maxdiff[id]= 18   2179      0 -29050.00 [54361]
      #big_vardiffs_with_big_inctax_diff=                3
TAXDIFF:ovar,#diffs,#1cdiffs,maxdiff[id]= 19   2347    168  -8027.50 [54361]
      #big_vardiffs_with_big_inctax_diff=                3
TAXDIFF:ovar,#diffs,#1cdiffs,maxdiff[id]= 22     46      0   -814.00 [78456]
TAXDIFF:ovar,#diffs,#1cdiffs,maxdiff[id]= 25    109    109      0.01 [15329]
TAXDIFF:ovar,#diffs,#1cdiffs,maxdiff[id]= 26   2317      0 -40000.00 [23353]
      #big_vardiffs_with_big_inctax_diff=                3
TAXDIFF:ovar,#diffs,#1cdiffs,maxdiff[id]= 27    736    715   7937.50 [54361]
      #big_vardiffs_with_big_inctax_diff=                3
TAXDIFF:ovar,#diffs,#1cdiffs,maxdiff[id]= 28    158     91  -7937.50 [54361]
      #big_vardiffs_with_big_inctax_diff=                3
TAXDIFF:ovar,#diffs,#1cdiffs,maxdiff[id]=  4    851    848    178.00 [2412]
                       #big_inctax_diffs=                3
```

Here is record 54361

```
input 
54361 2013 0 3 2 0 169600 0 3000 0 0 0 0 0 1900 36100 0 2000 1 0 -3000 -1000

taxsim
54361. 2013 0 34445.00 .00 19017.20 32.50 .00 2.90 171600.00 2000.00 .00 8950.00 11700.00 .00 .00 .00 150950.00 33005.00 .00 .00 .00 .00 .00 .00 171600.00 1830.00 32615.00 19017.20 174601.01 .00 .00 .00 .00 .00 .00 .00 .00 .00 .00 .00

simtax
[...]

PR #399, opened 2015-10-15
Update six .taxdiffs files to reflect changes caused by PR 394.

martinholmer:
The changes for the b1? and c1? files reflect pull request #394 improvements in the logic of the choice between standard and itemized deductions in the face of the AMT rules.  However, the changes for the a13 and a14 files, which contain only filing units with no itemized-deduction expenses, are puzzling.  One would think there should be no change in the a13.taxdiffs and a14.taxdiffs files, but that is not true.  While the total income tax liability (ovar=4) is exactly the same before and after pull request #394, the differences in four intermediate values (ovar 18, 19, 27, 28) have diverged from all essentially zero to a few having large nonzero differences.  These will be investigated in the future.


martinholmer:
@MattHJensen,  Please read the comment for pull request #399, which I have merged into the master branch.

MattHJensen:
Thanks @martinholmer, I'm seeing this now.

Issue #405, opened 2015-10-19
Bug in Travis-CI or in SimpleTaxIO calculate method


martinholmer:
Pull request #404 uncomments a single line of code in test_simpletaxio.py test_3.  The activated test_3 passes just fine on my Mac computer running Python 2.7 (as documented in the pull request #404 comment).

Why is the Travis-CI test failing?  Is it a bug in Travis-CI or a bug in the SimpleTaxIO calculate method?

The SimpleTaxIO calculate method is working fine when called from the validation/tests script on the computers of @MattHJensen and Martin Holmer, and perhaps others.

This is a critical test because its omission has already caused other check-ins to break the SimpleTaxIO class (for example, pull request #379 broke the SimpleTaxIO class and pull request #379 had to be repaired by pull request #389).  The whole point of the Travis-CI system is to prevent this sort of thing from happening.

Note: this problem was pointed out to @talumbau last week, but he got involved in the rush to release the TaxBrain webapp.

talumbau:
I was able to reproduce the failure on Mac and on Linux. It was surprising that I was able to reproduce on Mac. I'm running Mac OS 10.10.5. Here is a screen cap of my About This Mac window:

<img width="305" alt="screen shot 2015-10-19 at 12 30 25 pm" src="https://cloud.githubusercontent.com/assets/1891418/10585528/4729f520-765d-11e5-97f7-70175f4d1f15.png">

@martinholmer what version of Mac OS X are you running? Can you put in the screen cap for your machine?

talumbau:
Although I can't quantify exactly why, this fix:

https://github.com/OpenSourcePolicyCenter/Tax-Calculator/pull/406

resolves the issue. For some reason, the compiler does not it when certain record columns are int32 instead of int64. This occurs on a Linux 64 machine and a OS X 64 machine. I wondered if the tests were passing on a 32 bit machine, which is why I was asking about Marin's platform when running the tests locally.

talumbau:
One possibility that occurs to me is that the compiler might not like operations like this:

```
int32 OP float64 -> int32
```

in other words, arithmetic with a 32bit integer and a 64 bit floating point, which then gets stored to a truncated 32bit integer. The preferred size for the truncated integer would be 64 bits. I don't know for sure if that applies because I haven't looked closely at the operations for the code that triggers the error, but it seems reasonable to be concerned about mixed float/int operations where the size is different.

talumbau:
[...]

PR #406, opened 2015-10-19
Simtax8 fix

talumbau:


martinholmer:
@talumbau,  Thanks very much for finding a way to avoid the Travis-CI bug, but I'm mystified by your solution.  Remember, it was you who suggested I use the 'int32' dtype when specifying the pandas DataFrance in the SimpleTaxIO class.  The 'int32' dtype works just fine on my computer as part of simtax.py and as part of tests_simpletaxio.py test_3.  So, why doesn't it work with Travis-CI?  I'd like to learn something from this experience, but you've provided no explanation about why I need to specify 64-bit integer (rather than 32-bit integer) as the DataFrame dytpe.  What's going on with Travis-CI?
cc @MattHJensen

MattHJensen:
@martinholmer, check out the conversation over at #405. Still not a solved mystery, but there is more discussion there.

talumbau:
@martinholmer please see discussion in 
https://github.com/OpenSourcePolicyCenter/Tax-Calculator/issues/405

PR #410, opened 2015-10-21
Added Age attribute to records class

Amy-Xu:
This PR also added age as an attribute of the Record class. Could you review? @MattHJensen


Amy-Xu:
.......Why is Travis CI not running automatically?

martinholmer:
@Amy-Xu and @MattHJensen,  After a quick glance at this pull request, I have two concerns.

First, why is the ```income_type_string``` argument to the Calculator mtr method changed to a list?  There is no change to the documentation of the method and it is still called a string, which is a lie.  Why does it have to be a list?  Doesn't make any sense to me.

Second, if you want to remove the constraint that the ```income_type_string``` must be ```e00200p```, then you need to add **much** more logic than is included in this pull request.  When income type is e00200, you have to split the finite_diff between spouses when ```MARS==2``` (presumably) and not split it when ```MARS!=2```; you have to **not** increment ```e00200``` if the income type is not earnings (for example, dividends); and so on.

If you want to generalize the mtr method, you had better add a comprehensive test to make sure the generalized method is doing what you think it should do in **ALL** cases of the ```income_type_string```.

talumbau:
two comments:

- I agree that the doc string should be modified and the argument name should be changed to something that implies plurals if the argument is expected to be iterable (e.g. `income_types`)

- I think Travis CI might need to be pointed at this new repo since the org name changed. Will check it out and report back

Amy-Xu:
@martinholmer Sorry I was just generating data for our dynamic model and forgot to fix the doc string and add any explanation. 

The reason why `income_type_string` was changed to a list is that some income variables we want to calculate MTR for are actually calculated variables instead pure e-variables from PUF. Merely putting the variable name directly in the argument won't do the job properly. For example, the income we ran into yesterday was self-employed income of the primary filer (_sey_p), which is the sum of `e00900p` and `e02100p`. As you can see, if we simply replace `e00200p` with `_sey_p`, the mtr function would just add $1 on _sey_p before calculating any taxes and this $1 margin would be erased in the actual calculation because `e00900p` and `e02100p` remain the same. So one way to solve this problems is to add the margin to each e-variable component (I just realized this is a mistake when I wrote this. We need to change the logic for sure) and then do all the MTR calculations on top of that. This is the reason why we want this function to be able to take multiple e-variables.

Yes, I'll change the doc string and tests for this function. I was wondering other than e00200p and _sey_p, do you have any other case in mind that we need to cover in the tests? @martinholmer @MattHJensen Wage and self-employed income are the only two cases we ran into so far. I'm not sure what're the other possible cases.

Thanks @talumbau. Will do.

Amy-Xu:
A quick update on this PR:

- If one income type has several e-variables as components, we should add the $1 margin to one e-variable instead of all of them. Otherwise, it would be a marginal tax rate on a $n margin instead of $1 margin. So I added a break in each relevant for-loop to make sure only one e-variable involved.

- Several checks are added for the cases I can think of:
   - If the income type is any of the variables split into primary and secondary, only primary is allowed in this function
   - The income type needs to be a PUF variable at this point. This is accomplished through checking the first letter of variables. 

- The doc string and argument name has been updated.

cc @martinholmer  @MattHJensen @talumbau

Amy-Xu:
Just merged this PR with master. Right now if users want to calculate MTR on income types with several e-variable components, they can accomplish that by passing one component as `income_type_str`. For example, to calculate MTR for self-employed income, users can replace the default value of `income_type_str` with 'e00900p'. @martinholmer I added one more variable name to the list `mtr_valid_income_types `. Could you review?

This PR needs to be merged if our macro team ever wants to generate data themselves. Current master version is not enough for self-employed income MTR. @MattHJensen

martinholmer:
@Amy-Xu said:
> For example, to calculate MTR for self-employed income, users can replace the default value of income_type_str with 'e00900p'. @martinholmer I added one more variable name to the list mtr_valid_income_types. Could you review?
>
> This PR needs to be merged if our macro team ever wants to generate data themselves. Current master version is not enough for self-employed income MTR.

So are you saying that the macro model needs to compute the marginal tax rate on Schedule C self-employed income ('e00900') as a separate concept from the marginal tax rate on taxpayer earnings ('e00200p')?  If true, can you provide some background info on why that is so?

Amy-Xu:
@jdebacker I think Jason is in a better position to answer your question @martinholmer. I'm actually not quite sure why they are separate concepts. Would love to hear more about it.

Amy-Xu:
PR #449 dealt with all the issue with MTR function, and this PR currently only add the 'age' variable as an attribute to the records class. Merging.

cc @martinholmer @talumbau @MattHJensen

PR #411, opened 2015-10-22
update docs for substitution effect, income effect, and itemized deduction surtax switch. 

MattHJensen:
I added descriptions for the:

income effect:

"The taxpayer's taxable income changes according to the product of this parameter and the change in tax liability that would have occurred as a result of specified policy change in the absence of any behavioral response."

and the substitution effect:

"The taxpayer's taxable income changes according to the product of this parameter, the taxpayer's taxable income before the policy change, and the percent change in the after tax rate (1-MTR)." 

I also changed the title of the long_name of ID_BenefitSurtax_switch to 

"The surtax on itemized deduction benefits applies to the benefits derived from the itemized deductions specified with this parameter."

and I added a description of the ID_BenefitSurtax_switch:

"The surtax on itemized deduction benefits applies to the benefits derived from the itemized deductions specified with this parameter."

I think it would be helpful to add a discussion of the elasticity of taxable income more generally, but for now I just wanted to state exactly what the parameters do. 

cc @feenberg, @martinholmer

### DIFF
--- a/taxcalc/behavior.json
+++ b/taxcalc/behavior.json
@@ -1,7 +1,7 @@
 {   "_BE_inc":{
         "start_year": 2013,
         "long_name": "Income effect",
-        "description": "The taxpayer's taxable income changes according to the product of this parameter and the change in tax liability that would have occurred as a result of specified policy change in the absence of any behavioral response.",
+        "description": "Change in the demand for leisure (negative of labor supply) per penny of change in after tax income, evaluated at the PlanX level of income and deductions. That is, the work response to a lump sum tax.",
         "row_label": ["2013"],
         "col_label": "",
         "value": [0.0]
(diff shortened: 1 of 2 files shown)


feenberg:
On Thu, 22 Oct 2015, Matt Jensen wrote:

> 
> I added descriptions for the:
> 
> income effect:
> 
> "The taxpayer's taxable income changes according to the product of this parameter and
> the change in tax liability that would have occurred as a result of specified policy
> change in the absence of any behavioral response."

I think users may not understand that this is a change in effort, not a 
change in the tax code definition of taxable income. Maybe something like 
this:

"Change in the demand for leisure (negative of labor supply) per penny of 
change in after tax income, evaluated at the PlanX level of income and 
deductions. That is, the work response to a lump sum tax."


> 
> and the substitution effect:
> 
> "The taxpayer's taxable income changes according to the product of this parameter, the
> taxpayer's taxable income before the policy change, and the percent change in the
> after tax rate (1-MTR)."
>

Aren't we using "tax-home share" (mtr1-mtr0)/(1-mtr0)?

MattHJensen:
Updated and merging. Will continue conversation about take-home share in a minute.

PR #412, opened 2015-10-23
Add Python 3.5 to Travis CI testing

talumbau:


talumbau:
this worked with no changes (yay!) It seems reasonable to keep the 3.4 support as well.

PR #413, opened 2015-10-23
Internet Taxsim Validation

GoFroggyRun:
This PR deals with the issues in #394 and #399. In particular, by modifying the constraint over optimization of itemized deductions, those difference files for non-itemizers (a13 and a14) look consistent again when comparing to Internet Taxsim. It also seems to have very minor improvements toward intermediate variables in B group and C group. Please see diff files for details. There are still moderate validation errors on the total income tax liability (ovar=4) in group C (and only group C), which might due to certain minor policy changes. This will be further investigated if time permits. There are also some clean-ups for the rest of the files, which shall have no impact toward the results.
Essentially, this PR retains the optimization we have and meanwhile recovers the overall consistency against Internet Taxsim.  
@martinholmer @MattHJensen  Please take a look and let me know if you have any concerns, comments or remarks.


martinholmer:
Sean (aka @GoFroggyRun),  Very nice work!  This is exactly the sort of debugging I hoped others on the OSPC team would start doing with the SimpleTaxIO capability.

I hope @MattHJensen will allow you to continue this line of work next week.  It looks as if the biggest opportunity for improvement is in the ```c13.in``` sample where the filing unit with id 29224 has the largest tax liability difference relative to TAXSIM (ignore the ```c14.in``` sample for now).  See if you can figure out why the presence of positive childcare expenses (which are the only difference between the ```b13.in``` and the ```c13.in``` samples) causes such a large SIMTAX-vs-TAXSIM difference, being cognizant of the fact that TAXSIM could have the bug if your search for a Tax-Calculator bug is fruitless.  

It might be easier to debug the 29224 filing unit all by itself.  You can do this by temporarily editing the ```validation/tests``` script to loop through just ```c13``` and to skip the ```rm ...```  statement.  Then you'll have the ```c13.in``` file and you can extract the one filing unit as follows:
```awk "$1==29224" c13.in > case.in```
Then, ```python ../../simtax.py case.in``` will produce ```case.in.out-simtax```, which will contain just the output for filing unit 29224, which can be compared using the ```taxdiffs.tcl``` script if you extract the one line from the ```c13.in.out-taxsim``` file. The advantage of doing this is that when running just one filing unit through ```simtax.py```, you can insert ```print(cNNNNN)``` and ```print(eNNNNN)``` commands in the functions.py code.  This is a very crude style of debugging, but my impression is that the Tax-Calculator *decorator magic* implies that the Python debugger doesn't work on the code.  If I'm wrong about that, please let me know.  Also, if you find a more clever way to trace the tax calculation details of a single filing unit, please let me know.

Keep up the good work.

GoFroggyRun:
@martinholmer Thanks for your advice. I'll take your debug strategy and see if there's any possible improvements during the process.

martinholmer:
@MattHJensen,  Do you want me to merge pull request #413 into master?

MattHJensen:
@GoFroggyRun did you mean to include the first three commits and not just the fourth?

GoFroggyRun:
@MattHJensen Yes. The first three commits deal with some clean-up stuff, which shall have no impact toward the results. The forth commit modifies the constraint over optimization and improves the validation result.

MattHJensen:
Merging. Thanks @GoFroggyRun

PR #416, opened 2015-10-24
test that future-year values are overwritten by reform

MattHJensen:
Add an assertion to `test_multi_year_reform() to ensure that Policy reforms overwrite future-year values. 

I use the example that a 2015 reform of _CTC_c ought to overwrite the pre-specified values of _CTC_c for 2016-2018 that exist in current_law_policy.json.

This test passes in master because everything is working fine.


talumbau:
looks good.

PR #419, opened 2015-10-27
C_group Validation against Taxsim

GoFroggyRun:
This PR deals with issues regarding Child Tax Credit particularly. One extra intermediate value was assigned (by `ivar[17]`) to ensure the python calculator to work properly under simplified scenario. Earning splits were also implemented explicitly in function F2441, given that we now have `e00200` for both parties. One could see the improvements from the updated diff files, variable # 24 in particular, for details. 

After fixing the issue, by looking at one particular observation # 2827 in c2013 file:
```
2827 2013 0 3 2 0 4400 0 7000 0 0 0 0 0 891 127 11000 0 1 4682 0 2000,
```
the problem seems to become more interesting:
both calculators agree on the fact that this unit has zero AMT and also zero income tax, where the latter one corresponds to line 47 on 1040. In this case, [instruction for line 10 of f2441](https://www.irs.gov/pub/irs-pdf/i2441.pdf) suggests that one shall not take the [child care] credit regardlessly. The python calculator yields zero value for the credit and the Internet Taxsim, however, gives following result: (truncated)
```
2827. 2013 0 -210.00 .00 673.20 -15.00 .00 15.30 13400.00 .00 .00 8950.00 11700.00 .00 .00 .00 .00 .00 .00 .00 .00 210.00 2100.00 .00 13400.00 .00 .00,
```
where the child care credit is 2,100.00, instead of zero. If I understand the instruction correctly, the preceding example suggests that the Internet Taxsim might have minor issue on Child Care Credit. The issue has no impact toward the finial result, `var[4]`, for this particular unit since the unit has zero income tax, which might not be true in general. And I'm wondering if this could account for the reason why we are still having some difference for the remaining units. 
@martinholmer @MattHJensen, any comments, concerns or remarks? Thanks!

Attached: results under Python Calc:
```
2827. 2013 0 -210.00 0.00 673.20 -15.00 0.00 15.30 13400.00 0.00 0.00 0.00 11700.00 0.00 0.00 0.00 0.00 0.00 0.00 0.00 0.00 210.00 0.00 0.00 13400.00 0.00 0.00
```


GoFroggyRun:
@martinholmer Thanks again for the prompt enhancement.

So by playing around with `f2441`, I was able to achieve the consistency of total federal income tax liability (ovar[4]) against Internet Taxsim. It seems that two calculators have some dichotomous understanding of [qualifying person(s)](https://www.irs.gov/pub/irs-pdf/i2441.pdf) on form 2441, where I think our initial assumption, `n24=f2441`, is more reasonable. As mentioned above, there are other known but minor issues with child care credit, and I'm wondering are we happy with what we have at this moment. 

cc @MattHJensen

martinholmer:
Dan @feenberg,
The OSPC Tax-Calculator versus Internet TAXSIM validation work that Sean @GoFroggyRun has done in pull request #419 suggests that TAXSIM is approximating Form 2441 qualified persons as the total number of dependents in the tax filing unit rather than using the number of dependents under 17, which is TAXSIM input variable 19.  While I realize that both are approximations, Sean and I are wondering why TAXSIM uses all dependents to approximate Form 2441 qualified persons.
cc @MattHJensen

martinholmer:
Sean @GoFroggyRun,
No response from Dan yet.  While we wait to hear from him, why don't you get pull request #419 ready for a merge into the master branch.  This will mean syncing your master branch with the OSPC repo (so that the merged pull requests #422 and #423 are incorporated in your local master branch) and then merging master into your Cgroup_validation_Taxsim branch.  Double check that the new code produces the validation/tests results you are expecting and then push your revised branch to the OSPC repo.  It will be merged soon after you do the push.

Thanks again for the excellent validation work.

GoFroggyRun:
@martinholmer Sure. I'll let you know once this PR is ready to be merged.

GoFroggyRun:
@martinholmer All up-to-date and ready to be merged.

martinholmer:
@MattHJensen,
Do you want to review this pull request #419 and merge it into master branch?  Or should I merge?

It fixes a bug in the f2441 function and adds a couple lines of needed code to the simpletaxio.py file.
After these changes, there are no significant tax-liability differences between OSPC Tax-Calculator SIMTAX results and Internet TAXSIM results for 600,000 randomly-generated tax filing units.

Sean and I are still waiting to hear from Dan about why Internet TAXSIM uses **all** dependents (rather than dependents under 17) to approximate the number of Form 2441 qualified persons, for whom TAXSIM has a **child care expenses** dollar amount.  Depending on his response, I may add an ```--TAXSIM2441``` option to simtax.py that uses what we now think is inappropriate TAXSIM logic.  Using this option will allow SIMTAX to emulate TAXSIM in this respect, and therefore, allow SIMTAX to be used in validation work.  Not using the ```--TAXSIM2441``` option would use what Sean and I now think is the more sensible approximation of Form 2441 qualified persons.

MattHJensen:
>Do you want to review this pull request #419 and merge it into master branch? Or should I merge?

Could you merge if you have reviewed and agree with the changes? 

>It fixes a bug in the f2441 function and adds a couple lines of needed code to the simpletaxio.py file.
After these changes, there are no significant tax-liability differences between OSPC Tax-Calculator SIMTAX results and Internet TAXSIM results for 600,000 randomly-generated tax filing units.

This is excellent news. 

>Depending on his response, I may add an --TAXSIM2441 option to simtax.py that uses what we now think is inappropriate TAXSIM logic. Using this option will allow SIMTAX to emulate TAXSIM in this respect, and therefore, allow SIMTAX to be used in validation work. Not using the --TAXSIM2441 option would use what Sean and I now think is the more sensible approximation of Form 2441 qualified persons.

This makes sense to me. 

As an aside, do you think it would be helpful for our contributor guide to say something about making sure to use SIMTAX and include changes to the .taxdiff files when they make substantial changes?

martinholmer:
@MattHJensen said:
> As an aside, do you think it would be helpful for our contributor guide to say something about making sure to use SIMTAX and include changes to the .taxdiffs files when they make substantial changes?

Yes, probably so, but the more important question is how we should add (part of?) the validation/tests to the Travis-CI routine.  Now that we have achieved some kind of validation success (vis a vis Internet TAXSIM), the question is how do we guard against developers making changes that lead to regressions in the validation tests results.  Currently, there is no way for developers to know they have made a mistake because the unit tests are way too weak to detect something like this.

What are your thoughts on this matter?

PR #421, opened 2015-10-27
example celery module; flask celery status server; test for it

PeterDSteinberg:
celery_status_server.py  runs flask server on port 5070.  URLs:

#### Register a job with email, id, callback
/register_job?email=me@email.com&ticket_id=abc&callback=somewhereelse.com/callback
#### Get all statuses
/celery_status_server
#### Example async delay function
/example_async

test_celery_status_server.py tests the server above when it is running.  Requires celery and redis running.

example_async.py is just an example celery app with timer sleep function.

### DIFF
--- a/celery_status_server.py
+++ b/celery_status_server.py
@@ -1,0 +1,110 @@
+#celery_status_server.py
+
+from flask import Flask, request, make_response
+from celery import Celery
+from celery.app.control import Inspect
+import datetime
+import requests
+import os
+import json
+import time
+import sys
+import subprocess as sp
+from functools import partial
+import example_async
+"""
+nohup redis-server &
+nohup celery -A example_async worker -P eventlet -l info &
+"""
+server_app = Flask('celery_status_server')
+port = 5070
+server_url = "http://localhost:{0}".format(port)
+
+TICKET_IDS_WATCHED = {}
+m = '--module'
+if not m in sys.argv:
+    raise ValueError('needs --module that has the celery_app importable from here')
+else:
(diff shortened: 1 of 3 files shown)


celery_status_server.py
```diff
@@ -0,0 +1,110 @@
...
+
+@server_app.route('/example_async', methods=['GET'])
+def example():
+
+    ticket_id =  example_async.example_async.delay()
+    print('ticket_id', ticket_id)
+    url = '{0}/register_job?email={1}&ticket_id={2}&callback="http://google.com"'.format(server_url, ticket_id, ticket_id)
+    return json.dumps({'ticket_id': str(ticket_id)})
+
+@server_app.route("/register_job", methods=['GET'])
```
talumbau: I think this should be a POST, right?

PeterDSteinberg: Yes that's a better idea.  Will change and move soon.

talumbau:
Before moving on though, let's close this PR and open on webapp-public

celery_status_server.py
```diff
@@ -0,0 +1,110 @@
...
+                                                         'eta': eta,
+                                                         }
+                    else:
+                        pass
+    for key in TICKET_IDS_WATCHED:
+        if key not in return_val:
+            return_val[key] = {'inputs': TICKET_IDS_WATCHED[key].copy(),
+                               'status': 'done'}
+            to_pop.append(key)
+    response = {}
```
PeterDSteinberg: It should be possible here to call AsyncResult() to see if the job is really done or if the ticket id was never processed.  Currently either of those cases would show up here as 'done' status.

PR #422, opened 2015-10-27
Marginal tax rate function bug fix

Amy-Xu:
Variable `earnings_type` saves the original `e00200` (Wage and Salaries) values and remains the same throughout the MTR calculation. But this variable was considered as the changed `e00200` attribute value in current master. Thus one extra dollar was deducted in the downside scenario and another extra dollar was added in the reset step. This PR fixed this issue. @MattHJensen Could you review?

### DIFF
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -274,9 +274,9 @@ class Calculator(object):
 
         # Calculate the tax change with a marginal decrease in income.
         setattr(self.records, income_type_string,
-                income_type - 2 * finite_diff)
+                income_type - finite_diff)
         setattr(self.records, 'e00200',
-                earnings_type - 2 * finite_diff)
+                earnings_type - finite_diff)
         self.calc_all()
 
         _ospctax_down = copy.deepcopy(self.records._ospctax)
@@ -301,8 +301,8 @@ class Calculator(object):
 
         # Reset the income_type to its starting point to avoid
         # unintended consequences.
-        setattr(self.records, income_type_string, income_type + finite_diff)
-        setattr(self.records, 'e00200', earnings_type + finite_diff)
+        setattr(self.records, income_type_string, income_type)
+        setattr(self.records, 'e00200', earnings_type)
         self.calc_all()
 
         # Choose the more modest effect of either adding or subtracting income


MattHJensen:
@Amy-Xu, could you post some summary statistics? Medians by agi bin would be great, here is a script I've used in the past:


    import pandas as pd
    from taxcalc.policy import Policy
    from taxcalc.records import Records
    from taxcalc.calculate import Calculator
    from taxcalc.utils import add_income_bins
    
    mtr_FICA, mtr_IIT, mtr_combined = calc.mtr('e00200p')
    all_mtrs_df = pd.DataFrame({'mtr_FICA': mtr_FICA,
                                'mtr_IIT': mtr_IIT,
                                'mtr_combined': mtr_combined,
                                '_expanded_income': calc.records._expanded_income,
                                'weights': calc.records.s006})
    
    add_income_bins(all_mtrs_df)
    grpd = all_mtrs_df.groupby('bins')
    print grpd.median()[['mtr_FICA', 'mtr_IIT', 'mtr_combined']]

MattHJensen:
This LGTM by the way. I just want to see the summary statistics to ensure that you haven't uncovered anything else going on with the mtr function that should be dealt with now.

martinholmer:
@Amy-Xu,
Can you explain why you think the master branch code contains a bug?
cc @MattHJensen

Amy-Xu:
@MattHJensen do you just want to see the statistics from this branch? or a comparison with OSPC master?

MattHJensen:
Just this branch is fine.

Amy-Xu:
@martinholmer At the beginning I just found the sum of all e00200 was slightly higher after running mtr() than before, which shouldn't happen because mtr() is not supposed to change any e-variables. Then I traced all the way back to the few lines I revised here.

As you can see in the old version, we saved e00200 in `earnings_type` and used this `earnings_type` throughout the entire calculation. Say if the original e00200 is 100, what's supposed to happen is:

                                  e00200                  earnings_type
default                          100                           100
add $1 margin              101                           100
minus $1 margin           99                            100
reset                            100                            100

But what actually happened in `mtr()` was:

                                  e00200                  earnings_type
default                          100                           100
add $1 margin              101                           100
minus $1 margin           98                            100
reset                            101                            100

which is not quite right. After I make the change in this PR, it should be fixed. And the sum of e00200, e00900, e02100 remains the same after the MTR calculation.

martinholmer:
@Amy-Xu said:
>As you can see in the old version, we saved e00200 in earnings_type and used this earnings_type throughout the entire calculation. Say if the original e00200 is 100, what's supposed to happen is:

>e00200	earnings_type
>default	100	100
>add $1 margin	101	100
>minus $1 margin	99	100
>reset	100	100

>But what actually happened in mtr() was:

>e00200	earnings_type
>default	100	100
>add $1 margin	101	100
>minus $1 margin	98	100
> reset	101	100

>which is not quite right. After I make the change in this PR, it should be fixed. And the sum of e00200, e00900, e02100 remains the same after the MTR calculation.

OK, I'll take your word for it, but the ```-2*finite_diff``` code is straight out of the original version by @MattHJensen.  And your bug fix increases my confusion about how Python works.  Several weeks ago T.J. @talumbau and @MattHJensen assured me that even though the official Python documentation for ```getattr()``` says it returns a **value** that what it actually does is return a **pointer** to the attribute.  So, the ```-2* finite_diff``` code seemed sensible to me in that light because the first ```setattr()``` call raised the values of the attribute in the calc.records object.  And so to get below the original values, one would need to subtract one finite_diff to get back to the original and then subtract one more to get below the original values.

But your explanation seems to be saying that that it is a **value** that is returned from ```getattr()```.  If so, I am now thoroughly confused about how Python works.  Can somebody enlighten me on this topic, so I don't make errors in the future.

Amy-Xu:
@MattHJensen Here's the stat table. Could you take a look?
screen shot 2015-10-27 at 3 15 34 pm

martinholmer:
@Amy-Xu,
These aggregate tables are interesting, but you might want to run the taxcalc/validation/tests script on your branch with the mtr() bug fix.  It seems to show that the few FICA mtr differences between the OSPC Tax-Calculator and Internet TAXSIM go away (which is a good thing) but that the number of INCTAX mtr differences increase and the maximum difference is much larger (which is a bad thing).  Do you have any thoughts on why things would get worse in the Tax-Calculator vs TAXSIM comparison after fixing the mtr() bug?  Maybe you should look closely at some of the big differences in the validation/tests taxdiff results on INCTAX mtr, which is output variable 7.
cc @MattHJensen and Sean (@GoFroggyRun) Wang

Amy-Xu:
@martinholmer you're definitely right on that. This is a way simpler issue. If you take another look, the `- 2 * finite_diff` was deduced from `earnings_type` rather than `e00200`. And `earnings_type` stays at 100 all the time, and you would end up with (100 - 2) = 98. You were also having `e00200` in your mind, and thought the 2 was deducted from the previous `e00200` of 101. But this `earnings_type` is just a standalone constant. You see my point?

martinholmer:
@Amy-Xu said:
> you're definitely right on that. This is a way simpler issue. If you take another look, the - 2 * finite_diff was deduced from earnings_type rather than e00200. And earnings_type stays at 100 all the time, and you would end up with (100 - 2) = 98. You were also having e00200 in your mind, and thought the 2 was deducted from the previous e00200 of 101. But this earnings_type is just a standalone constant, with value equal to the original e00200 at 100. You see my point? 

Sorry to be so dense.  Yes, I think I do see your point, but your point seems to be that "earnings_type  is just a standalone constant [returned by the getattr() call]".  That's what I thought the Python documentation said, but I was told later that getattr() returns a pointer to the attribute (which is a constant address location in memory).  So, I'm still confused.

But this is my own problem, not yours.  If you have reliable evidence that the mtr() code on the master branch is changing the values of e00200, then by all means fix that error.

MattHJensen:
>If you have reliable evidence that the mtr() code on the master branch is changing the values of e00200, then by all means fix that error.

Merging, but don't mean to kill the conversation on getattr()

Amy-Xu:
@martinholmer I guess I didn't read your reply carefully. So you meant this line `earnings_type = getattr(self.records, 'e00200')` actually returns a pointer p instead of the value *p? Then I'm confused as well, even though my results confirmed me I should modify it as if it were *p.

@talumbau @MattHJensen any comments?

talumbau:
I don't think we will enhance understanding by talking about pointers, just for the simple reason that they don't exist in the Python language. I don't understand the full context of this discussion, but I can comment on certain lines of code.

For this conversation, it seems we are discussing this line of code:

```
earnings_type = getattr(self.records, 'e00200')
```

The first thing I wonder about is why this line of code is not written as follows:

```
earnings_type = self.records.e00200
```

Is it because of a warning from pylint? I'm just wondering because the typical use case for `getattr` is something like this:

```
x = getattr(myobj, someattr)
```
where `someattr` is a variable, the contents of which is the name of an attribute on `myobj`. Doing `getattr` with a string literal (e.g. `'e00200'`) is usually reserved for certain special cases. Anyway, for the line of code above (in either form) it appears that the question on the table is something like "Is `earnings_type` a value or a pointer?" The best way to answer this question is as follows: it is neither. However, I don't think that's very helpful. Here is a statement that I think is helpful:

After the execution of either of those lines of code, the variable `earnings_type` is a name for the value associated with the attribute `e00200` on the object `self.records`. Does that sentence help clarify the issue? If not, what is unclear? The best way to understand Python variables is to view them as names for objects. Creating a Python variable is like putting a name tag on a piece of data. It's a different concept from pointers and "references", which are a related concept in Java and C++.

Amy-Xu:
@talumbau I guess the format of `earnings_type = getattr(self.records, 'e00200')` is just a replication of the line before which is `income_type = getattr(self.records, income_type_string)`. The `getattr` is much needed in this `income_type` line because `income_type_string` is a user-specified attribute name. After this `income_type` line we were like, 'users most likely want to see the MTR from `e00200`. Let's calculate this no matter what'. So even though `getattr` is not necessary, it's still kept there.

I know I probably get quite stubborn and annoying if I keep following up on this again. But still couldn't resist: 
>After the execution of either of those lines of code, the variable earnings_type is a name for the value associated with the attribute e00200 on the object self.records.

The unclear part to me is whether this value association is maintained throughout the entire function or just at the time point when `earnings_type`'s defined and initialized. If `earnings_type` is a name for the value associated with `e00200` continuously in real time, it means whenever `e00200` changed, `earnings_type` would be associated to the new value automatically as well. If it's the second case, then the association broke right away and `earnings_type` stuck with the value of `e00200` at that particular time point. Although my testing results told the latter case is closer to what actually happened, still want to confirm from a theoretical perspective.

talumbau:
Oh I see. This is helpful. One thing that trips people up in this scheme of "variables are names" seems to be when we are dealing with immutable values like integers. So for example:

```
a = 14
b = a
a = a + 2
```
what is `b` then? The answer is that `b` is still 14. Here is the way to think about this line by line:
```
a = 14
```
The name `a` is associated with the immutable object we know as the integer 14.
```
b = a
```
The name `b` is associated with the immutable object we know as the integer 14
```
a = a + 2
```
After some math happens, the right hand side is an immutable object we know as the integer 16. `a` is now a name associated with that immutable object.
Nothing happened to `b` which is still a name pointing to the object associated with the integer 14. Does that help?

talumbau:
Let me focus on numpy arrays as well:

```
>>> import numpy as np
>>> x = np.array([1,2,3])
>>> y = x
>>> x = x + 2
>>> y
array([1, 2, 3])
```
Line by line this is as follows:

```
>>> x = np.array([1,2,3])
```
x is now a name for the numpy array [1,2,3]
```
>>> y = x
```
y is now a name for the numpy array [1,2,3]
```
>>> x = x + 2
```
We evaluate the right hand side and get a new numpy array [3,4,5]. `x` is now a name for the numpy array [3,4,5]
```
>>> y
```
y is still a name for the numpy array [1,2,3]

Amy-Xu:
Thanks T.J! @talumbau  This is much clearer to me now.

PR #424, opened 2015-10-28
GDP growth rates and inflation rates validation

Amy-Xu:
In this PR, 
- changed default target growth parameter to current real GDP growth rates from CBO projection
- add test to check that hard-coded GDP growth rates and inflation rates conform across all the objects


MattHJensen:
Bringing in @talumbau for test review as well.

MattHJensen:
@Amy-Xu, could you post the baselines when you add a point to baseline growth on this branch versus on master and explain any differences if there are any?

taxcalc/calculate.py
```diff
@@ -189,14 +189,16 @@ def calc_all_test(self):
 
     def increment_year(self):
         if self.growth.factor_adjustment != 0:
-            if np.any(self.growth._factor_target) != 0:
+            if np.array_equal(self.growth._factor_target,
```
talumbau: probably a better choice is:
```
if not np.array_equal((self.growth._factor_target, self.growth.REAL_GDP_GROWTH):
```

taxcalc/calculate.py
```diff
@@ -189,14 +189,16 @@ def calc_all_test(self):
...
+                              self.growth.REAL_GDP_GROWTH) is False:
                 msg = "adjustment and target factor \
                        cannot be non-zero at the same time"
                 raise ValueError(msg)
             else:
                 adjustment(self, self.growth.factor_adjustment,
                            self.policy.current_year + 1)
-        elif np.any(self.growth._factor_target) != 0:
+        elif np.array_equal(self.growth._factor_target,
```
talumbau: same here

talumbau:
I made a couple of style comments. It looks like this also fixes the issue of 'list of list' for the growth parameters. Looks good.

Amy-Xu:
This PR:
pr

master:
master

The difference I think is because I changed the inflation rate [here](https://github.com/open-source-economics/Tax-Calculator/pull/424/files#diff-9a3aee4aed7af3316665aed6977a7f1bR57) to make Stage I factors and our hard-coded CPI rates conform with each other.

Amy-Xu:
Thanks @talumbau. Just changed according to your notes!

Amy-Xu:
@MattHJensen Just found a bug and fixed it. Here's the diagnostic tables under 0.01 adjustment.

This PR:
pr 0 1 adjustment

Master:
mastser 0 1 adjustment

Amy-Xu:
@MattHJensen Here's the diagnostic tables under 0.04 targets

This PR:
pr 0 04 target

Master:
master 0 04 target

The difference is pretty huge on a 10-year span, because we changed the algebra of how we apply this target, aka from aggregate to per capita.

martinholmer:
@Amy-Xu,  Why are you accessing a **private** member of the Policy class from the Calculator code?  That is not good programming practice; you should be using only the **public** API of a class when writing code that uses that class.  (Now, I know that there is a long-standing and massive exception to this rule in the unit tests, but I'm OK with that.)  Why are you accessing ```policy._inflation_rate``` in the Calculator class?  If you need additional **public** access methods to Policy, add them yourself or, better yet, raise an issue about your need.

martinholmer:
@Amy-Xu, My apologies.  On further investigation I see that somebody else (not you) added the calculate.py code that accesses private members of the Policy class.  I'll fix this soon; you don't have to do anything with regards to this matter.

martinholmer:
@Amy-Xu,  I see that in this pull request #424 you are testing that the default inflation rates in the Policy class are equal to the inflation rates implied in the blowup factors.  That's good; thanks for doing this.

Amy-Xu:
@martinholmer No worries! Thanks for reviewing the code!

Amy-Xu:
I'm done with this PR. @MattHJensen

martinholmer:
@MattHJensen and @Amy-Xu,  Amy is done, but do these changes conflict with the changes in #433?

MattHJensen:
@martinholmer said: 

> but do these changes conflict with the changes in #433?

These changes may conflict with those in #433, but I advocate that we merge this first for the following reasons: 

- The approach in #433 has not been reviewed yet. 

- #424 provides two bug fixes, which I tend to think should be prioritized. 

- Review for #428, another bug fix, is waiting on #424 (because the two contain many of the same commits. 

@martinholmer, does that make sense to you? Afterwards, you could add a commit to #433 to fix merge conflicts or you could ask if @Amy-Xu would open a PR to your branch to fix merge conflicts with master. There could be a reason to do #433 first that I am missing, in which case we would hold off on #424 and #428 until #433 has been reviewed and committed.

MattHJensen:
Ahha, @martinholmer and @Amy-Xu, looks like #433 is ready to go. https://github.com/open-source-economics/Tax-Calculator/pull/433#issuecomment-153112976. @Amy-Xu, would you mind if we merged that first? I think it might be easier for you to deal with the conflicts in this PR and #428 rather than in that PR.

Amy-Xu:
I can do the merge.

Amy-Xu:
@MattHJensen this PR should be good now.

taxcalc/tests/test_growth.py
```diff
@@ -87,13 +86,13 @@ def test_target():
...
     # set target
-    factor_x = {2015: {'_factor_target': [[0.04]]}}
+    factor_x = {2015: {'_factor_target': [0.04]}}
     growth_x.update_economic_growth(factor_x)
 
     target(calc_x, growth_x._factor_target, IRATES, 2015)
 
-    assert calc_x.records.BF.AGDPN[2015] == 0.04 + 0.22 + 1
-    assert calc_x.records.BF.ATXPY[2015] == 1.2572996865676191
+    assert calc_x.records.BF.AGDPN[2015] == 1.0522542782635027
```
MattHJensen: Will this test fail if we ever change the default `pirates` or gdp growth rates? Do we want it to fail in that scenario? 

It seems it would be helpful to test more generally that if the user says GDP growth rate should be 4%, and that is x percentage points different than the current GDP growth rate, then the aggregate variables like wages becomes x percent different the next year than they otherwise would have been (could pick a few).

MattHJensen:
@Amy-Xu, before I merge this, I want to verify that my interpretation of what is happening with _factor_target is correct: 

If CBO says the GDP growth rate is 2% and the user says it should be 3%, then we population adjust the difference and add that population-adjusted difference to all of our other blowup factors. We can expect all *aggregate* variables to which a blowup factor is applied to then be 1% bigger the next year than they otherwise would have been. 

So the user is *not* thinking in terms of population-adjusted GDP growth, we just have to make a population adjustment in the background to implement a plain-old real gdp growth target. 

Is that correct?

feenberg:
On Tue, 3 Nov 2015, Matt Jensen wrote:

> 
> @Amy-Xu, before I merge this, I want to verify that my interpretation of
> what is happening with _factor_target is correct:
> 
> If CBO says the GDP growth rate is 2% and the user says it should be 3%,
> then we population adjust the difference and add that population-adjusted
> difference to all of our other blowup factors. We can expect all aggregate
> variables to which a blowup factor is applied to then be 1% bigger the
> next year than they otherwise would have been.

Not sure what you are saying, but I think we should assume that the user 
believes percapita growth will be larger, not population. If the 
population growth is fixed, then doesn't 3% aggregate growth mean we 
should increase all blowup factors by 1 point? Maybe I am missing some 
simple algebraic point?

dan

Amy-Xu:
@MattHJensen Yea that's what we are doing right now. If you add 1% to any current year, you can see the AGI difference of planX and planY of that year is approximately 1% of previous year planX AGI.

@feenberg I'm not quite sure whether I got your idea right or not, but our blow up factors are on per capita basis. Yes, when the users give us a aggregate GDP growth rate, they assume per capita growth will be larger. But this per capita growth is still smaller than the aggregate growth due to the population growth rate. For example, if the total GDP grows faster by 1%, per capita growth would be 1% over population growth rate larger. If we apply the 1% directly to blow-up factors (which is per capita basis), we are going to over-blow-up the aggregate by (1 + population growth).

feenberg:
I see. The difference is small but we might as well get it right.

dan

taxcalc/tests/test_growth.py
```diff
@@ -87,13 +86,22 @@ def test_target():
...
-    assert calc_x.records.BF.AGDPN[2015] == 0.04 + 0.22 + 1
-    assert calc_x.records.BF.ATXPY[2015] == 1.2572996865676191
+    distance = ((growth_x._factor_target[2015 - 2013] -
+                 growth_x.default_GDP_growth_rates(2015 - 2013)) /
+                calc_x.records.BF.APOPN[2015])
+    AGDPN_post = AGDPN_pre + distance
+    ATXPY_post = ATXPY_pre + distance
+
+    assert calc_x.records.BF.AGDPN[2015] == AGDPN_post
+    assert calc_x.records.BF.ATXPY[2015] == ATXPY_post
```
MattHJensen: This seems like an improvement for testing the blowup factors. Do you think it would also be useful to test the following: 

"if the user says GDP growth rate should be 4%, and that is x percentage points different than the current GDP growth rate, then the aggregate variables like wages becomes x percent different the next year than they otherwise would have been (could pick a few)."

Amy-Xu: Yes, that would be helpful. But the aggregate variables are approximately x percent different, instead of exactly x percent. I think senior growth rate is the reason why this is happening. Do you think we should make them exactly the same in the target function? It might take some time.

MattHJensen: Could you add a test that they are approx the same for now and open an issue that we should come back and make them exactly the same (also document in the test why it is approx not exact)?

Amy-Xu:
@MattHJensen have you ever run into negative aggregate AGI when using 91 puf?

taxcalc/tests/test_growth.py
Amy-Xu: Have you run into negative aggregate AGI when using 91 puf? If you take a look at the new test function I just added, somehow it's giving me negative AGI for 2013.

Amy-Xu:
Checking actual e-variables is not very feasible in our test suite because of 91 puf data and extrapolation. Just to list a few results from changing growth targets here. @MattHJensen 

If we change the target to .391 for 2015, which is exactly 0.01 higher than default GDP growth rate,

|                                  | 2014 baseline | 2015 baseline | 2015 1pts | Delta | as of 2014 |
| -------------------------- |:-----------------------:| --------------:| -------:| -------:| --------------:|
|Wage and Salaries |6.746e+12 | 7.059e+12 | 7.127e+12 | 0.068e+12| 0.0099|
|Unemployed Compensation | 5.839e+11 | 6.179e+11| 6.239e+11 | 0.060e+11| 0.0102|

MattHJensen:
That's close! Merging.

Issue #426, opened 2015-10-29
test against simple internet taxsim reforms


MattHJensen:
The SIMTAX testing facility allows us to test our result against internet TAXSIM's. We should test both the baseline as well as a few reforms. Good reforms to start with would be these:

```
18	
0	No surtax on top rate
x	Surtax for the top rate (such as .05)
19	
0	No surtax on the 2nd to last bracket
x	Surtax for the rate below the top rate
75	
Y	Add $Y to Standard Deduction (may be negative)
0	No change to Standard deduction.
```

http://users.nber.org/~taxsim/taxsim-calc9/plans.html

@GoFroggyRun, could you take the lead on this? 

cc @martinholmer, @feenberg

GoFroggyRun:
Sure. Got you.

MattHJensen:
Testing itemized deductions by discontinuing the phaseout and testing AMT by eliminating it would also be useful. 
```
3	
0	Itemized deductions phaseout continued after 2010
1	Itemized deductions phaseout discontinued after 2010
50	
0	Suppress AMT
1	Include AMT
```

martinholmer:
@MattHJensen and Sean (@GoFroggyRun) Wang,
I may be mistaken, but I don't think that TAXSIM can be used to simulate tax policy reforms.  You should check with Dan @feenberg before starting any work on this.  (There is a limited ```option``` code system in Internet TAXSIM, but, as far as I know, there is not a general capability of specifying policy reforms.)

MattHJensen:
@martinholmer, you are right that there is no general capability for specifying policy reforms in internet TAXSIM. The resolution of this issue would rely on using the internet TAXSIM `option` code system, and the numbers in my two comments above refer to the `options` that would need to be used to simulate the reform in question.

martinholmer:
@MattHJensen said:
> you are right that there is no general capability for specifying policy reforms in internet TAXSIM. The resolution of this issue would rely on using the Internet TAXSIM option code system, and the numbers in my two comments above refer to the options that would need to be used to simulate the reform in question.

OK, then that will work.

feenberg:
On Thu, 29 Oct 2015, Martin Holmer wrote:

> 
> @MattHJensen and Sean (@GoFroggyRun) Wang,
> I may be mistaken, but I don't think that TAXSIM can be used to simulate tax policy
> reforms. You should check with Dan @feenberg before starting any work on this. (There
[...]

PR #434, opened 2015-11-02
Add positive spouse earnings to input samples used in validation tests

martinholmer:
The original version of make-in.tcl generated input samples in which all married couples had spouse earnings set to zero.  This was done because at that time the Tax-Calculator did not correctly calculate FICA taxes for couples in which combined earnings were more than roughly $120,000, so spouse earnings would have generated many differences with Internet TAXSIM results that were caused by a known problem that was in the process of being corrected.

Now that the Tax-Calculator computes FICA taxes correctly, the make-in.tcl program has been revised to generate random samples of tax filing units that contain couples with positive taxpayer and positive spouse earnings.  Using these six new input samples, there are a number of significant income tax liability (TAXSIM output variable 4) differences that need to be investigated.  For example, for the a13.in sample there are now 30 (out of 100,000) differences with the largest being $315.75 (for unit with id 3776).  And there are 400 differences in the c13.in sample.  Before adding positive spouse earnings, there were no differences (greater than one cent in absolute value) in any of the six samples.


MattHJensen:
@GoFroggyRun, could you review this pr?

GoFroggyRun:
Allowing spouse to have positive income would definitely be a more general assumption. There are, however, some differences between two calculators. By taking a brief look at the diff files, I think [ovar] 27 might be able to explain the errors we are having. @martinholmer What do you think? 
@MattHJensen This PR looks good to me, but my concern is: should we merge this PR after #426 being taken care of? If not, then I think we might need to figure out the errors in this PR first before working on #426 .

MattHJensen:
>This PR looks good to me, but my concern is: should we merge this PR after #426 being taken care of? If not, then I think we might need to figure out the errors in this PR first before working on #426 .

@GoFroggyRun, this is up to you. Let us know what you think would be most efficient and we'll merge this when you tell us to.

GoFroggyRun:
@MattHJensen Given that I'm not sure about how long it might take to fix the issues in this PR, and I consider most policy reforms I'm looking at are independent of this earning split issue. I was hoping this PR to be merged after #426 looks good, which hopefully won't take too long if everything goes well.

martinholmer:
@MattHJensen and Sean (@GoFroggyRun) Wang,
If you want to do the reform work in #426 first, OK.  But it seems to me that if we can't even get current law tax liability correct when spouses have earnings, then it is not very convincing to say that several reforms were validated but only under the assumption that there are no spouse earnings.  Seems like the reform tests would have to be redone with samples that contained positive spouse earnings in order for those reform validation tests to be convincing.  So what's the point of doing the reform validation tests twice?

GoFroggyRun:
@martinholmer Yes you are right. We should deal with this issue before working on #426 .
@MattHJensen Please merge this PR.

MattHJensen:
Merging. Thanks @martinholmer and @GoFroggyRun

PR #435, opened 2015-11-02
Add parameter to exclude taxable interest income from AGI

Amy-Xu:
In this PR, one more haircut parameter is added for taxable interest income to exclude it from AGI and therefore regular tax. There're two relevant issues:

- Currently this parameter is named as `_ALD_Interest_HC`, which means above-the-line deduction haircut for interest income. But obviously interest income is not a deduction. On top of that, I'm not quite sure whether this is technically a haircut. Algebra is pretty simple for this parameter, but asking for naming advice. @MattHJensen @martinholmer @feenberg 

- @MattHJensen mentioned that all this kind of parameters would influence AGI amount. We want to maintain the AGI decile/bins the same as current law for easy comparison. To make sure our tables use the same set of AGI results, I'm going to create a new function calculating and storing current law AGI in a separate PR. Have not started to implement it yet. Welcome any comments/suggestions.

### DIFF
--- a/taxcalc/current_law_policy.json
+++ b/taxcalc/current_law_policy.json
@@ -1112,6 +1112,19 @@
         "col_label": "",
         "value":     [3]        
     },
+    "_ALD_Interest_HC":{
+        "long_name": "Interest income haircut",
+	"description": "This pencentage can be applied to limit the interest income included in AGI. ",
+        "irs_ref": "Form 1040, line 8a",
+        "notes": "The final taxable interest amount will be (1- Haircut) * Interest. ",
+        "start_year": 2013,
+        "col_var": "",
+        "row_var": "",
+        "row_label": ["2013"],
+        "cpi_inflated": false,
+        "col_label": "",
+        "value":     [0]        
+    },
     "_ALD_StudentLoan_HC":{
         "long_name": "Adjustment for student loan interest haircut",
 	"description": "This pencentage can be applied to limit the student loan interest adjustment allowed. ",
--- a/taxcalc/current_law_policy.json
+++ b/taxcalc/current_law_policy.json
@@ -1116,7 +1116,7 @@
         "long_name": "Interest income exclusion",
 	"description": "This pencentage can be applied to limit the interest income included in AGI. ",
         "irs_ref": "Form 1040, line 8a",
-        "notes": "The final taxable interest amount will be (1- Haircut) * Interest. ",
+        "notes": "The final taxable interest amount will be (1- exclusion) * Interest. Even though this portion of interest wouldn't be included in AGI, it still contributes to Net Investment Income Tax and Earned Income Tax Credit.",
         "start_year": 2013,
         "col_var": "",
         "row_var": "",
--- a/taxcalc/current_law_policy.json
+++ b/taxcalc/current_law_policy.json
@@ -1112,8 +1112,8 @@
         "col_label": "",
         "value":     [3]        
     },
-    "_ALD_Interest_HC":{
-        "long_name": "Interest income haircut",
+    "_ALD_Interest_ec":{
+        "long_name": "Interest income exclusion",
 	"description": "This pencentage can be applied to limit the interest income included in AGI. ",
         "irs_ref": "Form 1040, line 8a",
         "notes": "The final taxable interest amount will be (1- Haircut) * Interest. ",
--- a/taxcalc/functions.py
+++ b/taxcalc/functions.py
@@ -90,7 +90,7 @@ def Adj(e35300_0, e35600_0, e35910_0, e03150, e03210, e03600, e03260,
 
 
 @iterate_jit(nopython=True)
-def CapGains(e23250, e22250, e23660, _sep, _feided, FEI_ec_c, ALD_Interest_HC,
+def CapGains(e23250, e22250, e23660, _sep, _feided, FEI_ec_c, ALD_Interest_ec,
              ALD_StudentLoan_HC, f2555, e00200, e00300, e00600, e00700, e00800,
              e00900, e01100, e01200, e01400, e01700, e02000, e02100,
              e02300, e02600, e02610, e02800, e02540, e00400, e02400,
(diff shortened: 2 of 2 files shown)


martinholmer:
@Amy-Xu,  Pull request #435 passes all the unit test, but how about the validation/tests?  Is there no change in federal income tax liability when you run the validation/tests?  If not, are Tax-Calculator-versus-Internet-TAXSIM differences to be expected as a result of these changes?

feenberg:
On Mon, 2 Nov 2015, Amy Xu wrote:

> 
> In this PR, one more haircut parameter is added for taxable interest income to exclude
> it from AGI and therefore regular tax. There're two relevant issues:
>
>  * 
>
>     Currently this parameter is named as _ALD_Interest_HC, which means above-the-line
>     deduction haircut for interest income. But obviously interest income is not a
>     deduction. On top of that, I'm not quite sure whether this is technically a
>     haircut. Algebra is pretty simple for this parameter, but asking for naming
>     advice. @MattHJensen @martinholmer @feenberg

I would call it an "exclusion". A haircut with a 100% rate would be an 
exclusion.

>
>  * 
>
>     @MattHJensen mentioned that all this kind of parameters would influence AGI
>     amount. We want to maintain the AGI decile/bins the same as current law for easy
>     comparison. To make sure our tables use the same set of AGI results, I'm going to
>     create a new function calculating and storing current law AGI in a separate PR.
>     Have not started to implement it yet. Welcome any comments/suggestions.
>

Aren't all current law parameters available already? How else do we 
calculate changes?

dan

martinholmer:
@Amy-Xu,  OK, I now see that this is a REFORM parameter, so there will be no effect on results from the validation/tests, which all assume current-law tax policy.  Sorry for the confusion.

MattHJensen:
> I would call it an "exclusion". A haircut with a 100% rate would be an
exclusion.

In this case, the user has the ability to apply a haircut with a 100% rate or a haircut with < 100% rate, so I believe using 'haircut' is more appropriate than 'exclusion'. 

What is confusing, however, is that in the other cases we apply a haircut to limit the benefit from an adjustment. E.g., setting the 'adjustment for student loan interest haircut' to 1 *increases* the taxpayer's liability. In the case of this interest haircut, however, increasing the haircut *reduces* the taxpayer's liability.

MattHJensen:
@feenberg, is this a sensible sentence:  "50% of interest income is excluded"?, or more generally, "x% of interest income is excluded", where x<100?  

If those are sensible maybe we should call this an exclusion after all.

feenberg:
On Tue, 3 Nov 2015, Matt Jensen wrote:

> 
> @feenberg, is this a sensible sentence: "50% of interest income is
> excluded"?, or more generally, "x% of interest income is excluded", where
> x<100?
> 
> If those are sensible maybe we should call this an exclusion after all.

I do think those are clear constructs. There is precident in the phrase 
"Foreign Income Exclusion" calculated on form 2555. There is some 
ambiguity about the difference between an exclusion and a deduction, and 
the FIE is now more like a credit, but still called an exclusion. Insiders 
talk about "above the line deductions" when they mean adjustments and 
exclusions, but that is rather a long bit of text when "exclusion" is 
available.

dan

feenberg:
On Tue, 3 Nov 2015, Matt Jensen wrote:

>
>       I would call it an "exclusion". A haircut with a 100% rate
>       would be an
>       exclusion.
> 
> In this case, the user has the ability to apply a haircut with a 100% rate
> or a haircut with < 100% rate, so I believe using 'haircut' is more
> appropriate than 'exclusion'.

I would use "haircut" to signify the reduction in a credit based on 
income. So there might be a 1000 dollar child credit reduced by 100 for 
each 600 of AGI. The reduction would be called a "haircut".

Recall that long term gains were partially included in income, that was 
called the "capital gains deduction". I think we should reserve 
"deduction" for Schedule A items, though.

Credits with "clawbacks" are a new thing. "Clawback" is a good (if 
British) word for any reduction in a credit based on income. "Haircut" 
could be a synonym or we could reserve it for a percentage reduction in a 
credit (as opposed to a linear reduction).

dan

MattHJensen:
@amy-xu, it seems like 'exclusion' is the right word to use for this PR. 

-----

On the usage of "haircut": 

@feenberg, we currently are using the term "haircut" to describe the "HC" parameters in the following snippet. 

    c02900 = (e03150 + (1 - ALD_StudentLoan_HC) * e03210 + e03600 +
              (1 - ALD_SelfEmploymentTax_HC) * e03260 +
              (1 - ALD_SelfEmp_HealthIns_HC) * e03270 +
              (1 - ALD_KEOGH_SEP_HC) * e03300 + (1 - ALD_EarlyWithdraw_HC) *
              e03400 + (1 - ALD_Alimony_HC) * e03500 + e03280 + e03900 +
              e04000 + e03700 + e03220 + e03230 + e03240 + e03290)

This does not conform with your suggestion above:

> I would use "haircut" to signify the reduction in a credit based on
income. So there might be a 1000 dollar child credit reduced by 100 for
each 600 of AGI. The reduction would be called a "haircut".

Do you have a suggestion for how we might rename the "HC" parameters in this snippet? 

cc @martinholmer @aviard

feenberg:
On Tue, 3 Nov 2015, Matt Jensen wrote:

> 
> @amy-xu, it seems like 'exclusion' is the right word to use for this PR.
> 
> ______________________________________________________________________________________
> 
> On the usage of "haircut":
> 
> @feenberg, we currently are using the term "haircut" to describe the "HC" parameters
> in the following snippet.
> 
> c02900 = (e03150 + (1 - ALD_StudentLoan_HC) * e03210 + e03600 +
>           (1 - ALD_SelfEmploymentTax_HC) * e03260 +
>           (1 - ALD_SelfEmp_HealthIns_HC) * e03270 +
>           (1 - ALD_KEOGH_SEP_HC) * e03300 + (1 - ALD_EarlyWithdraw_HC) *
>           e03400 + (1 - ALD_Alimony_HC) * e03500 + e03280 + e03900 +
>           e04000 + e03700 + e03220 + e03230 + e03240 + e03290)
> 
> This does not conform with your suggestion above:
>
>       I would use "haircut" to signify the reduction in a credit based on
>       income. So there might be a 1000 dollar child credit reduced by 100 for
>       each 600 of AGI. The reduction would be called a "haircut".
> 
> Do you have a suggestion for how we might rename the "HC" parameters in this snippet?


I am stuck. I can't think of anything more suitable than HC.

dan

MattHJensen:
@Amy-Xu, the right way to think about this parameter is that it allows the user to exclude all or part of interest income from taxable income, but interest income will still be taxed under NIIT if the user doesn't explicitly set the NIIT rate to zero and interest income will still be used in the determination of eitc eligibility. Are there any other quirks? Could you add this sort of info to the meta data in current_law_policy.json?

feenberg:
What is NIIT?

dan

Amy-Xu:
@feenberg NIIT is Net Investment Income Tax aka Unearned Income Medicare contribution

Amy-Xu:
@MattHJensen Just changed the name and added a few notes. Could you take a look?

MattHJensen:
Thanks @Amy-Xu, merging!

PR #437, opened 2015-11-03
Fix problems in ParametersBase _update method

martinholmer:
Read the detailed commit messages for details.

@MattHJensen and T.J. @talumbau,  Please review and provide comments.


MattHJensen:
Everything in `test_behavior.py` looks great. @talumbau, could you look through the changes to ParametersBase? (The detailed commit messages were very helpful).

taxcalc/parameters_base.py
```diff
@@ -294,26 +293,75 @@ def _update(self, year_mods):
...
             nval = self.expand_array(values,
-                                     inflate=cpi_inflated,
+                                     inflate=indexed,
                                      inflation_rates=inf_rates,
                                      num_years=num_years_to_expand)
-            cval[(self.current_year - self.start_year):] = nval
-        self.set_year(self._current_year)
+            cval[(year - self.start_year):] = nval
+        # handle unused parameter names, which all end in _cpi
+        for name, values in year_mods[year].items():
```
talumbau: the logic of this function is now:

```
for all the names in the reform:
    if it ends in '_cpi'
        skip it
    else:
        do the work and keep track of the name

for all the names the names in the form:
    if I haven't seen the name before:
        # these are the names that end in '_cpi'
        do the necessary work
```

So we go through all the names twice. I don't understand why the logic is not:
```
for all the names in the reform:
    if it ends in '_cpi'
        keep track of it and move on
    else:
        do the work and keep track of the name

for all the names I kept track of (which by definition end in '_cpi'):
        do the necessary work
```

That way we don't go through the names twice and avoid extra checking.

taxcalc/parameters_base.py
```diff
@@ -294,26 +293,75 @@ def _update(self, year_mods):
...
+            if name not in used_names:
+                if not name.endswith('_cpi'):
+                    msg = 'unused parameter name {} does not end in _cpi'
+                    raise ValueError(msg.format(name))
+                pname = name[:-4]  # root parameter name
+                if pname not in self._vals:
+                    msg = 'root parameter name {} not in values dictionary'
+                    raise ValueError(msg.format(pname))
+                pindexed = values  # set indexing status
```
talumbau: Instead of calling `values` by a different name here (`pindexed`), it seems better to do:

```
# handle unused parameter names, which all end in _cpi
for name, pindexed in year_mods[year].items():
```

then we don't have to change the name midway through the for loop.

taxcalc/parameters_base.py
```diff
@@ -294,26 +293,75 @@ def _update(self, year_mods):
...
+                self._vals[pname]['cpi_inflated'] = pindexed  # remember status
+                pvals = getattr(self, pname)
+                pvalues = [pvals[year - self.start_year]]
+                cval = getattr(self, pname, None)
+                if cval is None:
+                    msg = 'parameter {} in year_mods for year [] is unknown'
+                    raise ValueError(msg.format(pname, year))
```
talumbau: we could never hit this exception because in the case that `cval` is set to None (that is, that the value of `pname` does not name an attribute on `self`), we would already have raised an exception in line 344 which is:

```
pvals = getattr(self, pname)
```

so maybe it is best to do this check on `pvals`.

taxcalc/parameters_base.py
```diff
@@ -294,26 +293,75 @@ def _update(self, year_mods):
...
+                    msg = 'unused parameter name {} does not end in _cpi'
+                    raise ValueError(msg.format(name))
+                pname = name[:-4]  # root parameter name
+                if pname not in self._vals:
+                    msg = 'root parameter name {} not in values dictionary'
+                    raise ValueError(msg.format(pname))
+                pindexed = values  # set indexing status
+                self._vals[pname]['cpi_inflated'] = pindexed  # remember status
+                pvals = getattr(self, pname)
+                pvalues = [pvals[year - self.start_year]]
```
talumbau: I don't understand why this line is:

```
 pvalues = [pvals[year - self.start_year]]
```
instead of:
```
pvalues = pvals[year - self.start_year]
```
It seems to me that the current logic would increase the "dimensionality" (for lack of a better term) of the underlying value. In other words, if `pvals` is a list, `pvalues` would then be a list with its one element being a list. If `pvals` is already a list of lists, then `pvalues` becomes a list with its one element being a list of lists. I don't think that's what we want to pass to `expand_array`. So, could you explain the logic to me?

taxcalc/parameters_base.py
```diff
@@ -294,26 +293,75 @@ def _update(self, year_mods):
...
+                cval = getattr(self, pname, None)
+                if cval is None:
+                    msg = 'parameter {} in year_mods for year [] is unknown'
+                    raise ValueError(msg.format(pname, year))
+                nval = self.expand_array(pvalues,
```
talumbau: if `pvalues` should actually be defined as I suggest above, then we could actually remove that variable and just change this line to:

```
nval = self.expand_array(pvals[year - self.start_year],
```

taxcalc/parameters_base.py
martinholmer: @talumbau, Good suggestion.  I made your suggested change to this:
```for name, pindexed in year_mods[year].items():```

File: taxcalc/parameters_base.py (hunk 4 above)
Comment (martinholmer) [inline taxcalc/parameters_base.py]: @talumbau,  If you remove the outer brackets as you suggested and use
```
pvalues = pvals[year - self.start_year]
```
you get the following error:
```
...
# handle unused parameter names, which all end in _cpi
        for name, pindexed in year_mods[year].items():
            if name not in used_names:
                if not name.endswith('_cpi'):
                    msg = 'unused parameter name {} does not end in _cpi'
                    raise ValueError(msg.format(name))
                pname = name[:-4]  # root parameter name
                if pname not in self._vals:
                    msg = 'root parameter name {} not in values dictionary'
                    raise ValueError(msg.format(pname))
                self._vals[pname]['cpi_inflated'] = pindexed  # remember status
                pvals = getattr(self, pname)
                pvalues = pvals[year - self.start_year]
                cval = getattr(self, pname, None)
                if cval is None:
                    msg = 'parameter {} in year_mods for year [] is unknown'
                    raise ValueError(msg.format(pname, year))
                nval = self.expand_array(pvalues,
                                         inflate=pindexed,
                                         inflation_rates=inf_rates,
                                         num_years=num_years_to_expand)
>               cval[(year - self.start_year):] = nval
E               ValueError: could not broadcast input array from shape (8) into shape (8,6)

parameters_base.py:353: ValueError
```
Does the code make more sense now?

taxcalc/parameters_base.py
martinholmer: @talumbau, Good point!  I've simplified the code as you suggested.

taxcalc/parameters_base.py
talumbau: Yes, I see what you are doing. Previously, we just expanded the user given value. Now, in the first loop you take any user-provided value and expand. Down here, we reach into the object itself, select out a single value, `pvals[year - self.start_year]` , and then expand that. Since `pvals[year - self.start_year]` is the value for a single year, it must be put into a list for the call to `expand_array`. Got it thanks

taxcalc/parameters_base.py
martinholmer: @talumbau, You are right to say that we should avoid the second loop when possible and, if it is not possible to avoid it altogether, should loop through only the unused names.  I made some code changes that do this.  Let me know what you think.

martinholmer:
@talumbau,  I've just now pushed all the changes I've made in response to your comments.  Have I missed any of your comments?  Let me know what you think about the revised pull request #437.

talumbau:
Looks good to me. +1

martinholmer:
@talumbau said:
> Looks good to me. +1
OK, should @MattHJensen or @talumbau merge this into the master branch?

PR #438, opened 2015-11-03
Known AMT Issue

GoFroggyRun:
This PR fixes part of the issue coming from #434 . In #434 we took the issue that allowing earnings for both parties would effect the federal income tax liability ([ovar4]). But this is actually not the fundamental reason. 
In fact, based on the independent sampling for primary earning and secondary earning, the total wages for some units got boosted, almost doubled in terms of expectation, which amplifies the originally hidden issue for AMT, and thus leads to differences between two calculators. The earning split part should have no bug based on the validation for group C, but the issue regarding AMT does exist and needs to be taken care of immediately. 
Given that it is not sure how long it might take to fix the AMT issue, I modified the sampling a little bit to get rid of the errors and was hoping these units could still be used to "validate" between two calculators before the real issue got identified and fixed. The original sampling will be retained after the AMT got fixed completely.
@martinholmer @MattHJensen Does this make sense? Should we merge this PR first and get another PR for AMT issues? Or do we want to keep the original sampling method until the AMT issue got fixed? Please advice, thanks!


martinholmer:
Sean (@GoFroggyRun) Wang,
This is a clever, and very fast, diagnosis of the problem; thanks for the insight.

My view is that we should figure out what the underlying problem is in the 
Tax-Calculator logic.  I'm fairly sure that the problem is in the Tax-Calculator
because using the a13.in sample (from the master branch) with the PENSIM
income tax module produces the same income tax liabilities as Internet TAXSIM
generates with the a13.in sample.  So the score is 2-1 against the Tax-Calculator,
so to speak.

Is there some AMT threshold parameter that is incorrect?  Looking through the
code, is there a place when e00200 is being used to see what the AMT should
be?  Perhaps try to tracing id 3776?  You've been efficient at finding bugs before,
so you decide on the best tactic.

@MattHJensen, do you agree on this strategy?  Doesn't make any sense to me
to patch the make-in.tcl program when the real problem seems to be elsewhere.

MattHJensen:
> @MattHJensen, do you agree on this strategy? Doesn't make any sense to me
to patch the make-in.tcl program when the real problem seems to be elsewhere.

@martinholmer, just to confirm what you mean:  fixing AMT will resolve any differences, even if wages are unrealistically large, and so it doesn't make sense to patch make-in.tcl (as in this PR) to make wages more realistic; instead is better to just focus on fixing AMT. 

That makes sense to me. @GoFroggyRun?

GoFroggyRun:
@martinholmer @MattHJensen  Sure. Then I'll continue to figure out the AMT issue in this PR before it gets merged.

martinholmer:
@MattHJensen said:
> just to confirm what you mean: fixing AMT will resolve any differences, even if wages are unrealistically large, and so it doesn't make sense to patch make-in.tcl (as in this PR) to make wages more realistic; instead is better to just focus on fixing AMT.

Yes, I think the working hypothesis should be "fixing AMT will resolve [all] differences".  Maybe it will not turn out that way, but the preliminary evidence Sean (@GoFroggyRun) produced is consistent with that hypothesis.

GoFroggyRun:
After a couple changes being made to the AMTI function, the issue becomes more interesting that, although there's no difference for A group at all, more differences seem to appear for samples in group B and C. I'm actually wondering why the enhancement for group A wasn't consistent when it comes to two other groups? Any ideas or insights? @martinholmer @MattHJensen

GoFroggyRun:
@martinholmer @MattHJensen 
So it turns out there're still some minor mistakes in AMTI, which now got completely fixed. I was planning to do some polishes and clean-ups in AMTI function to make it easier to read. But feel free to merge this PR first, if it looks good.

martinholmer:
Sean (@GoFroggyRun),
This work is impressive.  Thanks for your help; it would have taken me much longer to do this.

Given that the independently-developed PENSIM income tax module produces the same income tax liabilities as does Internet TAXSIM (at least for the a13.in sample, which is the only one I've checked), it appears as if your Tax-Calculator code changes are, in fact, bug fixes (rather than just changes aimed at matching TAXSIM output).

I'll leave it up to @MattHJensen and Dan @feenberg about how to proceed from here.  If you want to streamline the new code sometime later today as part of this pull request, that is fine with me.  That might be a good strategy because it would give Dan time to look over your code changes.

MattHJensen:
>Sean (@GoFroggyRun),
This work is impressive. 

Agreed! Did you also compare these changes against the tax forms?

GoFroggyRun:
@MattHJensen Yes. That's basically how I made these changes.

MattHJensen:
>@MattHJensen Yes. That's basically how I made these changes.

That's what I thought based on your standard (and efficient) workflow. Excellent.

GoFroggyRun:
I have just finished streamlining the code, and the results for those diff files remain consistent. I don't think I'll make further changes in this PR unless issues/bugs have been identified. Please take a look and let me know if you consider certain clauses suspicious. @martinholmer @MattHJensen @feenberg

martinholmer:
@MattHJensen and Dan @feenberg, Do you think pull request #438 should be merged into the master branch, or should we spend more time reviewing the changes Sean has made?

MattHJensen:
Lgtm for now. Dan will eventually review indirectly when Sean does another round of tax-calculator to taxcalc syncing.

MattHJensen:
@martinholmer, please go ahead and merge this when you are ready. This is a clear improvement that has been tested against the forms and internet taxsim and will receive more review later when we sync with taxcalc.

MattHJensen:
@GoFroggyRun, looks like there are some conflicts that need to be resolved before @martinholmer can merge.

GoFroggyRun:
@martinholmer @MattHJensen 
All set and ready to go.

PR #439, opened 2015-11-03
Move read-JSON-reform-file capabilities to Policy class

martinholmer:
These capabilities were originally developed as part of the SimpleTaxIO class development, but logically belong in the Policy class.

@MattHJensen and T.J. @talumbau,  Can you review and comment or merge into master?

### DIFF
--- a/taxcalc/parameters_base.py
+++ b/taxcalc/parameters_base.py
@@ -159,7 +159,7 @@ class ParametersBase(object):
             if num_values <= nyrs:
                 # val should be the single start_year value
                 rawval = getattr(ppo, name[1:])
-                if isinstance(rawval, np.core.ndarray):
+                if isinstance(rawval, np.ndarray):
                     val = rawval.tolist()
                 else:
                     val = rawval
@@ -361,11 +361,11 @@ class ParametersBase(object):
         If necessary, pad out additional years by increasing the last given
         year using the given inflation_rates list.
         """
-        if isinstance(x, np.core.ndarray):
+        if isinstance(x, np.ndarray):
             if len(x) >= num_years:
                 return x
             else:
-                ans = np.core.zeros(num_years, dtype='f8')
+                ans = np.zeros(num_years, dtype='f8')
                 ans[:len(x)] = x
                 if inflate:
                     extra = []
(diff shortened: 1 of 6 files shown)


talumbau:
Ah I see. This PR digs in to the `core` subpackage of the `numpy` package for basically every usage of numpy. This is likely done because of pylint warnings. As I mentioned [here](https://github.com/open-source-economics/Tax-Calculator/issues/30#issuecomment-126353910) I don't think that this code project should adhere to the standard of "no warnings from default pylint". There are many reasons to prefer something other than that standard. In this particular case, the reasons to prefer `np.xxx` to `np.core.xxx` are as follows:

- it is the idiomatic usage among all numpy-based Python projects that I know of
- to my knowledge, the API does not guarantee that `core` will always be around, and so future versions of numpy may relocate functions that we need. However, we know that `np.zeros` is not going anywhere, since the entire community depends on that.

As was mentioned in the [issue](https://github.com/open-source-economics/Tax-Calculator/issues/30) I referenced above, no one has explicitly stated that they wish to make the settings for this project 'no warnings from default pylint' but there has been no community discussion on this matter. I just opened #442 to discuss. For this PR, I advocate for not using `np.core.xxx` for the reasons mentioned above.

martinholmer:
@talumbau, Is this better?

talumbau:
I like the tests here and see no reason at this point not to merge. One style note: I think it is important to be judicious in the usage of `is_instance`. The reason is that Python code should usually just care that a value *behaves* in a certain way not that it *is* a certain type. For example, this PR now has more checking that certain values are of type `dict`. In reality, we just need some type that has key/value pairs - e.g. an `OrderedDict` object would probably do just as well as a `dict` object in most of our code. It is difficult to enforce a strict rule on these kinds of things - just a note for future work.

martinholmer:
T.J. @talumbau said:
> One style note: I think it is important to be judicious in the usage of is_instance. The reason is that Python code should usually just care that a value behaves in a certain way not that it is a certain type. For example, this PR now has more checking that certain values are of type dict. In reality, we just need some type that has key/value pairs - e.g. an OrderedDict object would probably do just as well as a dict object in most of our code. It is difficult to enforce a strict rule on these kinds of things - just a note for future work.

OK, I'll keep that in mind.  Thanks for the tip.

PR #440, opened 2015-11-03
Add function for AGI under current law

Amy-Xu:
As mentioned in #435, we want to have both distribution table and difference table use AGI or expanded income under current law as row label classifiers. This wouldn't be a problem at all if users ran the default plan, but if users ran any reforms on adjustments or more generally any AGI components, AGI or expanded income would be changed and therefore the results tables would be misleading in comparison.

In this PR, default row label classifier is set as `_expanded_income_currentlaw` to make both PlanX and PlanY have the same measure on row labels. Notice that the actual column for AGI or expanded income is still using the results for that plan, reform or current law.


feenberg:
On Tue, 3 Nov 2015, Amy Xu wrote:

> 
> As mentioned in #435, we want to have both distribution table and difference table use AGI or
> expanded income under current law as row label classifiers. This wouldn't be a problem at all
> if users ran the default plan, but if users ran any reforms on adjustments or more generally
> any AGI components, AGI or expanded income would be changed and therefore the results tables
> would be misleading in comparison.

I thought this was clear from the beginning - the tab variable is always 
based on PlanX and isn't adjustable by the user.

> 
> In this PR, default row label classifier is set as _expanded_income_currentlaw to make both
> PlanX and PlanY have the same measure on row labels. Notice that the actual column for AGI or
> expanded income is still using the results for that plan, reform or current law.
>

Aggregate AGI in PlanY is aggregate for PlanY, but it should be 
distributed across the AGI categories according to taxpayer's place in 
PlanX. That is, no one changes row location for any reason to do with 
PlanX.

dan

MattHJensen:
@Amy-Xu, could you show tables for a reform that changes AGI on master and a reform that changes AGI on this branch?

Amy-Xu:
@MattHJensen Here're the tables based on removing student loan adjustment (ALD_StudentLoan_HC == 1). I cannot use diagnostic tables because it doesn't show any bins or deciles, but the distribution tables are not formatted in taxcalc so I just pasted a few columns for each table. You can check the weighted sum of s006 for the classifier. 

Master: 
PlanX: 
calc1 master
calc1 tail master

PlanY, classifier using _expanded_income
calc2 master
calc2 tail master

PR:
PlanX (mysteriously not the same as master baseline, need to figure out what happened)
calc1 pr
calc1 tail pr

PlanY, classifier using _expanded_income_currentlaw
calc2 pr head
calc2 pr

Amy-Xu:
@feenberg 
>I thought this was clear from the beginning - the tab variable is always
based on PlanX and isn't adjustable by the user.

I thought the same, but didn't realize this might not be what actually happened until Matt pointed out.

MattHJensen:
> PlanX (mysteriously not the same as master baseline, need to figure out what happened)

Other than this, the tables look good. What I was checking for was that the bins were the same under both plan x and plan y for this PR but *not* for master. I confirmed that. Do you think we should add a test for this?

Amy-Xu:
replaced by #446. closing.

Issue #397, opened 2015-10-15
Try dropping the up and down calculations in MTR.


MattHJensen:
@feenberg has suggested that we explore whether we can avoid all the kinks tax schedule for the purpose of our marginal tax rate calculations by first adding a penny to income and then calculating mtr using a 1 cent finite diff. If all of the kinks are at dollar levels, this would work and we could avoid doing the 'up' and 'down' calculation and potentially save ourselves two calculator runs.

MattHJensen:
@feenberg,

> After Ohara is through with our data - is it still intergers?

Yes, but that is because we are unintentionally truncating everything. I guess when we reverse this truncation 'the add a penny to income first' idea won't work.

feenberg:
On Thu, 15 Oct 2015, Matt Jensen wrote:

> 
> @feenberg,
>
>       After Ohara is through with our data - is it still intergers?
> 
> Yes, but that is because we are unintentionally truncating everything. I guess when we
> reverse this truncation 'the add a penny to income first' idea won't work.

Rather instead of "won't work" we should say "won't be needed".

dan

MattHJensen:
Ohh, I see your point. That is good news.

martinholmer:
@MattHJensen and Dan @feenberg,

Using the master branch at 3:00pm 03-Nov-2015, I tried a few variations in mtr-calculation methods.  I dropped the finite_diff from one dollar to one dime and then to one cent.  And I tried to use just the up values of mtr when the finite_diff was one cent.  Here is what I found using simtax.py on the a13.in sample of 100,000 filing units with no itemized expenses and no childcare expenses in 2013.  (Note that
output variable 7 is marginal federal income tax rate on taxpayer earnings and output variable 9 is marginal FICA tax rate on taxpayer earnings, and the results below are summary tabulations of the **differences** between Tax-Calculator and Internet-TAXSIM percentage marginal tax rates.  If there is no summary information, it means there were no differences for that output variable between the two sets of results.)
```
ONE-DOLLAR-FINITE-DIFF + UP/DOWN-LOGIC:
TAXDIFF:ovar,#diffs,#1cdiffs,maxdiff[id]=  7    193     14    -26.33 [23437]
      #big_vardiffs_with_big_inctax_diff=               30
```
```
ONE-DIME-FINITE-DIFF + UP/DOWN-LOGIC:
[...]

PR #441, opened 2015-11-04
[WIP] Streamline logic of Calculator class mtr() method

martinholmer:
Change default finite_diff from one dollar to one cent and retain only the "up" calculations of marginal tax liabilities.  Also, revise the employer FICA adjustment logic in two ways: (1) make it can handle finite_diff values other than one dollar, and (2) fix problem with the calculation of the er_fica_adjustment variable, which previously used combined (not individual) earnings to determine whether over OASDI maximum taxable earnings level.

With regards to the marginal FICA tax rates in the new a13.taxdiffs file, see today's discussion under Issue #397.

@MattHJensen, can you check that I have revised the "employer FICA adjustment" logic correctly?
Also, check the long comment describing the rationale for this adjustment; I'm not sure I understand it well enough to have correctly described what is going on here.  Any alternative test you could suggest would be welcomed.

### DIFF
--- a/taxcalc/behavior.py
+++ b/taxcalc/behavior.py
@@ -52,10 +52,10 @@ def behavior(calc_x, calc_y, update_income=update_income):
     """
 
     # Calculate marginal tax rates for plan x and plan y.
-    _, _, combined_mtr_x = calc_x.mtr('e00200p')
+    _, _, combined_mtr_x = calc_x.mtr()
     mtr_x = combined_mtr_x
 
-    _, _, combined_mtr_y = calc_y.mtr('e00200p')
+    _, _, combined_mtr_y = calc_y.mtr()
     mtr_y = combined_mtr_y
 
     # Calculate the percent change in after-tax rate.
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -210,8 +210,8 @@ class Calculator(object):
     def current_year(self):
         return self.policy.current_year
 
-    def mtr(self, income_type_string,
-            finite_diff=1.0,
+    def mtr(self, income_type_str='e00200p',
+            finite_diff=0.01,
             wrt_adjusted_income=True):
         """
         Calculates the individual income tax, FICA, and combined marginal tax
@@ -221,7 +221,7 @@ class Calculator(object):
 
         Parameters
         ----------
-        income_type_string: string
+        income_type_str: string
             specifies an income attribute in the Records class.
 
         finite_diff: float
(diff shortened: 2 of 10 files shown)


MattHJensen:
@martinholmer, this is much cleaner. Thank you! 

As to the employer fica adjustment, here is an example that illustrates why we need it:

Let's say a taxpayer is subject to *employee*-fica at a 15.3% rate. He has $100 of wages and and has to pay $15.3 of taxes, leaving him with $84.7 after taxes. 

Another taxpayer is subject to *employer*-fica at a 15.3% rate; his employer wants to spend $100 on his compensation, too (solving the equation for x: .153*x + x = 100). This taxpayer takes home $86.73. 

While these two employees seem to have the same marginal tax rates (adding a dollar of wages to each will cause their tax liabilities to rise by 15.3 cents), the cost to the employer is clearly different and therefore the incentive effects are different. The employer-fica-adjustment captures this. 

Given this, I think it is a bit confusing to create an adjusted_finite_diff variable as in this PR; rather, it is clearer to add the adjustment factor directly to the denominator of the MTR calculation. 

I'll give some more thoughts to tests this afternoon. 

cc @feenberg

martinholmer:
@MattHJensen said:
> Let's say a taxpayer is subject to employee-fica at a 15.3% rate. He has $100 of wages and and has to pay $15.3 of taxes, leaving him with $84.7 after taxes.
>
> Another taxpayer is subject to employer-fica at a 15.3% rate; his employer wants to spend $100 on his compensation, too (solving the equation for x: .153*x + x = 100). This taxpayer takes home $86.73.
>
> While these two employees seem to have the same marginal tax rates (adding a dollar of wages to each will cause their tax liabilities to rise by 15.3 cents), the cost to the employer is clearly different and therefore the incentive effects are different. The employer-fica-adjustment captures this.

I'm confused about the points being made above.  I thought the idea of using marginal tax rates in the Behavior class was to simulate **employee** (that is, the individual income tax payer) responses to taxes.  If that is so, then what difference to the employee does it make what the total compensation cost for his/her employer is?  Sorry, if I am missing something obvious here.  But if I don't get it right away, then the comments in the code probably need to be revised to make the point you have in mind more clearly.

feenberg:
On Wed, 4 Nov 2015, Matt Jensen wrote:

> 
> @martinholmer, this is much cleaner. Thank you!
> 
> As to the employer fica adjustment, here is an example that illustrates why we need it:
> 
> Let's say a taxpayer is subject to employee-fica at a 15.3% rate. He has $100 of wages and
> and has to pay $15.3 of taxes, leaving him with $84.7 after taxes.
> 
> Another taxpayer is subject to employer-fica at a 15.3% rate; his employer wants to spend
> $100 on his compensation, too (solving the equation for x: .153*x + x = 100). This taxpayer
> takes home $86.73.
> 
> While these two employees seem to have the same marginal tax rates (adding a dollar of wages
> to each will cause their tax liabilities to rise by 15.3 cents), the cost to the employer is
> clearly different and therefore the incentive effects are different. The
> employer-fica-adjustment captures this.
> 
> Given this, I think it is a bit confusing to create an adjusted_finite_diff variable as in
> this PR; rather, it is clearer to add the adjustment factor directly to the denominator of
> the MTR calculation.
> 
> I'll give some more thoughts to tests this afternoon.

Recall the the underlying difficulty arises from the fact that half of the 
FICA tax (the employer share) does not show up in wages (or AGI). If we 
put total compensation in the denominator (wage plus .5*fica) and change 
in tax in the numerator, we get the correct MTR for the income tax. The 
MTR for FICA is right either way.

dan

martinholmer:
Dan @feenberg said:
> I recall that the underlying difficulty arises from the fact that half of the FICA tax (the employer share) does not show up in wages (or AGI). If we put total compensation in the denominator (wage plus .5*fica) and change in tax in the numerator, we get the correct MTR for the income tax. The MTR for FICA is right either way.

So, we are defining the marginal tax rate to be change in tax divided by change in total compensation caused by a small increase in taxpayer earnings.  Right?  If that is correct, and if ```d``` is the small increase in taxpayer earnings, and ```t``` is the marginal employee+employer FICA tax rate, and ```L``` is the change in tax liability caused by ```d```, then we have this:
```
mtr_income_tax = L / ( d + d * t/2 ) 
```
Is that correct?

What I don't understand is Dan's last sentence:
> The MTR for FICA is right either way.

I'm not sure what that means.  Perhaps I'm confused, but the mtr_fica would be different depending on whether the denominator is ```d*(1+t/2)``` or ```d```.  Can you provide more explanation?

cc @MattHJensen

feenberg:
On Wed, 4 Nov 2015, Martin Holmer wrote:

> 
> Dan @feenberg said:
>
>       I recall that the underlying difficulty arises from the fact that half of the
>       FICA tax (the employer share) does not show up in wages (or AGI). If we put total
>       compensation in the denominator (wage plus .5*fica) and change in tax in the
>       numerator, we get the correct MTR for the income tax. The MTR for FICA is right
>       either way.
> 
> So, we are defining the marginal tax rate to be change in tax divided by change in total
> compensation caused by a small increase in taxpayer earnings. Right? If that is correct, and
> if d is the small increase in taxpayer earnings, and t is the marginal employee+employer FICA
> tax rate, and L is the change in tax liability caused by d, then we have this:
> 
> mtr_income_tax = L / ( d + d * t/2 ) 
> 
> Is that correct?



See page 25 of

   http://www.nber.org/papers/w5155.pdf

> We also recognize that the taxayer's income is defined net of the 
> employer's share of the Social Security payroll tax. Each $100 of 
> taxable wage income up to a maximum taxable [amount] is subject to an 
> extra tax of $7.63. This indicates that the full gross wage is $107.65 
> per $100 of taxable wage income, implying that $100 of taxable wage 
> income corresponds to a marginal product of labor of $107.65.

So the "t" in the equation above would be the rate of the employer's share 
of FICA, not anything in the income tax itself.

dan

martinholmer:
@MattHJensen and Dan @feenberg,  Does the mtr() method look OK to you now?

MattHJensen:
Looks good to me. Thanks @martinholmer.

martinholmer:
@MattHJensen said:
> Looks good to me. Thanks @martinholmer

OK, do you want to merge it into master or should we wait for Dan @feenberg to look at it again?

MattHJensen:
Let's merge. This is a clear improvement over what what is in master that has been tested against both the tax forms and internet taxsim. It will get further review when we do the next sync with taxcalc.

MattHJensen:
Oops, thought this was the AMT fix. This one should be merged too, though, so I'm not reverting.

feenberg:
If you put the code in an email, I can look at it, but I may not 
understand it.

dan

PR #443, opened 2015-11-05
Tiny Enhancement of SIMTAX

GoFroggyRun:
While dealing with #438 , I found that having some intermediate variables printed out might be helpful in terms of debugging (please refer to the commit "debug strategy revealed" in that PR). And I was wondering can we add two extra optional variables in the output, namely [ovar29] and [ovar30]. Notice that I only kept one decimal point on purpose, hoping this would make them easier to be distinguished from the remaining 28 variables. 
@martinholmer What do you think? Does such enhancement make sense? Please take a look. 
(I was also thinking of incorporate such option into commend line, but my concern is that, in that case, the command might be a little bit tedious and most user might not be interested in such functionality)

### DIFF
--- a/taxcalc/simpletaxio.py
+++ b/taxcalc/simpletaxio.py
@@ -540,7 +540,7 @@ class SimpleTaxIO(object):
         recs.p22250[idx] = ivar[21]  # short-term capital gains (+/-)
         recs.p23250[idx] = ivar[22]  # long-term capital gains (+/-)
 
-    OVAR_NUM = 28
+    OVAR_NUM = 30
 
     @staticmethod
     def _extract_output(crecs, idx, ivar):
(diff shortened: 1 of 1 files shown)


martinholmer:
Sean (@GoFroggyRun),  Thanks for your insight on how to make SIMTAX a more flexible debugging tool.  Let me think about your specific suggestion and then we can figure out how best to proceed.

martinholmer:
Sean (@GoFroggyRun),  Given your experience using SIMTAX for debugging Tax-Calculator code, I'm curious as to why you think two extra debugging output variables is the right number.  Is that what was needed in your recent AMT debugging work?  Might more debugging output be needed in other situations?

GoFroggyRun:
@martinholmer Yes. The two extra variables I added here are arbitrary. There are definitely cases where more intermediate variables will be needed. But I'm afraid putting too many tail variables will make the rest 28 variables hard to be observed. So two is just a trade-off I have made here. And I definitely won't mind if you'd like to put a few more variables on top of that.

Issue #436, opened 2015-11-02
Move extrapolation and imputations to taxdata repo? 


MattHJensen:
Tax-calculator currently does more than calculate taxes for some given dataset. It also does quite a bit to prepare that dataset, including applying blowup factors, reweighing the data, making some imputations on the fly, and adjusting the data based user-given parameters (`growth.py`). 

I have been wondering whether it would be more sensible to move the extrapolation, imputations, and growth.py functionality to the taxdata repository. Tax-calculator would then purely calculate taxes, and tax-data would purely create tax databases. 

How this would actually be implemented, I'm not so sure.

This relates to #425.

@martinholmer, @talumbau, and @Amy-Xu, what do you think?

feenberg:
On Mon, 2 Nov 2015, Matt Jensen wrote:

> 
> Tax-calculator currently does more than calculate taxes for some given dataset. It
> also does quite a bit to prepare that dataset, including applying blowup factors,
> reweighing the data, making some imputations on the fly, and adjusting the data based
> user-given parameters (growth.py).
> 
> I have been wondering whether it would be more sensible to move the extrapolation,
> imputations, and growth.py functionality to the taxdata repository. Tax-calculator
> would then purely calculate taxes, and tax-data would purely create tax databases.

This is the practice I have mostly observered. I like it. I have a dataset 
for each year and to read all the years I concatenate them:

    cat /home/data/soi/taxsim/asc/x*.asc | a.out

We would need something a little different here.

dan

martinholmer:
After pull request #754 was merged into the master branch, all the data imputation logic is being done in the [taxdata repository final-preparation script](https://github.com/open-source-economics/taxdata/blob/master/final%20prep/puf-cps-processing.py).  The blow-up logic remains in the Tax-Calculator.

Is there any further work that should be done on this issue #436?  If not, should the issue be closed?

@MattHJensen @feenberg @talumbau @Amy-Xu

MattHJensen:
I wonder if, in the future, we will want to pass some sort of generic blowup dictionary to the tax-calculator rather than include blow-up logic that is semi-specific to the taxdata repository. For now, I think what we have is enough and this issue ought to be closed.

PR #444, opened 2015-11-06
Add support for numpy v1.10

talumbau:
I don't know if this is too "ugly" of a fix. I should probably also show the distribution tables for master and this branch as a double check that the math hasn't changed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ install:
   - export PATH="$HOME/miniconda/bin:$PATH"
   - conda config --set always_yes yes --set changeps1 no
   - conda update conda
-  - conda create -n taxcalcdev python=$TRAVIS_PYTHON_VERSION pytest setuptools numpy=1.9.3 pandas=0.16.2 numba toolz six
+  - conda create -n taxcalcdev python=$TRAVIS_PYTHON_VERSION pytest setuptools pandas=0.16.2 numba toolz six
   - source activate taxcalcdev
   - pip install pytest-pep8
   - python setup.py install
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -7,7 +7,7 @@ requirements:
     - pytest
     - setuptools
     - python
-    - numpy <=1.9.3
+    - numpy
     - pandas <=0.16.2
     - numba
     - toolz
@@ -16,7 +16,7 @@ requirements:
     - pytest
     - setuptools
     - python
-    - numpy <=1.9.3
+    - numpy
     - pandas <=0.16.2
     - numba
     - toolz
--- a/environment.yml
+++ b/environment.yml
@@ -3,7 +3,7 @@ dependencies:
 - pytest
 - setuptools
 - python
-- numpy <= 1.9.3
+- numpy
 - pandas <=0.16.2
 - numba
 - toolz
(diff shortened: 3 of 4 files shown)


talumbau:
Actually, I ran `test.py` with the latest PUF for this branch (with numpy 1.10 and numba 0.22 for numpy 1.10) and against master (with numpy 1.9.3 and numba 0.22 for numpy 1.9.3) and got the exact same results. Therefore, if this change is stylistically acceptable, it is good to be merged.

talumbau:
cc @Amy-Xu @MattHJensen. The urgency on this is not high yet, but in terms of integrating the micro/macro calculations, I think the best case scenario for deployment is to support numpy v1.10. Therefore, we should have a solution of some kind ready to go, if not the solution I propose in this PR.

MattHJensen:
I must admit that I haven't completely kept up with this issue. Could you let us know why the proposed solution is preferred:

    self.e00200 = self.e00200 + self.e00200*self.BF.AWAGE[year]

talumbau:
I think you mean:

```
self.e00200 = self.e00200*self.BF.AWAGE[year]
```

correct? Assuming that's what you meant, there are two differences between:

```
self.e00200 = self.e00200*self.BF.AWAGE[year]
```

and

```
a = self.e00200
b = self.BF.AWAGE[year]
np.multiply(a, b, out=a, casting='unsafe')
```

1. if `e00200` has a dtype in the `int` family (e.g. `int64`), which is the case upon first reading in the PUF, the second style retains the `int64` dtype, whereas the first calculation would product a new array with a floating point dtype, and that new array would be assigned to the attribute `e00200` on `self`.

2. (I think this one is more important) There are some values that are 'aliased', so two attributes that refer to the same value. If you use the first style above, the aliasing would be broken, because a new value is created and then assigned as an attribute. If you use the second style, the aliasing remains intact. I can't remember which values @Amy-Xu was concerned about in regards to the aliasing, but this was the original motivation to begin writing the operations as `a *= b`.

MattHJensen:
> I think you mean:
> self.e00200 = self.e00200*self.BF.AWAGE[year]

Yes. That's what I meant. 

>  the first calculation would produce a new array with a floating point dtype, and that new array would be assigned to the attribute e00200 on self.

This is actually somewhat preferred because of https://github.com/open-source-economics/Tax-Calculator/issues/397#issuecomment-148507440.

> If you use the second style, the aliasing remains intact. I can't remember which values @Amy-Xu was concerned about in regards to the aliasing, but this was the original motivation to begin writing the operations as a *= b. 

Ahhh, yes, this does seem more important than the previous point. 

-----

I suppose it is worth adding that if anything happens with #436, this issue will be revisited in the taxdata repo.

talumbau:
Seems like we do have competing concerns here. As we are building toward the microsim-dynamic model integration, the easiest thing from a deployment perspective is to move to numpy 1.10 for everything. I can work around this, but if we can live with the changes in this PR, it will make things easier. One thing is that the change is fairly localized, so one wouldn't have to look very hard for all of the occurrences of the issue if a different technique was desired later on.

MattHJensen:
I'm fine proceeding with this approach. We'll have an opportunity to revisit later during potential work on #436. 

@Amy-Xu, could you look this over and merge if it looks good to you?

Amy-Xu:
@talumbau I was concerned about p23250 and e23250 and just checked they are the same in this PR. Looks good to me. Merging.

Thanks @talumbau @MattHJensen

PR #446, opened 2015-11-07
Distribution table classifier

Amy-Xu:
This PR is replacing #440 which will be closed due to unexpected complications. In this PR, one more argument `baseline_calc` was added to function `create_distribution_table` in order to get the income measure from a baseline calculator. 

@talumbau The reason why we're doing this is that no matter what reform plans are passed in the calculator, we always want the row label classifiers to be the current law value. Even though many reforms won't change the AGI/Expanded Income values, some reforms (any plan involves haircut parameters) do change AGI and Expanded Income. Under these reforms, we should still group taxpayers using their original AGI/Expanded Income instead of the one under reform. 

So I added this `baseline_calc` to pass in the baseline data. I can definitely see this might increase the chance for users getting non-sense results, for example if they happen to pass in a non-baseline calculator or a baseline calculator not on the right year. I have only been able to add a check whether the current year attributes of two calculators are the same or not, but no clue how to check the 'baseline' part.

Anyways @MattHJensen Here're the results using the same reform `{2013:{"_ALD_StudentLoan_HC": [1]}}`. All the weights conform with master.

Baseline:
`create_distribution_table(calc1, groupby = "webapp_income_bins", result_type = "weighted_sum")`
calc1_head
calc1_tail

Reform:
`create_distribution_table(calc2, groupby = "webapp_income_bins", result_type = "weighted_sum", baseline_calc=calc1)`
calc2_tail

calc2_head

Welcome any comments/suggestions! Thanks!

### DIFF
--- a/taxcalc/tests/test_utils.py
+++ b/taxcalc/tests/test_utils.py
@@ -350,6 +350,32 @@ def test_diff_table_sum_row():
                           tdiff2[non_digit_cols][-1:])
 
 
+def test_row_classifier():
+    # create a current-law Policy object and Calculator calc1
+    policy1 = Policy()
+    records1 = Records(data=TAX_DTA, weights=WEIGHTS, start_year=2009)
+    calc1 = Calculator(policy=policy1, records=records1)
+    calc1.calc_all()
+    calc1_s006 = create_distribution_table(calc1,
+                                           groupby="webapp_income_bins",
+                                           result_type="weighted_sum").s006
+
+    # create a policy-reform Policy object and Calculator calc2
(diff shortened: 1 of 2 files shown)


MattHJensen:
LGTM! @talumbau?

talumbau:
This makes sense to me. I like the test as well. :+1:

Amy-Xu:
@MattHJensen If this looks good, could you merge it in?

MattHJensen:
Thanks @Amy-Xu and @talumbau. Merging.

PR #451, opened 2015-11-10
Fix calculation of long-term-capital-gains marginal income tax rate

martinholmer:
Fix accomplished by changing income_type_str for long-term capital
gains in Calculator class mtr() method from 'p23250' to 'e23250'.

@Amy-Xu, can you review and merge into master branch.  Before this fix all the ltcg mtr values were zero; now they look very reasonable and line up with Internet-TAXSIM ltcg mtr estimates.

### DIFF
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -255,12 +255,12 @@ class Calculator(object):
         'e00200p', taxpayer wage/salary earnings (which is the default value);
         'e00900p', taxpayer (Schedule C) self-employment income;
         'e00300',  taxable interest income;
-        'p23250',  long-term capital gains;
+        'e23250',  long-term capital gains;
         'e01700',  federally-taxable pension benefits; and
         'e02400',  social security (OASDI) benefits.
         """
         mtr_valid_income_types = ['e00200p', 'e00900p',
-                                  'e00300', 'p23250',
+                                  'e00300', 'e23250',
                                   'e01700', 'e02400']
         # check validity of income_type_str parameter
         if income_type_str not in mtr_valid_income_types:


talumbau:
I don't have the necessary expertise to understand the meaning of this change, but just a note that these two values (`p23250` and `e23250`) are supposed to alias each other in a `Records` object:

https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/records.py#L655

This change would seem to indicate that we get the wrong answer when we use one of the names and the right answer when we use the other. Is that correct? If so, that would seem to indicate to me that we are not correctly aliasing the variables since we can't use one in place of the other. No problem if I am way off-base, just saying what it looks like to me.

Amy-Xu:
Martin @martinholmer, actually I'm confused as well after @talumbau pointed this out. We had the problem where the two variables do not give the same values but right now they should be equivalent. Let me ponder why it's showing difference for mtr().

martinholmer:
T.J. @talumbau said:
> I don't have the necessary expertise to understand the meaning of this change, but just a note that these two values (p23250 and e23250) are supposed to alias each other in a Records object:
>
> https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/records.py#L655
>
> This change would seem to indicate that we get the wrong answer when we use one of the names and the right answer when we use the other. Is that correct? If so, that would seem to indicate to me that we are not correctly aliasing the variables since we can't use one in place of the other. No problem if I am way off-base, just saying what it looks like to me.

I don't pretend to understand the need for or nature of this aliasing business, but what is going on in the SimpleTaxIO class is different than what is going on when using the PUF data in the Records class.  If you want to investigate, look at the SimpleTaxIO logic with special reference to the calculate method, which is where the marginal tax rates are calculated.  If you do look at SimpleTaxIO, you will see that for calculating dollar tax liabilities, 'p23250' is used as the input for long-term capital gains and this usage produces results that are known to be correct.

I'm hoping the this question of whether or not there is a problem with the Records class variable aliasing will not delay the merge of this pull request.

martinholmer:
@Amy-Xu,  Read my response to T.J.'s comment.

Amy-Xu:
@talumbau I opened a issue for this PR. Indeed I feel this might confuse some users when they use these two 'supposed-to-be-the-same' variables, but we need to use this function right now so I'm going to merge this PR and keep investigating in the issue.

PR #454, opened 2015-11-11
Fix taxdiff.awk to work when taxdiffs.tcl --ovar4 option is used

martinholmer:
The taxdiffs.tcl --ovar4 option limits checking across two validation output files to output variable 4, which is the total federal individual income tax liability.


martinholmer:
Merge this pull request into the master branch because it has no effect on any other Tax-Calculator code.

PR #457, opened 2015-11-12
Parallel execution of validation/tests

martinholmer:


martinholmer:
Sean (@GoFroggyRun),  Could you look over this pull request?  Any suggestions for improvements would be appreciated.

GoFroggyRun:
@martinholmer This PR looks good to me. I was wondering, after this PR being merged, should we incorporate the `reform_tests` you proposed in #426 into an option (like `ALLTESTS ` maybe) in the `validation/tests` file, instead of executing `validation/reform_tests`? It seems more consistent to me if all the validations could go through one script. What do you think? Does this make sense?

martinholmer:
Sean (@GoFroggyRun) said:
> I was wondering, after this PR being merged, should we incorporate the reform_tests you proposed in #426 into an option (like ALLTESTS maybe) in the validation/tests file, instead of executing validation/reform_tests? It seems more consistent to me if all the validations could go through one script. What do you think? Is this what you were also thinking of?

Good idea!  I'll work on that later today.  Meanwhile, can you prepare a pull request that does three kinds of things: (a) adds each trimmed TAXSIM reform output file to out-taxsim.zip; (b) adds to git version control each of the reform JSON files you've created; and (c) adds to git version control each of the abbreviated (ovar4 only) .taxdiffs files you have created?  Have I missed anything at your end?  Then, after I redo the validation tests script to handle reform tests, I can see if I can pass all the reform tests (just as you did).  Does this make sense as a course of action?

GoFroggyRun:
@martinholmer That's pretty much it. I'll be working on a PR right away.

PR #453, opened 2015-11-11
Added paramter for CTC reform

Amy-Xu:
In this PR, parameter '_CTC_additional' is added for a partially refundable Child Tax Credit, on top of all the child/dependents credit available. This credit amount is limited by the sum of total income and payroll tax liabilities.

Here're the baseline table , and the diagnostic table with an additional credit of $2500 for each qualifying child.

Baseline:
master

$2500 credit for each qualifying child:
screen shot 2015-11-11 at 3 29 30 pm

@MattHJensen Could you review?

PR #459, opened 2015-11-12
Added function & parameters for personal refundable credit

Amy-Xu:
This personal refundable credit's designed to replace current standard deduction and therefore eliminate the bifurcate filing routine. All relevant parameters are set to zeros by default. Any reform values need to be input from user-defined json nodes.

For example, if we want to have a credit of $2,000 for individual and $4,000 for married filing jointly, phaseout starting from $150,000 (individual) / $300,000 (married filing jointly) at the rate of 0.04 per dollar of AGI exceeding the thresholds, the json nodes would be like:

```
reform = {2015: {"_II_credit": [[2000, 4000, 2000, 2000]],
                 "_II_credit_ps": [[150000, 300000, 150000, 150000]],
                 "_II_credit_prt": [0.04]}}
```

The results would look like this using diagnostic tables:

Baseline (current law)

Reform Scenario1 (add the refundable credit and keep standard deduction)

Reform Scenario2 (add the refundable credit and repeal standard deduction)


MattHJensen:
@amy-xu, could you add a table that adds the new credit and repeals both the standard deduction and the personal exemption?

Amy-Xu:
@MattHJensen This is the table when standard deduction and personal exemption repealed.

screen shot 2015-11-13 at 9 54 06 am

Also, I haven't added this credit to refundable credit yet. Because if I add it to refundable credit totals before deducting refundable credit from `09200`, the individual income tax would just be extra low due to the double deduction of this credit. Do you think we should still add this credit to refundable credit but after calculating income tax?

MattHJensen:
> Also, I haven't added this credit to refundable credit yet. Because if I add it to refundable credit totals before deducting refundable credit from 09200, the individual income tax would just be extra low due to the double deduction of this credit. 

Are you treating this provision as a deduction against income rather than a credit against tax, or am I misinterpreting? I thought it was just a simple refundable credit; am I wrong?

Amy-Xu:
I see where's confusing. I thought even though it is a credit, it should be deducted from taxable income before calculating tax liabilities in order to get right brackets and rates for taxable income. So it wouldn't be treated entirely as a refundable credit (deducted right before final results). But actually the taxable income are now turned into something more like AGI, no deduction or exemption or credits at all. Is that correct?

MattHJensen:
> But actually the taxable income should be something more like AGI, without any deduction or exemption or credits at all. Is that correct? 

This new provision should just be a refundable credit based on MARS. If the user also turns off standard deduction and personal exemption, then taxable income would look more like AGI; those reforms are likely to be coupled with the use of this new provision, but they are separate from it.

Amy-Xu:
After moving the credit to refundable credits, baseline remains the same. Here're the results for the two reform scenarios.

Reform scenario1 (add credits and keep standard deduction and personal exemption):
screen shot 2015-11-13 at 11 07 44 am

Reform Scenario2 (add the credit and repeal both standard deduction and personal exemption):
screen shot 2015-11-13 at 11 07 52 am

feenberg:
Would someone like to describe the proposal?

dan

Amy-Xu:
@feenberg Here's a brief description I copied:

>
Creates a new $2,000 (individual) / $4,000 (married filing jointly) refundable personal tax credit in place of the standard deduction: Credit phases out beginning above $150,000 (individual) / $300,000 (married filing jointly) and would be unavailable to taxpayers with an annual income in excess of $200,000 (individual) / $400,000 (married filing jointly).

And a bit more here:
screen shot 2015-11-13 at 4 28 08 pm

MattHJensen:
Amy, I have the same question here as for #453:

> should only the portion of the credit that takes _iitax liabilities below zero be in the refundable credits row, with the rest in the nonrefundable credits row? Are other refundable credits already in the code decomposed in some way?

MattHJensen:
Thanks @Amy-Xu, merging.

PR #465, opened 2015-11-17
Add consider_imputations argument to Records class constructor

martinholmer:
The default value of the new argument is True, which was the implied logic before the addition, so that all code that calls the Records class constructor will work exactly as before.  This new constructor argument has been added to make explicit whether or not to do the variable imputations.  The addition of this argument makes the Records class constructor argument consider_imputations analogous to the Calculator class constructor argument sync_years.

Also, use this new argument to streamline logic of SimpleTaxIO class.

@MattHJensen and @Amy-Xu,  Please review to confirm that this addition to the Records class constructor does not distrupt any existing uses of the Records class.

### DIFF
--- a/taxcalc/records.py
+++ b/taxcalc/records.py
@@ -15,24 +15,30 @@ class Records(object):
     Parameters
     ----------
     data: string or Pandas DataFrame
-        string describes CSV file in which records data reside
-        DataFrame already contains records data
+        string describes CSV file in which records data reside;
+        DataFrame already contains records data;
         default value is the string 'puf.csv'
 
     blowup_factors: string or Pandas DataFrame
-        string describes CSV file in which blowup factors reside
-        DataFrame already contains blowup factors
+        string describes CSV file in which blowup factors reside;
+        DataFrame already contains blowup factors;
         default value is filename of the default blowup factors
 
     weights: string or Pandas DataFrame
-        string describes CSV file in which weights reside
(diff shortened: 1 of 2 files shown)


martinholmer:
@talumbau,  Is Travis-CI broken?  I can pass all the unit tests on my computer.  What's up (or down)?

talumbau:
Hmm. For some reason the 'wget' of a necessary file was unsuccessful during the build. I restarted it and everything was fine. Travis CI had some problems earlier today (at least, according to their status page). Seems like just a fluke.

talumbau:
oh, no need to do a 'fake' push to trigger a re-test. When you click on the red 'X' or green check on a Travis CI build, you get to a status page that will have a section like this:

screen shot 2015-11-16 at 10 13 24 pm


The 'circle arrow' button on the right will restart the build in case there was some kind of transient error.

martinholmer:
T.J. @talumbau said:
> The 'circle arrow' button on the right will restart the build in case there was some kind of transient error.

Thanks for the tip!

martinholmer:
T.J. @talumbau,  Travis-CI seems to be broken again.

MattHJensen:
@Amy-Xu, could you review and merge if everything looks good to you?

Amy-Xu:
LGTM. Merging.

Thanks @martinholmer!

PR #466, opened 2015-11-17
moved 09 extrapolation to records.py

Amy-Xu:
Previously if any user use Records object directly for any testing purposes without setting up a calculator,  many variables, including the weights, wouldn't have the 09 extrapolated values. For example, the weighted sum of wages & salaries (e00200) ended up at unseasonably high level due to the wrong weights variable. In this PR, 2009 extrapolation is moved from calculate.py to records.py to solve this issue.

@MattHJensen @martinholmer Could you review?

### DIFF
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -66,9 +66,6 @@ class Calculator(object):
             print("You loaded data for " +
                   str(self._records.current_year) + '.')
 
-            if self._records.current_year == 2009:
-                self.records.extrapolate_2009_puf()
-
             while self._records.current_year < self._policy.current_year:
                 self._records.increment_year()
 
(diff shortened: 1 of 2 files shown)


martinholmer:
@Amy-Xu said:
> Could you review?

Looks like a good change to me except for one thing:  shouldn't ```extrapolate_2009_puf()``` now be a PRIVATE method of the Records class?   If so, it should be renamed ```_extrapolate_2009_puf()``` and moved BELOW the line that says ```# --- begin private methods of Records class --- #```.  And the call to it needs to add the leading underscore.

martinholmer:
@Amy-Xu,  Looks good.  I'll merge as soon as Travis-CI (which is moving a the speed of a glacier) finishes checking.

PR #468, opened 2015-11-17
update fica param metadata

MattHJensen:
The fica params metadata in `current_law_policy.json` indicated that we do not calculate payroll taxes. Now we do, so I updated these.


martinholmer:
@MattHJensen,  Good that you caught this.

PR #471, opened 2015-11-18
Add set_current_year method to Records class

martinholmer:
Also revise and add documentation for Records class.

### DIFF
--- a/taxcalc/records.py
+++ b/taxcalc/records.py
@@ -420,9 +420,8 @@ class Records(object):
         Sets current year to specified value and updates FLPDYR variable.
         Unlike increment_year method, blowup and reweighting are skipped.
         """
-        diff_years = new_current_year - self._current_year
-        self.FLPDYR += diff_years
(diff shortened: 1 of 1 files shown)


MattHJensen:
@Amy-Xu, could you review this?

Amy-Xu:
LGTM!

martinholmer:
@Amy-Xu said:
> LGTM! Do we need to add a test for this new set_current_year() function to verify extrapolation's off?

I'm very bid on testing, but the simplicity of the set_current_year() method --- three lines of code --- seems to make it very clear that there is no step 1 or step 2 extrapolation as there is in the increment_year() method just above.  Seems as if adding a test is a bit of overkill, but maybe I'm missing something.
cc @MattHJensen

MattHJensen:
@Amy-Xu, could you merge when you are ready?

Amy-Xu:
LGTM. Merging.

Issue #469, opened 2015-11-17
Earning Split Issue?


GoFroggyRun:
I might be terribly wrong, but while working on the validation  against sas calculator , after implementing the earning split into sas calculator, I found that, by printing `e00200p` and `e00200s` outputs/variables in the `EI_FICA` function, the earning split functionality seems to be null in our calculator, in the way that they won't simultaneously take positive values.
I'm using the `puf` file of size 129682845 bytes and executing `test.py` under the tax-calculator directory on current OSPC master.  I was wondering am I the only person experiencing such issue? If so, please advice. Much appreciated. 
@Amy-Xu @martinholmer @MattHJensen

martinholmer:
Sean (@GoFroggyRun),
I'm working on a different project (that involves reading in the puf.csv file) and I'm experiencing some problems as well.  My problem is a python crash.  But it is too early to tell exactly what is the cause of my problem; it may well be my misuse of the basic Tax-Calculator classes.  After I've found out what is causing my problem, I'll report back here.  It may be related or it may be completely unrelated.
cc @Amy-Xu and @MattHJensen

GoFroggyRun:
@martinholmer 
Thanks for the updates. I'll be trying to fix this "issue" myself before it got confirmed.

MattHJensen:
@GoFroggyRun, notice that there are some differences between `calc_all()` and `calc_all_test()` in `calculate.py`. 

`test.py` calls `calc_all_test()`, but every other application of the calculator that I know of uses `call_all()`. For one thing, `calc_all_test()` does not execute the proper itemized filing optimization. It also doesn't allow for the itemized deduction surtax. 

It might be that this difference is causing your issues or it might not be, I haven't had time to check further. But I would suggest that you look at test_calculate.py for more standard uses of the calculator class. 

As an aside, I will be opening an issue suggesting that we remove the `test.py` script and `calc_all_test()` entirely (so we don't need to have a discussion about the usefulness of those scripts here).

GoFroggyRun:
@MattHJensen I see. Thanks for the advice. 
It is quite puzzling, however, that the `_earning_split` variable takes either 0 or 1, where it takes 0 when the conditions got satisfied, for each observation using `test.py` script. And I'm actually pretty confused by why this is happening. Shouldn't its value be constant if this part doesn't get executed at all? The situation right now is that the switch which assigns values to `_earning_split` seems to be valid while `_earning_split` by no means take the right value when the switch is on.
[...]

PR #474, opened 2015-11-18
Earning split issue fixed

GoFroggyRun:
This PR deals with issue #469 . The `_earning_split` variable was previously of integer type so that the earning split functionality was correctly implemented in the calculator. By multiplying a constant of float type, we are now doing the division in the right way. See the table for some simple verification:

|Observation|Earning Split Value|
|-----|------|
|0|    1.000000|
|1  |  1.000000|
|2  |  1.000000|
|3 |   0.619925|
|4  |  1.000000|
|5  |  0.619925|
|6 |   1.000000|
|7    |0.022172|
|8  |  0.741935|
|9 |   1.000000|

@MattHJensen @martinholmer Please take a look.

### DIFF
--- a/taxcalc/records.py
+++ b/taxcalc/records.py
@@ -739,7 +739,7 @@ class Records(object):
                          self.wage_head + self.wage_spouse,
                          0)
         self._earning_split = np.where(total != 0,
-                                       self.wage_head / total, 1)
+                                       (self.wage_head * 1.0) / total, 1.0)
         self.e00200p = self._earning_split * self.e00200
         self.e00200s = (1 - self._earning_split) * self.e00200
 


martinholmer:
Sean @GoFroggyRun,  This looks like an improvement to me.  It will be interesting to see by how much aggregate FICA taxes go up after this bug fix?  What do you think, @MattHJensen, about this fix?

MattHJensen:
@martinholmer and @GoFroggyRun, I agree that this is an improvement. Great catch @GoFroggyRun.  Our dynamic team is using these results in a presentation tomorrow, so I am going to merge this. @GoFroggyRun, it would be helpful for you to look into @martinholmer's query when you have a chance:

> It will be interesting to see by how much aggregate FICA taxes go up after this bug fix?

Issue #452, opened 2015-11-10
p23250 & e23250


Amy-Xu:
I thought the two variables are equivalent in the current master, but from PR #451 it seems they are not exactly interchangeable in every single cases due to SimpleTaxIO. Merging the PR and posting this issue to take one more look. 

cc @talumbau @martinholmer

martinholmer:
Here are the SImpleTaxIO results that generated this issue.  These results were generated on a local branch where the mtr() code had been revised to accept either 'e23250' or 'p23250' as the income_type_str.  This branch also changed SimpleTaxIO to compute marginal tax rates with respect to long-term capital gains (instead of taxpayer earnings).  The two aliases for long-term capital gains do not produce the same results as can be seen below.
```
==> when income_type_str='e23250':
validation> bash test c13 . save
M	taxcalc/validation/c13.taxdiffs
validation> awk '{n[$7]++}END{for(i in n)print i,n[i]}' c13.in.out-simtax | awk '$2>3000' | sort
0.00 36809
12.75 4405
15.00 19411
25.00 5471
25.80 4420
27.75 4053
validation>

==> when income_type_str='p23250':
validation> bash test c13 . save
M	taxcalc/validation/c13.taxdiffs
validation> awk '{n[$7]++}END{for(i in n)print i,n[i]}'
c13.in.out-simtax | awk '$2>3000' | sort
0.00 100000
validation>
```

Amy-Xu:
It make more sense to me now. It seems there's another function '_read_input'  mainly focusing on validation of data type and formats in `simpletaxio,py', without the [alias setup](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/records.py#L658). But I'm still a bit confused on why it is p23250 that is all zeros but not e23250, because both 09 PUF and cps-puf should only have p23250 instead of e23250. Anyways in my opinion this issue comes in two parts.

- Do we need to make the testing and a regular calculator consistent in handling the variable names? In a regular calculator, because most variables have a leading e, we do not want to add any confusion for users on this kind of detailed naming stuff -- they've already had a lot to digest thinking about the 5-digit variable indexes. If any of the users is expert enough, they should be able to find info from IRS 09 PUF booklet. I'm aware that imputations and extrapolation is not quite necessary for this testing, but how about variable name consistency? 

[...]

Issue #425, opened 2015-10-29
calculator should produce sensible results with any subset of input variables


MattHJensen:
The calculator should be able to produce sensible results with any subset of input variables, especially when the extrapolation is turned off. Variables that are missing should assumed to be zeros. (Sometimes an entire grouping of variables may need to be assumed to be zero if any one of the variables in the grouping is missing). 
 
I believe satisfying this requirement will require the following changes to both the `Records` class and to `functions.py`: in `Records`, arrays of zeros will need to be created based on the presence of the variable in the data rather than based on whether the variable is listed in `ZEROED_NAMES`. In `functions.py`, logic will need to be controlled by the presence of variables (or presence of all variables in a groups of variables) rather than by the value of `puf` or `cmp`. 

cc @martinholmer, @feenberg, @talumbau, @Amy-Xu

MattHJensen:
This is related to https://github.com/open-source-economics/Tax-Calculator/issues/319#issuecomment-122541546.

martinholmer:
@MattHJensen said:
> The calculator should be able to produce sensible results with any subset of input variables, especially when the extrapolation is turned off. Variables that are missing should assumed to be zeros. (Sometimes an entire grouping of variables may need to be assumed to be zero if any one of the variables in the grouping is missing).
>
> I believe satisfying this requirement will require the following changes to both the Records class and to functions.py: in Records, arrays of zeros will need to be created based on the presence of the variable in the data rather than based on whether the variable is listed in ZEROED_NAMES. In functions.py, logic will need to be controlled by the presence of variables (or presence of all variables in a groups of variables) rather than by the value of puf or cmp.

@Amy-Xu and Sean @GoFroggyRun,
Matt has asked me to coordinate the work involved in resolving this issue #425 along with issues #452 and #436.  In this comment, I describe a strategy for resolving these issues for your comment and discussion.  After reaching a consensus on strategy, we will break the strategy down into smaller pieces of work and assign them to you.  But both Matt and I feel it is important for you to be involved in the strategic discussion.

As background, part of Matt's motivation for raising this issue is that in the coming months we might be working with other organizations that have their own tax models to conduct cross-model validation work.  So, the strategy I describe below tries to move the Tax-Calculator code in a direction that will make that collaboration easier.

[...]

PR #475, opened 2015-11-19
Add IncomeTaxIO class for cross-model validation using PUF

martinholmer:
The IncomeTaxIO class is a more complex version of the SimpleTaxIO class.  This initial version reads csv-formatted PUF data and writes Internet-TAXSIM output.  All the FICA-related output variables are set to zero to accomodate situations where the input PUF data do not include information about the split between head and spouse earnings.  There are two unit tests of this new class in the test_incometaxio.py file.  This pull request also includes a command-line interface to the IncomeTaxIO class called inctax.py.

When using puf.csv as input, the results generated by this inctax.py script, and the new test.py script, will help guard against performance regressions as we make progress on Issues #425, #436, and #452.


martinholmer:
@MattHJensen, @Amy-Xu, Sean @GoFroggyRun,
Do you want to look this over before it gets merged?  There is nothing here the affects the logic of the Tax-Calculator itself, but sometimes it is easier to see what code is being added in the context of a pull request.  Any questions, comments or suggestions would be welcome.

GoFroggyRun:
@martinholmer The IncomeTaxIO class looks good to me. It might be hard for me to get any suggestions offhand, by looking at the code. We can definitely figure out further improvements afterwards though.

MattHJensen:
This looks great and I plan to merge it. @Amy-Xu, any objections? 

@martinholmer, my next step will be to run the same data through this and SimpleTaxIO (modifying headers as needed) to make sure that I understand everything well enough to get the same output.

Amy-Xu:
No objections.

MattHJensen:
Merging. Thanks @martinholmer!

martinholmer:
@MattHJensen said:
> my next step will be to run the same data through this and SimpleTaxIO (modifying headers as needed) to make sure that I understand everything well enough to get the same output.

That's an interesting idea I hadn't thought about.  As you work with these two capabilities remember that (right now, at least) the IncomeTaxIO class calls the impute_variables() method in the Records class, while SimpleTaxIO does not call the impute_variables() method.  I'll be interested in what your experience is like.

MattHJensen:
> As you work with these two capabilities remember that (right now, at least) the IncomeTaxIO class calls the impute_variables() method in the Records class, while SimpleTaxIO does not call the impute_variables() method.

Thanks for this reminder. I'll set consider_imputations = False in IncomeTaxIO when testing against SimpleTaxIO. For our initial round of cross-model validation with TPC we'll likely want to skip the imputations as well.

martinholmer:
@MattHJensen said:
> Thanks for this reminder. I'll set consider_imputations = False in IncomeTaxIO when testing against SimpleTaxIO. For our initial round of cross-model validation with TPC we'll likely want to skip the imputations as well.

It will be interesting to see how that goes, but my initial impression was that there was not an easy and sensible way to skip the impute_variables() method.  Mostly it is an issue of the ```_numextra``` variable being a key parameter and not being in the IRS-SOI-PUF file.  Maybe I'm confused about what you mean by "use any subset of the PUF data."  Do you mean any subset of the IRS-SOI-PUF data or any subset or the OSPC-generated ```puf.csv``` file (which is actually a PUF-CPS merge)?

MattHJensen:
> Do you mean any subset of the IRS-SOI-PUF data or any subset or the OSPC-generated puf.csv file (which is actually a PUF-CPS merge)?

Ideally this distinction shouldn't matter. When I say "use any subset of the PUF data," I suppose I really mean "use any subset of input data." It could be that we need to specify some minimum set of variables that could possibly include `_numextra,`  but I haven't yet confirmed that there is a problem with having `_numextra` as all zeroes, which is how we've talked about treating missing variables.

PR #477, opened 2015-11-20
Added epsilon to weighted_share_of_total function for diff tables

Amy-Xu:
This is similar to what we have in drop q, just to make sure that even when the tax liabilities aren't big enough, diff tables can still be generated.

@MattHJensen Have a few questions for this function:
- Is calc1 in the create_difference_table always the baseline? Is there any scenario where users want to compare diffs between two reform plans?

- In either case, which calculator should row label classifier come from, calc1 or calc2?

- Do we need to add any docs to let users know that if they ever want to see the diff in the reform year, they need to do increment_year accordingly and then calc_call()?

@talumbau just a heads up, Matt wants to have this merged before tomorrow's release. I should be able to have things fixed quickly after knowing what to do with the issues above. I'll let you know when I finish everything. Thanks!

### DIFF
--- a/taxcalc/tests/test_utils.py
+++ b/taxcalc/tests/test_utils.py
@@ -15,6 +15,7 @@ from pandas.util.testing import assert_series_equal
 from numba import jit, vectorize, guvectorize
 from taxcalc import *
 from csv_to_ascii import ascii_output
+from taxcalc.utils import EPSILON
 
 # use 1991 PUF-like data to emulate current PUF, which is private
 TAX_DTA_PATH = os.path.join(CUR_PATH, '../../tax_all1991_puf.gz')
@@ -226,7 +227,7 @@ def test_weighted_share_of_total():
     df = DataFrame(data=data, columns=['tax_diff', 's006', 'label'])
     grped = df.groupby('label')
     diffs = grped.apply(weighted_share_of_total, 'tax_diff', 42.0)
-    exp = Series(data=[16.0 / (42. + 1e-3), 26.0 / (42.0 + 1e-3)],
+    exp = Series(data=[16.0 / (42. + EPSILON), 26.0 / (42.0 + EPSILON)],
                  index=['a', 'b'])
     exp.index.name = 'label'
     assert_series_equal(exp, diffs)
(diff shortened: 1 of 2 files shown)


MattHJensen:
@Amy-Xu, nice bug catch and thanks for fixing this. 

> Is calc1 in the create_difference_table always the baseline? Is there any scenario where users want to compare diffs between two reform plans?

For now let's assume calc1 is always calc_x. 

Users may want to compare diffs between two reform plans, but I don't think it is worth the effort right now to allow the user to specify a third (baseline) calculator for the row label classifiers. There might be other ways around this issue that we haven't thought of. Perhaps this should be an issue. 

> In either case, which calculator should row label classifier come from, calc1 or calc2?

Row label classifier for distributional tables should come from calc_x (calc1). cc @feenberg. 

> Do we need to add any docs to let users know that if they ever want to see the diff in the reform year, they need to do increment_year accordingly and then calc_call()?

That does seem like it would be helpful.

Amy-Xu:
@MattHJensen Could you take a look again? I finished editing everything I had in mind.

MattHJensen:
This lgtm. @talumbau, could you review and merge if all looks good?

taxcalc/tests/test_utils.py
```diff
@@ -226,7 +226,8 @@ def test_weighted_share_of_total():
     df = DataFrame(data=data, columns=['tax_diff', 's006', 'label'])
     grped = df.groupby('label')
     diffs = grped.apply(weighted_share_of_total, 'tax_diff', 42.0)
-    exp = Series(data=[16.0 / 42., 26.0 / 42.0], index=['a', 'b'])
+    exp = Series(data=[16.0 / (42. + 1e-3), 26.0 / (42.0 + 1e-3)],
```
talumbau: If you added code like this:

```
from taxcalc.utils import EPSILON
```

and then did:

```
exp = Series(data=[16.0 / (42. + EPSILON), 26.0 / (42.0 + EPSILON)],
```

then you would never have to change this if `EPSILON` changed.

Amy-Xu:
Thanks T.J.! @talumbau Just fixed the hard-coded number in the test.

talumbau:
:+1:

PR #478, opened 2015-11-20
Added haircut parameter for all taxes paid to state and local government

Amy-Xu:
There are two issues involved in this PR:

- Current master version of state and local haircut parameter is not applied previously, in that it's applied to the combo of income taxes, sales taxes, real estate and personal property taxes under regular tax, but only to income taxes and sales taxes under AMT. We need to make this parameter applied consistently.

- The individual income tax (IIT) liabilities can change quite differently, depending on what variables this haircut parameter is applied to.
     - if applied to the combo of income taxes and sales taxes (e18400), real estate (e18500) and personal property taxes (e18800), the IIT change is about at 100 billion level on the out years. 
screen shot 2015-11-19 at 11 04 27 am
     - however, if applied to just income and sales tax, the IIT change is only half of that size.
screen shot 2015-11-19 at 12 00 58 pm

To solve the issues mentioned above, in this PR I added one more haircut parameter `ID_AllTaxPaid_HC`, to give users the option either repealing the combo deduction or just the income and sales taxes. If this parameter is set to one, then IIT change is the same as first table presented above. If `ID_StateLocalTax_HC` is set to one, then IIT change would be the second table.

### DIFF
--- a/taxcalc/current_law_policy.json
+++ b/taxcalc/current_law_policy.json
@@ -983,7 +983,7 @@
         "value":     [0.8]        
     },
     "_ID_StateLocalTax_HC":{
-        "long_name": "Deduction for state and local taxes haircut.",
+        "long_name": "State and local taxes (Income and Sales) deduction haircut.",
         "description": "This percentage reduces the state and local tax deduction." ,
         "irs_ref": "",
         "notes": "This parameter allows for the implementation of Option 51 from https://www.cbo.gov/sites/default/files/cbofiles/attachments/49638-BudgetOptions.pdf.",
@@ -995,11 +995,11 @@
         "col_label": "",
         "value": [0]
     },
-    "_ID_AllTaxPaid_HC":{
-        "long_name": "Deduction for all taxes paid at state and local level (Income, Sales, Real Estate, and Personal Property).",
-        "description": "This percentage reduces all taxes paid deduction." ,
+    "_ID_RealEstate_HC":{
+        "long_name": "Real estate taxes deduction haircut.",
+        "description": "This percentage reduces real estate taxes paid eligible to deduct in itemized deduction." ,
         "irs_ref": "",
-        "notes": "This parameter is currently used to eliminate this part of itemized deduction.",
+        "notes": "This parameter is currently used to eliminate real estate taxes paid itemized deduction.",
         "start_year": 2013,
         "col_var": "",
         "row_var": "",
--- a/taxcalc/current_law_policy.json
+++ b/taxcalc/current_law_policy.json
@@ -403,7 +403,7 @@
         "row_var": "FLPDYR",
         "row_label": ["2013"],
         "cpi_inflated": true,
-        "col_label": ["Medical", "StateLocal", "Casualty", "Miscellaneous", "Interest", "Charity"],        
+        "col_label": ["Medical", "AllTaxPaid", "Casualty", "Miscellaneous", "Interest", "Charity"],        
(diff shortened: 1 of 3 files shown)


MattHJensen:
@Amy-Xu, this is great news. Excellent detective work! 

As far as parameter naming, I think it would make sense to do something like the following:

`ID_StateLocalTax_HC` -> `ID_StateLocalTaxExRealEstate_HC
`ID_AllTaxPaid_HC` -> `ID_StateLocalTax_HC`

This would have the added advantage of not requiring any changes to the webapp in order for the ID_StateLocalTax_HC to continue working as expected. Meta data would need to be updated as well, of course. 

It might also be worth adding a note to the (renamed) `ID_StateLocalTax_HC` that it includes real estate taxes, matching the CBO Budget Option estimate but not the JCT Tax Expenditure estimate. Also adding a note to the `ID_StateLocalTaxExRealEstate_HC` that it is a closer match to the JCT Tax Expenditure estimate.

What do you think?

MattHJensen:
You also will likely need to run the validation tests (see the contributor guide at taxcalc.readthedocs.org). I wonder whether this will create some mismatches in our reform comparisons against internet taxsim.

Amy-Xu:
I was hesitant to change the names as well, and do want to keep the original names for less trouble on the testing side. But currently in this PR, `ID_StateLocalTax_HC` covers what so-called as state and local taxes (e18400) in puf, which includes income taxes and sales taxes but no real estate taxes. While `ID_AllTaxPaid_HC` covers not only the state and local taxes, but also real estate taxes (e18500), personal property taxes (e18800, non-puf), and other taxes (e18900). How do you think to rename the two haircut parameters?

MattHJensen:
Real estate taxes (e18500) and personal property taxes (e18800) are almost entirely paid to states. There is some foreign real estate taxes paid in e18500, but I expect that is negligible. Therefore If we removed e18900 from ID_AllTaxPaid_HC, I think it would still be defensible to call this ID_StateLocalTax_HC and it would fairly closely match the CBO budget option for "eliminate the deduction of state and local taxes" https://www.cbo.gov/budget-options/2013/44799. What do you think? Perhaps the meta data for the param should indicate that it also eliminates the deduction for foreign real estate taxes paid, but that that is likely small.

Amy-Xu:
Yea that totally makes sense, but how about this one `ID_StateLocalTaxExRealEstate_HC` you suggested? It doesn't include real estate taxes, and perhaps we shouldn't have the real estate part in the name. But if real estate's removed, it's going to me the same as the other one.

MattHJensen:
@Amy-Xu, can you confirm when you set the real estate and state and local haircuts to one you get the same thing as when you set the old ID_AllTaxPaid_HC to 1?

Amy-Xu:
Thanks for the reminder! @MattHJensen Here're the tax liability changes from applying only `ID_StateLocalTax_HC`, and the combo of `ID_StateLocalTax_HC` and `ID_RealEstate_HC`. 

screen shot 2015-11-20 at 2 06 23 pm

With the original diagnostic tables attached here:

Baseline
baseline

Only `ID_StateLocalTax_HC`
statelocal

Both `ID_StateLocalTax_HC` and `ID_RealEstate_HC`
screen shot 2015-11-20 at 1 57 58 pm

Lastly, a screen shot running validation test against TAXISM
screen shot 2015-11-20 at 2 09 42 pm
It seems to be good.

MattHJensen:
Looks great @Amy-Xu, merging. I think last todo on this is to open a webapp issue to get the real estate hc included there.

PR #479, opened 2015-11-20
Streamline handling of policy parameter indexing rates

martinholmer:
The commit messages describe what has been changed.  A new public Policy wage_growth_rates() method was added to be analogous to the existing inflation_rates() method.  Use both those rates methods in the indexing_rates_for_update() method in the parameters_base.py file.  Also, add ParametersBase methods for inflation_rates and wage_growth_rates, which both return None, in order to better conform with object-oriented design principles, and to simplify the coding of the indexing_rates() and indexing_rates_for_update() methods in the ParametersBase class.


martinholmer:
T.J. @talumbau,  Did I miss any use of the Policy class inflation_rates() method in the several TaxBrain-related repos?

talumbau:
thanks for checking! I double-checked and it is an unused function in the relevant repos. I believe this function was used at one point in the past but it has been superseded by much more advanced capabilities.

taxcalc/current_law_policy.json
```diff
@@ -433,10 +433,10 @@
         "value":    [[0.0765, 0.1598, 0.2106, 0.2106]]        
     },
     "_SS_Earnings_c":{
-        "long_name": "Max Taxable Earnings for Social Security",
-	"description": "Only earnings lower than this maximum amount are subjected to Social Security tax (OASDI tax).",
+        "long_name": "Maximum Taxable Earnings for Social Security",
+	"description": "Only individual earnings below this maximum amount are subjected to Social Security (OASDI) payroll tax.",
         "irs_ref": "W-2, Box 4, instructions",
-        "notes": "",
+        "notes": "This parameter is indexed by the rate of growth in average wages, not by the inflation rate.",
```
MattHJensen: Helpful changes!

Amy-Xu:
Do we need to give users a public method showing them the inflation rates used tho?

martinholmer:
@Amy-Xu said:
> Do we need to give users a public method showing them the inflation rates used, though?

That was my thought when starting out this morning with the objective of adding a public wage growth rates method to make it comparable with inflation rates.  But then I found inflation rates was not being used anywhere in the Tax-Calculator or TaxBrain code, so I went in the other direction.  Maybe I should have stuck with my original intention.  What do you think?

Amy-Xu:
Does [this line](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/calculate.py#L172) count? @martinholmer

martinholmer:
@Amy-Xu said:
> Does this [middle] line count?
```
            target(self, self.growth._factor_target,
                   self.policy.inflation_rates,
                   self.policy.current_year + 1)
```

Not really because in pull request #480 the inflation_rates argument to the ```target``` function has been removed because it is unused in the function.

But the more I think about your original point, the more I think you are correct.  I'll add the public ```inflation_rates``` method back in and add an analogous ```wage_growths_rates``` method.  That makes more sense even if the current codebase is using neither of them.

Amy-Xu:
Totally. I just looked though #480 and tested the target function. Looks great.

MattHJensen:
@amy-xu, could you go ahead and merge if all looks good?

Amy-Xu:
LGTM. Thanks @martinholmer. Merging.

PR #483, opened 2015-11-24
Added real estate taxes paid deduction haircut to the benefit surtax

Amy-Xu:
As in previous PR #478, we separated the previous general state and local taxes paid into State & Local and Real Estate taxes paid. We need make the corresponding change to benefit surtax as well. In this PR, added one more element to the switch parameter and adjusted `BenefitSurtax()` function as well.

### DIFF
--- a/taxcalc/current_law_policy.json
+++ b/taxcalc/current_law_policy.json
@@ -403,8 +403,8 @@
         "row_var": "FLPDYR",
         "row_label": ["2013"],
         "cpi_inflated": true,
-        "col_label": ["Medical", "AllTaxPaid", "Casualty", "Miscellaneous", "Interest", "Charity"],        
-        "value":    [[true, 	   true,           true,          true,           true,      true]]
+        "col_label": ["Medical", "State & Local", "Real Estate", "Casualty", "Miscellaneous", "Interest", "Charity"],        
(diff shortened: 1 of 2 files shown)


Amy-Xu:
Here're the diagnostic tables imposing a surtax on both state and local, and real estate.
```
reform = {2015:{"_ID_BenefitSurtax_Switch": [[True, False, False, True, True, True, True]],
               "_ID_BenefitSurtax_crt": [0.8],
               "_ID_BenefitSurtax_trt": [0.2]}}
```

Baseline
screen shot 2015-11-24 at 11 23 31 am

Reform:
screen shot 2015-11-24 at 11 23 38 am

talumbau:
this looks good to me. The meaning of each index in `ID_BenefitSurtax_Switch` is assigned in only one place, which is where Amy made her changes, so that seems good. A better situation might be a test that tests that each index of `ID_BenefitSurtax_Switch` (as in, setting some flags and running a reform and getting certain expected results), but that seems difficult to do w/o using the PUF. +1 from me.

MattHJensen:
One nice test might be making sure that if you make the HC params 1 you get the same results as an equivalent surtax.

talumbau:
I'll leave it to @Amy-Xu or @MattHJensen to create a follow-up issue. Planning on merging this now so we can launch this feature on TaxBrain.

Issue #489, opened 2015-11-29
use _combined instead of _iitax for calculating income effect 


MattHJensen:
Currently we calculate the income effect like this: 

    income_effect = calc_y.behavior.BE_inc * (calc_y.records._iitax -
                                              calc_x.records._iitax)

I believe it should instead be like this: 

    income_effect = calc_y.behavior.BE_inc * (calc_y.records._combined -
                                              calc_x.records._combined)

feenberg:
On Sun, 29 Nov 2015, Matt Jensen wrote:

> 
> Currently we calculate the income effect like this:
> 
> income_effect = calc_y.behavior.BE_inc * (calc_y.records._iitax -
>                                           calc_x.records._iitax)
> 
> I believe it should instead be like this:
> 
> income_effect = calc_y.behavior.BE_inc * (calc_y.records._combined -
>                                           calc_x.records._combined)
>

Yes, the combined it and fica is the correct number to use.

dan

MattHJensen:
Thans Dan. Fixed in #491, so closing.

PR #491, opened 2015-11-29
calculate income effect w _combined instead of _iitax

MattHJensen:
As per #489, this pr calculates income effect with _combined instead of with _iitax. 

```
income_effect = calc_y.behavior.BE_inc * (calc_y.records._combined -
                                          calc_x.records._combined)
```


MattHJensen:
Based on @feenberg's review of the proposed code change in #489, I am merging.

PR #492, opened 2015-12-01
[WIP]Short term solution for capital income MTR

Amy-Xu:
A new function for capital income MTR is added in this PR due to the following two reasons.

- The one-cent margin needs to be weighted split among all capital income sources before calculation. This step is only needed for these capital income sources but not any other type of income.

- Not much validation needed at this point since we only calculate mtr based on a certain set of capital income sources.

This is still just a short-term solution, as it's easy to implement and relatively efficient computation wise, without messing up the original mtr function which is currently used by other team members for testing purposes. In long run, we need to combine the two mtr functions. 

@MattHJensen @martinholmer Could you take a look at this function? I haven't added any test yet, but will do once you're happy with the function.

### DIFF
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -276,13 +276,14 @@ class Calculator(object):
         return (mtr_fica, mtr_iit, mtr_combined)
 
     def capital_mtr(self):
-        capital_income_sources = ['e00300', 'e00400', 'e00600', 'e00650', 'e01000',
-                                  'e01400', 'e01500', 'e01700', 'e02000', 'e23250']
-        
+        capital_income_sources = ['e00300', 'e00400', 'e00600', 'e00650',
+                                  'e01000', 'e01400', 'e01500', 'e01700',
+                                  'e02000', 'e23250']
+
         length = len(self.records.s006)
         income_sum = np.zeros(length)
         originals = {}
-        
+
         # get the sum of all capital income and save the values for later use
         for capital_income in capital_income_sources:
             income_sum += getattr(self.records, capital_income)
@@ -295,7 +296,9 @@ class Calculator(object):
         finite_diff = 0.01  # a one-cent difference
         epsilon = 10e-20
         for capital_income in capital_income_sources:
-            income_up = originals[capital_income] * (1 + finite_diff/(income_sum + epsilon))
+            margin = finite_diff * originals[capital_income]/(income_sum +
+                                                              epsilon)
+            income_up = originals[capital_income] + margin
             setattr(self.records, capital_income, income_up)
 
         self.calc_all()
@@ -303,7 +306,7 @@ class Calculator(object):
 
         # delta of the results with margin added
         iitax_delta = iitax_up - iitax_base
-        
+
         # calculates mtrs
         mtr_iit = iitax_delta / finite_diff
 
@@ -311,7 +314,7 @@ class Calculator(object):
         for capital_income in capital_income_sources:
             setattr(self.records, capital_income, originals[capital_income])
         self.calc_all()
-        
+
         return mtr_iit
 
     def diagnostic_table(self, num_years=5):
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -281,14 +281,18 @@ class Calculator(object):
                                             'e01500', 'e01700', 'e02000',
                                             'e23250']):
 
+        num_sources = len(capital_income_sources)
         length = len(self.records.s006)
         income_sum = np.zeros(length)
+        no_capital_income = np.zeros(length)
         originals = {}
 
         # get the sum of all capital income and save the values for later use
         for capital_income in capital_income_sources:
             income_sum += getattr(self.records, capital_income)
             originals[capital_income] = getattr(self.records, capital_income)
+            no_capital_income = np.where(originals[capital_income] != 0,
+                                         1, no_capital_income)
 
         self.calc_all()
         iitax_base = copy.deepcopy(self.records._iitax)
@@ -301,6 +305,12 @@ class Calculator(object):
                                                                 epsilon)
             income_up = originals[capital_income] + margin
             setattr(self.records, capital_income, income_up)
+
+        # distribute the margin to interest if taxpayer has no capital income
+        interest = np.where(no_capital_income == 0,
+                            finite_diff,
+                            self.records.e00300)
+        setattr(self.records, 'e00300', interest)
 
         self.calc_all()
         iitax_up = copy.deepcopy(self.records._iitax)
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -289,9 +289,7 @@ class Calculator(object):
             originals[capital_income] = getattr(self.records, capital_income)
 
         self.calc_all()
-        fica_base = copy.deepcopy(self.records._fica)
         iitax_base = copy.deepcopy(self.records._iitax)
-        combined_taxes_base = iitax_base + fica_base
 
         # add the one-cent margin
         finite_diff = 0.01  # a one-cent difference   
@@ -300,19 +298,13 @@ class Calculator(object):
             setattr(self.records, capital_income, income_up)
 
         self.calc_all()
-        fica_up = copy.deepcopy(self.records._fica)
         iitax_up = copy.deepcopy(self.records._iitax)
-        combined_taxes_up = iitax_up + fica_up
 
         # delta of the results with margin added
-        fica_delta = fica_up - fica_base
         iitax_delta = iitax_up - iitax_base
-        combined_delta = combined_taxes_up - combined_taxes_base
         
         # calculates mtrs
-        mtr_fica = fica_delta / finite_diff
         mtr_iit = iitax_delta / finite_diff
-        mtr_combined = combined_delta / finite_diff
 
         # reset everything
         for capital_income in capital_income_sources:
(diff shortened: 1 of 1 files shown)


taxcalc/calculate.py
```diff
@@ -275,6 +275,48 @@ def mtr(self, income_type_str='e00200p',
         # return the three marginal tax rate arrays
         return (mtr_fica, mtr_iit, mtr_combined)
 
+    def capital_mtr(self):
+        capital_income_sources = ['e00300', 'e00400', 'e00600', 'e00650',
```
MattHJensen: It might be easier to test for consistency against the other mtr function (until we merge the two) if this list were a default argument. Then you could have a couple of kinds of tests. 

1. pass `['e00300']` to this function and make sure you get the same as when you use the preexisting `mtr()`. 

2. pass `['e00300', 'e2350']` to this function and make sure that you get the same thing as if you calculate the mtrs separately for those variables via the preexisting `mtr()` and then weight the mtrs after the fact.

Amy-Xu: Got you. Will do.

Amy-Xu:
@jdebacker In this PR you can see the new function I added for capital income sources. It includes 10 types of income you mentioned and split the one-cent margin according to each income type in proportion to the total capital income of that taxpayer. For taxpayers without any capital income, they get zero margin added to their capital income variables and therefore end up with zero MTR. Is that what you expect? Could you take a look at this function as well? Please let me know if anything doesn't make sense.

feenberg:
On Wed, 2 Dec 2015, Amy Xu wrote:

> 
> @jdebacker In this PR you can see the new function I added for capital income sources.
> It includes 10 types of income you mentioned and split the one-cent margin according
> to each income type in proportion to the total capital income of that taxpayer. For
> taxpayers without any capital income, they get zero margin added to their capital
> income variables and therefore end up with zero MTR. Is that what you expect? Could
> you take a look at this function as well? Please let me know if anything doesn't make
> sense.

I don't think this is going to be right, but I am not sure of the details.

If you split the penny in proportion to capital income shares, there is a 
problem if there in no capital income - the formula would have you 
dividing by zero, since the share of each form of capital income is that 
amount of income divided by the total capital income. It is not right to 
substitute zero for the undefined result. Consider the effect, which is to 
mix a bunch of zeroes in for taxpayers who would in fact pay tax on 
capital income.

If what we want is the average marginal rate in the income clasess, the 
the right thing to do is to calculate the tax, then multiply every capital 
income item by 1.01 and recalculate the tax. Then divide the aggregate 
change in tax by the aggregate change in income. This gives the average 
marginal rate of tax on capital income (which is not the same as the 
average of the taxpayer marginal rates).

If what we want is a rate to use in a behavorial equation, then we need a 
rate for taxpayers with no capital income and we should spread the penny 
accross the capital income items in some plausible but arbitrary manner. 
For example, in proportion to the average across taxpayers with such 
income.

If this isn't clear, we should have a phone call.

dan
617-863-0343

Amy-Xu:
Just spoke with Dan, here's a quick update. Among all the issues Dan pointed out, only one thing left is the mtr function for taxpayers without any capital income. It's incorrect to simply give them zero marginal tax rates, which is what I'm doing right now. We need to find a way to split the one-cent margin properly. This would make a pretty huge difference. @MattHJensen @jdebacker FYI

jdebacker:
Excellent. Dan's suggestions are great.

Re the zero capital income filers - I check the PUF and filers with capital income in [0, $100) have an average of 95% of their capital income from interest income.  

Shall we just round and assume that additional capital income for those with none comes entirely from interest income?  This looks to be a pretty good approximation.

@Amy-Xu

Amy-Xu:
Ahhhhh that's exactly what Dan suggested. Forgot to add that in my note. Will do asap.

Sent from my iPhone

On Dec 2, 2015, at 5:18 PM, Jason DeBacker <notifications@github.com<mailto:notifications@github.com>> wrote:


Excellent. Dan's suggestions are great.

Re the zero capital income filers - I check the PUF and filers with capital income in [0, $100) have an average of 95% of their capital income from interest income.

Shall we just round and assume that additional capital income for those with none comes entirely from interest income? This looks to be a pretty good approximation.

@Amy-Xu<https://github.com/Amy-Xu>

feenberg:
On Wed, 2 Dec 2015, Jason DeBacker wrote:

> 
> Excellent. Dan's suggestions are great.
> 
> Re the zero capital income filers - I check the PUF and filers with capital income in
> [0, $100) have an average of 95% of their capital income from interest income.
> 
> Shall we just round and assume that additional capital income for those with none
> comes entirely from interest income? This looks to be a pretty good approximation.

I agree.

dan

Amy-Xu:
@jdebacker @feenberg How should we split the margin if the taxpayer's different income sources cancelled out in the sum? I'm testing the scenario where includes two capital income, e00300 and e23250, in the MTR function. One taxpayer happens to have 173 for e00300 and -173 for e23250. The sum of the two is zero and I added an epsilon to the sum, which makes the split super bizarre, but I'm not sure how to solve it either.

Amy-Xu:
An update on the difference between two methods, one splits the margin and calculates mtr and another calculates mtr for each item first and weighted sum in the end. Out of all five records showing difference, four of them are due to the fact that they fall on the kinks of capital tax rate function. Specifically, after stacking their capital gains and dividends on their ordinary income, you can see that the total (c24530) sits at the second to the lowest tax bracket, which separates zero and 15% percent rate for capital gains. The last record, even thought I'm not 100% sure at this point, I guess the diff is from the issue I post above. Will update again on this.

jdebacker:
I think in general you want to think about how to compute the MTR across
all components if any of the components in negative.  E.g. if the component
is negative do we want to multiply it by 1.01?  That doesn't seem quite
right.  It seems like you'd want to multiply by -1.01 in that case.  But I
haven't thought about this too much and am happy to defer to @feenberg on
this.

Amy-Xu:
@jdebacker yea good call! Definitely haven't thought about all negative values yet. But let me explain this first, I also explained to Dan yesterday but didn't summarize it here. Right now we are not adding the margin by multiplying income by 1.01, which would matter if we average mtrs across all taxpayers. Instead, we simply added one cent to their income. This should be equivalent to the multiply method in our case.

feenberg:
Unless the kink produces a wild rate, no worries.

dan

Amy-Xu:
@feenberg The kinks indeed produces wild rates for some records, like 10e+22 level. I don't think that's normal for MTRs.

Another important issue worth your attention is how we should deal with negative capital income in general. The case I ran into was two kinds of capital income cancelled each other, which makes it difficult to split the margin before MTR calculation. How do you think we should deal with that?

feenberg:
On Thu, 3 Dec 2015, Amy Xu wrote:

> 
> @feenberg The kinks indeed produces wild rates for some records, like 10e+22 level. I
> don't think that's normal for MTRs.

You get that number from adding only a penny? That would be a tax 
difference of 10e20 dollars, which is pretty big. Can you find where in 
the code the notch occurs? Maybe the taxpayer has a million dollars of one 
type of income and only a dollar of another. Then the finite difference 
for the smaller amount is only a millionth of a cent and a small notch 
could give a large MTR. Often a notch indicates a programming error. If 
you look at the intermediate results, it may be clear where something big 
and wrong occurs. Legislative notches are typically only a few dollars.

> 
> Another important issue worth your attention is how we should deal with negative
> capital income in general. The case I ran into was two kinds of capital income
> cancelled each other, which makes it difficult to split the margin before MTR
> calculation. How do you think we should deal with that?

The macro guys should say what they need, but I wouldn't calculate the 
rates separately for each income type if it causes trouble as mentioned 
above. For a taxpayer with losses, adding a penny to income does produce 
the correct measure of the wedge in capital returns for that taxpayer.

dan

jdebacker:
@Amy-Xu Ok - then let's just do what you are doing for negative income (adding a positive amount) and use the implied MTRs even changes cancel out - this shouldn't result in cancelations, right?  If so, they seem legit and we are ok with MTR = 0 in those cases.

Re the large MTRs.  Until you can determine the cause and whether it's due to what Dan suggests, we can still proceed with what you have - we "clean" the data anyway - dropping really large ATR/MTRs from the data before we fit our functions to the data.

Amy-Xu:
Just figured out where the problem is for the extremely large MTRs. They overlap with those records that have two income variables canceled out. The reason why they end up with unreasonably large value is that I added an epsilon (10e-20) to the sum, which's supposed to be zero for those records. And then used the sum to calculate ratios of each income source in the sum for margin split. So the ratios ended up extremely large. But I figure this problem probably only exist when we use two variables as input in the testing scenarios. I'll see how likely this is going to happen when we use all ten types of capital income.

feenberg:
On Thu, 3 Dec 2015, Jason DeBacker wrote:

> 
> @Amy-Xu Ok - then let's just do what you are doing for negative income (adding a
> positive amount) and use the implied MTRs even changes cancel out - this shouldn't
> result in cancelations, right? If so, they seem legit and we are ok with MTR = 0 in
> those cases.
> 
> Re the large MTRs. Until you can determine the cause and whether it's due to what Dan
> suggests, we can still proceed with what you have - we "clean" the data anyway -
> dropping really large ATR/MTRs from the data before we fit our functions to the data.

plus or minus 100% would be a reasonable MTR filter. Labor income is often 
-.45 MTR (because of the EITC), and there are possibly other interactions 
that might make for even more extreme rates, so don't exclude negative 
rates.

dan

jdebacker:
Thanks, Dan!  Will do.

Amy-Xu:
Closing, replace by #497

PR #497, opened 2015-12-09
Added variables to current mtr function, replacing PR 492

Amy-Xu:
After further discussion with Dan, Jason and Matt, it seems the best way to deal with mtr weights is leaving the decision to our users -- they could choose to use either the negative capital income directly or the absolute values. Even though this method doesn't seem to be very computationally efficient, it would be more tricky to include the weighting step in MTR function given how many assumptions need to be made.

As of now, the capital income part of this function is just to connect Tax-calculator with our macro model. I assume there would be a script in the macro repo calling this mtr function for each item of capital income and then manually weighted average these mtrs right afterward. @jdebacker @talumbau  Does this sound right. If so, let me know you need me to provide a more update script for this function. I would love to read the current script if needed.

@martinholmer Do you think we need to add more warnings/tests regarding the new capital income variables added in the list? I wasn't sure as it seems we've already had the most essential ones.

cc @MattHJensen

### DIFF
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -222,9 +222,8 @@ class Calculator(object):
                                   'e00300', 'e23250',
                                   'e01700', 'e02400',
                                   'e00400', 'e00600',
-                                  'e00650', 'e01000',
-                                  'e01400', 'e01500',
-                                  'e02000']
+                                  'e00650', 'e22250',
+                                  'e01400', 'e02000']
         # check validity of income_type_str parameter
         if income_type_str not in mtr_valid_income_types:
             msg = 'mtr income_type_str="{}" is not valid'
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -220,7 +220,11 @@ class Calculator(object):
         """
         mtr_valid_income_types = ['e00200p', 'e00900p',
                                   'e00300', 'e23250',
-                                  'e01700', 'e02400']
+                                  'e01700', 'e02400',
+                                  'e00400', 'e00600',
+                                  'e00650', 'e01000',
+                                  'e01400', 'e01500',
+                                  'e02000']
         # check validity of income_type_str parameter
         if income_type_str not in mtr_valid_income_types:
             msg = 'mtr income_type_str="{}" is not valid'


martinholmer:
@Amy-Xu, 
I think adding these capital income types to the existing mtr() function is the most sensible approach.

And, yes, I do think we should be adding tests because there is probably more to be done for some of these new income types (for example, qualified dividends e00650 are part of dividends e00600 on Form 1040).  I will be happy to develop a set of mtr() tests in test_calculate.py after your pull request is merged into the master branch.  The objective of those tests would be to see if we are getting a reasonable range of marginal tax rates for each income type.

However, before you merge, let's discuss some of the seven income types you've added.
(1) What exactly is e01000?  If short-term gain/loss, why not use e22250 instead?
(2) How does e01500 differ from e01700 which is already in mtr()?

cc @MattHJensen and Jason @jdebacker

Amy-Xu:
Sounds good to me @martinholmer.

For your questions, I think Jason @jdebacker probably is in a better position to answer what the nuances between those variables are and why the macro model needs one or the other.

talumbau:
@Amy-Xu can you explain how I can use this change to do the equivalent of what Jason is doing here:

https://github.com/open-source-economics/OG-USA/pull/172/files#diff-5ed3b0443115d0443cd8aa1535ac0d79R69

including the call to the `capital_mtr` method (which is now gone). The goal is to make the changes in this PR available to the `ogusa` package and with this change, I'm not sure what to do. cc @jdebacker

Amy-Xu:
Sure. I think that line would be replaced by calculating each variable's MTR separately and using each income as weight to average all those MTRs.

```
# calculating MTRs separately
[mtr_fica_extra, mtr_iit_300, mtr_combined_extra] = calc1.mtr('e00300')
[mtr_fica_extra, mtr_iit_400, mtr_combined_extra] = calc1.mtr('e00400')
[mtr_fica_extra, mtr_iit_600, mtr_combined_extra] = calc1.mtr('e00600')
[mtr_fica_extra, mtr_iit_650, mtr_combined_extra] = calc1.mtr('e00650')
[mtr_fica_extra, mtr_iit_1000, mtr_combined_extra] = calc1.mtr('e01000')
[mtr_fica_extra, mtr_iit_1400, mtr_combined_extra] = calc1.mtr('e01400')
[mtr_fica_extra, mtr_iit_1500, mtr_combined_extra] = calc1.mtr('e01500')
[mtr_fica_extra, mtr_iit_1700, mtr_combined_extra] = calc1.mtr('e01700')
[mtr_fica_extra, mtr_iit_2000, mtr_combined_extra] = calc1.mtr('e02000')
[mtr_fica_extra, mtr_iit_23250, mtr_combined_extra] = calc1.mtr('e23250')

# weighted average all those MTRs
if every item in capital_income_sources == 0: # no capital income taxpayers
      capital_gain_mtr = mtr_iit_300 # give all the weight to interest income
else:
      capital_gain_mtr = (e00300 * mtr_iit_300 + e00400 * mtr_iit_300 + ... + e23250 * mtr_iit_23250) / sum_of_all_ten_variables

```

cc @jdebacker

talumbau:
Awesome, thanks! I will add this change to https://github.com/open-source-economics/OG-USA/pull/172 (rather, I will create a new PR based on that one with this change)

feenberg:
Isn't this going to be costly to calculate when embedded in a multi-period 
GE model?

dan

martinholmer:
@Amy-Xu , 
I've created a new testmtr.py program (see pull request #498) that can be expanded once the new capital income types in this pull request are added to the mtr() function.  Any suggestions about how to make the testmtr.py program better?

cc @MattHJensen and Sean (@GoFroggyRun) Wang

martinholmer:
Jason @jdebacker,
Can you provide some answers to the questions I originally asked @Amy-Xu?
> However, before you merge, let's discuss some of the seven income types you've added.
> (1) What exactly is e01000? If short-term gain/loss, why not use e22250 instead?
> (2) How does e01500 differ from e01700 which is already in mtr()?

cc @MattHJensen

Amy-Xu:
@feenberg Dan, yes it is compared to the previous version. But Matt and I still feel more inclined to go with this approach because 1) the weighting for negative capital income is tricky and including the weighting process inside MTR function needs to make a lot assumptions 2) the time spent in this this function is probably still not significant compared to the actual macro model running time. Our users are not waiting right in front of the screen anyways. They are going to receive an email when the results are ready.

That being said, we still want to improve the speed based on this approach. You mentioned earlier this week several income type in the list might give us the same MTRs. If this is true, then we do not need to run MTR for every single item, which could save us some time. The full list is here:

```
 ('e00300', 'e00400', 'e00600',
  'e00650', 'e01000', 'e01400',
 'e01500', 'e01700', 'e02000',
 'e23250')
```

Which ones do you think they will generate equivalent MTRs? I'll check them later today so we might be able to improve the efficiency a bit this way. Thank you Dan!

martinholmer:
@Amy-Xu,
I think the macro-model needs short-term capital gains E22250 rather than E01000 (which mixes short-term and long-term capital gains and is constrained to be no smaller than -3000 dollars when both E22250 and E23250 are negative).  The current mtr() method already handles long-term capital gains, which have a different tax treatment than short-term capital gains, using the E23250 variable.

Can you fix your pull request accordingly?  But if I have missed something here, let's discuss.

@MattHJensen and Jason @jdebacker and Dan @feenberg

martinholmer:
@Amy-Xu,
I think you need to remove e01500 (__all__ pension benefits and annuities received) from the list of new capital income types.  This variable is the sum of (a) federally-taxable pension benefits and annuities received --- which is already being handled by the mtr() method using e01700 --- and (b) pension benefits and annuities received that are not taxable under the federal income tax.  Obviously the marginal tax rate on non-taxable pension income is zero.  But the mtr() method has no idea about how much of a small increase in e01500 is taxable.  If none of it is taxable, then the mtr is zero by definition.  If all of it is taxable, then the mtr is exactly the same as for e01700, which is already handled in the mtr() method.  If some is taxable and the rest is non-taxable, the mtr() method doesn't have enough information to calculate a marginal tax rate.

It seems like weighting the zero mtr for non-taxable pension income with the positive mtr for e01700 --- which is already provided by the mtr() method --- should be part of what the macro model is doing with the marginal tax rates provided by the mtr() method.

cc @MattHJensen and Jason @jdebacker and Dan @feenberg

feenberg:
On Thu, 10 Dec 2015, Amy Xu wrote:

> 
> @feenberg Dan, yes it is compared to the previous version. But Matt and I still feel
> more inclined to go with this approach because 1) the weighting for negative capital
> income is tricky and including the weighting process inside MTR function needs to make
> a lot assumptions 2) the time spent in this this function is probably still not
> significant compared to the actual macro model running time. Our users are not waiting
> right in front of the screen anyways. They are going to receive an email when the
> results are ready.
> 
> That being said, we still want to improve the speed based on this approach. You
> mentioned earlier this week several income type in the list might give us the same
> MTRs. If this is true, then we do not need to run MTR for every single item, which
> could save us some time. The full list is here:
>
>  ('e00300', 'e00400', 'e00600',
>   'e00650', 'e01000', 'e01400',
>  'e01500', 'e01700', 'e02000',
>  'e23250')

I think there are only two rates:

    1) long term gains, qualified dividends (e23250,e00650)
    2) all other, including interest, schedule e, non-qualified dividends

dan

jdebacker:
I like the approach to just allowing each capital income item to be handled by the current mtr() method.  In the grand scheme of the computational model, this will not cost much time and the multiple calls to the mtr() method have the potential to be easily parallelized if we need more speed.

Re @martinholmer 's comments about the capital income sources - I agree with avoiding variables that give combined income sources.  @Amy-Xu, let me know if I need to provide you an updated list for the use of the mtr() method - or if you'll just include all capital income items that are taxable and are not combined items.

Re the approach of using the same rates for different items.  One would have to be careful with this under the policy proposal.  I think the key is that the mtr method() be able to handle the full potential of capital income items and one could write code specific to the baseline/policy tax law if one wanted to avoid excessive calls to the mtr() method (e.g. making use of the fact that there are only two rates under current law).

martinholmer:
Jason @jdebacker said:
>  I like the approach to just allowing each capital income item to be handled by the current mtr() method. In the grand scheme of the computational model, this will not cost much time and the multiple calls to the mtr() method have the potential to be easily parallelized if we need more speed.
>
> Re @martinholmer 's comments about the capital income sources - I agree with avoiding variables that give combined income sources. @Amy-Xu, let me know if I need to provide you an updated list for the use of the mtr() method - or if you'll just include all capital income items that are taxable and are not combined items.

So, the list of marginal tax rate income types in this pull request should be changed to the following:
```
mtr_valid_income_types = ['e00200p', 'e00900p',
                          'e00300', 'e00400',
                          'e00600', 'e00650',
                          'e22250', 'e23250',
                          'e01400', 'e01700',
                          'e02000', 'e02400']
```
Right?

After this list to changed to be what is shown above, I think this pull request is ready to be merged into the master branch.  After the merge I will revise the testmtr.py program to include these new capital income types.  The results of testmtr.py will show whether the marginal tax rates generated for the new income types have plausible distributions.

@MattHJensen and Dan @feenberg, do you see a better way to proceed?

martinholmer:
Jason @jdebacker said:
> Re the approach of using the same rates for different items. One would have to be careful with this under the policy proposal. I think the key is that the mtr method() be able to handle the full potential of capital income items and one could write code specific to the baseline/policy tax law if one wanted to avoid excessive calls to the mtr() method (e.g. making use of the fact that there are only two rates under current law).

I agree with Jason that any computational shortcuts are likely to produce incorrect marginal tax rates in many kinds of tax reforms.  Computational shortcuts with respect to mtr calculation seem very risky or, if done right, very complex.

Amy-Xu:
So it seems everyone's on the same page with the variable list. Could you review and merge the PR? @martinholmer 

cc @MattHJensen @jdebacker

PR #498, opened 2015-12-10
Add testmtr.py to check the distribution of marginal tax rates

martinholmer:
The point of this program is to check that the distribution of marginal tax rates for each income type is reasonable.

This program is analogous to the test.py program, in that both use the private puf.csv file as input.  This means neither program can be part of the Tax-Calculator py.test suite.

Currently the output of this program is as follows:
```
You loaded data for 2009.
Total number of data records = 219814
FICA mtr histogram bin edges:
     [0.0, 0.02, 0.04, 0.06, 0.08, 0.1, 0.12, 0.14, 0.16, 0.18, 1.0]
IIT mtr histogram bin edges:
     [-1.0, -0.3, -0.2, -0.1, 0.0, 0.1, 0.2, 0.3, 0.4, 0.5, 1.0]
FICA and IIT mtr histogram bin counts for e00200p:
219814 :      0  31950      0      0      0      0      0 183680   4184      0
219814 :   4578    108   1738  13632  38296  63674  47129  28832  21211    616
FICA and IIT mtr histogram bin counts for e00900p:
219814 :  14116  29235      0      0      0   1343      0 175120      0      0
219814 :   4577    108   1739  13632  38194  64120  47414  28203  21211    616
FICA and IIT mtr histogram bin counts for e00300:
219814 : 219814      0      0      0      0      0      0      0      0      0
219813 :      0      0      0      0  58388  64780  42206  30051  23767    621
WARNING: sum of bin counts is too low: max(mtr)=12614.95
FICA and IIT mtr histogram bin counts for e01700:
219814 : 219814      0      0      0      0      0      0      0      0      0
219814 :      0      0      0      0  58454  64725  46791  28019  21209    616
FICA and IIT mtr histogram bin counts for e02400:
219814 : 219814      0      0      0      0      0      0      0      0      0
219814 :      0      0      0      0  98938  40042  52535  27690    437    172
FICA and IIT mtr histogram bin counts for e23250:
219814 : 219814      0      0      0      0      0      0      0      0      0
219813 :      0      0      0  17145 125271  37963  27010   9525   2831     68
WARNING: sum of bin counts is too low: max(mtr)=12614.95
```

### DIFF
--- a/inctax.py
+++ b/inctax.py
@@ -4,6 +4,7 @@ INCome tax input-output capabilities for TAX-calculator.
 # CODING-STYLE CHECKS:
 # pep8 --ignore=E402 inctax.py
 # pylint --disable=locally-disabled inctax.py
+# (when importing numpy, add "--extension-pkg-whitelist=numpy" pylint option)
 
 import argparse
 import sys
--- a/simtax.py
+++ b/simtax.py
@@ -4,6 +4,7 @@ SIMple input-output capabilities for TAX-calculator.
 # CODING-STYLE CHECKS:
 # pep8 --ignore=E402 simtax.py
 # pylint --disable=locally-disabled simtax.py
+# (when importing numpy, add "--extension-pkg-whitelist=numpy" pylint option)
 
 import argparse
 import sys
--- a/test.py
+++ b/test.py
@@ -5,6 +5,7 @@ writes results to the 'results_puf.csv' file.
 # CODING-STYLE CHECKS:
 # pep8 --ignore=E402 test.py
 # pylint --disable=locally-disabled test.py
+# (when importing numpy, add "--extension-pkg-whitelist=numpy" pylint option)
 
 import pandas as pd
 from taxcalc import Policy, Records, Calculator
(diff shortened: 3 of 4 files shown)


Amy-Xu:
Let me summarize important issues here after talking with Martin:

- Having the script here separate from the test suite gives us the benefit of implementing the tests based on real data (PUF) and checking on real MTRs, which is difficult in the test suite. Printing results in this way also make it possible for users to make judgement calls on bins and outliers, which is also not something aimed at in the test suite. That said, this might be difficult for anyone outside the develop team who doesn't have access to PUF. We'll try to add a limited version of this test in the test suite in near future. I'll open an issue on that. 

- As mentioned above the fica and iit bins might vary depending on the input data and input variables. Quoting Martin: ''It's an art but not science." So I guess having a bit documentation on what those bins for and how to use those bins would be very helpful for people with enough experience on MTR distributions. Martin is going to add docs for this very soon.

- The max of MTRs (12614) is still too high. It seems to be an outliner, may or may not because of EITC eligibility (Martin's hunch). I'm open an issue keep investigating on that. 

cc @MattHJensen @feenberg

martinholmer:
@Amy-Xu,  Does this added documention make sense to you?

Amy-Xu:
Yep looks good to me. Merging.

PR #501, opened 2015-12-12
Add mtrinc and blowup options to inctax.py program

martinholmer:
Enhancements allow control from the command-line over the type of
income for which marginal tax rates are computed and over whether or
not the input variables are aged from the input data year to the
inctax.py TAXYEAR.  Using neither of these two new options results in
the exact same inctax.py behavior and results as before these changes.


martinholmer:
The changes in this pull request have no effect on core Tax-Calculator logic or results.

Issue #503, opened 2015-12-13
Very high marginal tax rates on long-term capital gains


martinholmer:
The ```testmtr.py``` program was recently added to the Tax-Calculator repository.  This program simply computes 2013 marginal tax rates (mtr) with respect to each income type for records in the ```puf.csv``` file and then summarizes those mtr into two histograms (one for FICA and one for income taxes) for each income type.  The top edge of the highest mtr histogram bin is 1.0 (that is, a 100 percent marginal tax rate).  **This issue identifies which ```puf.csv``` records have mtr values above 1.0 (or 100 percent).**

**Before pull request #484 was merged into the master branch,** there was only one ```puf.csv``` record (out of 219,814) that had mtr above one: recid 126452, which had a mtr with respect to interest income of 12612.69 and also had a mtr with respect to long-term capital gains of 12612.69.  Here is this record's 2013 Tax-Calculator output (in Internet-TAXSIM format) where the mtr is computed with respect to long-term capital gains (rather than taxpayer earnings):
```
> cp ../tax-calculator-data/puf.csv .
> python inctax.py --blowup --mtrinc e23250 puf.csv 2013
You loaded data for 2009.
Your data have been extrapolated to 2013.
> awk '$1==126452' puf-13.out-inctax
126452. 2013 0 433.87 0.00 0.00 1261268.73 0.00 0.00 18027.30 0.00 0.00 0.00 19500.00 0.00 0.00 0.00 0.00 0.00 0.00 0.00 0.00 1000.00 0.00 126.13 18027.30 0.00 0.00
```
Notice that output variable 25 (EITC) is $126.13 before the one cent is added to long-term capital gains.
The EITC has an unearned income eligibility limit, beyond which all EITC payments are lost.  The extra one cent of long-term capital gains apparently pushed this record over that limit causing a complete loss in EITC payments.  So, the mtr would be 126.13/0.01, which equals 12613, which is within rounding error of the 12612.6873 calculated by Tax-Calculator.  This inference is supported by the fact that this record has the exact same mtr with respect to taxable interest income, e00300.

So, in summary, the very high marginal tax rates calculated for recid 126452 are accurate representations of how the income tax works under current law.

[...]

PR #509, opened 2015-12-18
Add script fixing recid

Amy-Xu:
The RECID wasn't unique anymore after the CPS match. The script added in this PR fixed the key as Martin suggested in #503. The only difference is that I also added a ending zero for records without any duplicates.

@martinholmer Could you review the script? Thank you!

### DIFF
--- a/recid_transformation.py
+++ b/recid_transformation.py
@@ -1,0 +1,47 @@
+# The recid (key) for each record is not unique after CPS match
+# because 1) original PUF records were split in some cases
+# 2) non-filers were added
+
+# This script fixes the recid issue by
+# 1) adding a digit at the end of recid for all original PUF records
+#      - if no duplicates, add zero
+#      - otherwise, differentiate duplicates by numbering them with increment integers starting from zero
+# 2) setting recid for all non-filers (4000000 - 4005693)
+
+
+
+import pandas
+
+# import the old CPS matched PUF
+data = pandas.read_csv("cps-puf.csv")
+
+# sort all the records based on old recid
+sorted_dta = data.sort(columns='recid')
+
+# count how many duplicates each old recid has
+# and save the dup count for each record
+seq = sorted_dta.index
+length = len(sorted_dta['recid'])
+count = [0 for x in range(length)]
+for index in range(1, length):
+    num = seq[index]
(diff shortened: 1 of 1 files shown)


martinholmer:
@Amy-Xu,
This does produce exactly what we're looking for: unique integer RECID values in the ```puf.csv``` file.
Good work.

I have two important recommendations:

(1) Move this new script out of the Tax-Calculator repository and propose a pull request to add it to the taxdata repository.  My understanding is that the input file for this script ```puf-cps.csv``` is generated by the SAS programs in that repository and the point of your new script to do some post-processing of that file and write out the final version as the ```puf.csv``` file.  @MattHJensen has laid out a vision in issue #436 that would place **all** data preparation work in the taxdata repository, leaving the Tax-Calculator repo to contain only tax calculation code.  So, we want to follow this vision as we add new capabilities, otherwise there will just be more work to do in the future.

(2) Rename the script something like ```puf-cps-processing.py``` to make its name broader than RECID manipulation.  The reason for this recommendation is that we are going to be adding other post-processing logic to this script (beyond making the RECID values unique) in the coming days as part of resolving issue #425.  Pick your own name for the new script; just make it reflect what will soon be its broader purpose.

Also, I see there is some editing that needs to be done to achieve PEP8 compliance.

As far as I'm concerned, after this new script is moved to the taxdata repo and renamed, it is ready to be merged into the taxdata repository.

Dan @feenberg  and Sean @GoFroggyRun, do you have any questions or concerns?

GoFroggyRun:
@martinholmer Look good to me after the concerns you've mentioned being dealt with.

martinholmer:
@Amy-Xu,
Forgot to mention something you may have already done.  Given your excellent fix of the RECID problem, there is no need for John O'Hare to change anything in his problems pertaining to RECID values.  In fact, if he did change any of the RECID values in the ```puf-cps.csv``` file, it would break your fix.

Tell him to focus exclusively on making the X* variables consistent for the CPS records (in the same way as they are consistent in the SOI records).

cc @MattHJensen @feenberg @GoFroggyRun

Amy-Xu:
Thank you Martin! Closing.

PR #512, opened 2015-12-31
Expand extrapolation to 2026

Amy-Xu:
Added two more years (25-26) of weights and stage I factors in this PR. The stage I factors and stage II targets are simply derived from the last year value and growth rates in current budget window. A few other files in the code base are also modified to accommodate these changes.

DT looks like this:
screen shot 2015-12-30 at 3 03 46 pm

@MattHJensen @martinholmer @talumbau Could you review? Thanks!


martinholmer:
@Amy-Xu,
These changes look good except you need to do one more thing: replace all the "magic" numbers (now 14) with ```Policy.DEFAULT_NUM_YEARS```.  Doing this will eliminate much error-prone editing in the future when we add extra years.  There are also numbers like 13 and 11 that should be replaced by ```Policy.DEFAULT_NUM_YEARS - 1``` and ```Policy.DEFAULT_NUM_YEARS - 3```, respectively.

If you are not familiar with why such hard-coded "magic" numbers are undesirable, I suggest you read this short [Wikipedia entry](https://en.wikipedia.org/wiki/Magic_number_(programming)#Unnamed_numerical_constants).

I'll merge this pull request after you take this extra step.

Amy-Xu:
@martinholmer Thanks! Totally agree. Just made the changes to remove all those hard coded years.

martinholmer:
Look fine to me; merging into master branch and closing pull request.

PR #513, opened 2016-01-05
Remove logic & comments that change value of XTOT in functions.py

martinholmer:
The comment that constructs XTOT from its component parts is unnecessary because XTOT in the forthcoming puf.csv file will contain XTOT values that equal the sum of its component parts for all filing units (even the CPS records).

The code that changes the value of XTOT is confusing because the value of XTOT is not returned from the function where it is revised.  This is probably because XTOT is not used again in the functions.py code, but it is still confusing and potentially dangerous if the code were ever to use XTOT in another subsequent calculation.

These changes are in the spirit of issue #436 which views functions.py as containing only tax-calculation code and containing no data imputation or extrapolation logic.

### DIFF
--- a/taxcalc/functions.py
+++ b/taxcalc/functions.py
@@ -178,11 +178,10 @@ def SSBenefits(SSIND, MARS, e02500, _ymod, e02400, SS_thd50, SS_thd85,
 
 
 @iterate_jit(nopython=True, puf=True)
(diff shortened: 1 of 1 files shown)


Amy-Xu:
Diagnostic table remains exactly the same. 
screen shot 2016-01-05 at 1 50 12 pm

LGTM. Merging. Thank you Martin!

PR #514, opened 2016-01-05
Remove unused SOIYR variable from the Tax-Calculator

martinholmer:
SOIYR was set in records.py but never used anywhere in the Tax-Calculator code.


Amy-Xu:
LGTM. Merging. Thanks Martin!

PR #516, opened 2016-01-05
Add code coverage report generation

talumbau:


talumbau:
This PR resolves https://github.com/open-source-economics/Tax-Calculator/issues/346 by adding code coverage via the `coverage` tool and integrating with Travis CI via codecov.io. The statement above that this PR would drop coverage by 10% is incorrect because I had misconfigured the initial set up of the tool. By not including code in taxcalc/tests/, we have a more accurate assessment of how our tests cover the code in taxcalc/*py, which is what we are interested in testing. This tool is a reporting tool and does not act as any kind of control on merges of PRs. Although, we might have some kind of policy that PRs should not decrease code coverage (so people have incentive to write tests when they add new code). 

cc @MattHJensen @martinholmer @Amy-Xu @GoFroggyRun

talumbau:
When merged, this PR would make it so that the current test coverage level is displayed in the README, and subsequent PRs will report test coverage level upon a successful build on Travis CI.

talumbau:
This [link](https://codecov.io/github/open-source-economics/Tax-Calculator/taxcalc/growth.py?ref=f7b6ebb0588574a2135bf30d63919d22ff57da7e) shows a nice example of what the tool can do. We can see that in `growth.py`, we need a test for the case of giving bad input to the `__init__` of the `Growth` class:

screen shot 2016-01-05 at 3 15 47 pm

Amy-Xu:
:+1: WOW this is super helpful, (especially for people like me being super lazy writing tests :P). Thanks @talumbau. Will use this as a guide to complete the tests needed.

martinholmer:
@talumbau,  Looks like a very helpful addition.  Thanks.

talumbau:
Merging this. One thing to mention is that the coverage results aren't really as bad as them seem. The code in `functions.py` is essentially not considered, but the reality is that `@iterate_jit` does so much to the written code before it is actually executed (including `numba`-izing it) that there is little hope that a coverage tool could automatically track the execution of those functions. So, it might make sense to exclude that file in the `[omit]` section of the `.coveragerc`.

Issue #517, opened 2016-01-05
What is role of the FLPDYR variable read from puf.csv file?


martinholmer:
The values for the FLPDYR variable in the puf.csv file are as follows: 
```
> awk -F, 'NR==1{for(i=1;i<=NF;i++)if($i~/FLPDYR/)print i,$i}' puf.csv
15 FLPDYR
> awk -F, 'NR>1{t++;n[$15]++}END{print t;for(y in n)print y,n[y]}' puf.csv
219814
2009 212640
2008 5205
2007 1408
2006 561
```
So, about 3.2 percent of the filing units in puf.csv have tax filing years before 2009.

I have two questions about the role of the FLPDYR variable in the Tax-Calculator code given that it seems as if the only uses of the FLPDYR variable are: (a) to compute the _agep and _ages variables when the date of birth variables for the taxpayer (DOBYR) and spouse (SDOBYR) are available, which is not the case with puf.csv input; and (b) to specify Records.current_year when using puf.csv input.

(1)  There is about a three percent chance that (b) will not work because there is no guarantee that the first filing unit in the puf.csv file has a value of 2009.  It turns out that in the current version of the puf.csv file the first record has FLPDYR equal to 2009, but if the records were sorted in a different way that might not be true.  Perhaps we should add a test to test_records.py to ensure that the first record in puf.csv always has the value Records.PUF_YEAR.

(2) In the case where the DOBYR and SDOBYR variables are available, I'm confused about why the Records.increment_year() method increments FLPDYR (```self.FLPDYR += 1```).  It seems as if that would create an unrealistic age distribution that would become more unrealistic with each call to increment_year().  After 10 or more years, there would be no filing units in their twenties and too many elderly filing units that have unrealistic incomes (that is, no social security because they were in their fifties in the puf.csv file).  Why no remove the ```self.FLPDYR += 1``` statement in increment_years()?

martinholmer:
@MattHJensen, @Amy-Xu, @GoFroggyRun,
Have I missed something in my FLPDYR analysis in issue #517?

If not, I'd be happy to prepare a pull request that implements the test I suggest and drop the incrementing of FLPDYR in the Records increment_year() method.

feenberg:
On Tue, 5 Jan 2016, Martin Holmer wrote:

> 
> The values for the FLPDYR variable in the puf.csv file are as follows:
> 
[...]

PR #518, opened 2016-01-06
Revise FLPDYR logic along lines suggested by Dan Feenberg

martinholmer:
When not specifying a start_year in the Records constructor, the current_year and FLPDYR variable are now set to PUF_YEAR, so that ages will be correctly calculated.  Also, remove the increase of FLPDYR by one in the increment_year() method in order to calculate ages correctly.  These changes imply that FLPDYR no longer needs to be included in the MUST_READ_VARS list.

This pull request resolves issue #517 in a way that assumes the Tax-Calculator will always be using the most recently available IRS/SOI PUF data when two of the arguments of the Records class constructor are at their default values: that is, when data='puf.csv' and start_year=None.

As before these changes, data other than the most recent IRS/SOI PUF data can be read by the Records class as long as the name of the file is supplied as the data argument of the Records constructor and the year of the data is specified in the start_year argument of the Records constructor.

cc @MattHJensen @feenberg @Amy-Xu @GoFroggyRun

### DIFF
--- a/taxcalc/records.py
+++ b/taxcalc/records.py
@@ -26,7 +26,7 @@ class Records(object):
         default value is the string 'puf.csv'
         For details on how to use your own data with the Tax-Calculator,
         look at the test_Calculator_using_nonstd_input() function in the
-        tests/test_calculator.py file.
+        tests/test_calculate.py file.
 
     blowup_factors: string or Pandas DataFrame
         string describes CSV file in which blowup factors reside;
@@ -47,7 +47,7 @@ class Records(object):
               custom data's calendar year.  For details on how to
               use your own data with the Tax-Calculator, look at the
               test_Calculator_using_nonstd_input() function in the
-              tests/test_calculator.py file.
+              tests/test_calculate.py file.
 
     consider_imputations: boolean
         True implies that if current_year (see start_year above) equals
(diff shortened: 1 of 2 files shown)


Amy-Xu:
The code change looks good to me. I was just wondering whether we should remind users that if they do have a file with different FLPDYR they want to use as _current_year, they have to do that manually for Record object. Do we need to add one line of comment somewhere?

martinholmer:
@Amy-Xu said:
> The code change looks good to me.  I was just wondering whether we should remind users that if they do have a file with different FLPDYR they want to use as _current_year, they have to do that manually for Record object.  Do we need to add one line of comment somewhere?

Probably a good idea.  It seems appropriate to put that comment in the doc-string for the Record class constructor, which is where users would specify a non-PUF filename and a non-PUF start_year.  Does that seem sensible?  If so, I'll do that soon.  Thanks for the suggestion, Amy.

martinholmer:
@Amy-Xu,  I've added comments to Records class constructor documentation as well as point users who want to use their own data to the test_Calculator_using_nonstd_input() function in test_calculate.py file.

Amy-Xu:
Looks good to me. Thanks Martin!! Merging.

PR #522, opened 2016-01-08
[WIP] Updates on capital gain for Sch D

GoFroggyRun:
This PR deals with capital gain worksheet for Schedule D. The changes are made according to the latest worksheet on the instruction for Sch D, please see the last two pages [here](https://www.irs.gov/pub/irs-pdf/i1040sd.pdf). I have asked @Amy-Xu to help me verify these changes.

The diagnostic table for this PR now looks like: 
screen shot 2016-01-08 at 10 55 11 am
@Amy-Xu, could you take a look?

The issue here is that, although the `py.test` passes, the tax logic is now inconsistent against Internet Taxsim, please refer to the code below:
```
ITs-MacBook-Air:validation Sean.Wang$ ./tests
STARTING WITH VALIDATION TESTS : Fri Jan  8 10:58:21 EST 2016
M	taxcalc/validation/c14.taxdiffs
M	taxcalc/validation/c13.taxdiffs
FINISHED WITH VALIDATION TESTS : Fri Jan  8 11:00:42 EST 2016
VALIDATION TESTS RESULTS: any lines (starting with M) between the
                          STARTING... and FINISHED... lines above
                          indicate a test failure caused by the actual
                          *.taxdiffs results being different from the
                          expected *.taxdiffs results;
                          no M lines implies that all tests passed.
ITs-MacBook-Air:validation Sean.Wang$ git status
On branch Sch_D
```
And the diff files for c13 and c14 respectively:
```
[13] TAXDIFF:ovar,#diffs,#1cdiffs,maxdiff[id]=  4  53660    700   2250.00 [1961]
                       #big_inctax_diffs=            52960
[14] TAXDIFF:ovar,#diffs,#1cdiffs,maxdiff[id]=  4  54029    714   2250.00 [4296]
                       #big_inctax_diffs=            53315
```
@MattHJensen @martinholmer Any comments, concerns or remarks would be appreciated.

### DIFF
--- a/taxcalc/functions.py
+++ b/taxcalc/functions.py
@@ -644,14 +644,16 @@ def TaxGains(e00650, c01000, c04800, e01000, c23650, p23250, e01100, e58990,
         # line 27 on the latest worksheet
         _LINE27 = max(0., (min(_taxinc, CG_thd2[MARS - 1]) - c24540 - c24534))
         c24598 = CG_rt2 * min(c24597, _LINE27)  # actual 15% tax
+        _LINE30 = min(0, c24534) + min(c24597, _LINE27)
+        _LINE31 = _dwks21 - _LINE30
         # income subject to 20% tax, named _addtax
-        _addtax = CG_rt3 * (_dwks21 - min(0, c24534) - min(c24597, _LINE27))
+        _addtax = CG_rt3 * _LINE31
         _dwks25 = min(_dwks9, e24515)
         _dwks26 = c24516 + c24540
         _dwks28 = max(0., _dwks26 - _taxinc)
         c24610 = max(0., _dwks25 - _dwks28)
         c24615 = 0.25 * c24610
-        _dwks31 = c24540 + c24534 + _dwks21 - min(0, c24534) + c24610
+        _dwks31 = c24540 + c24534 + min(c24597, _LINE27) + _LINE31 + c24610
         c24550 = max(0., _taxinc - _dwks31)
         c24570 = 0.28 * c24550
 
--- a/taxcalc/functions.py
+++ b/taxcalc/functions.py
@@ -645,7 +645,7 @@ def TaxGains(e00650, c01000, c04800, e01000, c23650, p23250, e01100, e58990,
         _LINE27 = max(0., (min(_taxinc, CG_thd2[MARS - 1]) - c24540 - c24534))
         c24598 = CG_rt2 * min(c24597, _LINE27)  # actual 15% tax
         _LINE30 = min(0, c24534) + min(c24597, _LINE27)
-        _LINE31 = _dwks21 - _LINE30
+        _LINE31 = min(0, _dwks21 - _LINE30)
         # income subject to 20% tax, named _addtax
         _addtax = CG_rt3 * _LINE31
         _dwks25 = min(_dwks9, e24515)
@@ -655,7 +655,7 @@ def TaxGains(e00650, c01000, c04800, e01000, c23650, p23250, e01100, e58990,
         c24615 = 0.25 * c24610
         _dwks31 = c24540 + c24534 + min(c24597, _LINE27) + _LINE31 + c24610
         c24550 = max(0., _taxinc - _dwks31)
-        c24570 = 0.28 * c24550
+        c24570 = 0.2 * c24550
 
         # if c24540 > CG_thd2[MARS - 1]:
         #    _addtax = (CG_rt3 - CG_rt2) * c24517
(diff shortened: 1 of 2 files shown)


martinholmer:
Sean,  I have several questions about this pull request.  I'll pose the questions one at a time, rather than asking them all at the same time.

First question.  What is the motivation for making these revisions to the TaxGains function?
I haven't seen a issue discussion about the need to do this, and the conversation with this pull request is not clear about the motivation.  Can you provide us with more background?

cc @MattHJensen @feenberg @GoFroggyRun @Amy-Xu

GoFroggyRun:
@martinholmer Sure. 
So previously, we were testing our tax-logic against some reform options that CBO has also done. Please see here [page 111](https://www.cbo.gov/sites/default/files/cbofiles/attachments/44715-OptionsForReducingDeficit-3.pdf) for details regarding the capital gain reform we were testing, which raises all capital gain and dividends rates by 2%. 

The resulting difference of individual income tax on current master was somewhat different from what CBO has: 
screen shot 2016-01-10 at 11 01 50 am,
and we'd like to see why we are having slightly large difference.

The only two parts that involve CG and dividends rates would be form Sch D and form 6251. Given I have recently verified AMT part against forms, only the Sch D remains to be checked. 
After checked the tax logic against the latest instruction for Sch D, [last two pages here](https://www.irs.gov/pub/irs-pdf/i1040sd.pdf), it seems like our current code wasn't consistent against the current forms, at least from my perspective. And thus the changes in this PR were proposed.

This PR, however, fails the consistency test against Internet Taxsim. I was trying to see which particular part causes the inconsistency, but wasn't able to do so since this part basically got re-written and it's hard to separate it.

Attached is the new comparison table for reform: 
screen shot 2016-01-08 at 2 01 19 pm.
Please note that there are a couple other reasons that may also have resulted in such difference, and we're working on verifying them as well.

GoFroggyRun:
So it turns out that, based on filing some simple unit, Internet Taxsim yields the correct result. 
Basically what our current master is doing right now , and what the SAS calculator is doing, is that we are using the Worksheet on Sch D to override the Qualified Dividends and Capital Gain Tax Worksheet for Line 44 on Form 1040, and taking desired intermediate results in calculating the taxes. 
My guess is that, the two worksheets share the same tax logic where worksheet on Sch D that we are currently using is more general and thus we are ignoring the other worksheet. 
I'm not sure if it's worthwhile to recover the complete algebra on the tax forms, since, even if it makes a difference, the difference would be minor given that not many people are eligible for **28% Rate Gain** and **Unrecaptured 1250 Gain**. This issue will be investigated later.

feenberg:
On Tue, 12 Jan 2016, Sean.Wang wrote:

> 
> So it turns out that, based on filing some simple unit, Internet Taxsim yields the
> correct result.
> Basically what our current master is doing right now , and what the SAS calculator is
> doing, is that we are using the Worksheet on Sch D to override the Qualified Dividends
> and Capital Gain Tax Worksheet for Line 44 on Form 1040, and taking desired
> intermediate results in calculating the taxes.
> My guess is that, the two worksheets share the same tax logic where worksheet on Sch D
> that we are currently using is more general and thus we are ignoring the other
> worksheet.
> I'm not sure if it's worthwhile to recover the complete algebra on the tax forms,
> since, even if it makes a difference, the difference would be minor given that not
> many people are eligible for 28% Rate Gain and Unrecaptured 1250 Gain. This issue will
> be investigated later.

I think it is perfectly reasonable to ignore minor issues if they are 
difficult to correct. The main advantage of correcting them is to avoid
having them get in the way when testing the calculator. If they are not
corrected, keeping a list of "won't fix" errors is probably a good idea,
perhaps even make it a function that could be called to exclude certain
records when testing.

dan

GoFroggyRun:
@feenberg Thanks for your suggestion. I definitely agree that we could skip the parts that are considered to be minor. 
To keep track of the "won't fix" "errors" we had or might have, what I have in mind is that we could possibly make some accessible list, provided with compelling explanations correspondingly on why our assumptions are valid or reasonable. What do you think? 

cc @martinholmer @MattHJensen  @Amy-Xu

feenberg:
On Tue, 12 Jan 2016, Sean.Wang wrote:

> 
> @feenberg Thanks for your suggestion. I definitely agree that we could skip the parts
> that are considered to be minor.
> To keep track of the "won't fix" "errors" we had or might have, what I have in mind is
> that we could possibly make some accessible list, provided with compelling
> explanations correspondingly on why our assumptions are valid or reasonable. What do
> you think?

yes, that is the way to go. The explanations will be repetitive, so plan 
on using abbreviations with a dictionary elsewhere.

dan

Issue #507, opened 2015-12-16
Calculator to provide extrapolated "inputs"


GoFroggyRun:
It seems that, if we were turning on the "stage one" and "stage two" extrapolations. Then the actual input file might not be useful, since what is being used that runs through the tax logic is the extrapolated input. And those values being used in calculation were hard to track.

Given my experience in dealing with #506, I think it would be helpful for our calculator to be able to produce a complete list of extrapolated values for each inputed variable, possibly as an alternative option. 

PS: I was able to get the extrapolated "inputs" manually, with the help of ipython-notebook, where the manual part involves modifying the uppercase and lowercase back and forth. Thus I think resolving the uppercase and lowercase struggles when recording certain inputs, as mentioned in #425, would be a first step.

martinholmer:
Sean Wang ( @GoFroggyRun ) said:
> Given my experience in dealing with #506, I think it would be helpful for our calculator to be able to produce a complete list of extrapolated values for each inputed variable, possibly as an alternative option.
>
> PS: I was able to get the extrapolated "inputs" manually, with the help of ipython-notebook, where the manual part involves modifying the uppercase and lowercase back and forth. Thus I think resolving the uppercase and lowercase struggles when recording certain inputs, as mentioned in #425, would be a first step.

Two thoughts about this issue:

(1) I agree completely that the uppercase/lowercase differences in variable names between the input file ```puf.csv``` and the Python code are confusing and serve no purpose.  Eliminating these differences is part of the plan in resolving issue #425.

(2) I wonder if the debugging option you requested for the simtax.py and inctax.py programs would meet your needs for blown-up values of various input variables.  All you need to do is add the debugging variable names you want to add the output from simtax.py or inctax.py in the following SimpileTaxIO code:
```
    OVAR_NUM = 28
    DVAR_NAMES = [  # OPTIONAL DEBUGGING OUTPUT VARIABLE NAMES
        # '...',  # ovar[OVAR_NUM+1]
        # '...',  # ovar[OVAR_NUM+2]
        # etc.
        # '...'   # ovar[OVAR_NUM+n]
    ]
```
which then automatically activates the following logic:
```
        # add optional debugging output to ovar dictionary
        num = SimpleTaxIO.OVAR_NUM
        for dvar_name in SimpleTaxIO.DVAR_NAMES:
            num += 1
            dvar = getattr(crecs, dvar_name, None)
            if dvar is None:
[...]

PR #524, opened 2016-01-08
Cosmetic revisions to functions.py code

martinholmer:
These functions.py code revisions cause no changes in py.tests or validation/tests results, and no changes in test.py, inctax.py, and test_mtr.py results when using the proprietary puf.csv input file.  In other words, there is no substantive change in the operation of the Tax-Calculator.

Why were these changes made?  For two main reasons:

First, to simplify somewhat the code in the functions.py file.  For example, there were scores of function arguments that were never used in the function they were passed to.  This made the already complex code even more difficult to understand.

Second, and more importantly, the code changes imply that there are now no pylint warning messages other than those indicating the four TODO comments in the code (which presumably indicate further work that needs to be done).  While the use of the pylint tool is optional in this project, it can now be used effectively to detect questionable coding style in future development work.  Before these changes, there was no practical way to use pylint to detect questionable coding in pull requests because the pylint warnings on added code would be buried among the hundreds of warnings associated with existing code.  

Before these changes, pylint gave the functions.py code an overall rating of 3.25 out of 10.
After these changes, pylint gives the functions.py code an overall rating of 9.94 out of 10.

Another reason for making these changes is that they serve as an introduction to some of the kinds of Python coding that pylint considers questionable.   Matt Jensen had asked for this kind of information at the end of the discussion of issue #30.  In order to provide some of that kind of information and also to provide rationales for the particular functions.py changes included in this pull request, I have included below the working notes I made during the development of this pull request.

cc @MattHJensen @feenberg @Amy-Xu @GoFroggyRun @talumbau 

```
<08-Jan-2016>

(*) Begin work on cleanup-functions branch using current version of
functions.py, which was last revised on 05-Jan-2016 as seen by this:
~/work/OSPC/tax-calculator> git log -- taxcalc/functions.py
The objective of this work is to raise the pylint score toward ten so
that new pylint warnings can inform of problems with code revisions.

(0) Here are pylint results with the current version of functions.py:

Compilation started at Fri Jan  8 09:13:46

pylint --disable=locally-disabled --extension-pkg-whitelist=numpy functions.py 
************* Module taxcalc.functions

<complete list of warnings are in ./functions-0-pylint>

Report
======
676 statements analysed.

...

Messages
--------

+---------------------------+------------+
|message id                 |occurrences |
+===========================+============+
|invalid-name               |240         |
+---------------------------+------------+
|unused-argument            |75          |
+---------------------------+------------+
|too-many-arguments         |27          |
+---------------------------+------------+
|missing-docstring          |26          |
+---------------------------+------------+
|bad-continuation           |20          |
+---------------------------+------------+
|too-many-locals            |18          |
+---------------------------+------------+
|unused-wildcard-import     |17          |
+---------------------------+------------+
|superfluous-parens         |11          |
+---------------------------+------------+
|protected-access           |6           |
+---------------------------+------------+
|fixme                      |4           |
+---------------------------+------------+
|unused-variable            |3           |
+---------------------------+------------+
|too-many-branches          |3           |
+---------------------------+------------+
|too-many-statements        |2           |
+---------------------------+------------+
|wildcard-import            |1           |
+---------------------------+------------+
|unused-import              |1           |
+---------------------------+------------+
|too-many-lines             |1           |
+---------------------------+------------+
|pointless-string-statement |1           |
+---------------------------+------------+


Global evaluation
-----------------
Your code has been rated at 3.25/10

(1) The most common warning is invalid-name, which is caused by the
use of uppercase characters: a practice that is not considered
Pythonic, but which is deeply embedded in the Tax-Calculator code.
Solution: file-wide disabling of this pylint warning.
Results: Your code has been rated at 6.80/10 (previous run: 3.25/10, +3.55)

Messages
--------

+---------------------------+------------+
|message id                 |occurrences |
+===========================+============+
|unused-argument            |75          |
+---------------------------+------------+
|too-many-arguments         |27          |
+---------------------------+------------+
|missing-docstring          |26          |
+---------------------------+------------+
|bad-continuation           |20          |
+---------------------------+------------+
|too-many-locals            |18          |
+---------------------------+------------+
|unused-wildcard-import     |17          |
+---------------------------+------------+
|superfluous-parens         |11          |
+---------------------------+------------+
|protected-access           |6           |
+---------------------------+------------+
|fixme                      |4           |
+---------------------------+------------+
|unused-variable            |3           |
+---------------------------+------------+
|too-many-branches          |3           |
+---------------------------+------------+
|too-many-statements        |2           |
+---------------------------+------------+
|wildcard-import            |1           |
+---------------------------+------------+
|unused-import              |1           |
+---------------------------+------------+
|too-many-lines             |1           |
+---------------------------+------------+
|pointless-string-statement |1           |
+---------------------------+------------+

(2) Next fix the 26 missing doc-string warnings by adding simple
doc-strings that can be expanded with more information.
Results: Your code has been rated at 7.20/10
Commit changes so far; alltests OK

(3) The current version generates these import-related warnings:
+-----------------------+------------+
|unused-wildcard-import |17          |
+-----------------------+------------+
|wildcard-import        |1           |
+-----------------------+------------+
|unused-import          |1           |
+-----------------------+------------+
After revising import statements, the above warnings are eliminated.
Results: Your code has been rated at 7.48/10
Commit changes so far; alltests OK

(4) Almost all the functions have more arguments than the pylint
default maximum recommended number of function arguments.
+--------------------+------------+
|too-many-arguments  |27          |
+--------------------+------------+
But this is an essential part of the decorator magic that makes
Tax-Calculator execute much faster.  Solution is to disable this
warning in the whole file with the following comment at the top:
# pylint: disable=too-many-arguments
Results: Your code has been rated at 7.88/10

(5) There are many stylistic pylint warnings including:
+--------------------+------------+
|bad-continuation    |20          |
+--------------------+------------+
These warn about the indentation of second or subsequent lines of a
single statement.  Solution is to fix the indentation to comply with
pylint coding style.
Results: Your code has been rated at 8.18/10
Commit changes so far; alltests OK

(6) The functions.py file contains many superfluous parentheses, which
generate the following pylint warnings:
+---------------------------+------------+
|superfluous-parens         |11          |
+---------------------------+------------+
Solution is to remove the unneeded parentheses.
Results: Your code has been rated at 8.34/10
Commit changes so far; alltests OK

(7) The functions.py file contains (relative to the default pylint
maximums) too many lines in total, too many statements in two functions,
and too many branches in three functions.  Hence the following count of
pylint warnings:
+--------------------+--------+
|too-many-lines      |1       | 1756/1000 in functions.py file
+--------------------+--------+
|too-many-statements |2       | TaxGain: 128/50 ; AMTI: 130/50
+--------------------+--------+
|too-many-branches   |3       | TaxGain: 17/12 ; AMTI: 41/12 ; NumDep: 14/12
+--------------------+--------+
These pylint warnings are suggesting refactoring to make the code more
comprehensible.  That would be the ideal thing to do, but meanwhile we
can simply locally disable these warnings for the offending functions.
This means that if future additions to other functions make them too
complex, the pylint warnings will indicate the desirability of refactoring.
Results: Your code has been rated at 8.43/10

(8) The functions.py file contains (relative to the default pylint
maximum of 15) too many local variables in almost all the functions.
Hence the following count of the pylint too-many-locals warning:
+-----------------+------------+
|too-many-locals  |18          |
+-----------------+------------+
This warning is also suggesting code refactoring.  Since most of the
functions have far more than 15 local variables, this pylint warnings
is suppressed by adding the following comment at the top of the
functions.py file.
Results: Your code has been rated at 8.69/10

(9) The BenefitSurtax function contains several statements that either
read or change the value of the private Records class members: _iitax
and _surtax.  This generates the following pylint warning count:
+-----------------+------------+
|protected-access |6           |
+-----------------+------------+
It is not immediately clear why this function needs to violate the
private nature of these Records class members when all the other
functions avoid this kind of violation.
Solution: disable=protected-access warings in just BenefitSurtax.
Results: Your code has been rated at 8.78/10

(10) The functions.py file contains several cases where local
variables are defined in a function, but then never used.  These cases
occur in AMTI, NumDep, and NonEdCr.  Hence, this warning summary:
+----------------+------------+
|unused-variable |3           |
+----------------+------------+
Solution: comment out the unused local variables with the preface
# UNUSED-LOCAL-VAR:
Results: Your code has been rated at 8.80/10
Commit changes so far; alltests OK

(11) After the above changes to functions.py code, we have this
summary of pylint warning messages:
+----------------+------------+
|message id      |occurrences |
+================+============+
|unused-argument |76          |
+----------------+------------+
|fixme           |4           |
+----------------+------------+
* See the discussion of the unused-argument warning messages below in
  item (12).
* The fixme warnings are reminding us that there are four TODO comments,
  which should not be suppressed or disabled because they are in the
  code to remind us that somebody felt there was still work to be done
  in that area of the code.  The way to eliminate these fixme warnings
  is to do whatever work needs to be done and then remove the TODO
  comment.

(12) The unused-argument warnings messages are indicating that pylint
has found 76 cases in which a parameter passed into a function is
never used in that function.
Solution: these unused function arguments need to be eliminated in
order to simplify the code.  As this was done, pylint found more
unused-variable and unused-argument cases, which were also fixed
(with unused variables being commented out and unused argument being
removed from the function definition).
Results: Your code has been rated at 9.94/10
+----------------+------------+
|message id      |occurrences |
+================+============+
|fixme           |4           |
+----------------+------------+
Commit changes so far; alltests OK
```


martinholmer:
Oh, I see there is another minor reason for merging the changes in pull request #524:
the overall code coverage percent would rise by 0.22 percentage points to 63.29%.

cc @MattHJensen @feenberg @Amy-Xu @GoFroggyRun @talumbau

martinholmer:
Using the new ```testagg.py``` program that writes ten-year aggregate diagnostic table results based on ```puf.csv``` input to the ```testagg-results.txt``` file (as described in pull request #525), it can shown that the functions.py changes in this pull request produce exactly the same ten-year diagnostic table as does the master branch.  This is further confirmation that the cosmetic changes in this pull request have no effect on any results produced by the Tax-Calculator.

martinholmer:
Add ```test_functions.py``` file that contains a single test which checks that each function argument is used at least once in the body of its function before the function's return statement.  This test was added because pylint does not warn about the situation in which a function argument is used only in the function's return statement.

Remove unused function arguments so that the new test in ```test_functions.py``` does not fail.

taxcalc/tests/test_functions.py
```diff
@@ -0,0 +1,40 @@
...
+import sys
+import re
+CUR_PATH = os.path.abspath(os.path.dirname(__file__))
+
+
+def test_function_args_usage():
+    """
+    Checks each function argument in functions.py for use in its function body.
+    """
+    funcfilename = '{}/../functions.py'.format(CUR_PATH)
```
talumbau: something like:
```
os.path.join(CUR_PATH, '..', 'functions.py')
```
is probably best and is cross-platform (handles `\` for path separator for Windows platform

taxcalc/tests/test_functions.py
```diff
@@ -0,0 +1,40 @@
...
+        fcontent = funcfile.read()
+    fcontent = re.sub('#.*', '', fcontent)  # remove all '#...' comments
+    fcontent = re.sub('\n', ' ', fcontent)  # replace EOL character with space
+    funcs = fcontent.split('def ')  # list of function text
+    for func in funcs[1:]:  # skip first item in list, which is imports, etc.
+        fcode = func.split('return ')[0]  # fcode is between def and return
+        match = re.search(r'^(.+?)\((.*?)\):(.*)$', fcode)
+        if match is None:
+            sys.stdout.write('==========\n{}\n==========\n'.format(fcode))
+            assert 'match' == 'None'
```
talumbau: You could just assert `False` here or raise an exception with an error message indicating what likely occurred to get to this state.

taxcalc/tests/test_functions.py
```diff
@@ -0,0 +1,40 @@
...
+import re
+CUR_PATH = os.path.abspath(os.path.dirname(__file__))
+
+
+def test_function_args_usage():
+    """
+    Checks each function argument in functions.py for use in its function body.
+    """
+    funcfilename = '{}/../functions.py'.format(CUR_PATH)
+    with open(funcfilename, 'r') as funcfile:
```
talumbau: To go down this path, I would suggest making use of the `inspect` module, documented [here](https://docs.python.org/2/library/inspect.html) - in particular, `inspect.getsource`. I do have a concern about this being a somewhat fragile test, not because I have a concrete counter example, but rather just that having a test pass based on matching a regular expression to the source code of a function makes me somewhat nervous. I suspect someone could do something very innocuous, end up with a failed test, and have a very difficult to understand error message.

talumbau:
Looks like good changes to the `functions.py` file but I made comments on `test_functions.py` because I suspect there is a better approach.

taxcalc/tests/test_functions.py
martinholmer: @talumbau said:
> something like ```os.path.join(CUR_PATH, '..', 'functions.py')``` is probably best and is cross-platform (handles \ for path separator for Windows platform).

OK, I followed your suggestion, but you should note that most of the other test_*.py files violate your recommendation.   For example, in test_calculate.py, we have the following code:
```
TAX_DTA_PATH = os.path.join(CUR_PATH, '../../tax_all1991_puf.gz')
```
Are you saying that test_calculate.py will fail (because of the forward slashes) when run under Windows?  Is anybody on the Tax-Calculator development team working under Windows?

cc @MattHJensen @Amy-Xu @GoFroggyRun

talumbau: Windows is more resilient to this than it has been in the past, and it's definitely true the repo could use a once over to catch these kinds of things. Another choice would be something like:

```
os.path.sep.join([CUR_PATH, '..', '..', 'functions.py'])
```

just a general style note about creating paths.

taxcalc/tests/test_functions.py
martinholmer: @talumbau said:
> You could just assert False here or raise an exception with an error message indicating what likely occurred to get to this state.

Good suggestion.  I've tried to make the error message more informative.

taxcalc/tests/test_functions.py
martinholmer: @talumbau said:
> To go down this path, I would suggest making use of the inspect module, documented here - in particular, inspect.getsource().

Thanks for the suggestion, but after looking at the inspect module, it wasn't obvious how its use would simplify matters or make the new test more robust.  In particular, getsource() requires the name of the function whose source code you want.  So, it would seem, the functions.py text would have to be parsed for a list function names.  In addition, because I don't want the full code of each function (but just the code before the return statement), I would still have to manipulate the results from inspect.getsource() in order to remove the return statement.

It may turn out that your concerns are well founded, and I'll have to do further work on this new test.  But why don't we take a wait-and-see approach?

MattHJensen:
@GoFroggyRun, could you look this over and provide feedback? If everything looks good, @martinholmer or I could merge. Do you think all of the pylint-inspired changes are improvements?

taxcalc/functions.py
```diff
@@ -501,46 +516,31 @@ def StdDed(DSI, _earned, STD, p04470, e00100, e60000,
...
-
-    # c04500 = c00100 - max(c04470, max(c04100, _standard + e37717))
+           II_brk4, II_brk5, II_brk6, c02700):
+    """
+    TaxInc function: ...
+    """
     c04500 = max(0, c00100 - max(c21060 - c21040, _standard + e37717))
 
-    # Check with Dan whether this is right!
-    # if c04470 > _standard:
```
GoFroggyRun: @MattHJensen The rest of the changes look good to me except this part. @martinholmer Can we still keep part of the comments here? Although we are doing the optimization of itemizing in `calculate.py`, it would still be helpful to keep use this when performing the SAS-validation. Specifically, can we keep:
```
# if c04470 > _standard:
#    _standard = 0.
# if c04470 <= _standard:
#    c04470 = 0.
```

martinholmer: Sean said:
> @MattHJensen  The rest of the changes look good to me except this part.
> @martinholmer  Can we still keep part of the comments here? Although we are doing the optimization of itemizing in calculate.py, it would still be helpful to keep use this when performing the SAS-validation. Specifically, can we keep [the following commented out code]:
```
# if c04470 > _standard:
#    _standard = 0.
# if c04470 <= _standard:
#    c04470 = 0.
```

I don't understand what you mean when you say "performing the SAS-validation."  To me validation, means that, when using the same input, two models produce the same output.  If you uncomment these four lines, then you don't have the same logic as the Tax-Calculator, you have some third model.  (If this is not true, please explain why you have to uncomment these lines to do a "validation".)   How can comparing output from a third model with that from the SAS program be considered a SAS-vs-Tax-Calculator validation?  I guess my confusion about this is rooted in the fact that the SAS validation work is all being done in secret, which is contrary to open-source principles.  Nobody, not even those on the core development team, knows how the SAS validation work is being done.  We need much more explanation and transparency in order to understand your request.

cc @MattHJensen @feenberg @GoFroggyRun @Amy-Xu

GoFroggyRun: @martinholmer Sorry that the validation made you confused. 

First of all, most core team members do understand what's going on in the validation work, as I was tutored by them initially: the validation work has been performed prior to the time when I join the team. Also didn't mean to keep it secret or nontransparent, but there's nothing much to report if the empirical results were matched completely. 

When I mentioned SAS validation, it means that I'm comparing the **tax logic** of two calculators, and purely tax logic. I consider the extrapolation and imputation of Tax-Calculator, and the optimization of itemizing tax-logic-non-related. If we turn on the extrapolation and then comparing the two calculators, the work is meaningless, as we are feeding the two calculators (tax logic to be specific) with different inputs. To make sure we are really comparing tax logic of two calculators, there are a couple preliminaries to be done before the validation begins:

1. Extrapolation and imputation need to be switched off in Tax-calculator.
2. Optimization of itemization needs to be modified, as the code commented out here. 

Just like you mentioned, we need to feed the calculator with the **same** input, and if they yield the same output, the work is considered to be done. Without these two steps, the two calculator substantially won't have the same input. If we were not feeding the calculators with the same input, there's no way to perform the validation:

The first item listed above should be pretty straightforward that, after extrapolation and imputation, the real input for python calculator will be different from current `puf`, see #507 for details.

The second item might be a little twisted: optimizing in different ways will make some intermediate "input" to be different for two calculators. SAS calculator and python calculator are applying a different way on deciding whether a unit should itemize or not so that the unit could be better-off, which is considered an optimization. The way python is doing might be computationally expensive for SAS, since python tax logic is divided into many pieces of functions. PR #413 would be an example of how different ways of optimizing will yield various results. 

The two models, SAS calculator and python calculator to be specific, are somewhat different in terms of certain functionalities, and thus in general will not generate the same output even fed with the same input. What I'm trying to do here is to make sure two calculators to share the same functionalities, in order to verify and compare their tax logic, and purely tax logic. 

@MattHJensen @Amy-Xu Let me know if I missed something or said something wrong.

MattHJensen: @martinholmer and @GoFroggyRun, this might be useful information for this discussion: I just confirmed with @feenberg that he plans to make the SAS code available publicly in the Open Source Economics GitHub organization in the next few weeks.

MattHJensen: I also just heard from @feenberg that he has been testing the SAS code on 2012 and 2013 data as well as 2011 data at IRS. This is good news because there have not been major tax-law changes since 2013, but there have been since 2011.

MattHJensen:
This looks great. Thank you @martinholmer. Merging.

PR #529, opened 2016-01-19
Added Tax Logic for Hope Credit

GoFroggyRun:
Per requests from @MattHJensen, in order to analyze the reform that "make American Opportunity tax credit for higher education permanent”, the calculation for Hope Credit was incorporated.

Function `HopeCredit` was written using @feenberg's SAS calculator, which has the tax logic for Hope Credit, and the actual 2009 [Form 8863](http://www.unclefed.com/IRS-Forms/2009/f8863.pdf) as references. Also, the function only deals with the latest tax logic for Hope Credit, that is from year 2005 to 2009. (I assume this is how the Hope Credit to be retained after 2017)

When fed with proper inputs, the function should produce desire results. Since the function `HopeCredit` is called nowhere in `calculate.py`, `py.test` might not be able to detect potential errors. Thus, `py.test` was also executed when `HopeCredit` is called in `calculate.py`, in `self.calc_one_year()`.

@MattHJensen @feenberg @Amy-Xu Any comments, concerns or remarks would be appreciated.


feenberg:
Are the missed statements behind _cmp guards? Should we think about making 
some random INSOLE data for testing those?

dan

GoFroggyRun:
@feenberg:
Thanks for your comments. I actually don't think the code coverage issue (missing statements) is due to data or `_cmp` scenario though. I'll check with TJ on this part.

In order to get some comparison, however, we should definitely come up with a way to make-up the data for the Hope Credit function. Given that we are interested in comparing American Opportunity Credit and Hope Credit, where I believe we have complete data for neither of them, I was wondering wether we can simulate a few samples, with four independent identical distributed variables for each, and evaluate them in both American Opportunity Credit function and Hope Credit function? 
It seems to me that although `p87482`, `e87487`, `e87492` and `e87497` (in American Opportunity Credit), and `e87480`, `e87485`, `e87490` and `e87495` (in Hope Credit) have different names, equating arbitrary pair from each side would be appropriate, since within each function the variables are treated symmetrically. And thus we will be able to compare the effect from the proposal "make American Opportunity tax credit for higher education permanent”. What do you think?

GoFroggyRun:
@talumbau:
Hi TJ, sorry that the Travis CI fails multiple times, but do you have any ideas why `codecov` is decreasing, which I believe should be credited to changes in `functions.py`? Is there a way to fix this? I'm quite confused by this, so please advise. Thanks!

talumbau:
Hi @GoFroggyRun, the code coverage report is pretty naive. It just does a ratio of lines of code that are "hit" (i.e. executed) by the test suite vs. total lines of code in the `taxcalc` package. The general idea is that you always want to increase (or at least maintain) the ratio of code covered vs. total code.

The problem is that `functions.py` contains functions that are decorated with `@iterate_jit`, which does a lot of code transformation on the actual written source code before it is executed. Therefore, every line of code in `functions.py` is considered "missed." So, adding to `functions.py` is equivalent to "adding additional missed code" from the perspective of codecov.

A couple of solutions:

 - merge this PR anyway. We haven't stated as a project that PRs must either increase or at least not decrease code coverage. That's a reasonable long term goal, but we need to make sure codecov is doing exactly what we want before setting up a rule like that.

- exclude `functions.py` from the calculation on code coverage. I had been thinking about doing this anyway, since it's not an accurate measurement when it includes the code in `functions.py`. This would involved adding the line:

```
    taxcalc/functions.py
```
to the .coveragerc file:

https://github.com/open-source-economics/Tax-Calculator/blob/master/.coveragerc


Either one seems like a good way to go to me.

GoFroggyRun:
@talumbau Thanks for your suggestions! They are really helpful. 
Since the codecov test won't measure the `functions.py` in the right way whatsoever, I took the second solution that excludes `functions.py` while conducting code coverage test.

talumbau:
OK great. I have also ignored `functions.py` as well as the JSON files in #531, so we can merge in either order.

MattHJensen:
@GoFroggyRun, could you fix the merge conflicts on this branch, rerun py.test (which will include the new puf-based tests), ensure that it doesn't change results on master, and then I'll merge?

GoFroggyRun:
@MattHJensen 
Sure. But my local machine is running into some weird errors for py.test. I'll resolve them first before fixing the conflicts.

MattHJensen:
Turns out the Dec 2015 budget deal extended the AOC beyond 2017, and so we don't actually need the hope credit anymore. @GoFroggyRun, could you confirm, and then we can close this PR?

https://www.whitehouse.gov/blog/2015/12/22/tax-agreement-makes-permanent-expansions-important-tax-credits-working-families

GoFroggyRun:
@MattHJensen Yes. Confirmed and closed. Thanks for the information.

PR #531, opened 2016-01-19
Omit certain files from test coverage consideration

talumbau:
Omits functions.py (due to @iterate_jit considerations) and all JSON
files

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -4,3 +4,5 @@ omit =
     conda.recipe/*
     csv_to_ascii.py
     taxcalc/_version.py
+    taxcalc/functions.py
+    taxcalc/*.json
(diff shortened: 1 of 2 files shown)


MattHJensen:
Nice new score!

MattHJensen:
Do you think we should implement a policy of maintaining coverage?

talumbau:
I looked at the latest report with this PR:

https://codecov.io/github/open-source-economics/Tax-Calculator/taxcalc?ref=72fffb9812735da9e1064c6fad47c3138106dce6

and the remaining files we consider and resulting statistics look legitimate to me. The coverage report now looks to me like if someone were to decrease coverage, it would very likely be adding untested code. This would lead me to support the idea that all future PRs would be expected to have non-decreasing coverage reports from codecov.

MattHJensen:
@martinholmer, could you merge this if it seems like the right way to proceed to you?

PR #541, opened 2016-01-22
Improve IncomeTaxIO class logic

martinholmer:
Revise IncomeTaxIO class constructor logic in light of recent Records changes.  These revisions make it easier to use the inctax.py program in one of two ways: (1) with raw input data that are not adjusted in any way (that is, no imputation and no blowup or aging), or (2) with proprietary 2009 puf.csv data that is imputed and blown up in the standard manner.  The first way may be useful in cross-model-validation work, while the second way may be useful in debugging Tax-Calculator income tax logic.

The inctax.py help has been updated to stress that the input csv file ay contain any subset of variable names in the Records.VALID_READ_VARS set, which contains all the variables that Tax-Calculator knows how to use in tax calculations.

All the py.test suite of tests and the validation/tests continue to pass after these changes were made.

### DIFF
--- a/inctax.py
+++ b/inctax.py
@@ -54,12 +54,6 @@ def main():
                               '//-comments. No REFORM filename implies use '
                               'of current-law policy.'),
                         default=None)
-    parser.add_argument('--mtrinc',
-                        help=('MTRINC is name of income type for which '
-                              'marginal tax rates are computed. No MTRINC '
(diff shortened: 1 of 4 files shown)


martinholmer:
Added tests to test_incomtaxio.py to increase code coverage.
Merging because this pull request does not effect the normal operation of the Tax-Calculator.

PR #542, opened 2016-01-23
fix mutation of base_calc from diagnostic table

MattHJensen:
The new diagnostic tables allow for the incorporation of behavioral effects by allowing you to pass a base calculator to the method call: `calc_y.diagnostic_table(base_calc = calc_x)`. However, calc_x is mutated by the method; in particular, the year is moved forward in multi-year diagnostic tables. 

This PR creates a hardcopy of base_calc at the top of the method and mutates the copy rather than the passed calculator. I also add a test to ensure that calc_x and calc_y are never advanced in the future. 

@amy-xu, could you please review?

### DIFF
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -382,6 +382,7 @@ class Calculator(object):
         table = []
         row_years = []
         calc = copy.deepcopy(self)
+        base_calc = copy.deepcopy(base_calc)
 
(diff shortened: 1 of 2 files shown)


Amy-Xu:
Thanks for the fix. LGTM. Merging.

MattHJensen:
Thanks @Amy-Xu.

PR #543, opened 2016-01-23
Add --weights option to inctax.py program

martinholmer:
This enhancement allows the inctax.py program to do the following sort of thing: use ```puf.csv``` as INPUT, impute and age INPUT to the TAXYEAR (using the ```--blowup``` option), and write out the federal income tax liability and various intermediate values (AGI, etc.) and the aged sample weight in the TAXYEAR as output variable 29 (using the new ```--weights``` option).  The output file can then be tabulated (using the weights) to confirm that aggregate results for TAXYEAR are the same as those reported by the Tax-Calculator diagnostic_table() and by TaxBrain.  Having individual filing unit information that aggregates to know totals may be useful in investigating the questionable looking aggregate results from Tax-Calculator and/or TaxBrain.

### DIFF
--- a/inctax.py
+++ b/inctax.py
@@ -24,7 +24,7 @@ def main():
                      'the Tax-Calculator. '
                      'The INPUT file is a CSV-formatted file that contains '
                      'variable names that are a subset of the '
-                     'Records.VALID_READ_VARS set.  The OUTPUT file uses '
+                     'Records.VALID_READ_VARS set.  The OUTPUT file is in '
                      'Internet-TAXSIM format.  The OUTPUT filename is the '
                      'INPUT filename (excluding the .csv suffix or '
                      '.gz suffix, or both) followed by '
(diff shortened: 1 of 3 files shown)


martinholmer:
Here is an example of this pull request in action:
```
$ python inctax.py puf.csv 2015 --blowup --weights
You loaded data for 2009.
Your data have been extrapolated to 2015.
$ ls -l puf-15.out-inctax
-rw-r--r--  1 mrh  staff  38648255 Jan 23 12:22 puf-15.out-inctax
$ awk '{agi+=$10*$29; iitax+=$4*$29}END{print agi,iitax}' puf-15.out-inctax
1.02339e+13 1.44069e+12
```
The tabulated AGI is 10,233.9 billion, which is the same as the 2015 value in the diagnostic table.
The tabulated iitax is 1,440.69 billion, which rounds to the table value of 1,440.7 billion.

martinholmer:
The changes in this pull request have no effect on the core logic of the Tax-Calculator, so the changes are being merged into the master branch without any external review.

MattHJensen:
This looks very helpful. Thanks @martinholmer!

PR #552, opened 2016-01-25
Delete accuracy.py

MattHJensen:
We no longer use accuracy.py for testing against the SAS code. @gofroggyrun, please confirm before I merge this.


GoFroggyRun:
@MattHJensen
LGTM `Accuracy.py` is indeed no longer used.

MattHJensen:
Thanks @GoFroggyRun. Merging this deletion of accuracy.py. 

cc @martinholmer.

PR #560, opened 2016-01-27
Fix bug in EI_FICA function wrt AGI deduction for Self-Employment (FICA) Tax

martinholmer:
This pull request creates ```test_1``` in the new ```test_functions.py``` file that demonstrates a bug in the ```EI_FICA``` function.  The first of the two filing units in the test should have an AGI of $391,121.85 but the master branch logic produced an AGI of $397,321.85, a difference of $6,200.

This pull request also revises the ```EI_FICA``` function logic to correct this bug.

And finally, this pull request also revises the two ```pufcsv_*_expect.txt``` results files because the aggregate and marginal tax rate results are slightly different after the bug fix.

@MattHJensen @feenberg @GoFroggyRun @Amy-Xu


GoFroggyRun:
@martinholmer 
Thanks for the prompt fix! The patch looks good to me (and also really clear), after verifying against the actual form.

cc @MattHJensen

MattHJensen:
Thanks for reviewing this @GoFroggyRun. Could you please add a todo to look into what implications this has for the SAS calculator or the SAS-Python testing scheme?

Merging this now.

MattHJensen:
Thanks @martinholmer!

PR #570, opened 2016-01-31
Fix bug in AddCTC function wrt FICA calculation

martinholmer:
The following statement in the AddCTC function has at least two problems.
```
c82900 = 0.5 * (FICA_mc_trt + FICA_ss_trt) * min(min(SS_Earnings_c, c82880), e00200)
```
First, it applies the maximum taxable earnings (MTE) parameter (SS_Earnings_c) to the HI (Medicare) FICA tax.  And second, it fails to understand that the MTE is applied to **individual** earnings (not to the filing unit's wage-and-salary earnings total, e00200).

This pull request adds test_2 to test_functions.py, which fails when using the statement above, and then revises the Tax-Calculator logic to eliminate this test failure in a way that corrects both problems.  The aggregate results generated using the puf.csv file change hardly at all and there are only a few filing units whose marginal tax rates differ enough to put them in different histogram bins.


MattHJensen:
This looks great. Merging. Thanks @martinholmer.

feenberg:
On Sun, 31 Jan 2016, Martin Holmer wrote:

> 
> The following statement in the AddCTC function has at least two problems.
> 
> c82900 = 0.5 * (FICA_mc_trt + FICA_ss_trt) * min(min(SS_Earnings_c, c82880), e00200)
> 
> First, it applies the maximum taxable earnings (MTE) parameter (SS_Earnings_c) to the
> HI (Medicare) FICA tax. And second, it fails to understand that the MTE is applied to
> individual earnings (not to the filing unit's wage-and-salary earnings total, e00200).
> 
> This pull request adds test_2 to test_functions.py, which fails when using the
> statement above, and then revises the Tax-Calculator logic to eliminate this test
> failure in a way that corrects both problems. The aggregate results generated using
> the puf.csv file change hardly at all and there are only a few filing units whose
> marginal tax rates differ enough to put them in different histogram bins.

How can this be fixed when individual earnings are not in the PUF? Or are 
they imputed somehow?

dan

martinholmer:
Dan @feenberg said:
> How can this be fixed when individual earnings are not in the PUF? Or are they imputed somehow?

We have been using for some months now a ```puf.csv``` file that has imputed taxpayer and spouse earnings.  I guess when we started using that file, only some of the tax-calculation code got changed to use the individual earnings variables in the new file.  This change should have been done then, but got overlooked, I guess.

martinholmer:
Dan @feenberg,
Notice that the code that was fixed in pull request #570 was wrong even if individual earnings were not available.  It incorrectly capped Medicare FICA taxes using the maximum taxable earnings parameter.

PR #527, opened 2016-01-13
Extrapolation Factor Modified for Capital Gain 

GoFroggyRun:
The taxable net capital gain for 2013 is approximately $439 billion on current master, while the data from SOI suggests that the real net capital gain for 2013 should be about $504 billion. This indicates that we might want to modify the extrapolation factor of capital gain for year 2013 and afterwards.  
This PR modifies the extrapolation factor of capital gain in 2013 to match the SOI data on capital gain, that is setting the factor to be `2.198301`. Factors for years later on are also modified iteratively, by applying the growth rate to previous year's factor. And we have:
```
[2013] 2.198301 2.964478 3.450607 3.508749 3.619760 3.746481 3.899808 4.058296 4.222059 4.396487 4.576191 4.766445 4.964612 5.171036
```

The diagnostic table for this PR:
screen shot 2016-01-12 at 6 58 20 pm

Please note that applying the CBO option that increases capital gains and dividends rate by 2% on top of this PR would amplify the previous difference we have by roughly $2 billion, instead of the $1 billion that I might have mentioned by mistake: 
screen shot 2016-01-12 at 8 25 48 pm

@MattHJensen @Amy-Xu Any comments, concerns or remarks would be appreciated.

PR #575, opened 2016-02-02
Update factors and weights with Jan 16 CBO economic projections

Amy-Xu:
With the new factors and weights in this PR, the reform results almost stay at the same place but the baseline drops significantly, mainly due to the two following reasons.

- In PR #527, capital gain extrapolation factor was updated by @GoFroggyRun, by comparing our 2013 weighted cap gain and SOI 2013 stats. It was partially my fault that I forgot to tell Sean we should be using the net gain, which is the weighted positive part of either `e01100` or `p23250 + p22250`, instead of the total weighted sum. Thus after this PR, the net gain (2015) was over blown to 907b from previously 791b, about 116b increase, and baseline in 2015 was increased from 1410b to 1440b. Two takeaway points: 1) current baseline in 2015 should still stay at 1410b, and 2) 116b net capital gain diff leads to a 30b individual income tax jump.

- Switching from 2014 economic projection to the latest Jan 2016 version, the biggest difference still comes from capital gain section. CBO ([Detailed Revenue Projections, realized capital gain](https://www.cbo.gov/about/products/budget_economic_data#3)) changed their estimates for 13-15 significantly in this version, increasing 2013 realized gain from 416b to 511b, and 15 gain from 657b to 679b. As we are only using the growth rate instead of the actual values from CBO, our targets for 2015 drops by about 101b. Thus, this cap gain drop results in another decrease on individual income tax for about 25 billion.

This updates seems to drag our baseline further away from CBO's number, which is sad, but it still seems reasonable enough to update it anyways. Any feedback/questions on the reasoning above @feenberg @martinholmer @MattHJensen @GoFroggyRun?

There's still a small problem on the line of test I comment out in test_policy.py. I'll keep working on that.


Amy-Xu:
Before my latest commit, I'm not sure whether I messed anything up or not, but policy keeps giving me 41702 for the value of that parameter, but the actually manually-calculated result is 41702.695.

@martinholmer @talumbau Was wondering when we expanded parameters using CPI, are we rounding all numbers downward? If so, where's the code?

martinholmer:
@Amy-Xu said:
> Before my latest commit, I'm not sure whether I messed anything up or not, but policy keeps giving me 41702 for the value of that parameter, but the actually manually-calculated result is 41702.695.
>
> @martinholmer @talumbau Was wondering when we expanded parameters using CPI, are we rounding all numbers downward? If so, where's the code?

Amy,  Just let the test_policy.py code stay the way it is, and after this pull request is merged into the master branch, I'll review all the tests in test_policy.py to see which need to be updated to reflect the recent addition of historical 2015 values for policy parameters.

cc @MattHJensen @GoFroggyRun

martinholmer:
What do people think about pull request #575?  It seems ready for merging into the master branch.  Do others agree or not?  If no objections, I'll merge later today.

cc @MattHJensen @GoFroggyRun @Amy-Xu

talumbau:
Hi @Amy-Xu just to answer your question, currently when we expand parameters, it looks to me like we are not rounding down per year. The code for this is in `parameters_base.py` in the `expand_1D` and `expand_2D` functions. So here:

expand_2D:
https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/parameters_base.py#L456

expand_1D:
https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/parameters_base.py#L405

So each value stays as a floating point value as we expand and build the new values. Then before returning, we cast to the `dtype` of what we were given. So if it was originally `int`, at the end, just before returning from the function, we would round down (or really, just truncate) to integer values. As shown in this line:

https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/parameters_base.py#L412

MattHJensen:
This looks good to me, too, @martinholmer. Thanks @Amy-Xu! Merging.

PR #585, opened 2016-02-05
Remove redundant test from test_calculate.py file

martinholmer:
This complex, and difficult to maintain, test looks redundant.  That is, it appears as if removing it will leave the code coverage of the tests unchanged.  But I'm not sure of that because I have not been able to figure out how to run the code coverage report on my local computer.  Hence, I'm creating this pull request to get the needed code coverage report that will confirm or reject my hunch that removing this one test will leave code coverage unchanged.


martinholmer:
So, the code coverage report (see above) confirms the hunch that the removed test was, in fact, redundant.  This pull request does not change any tax-calculation logic or any other tests (other than the removal of the one redundant test).

PR #587, opened 2016-02-08
fix bug in personal refundable credit phaseout rate 

MattHJensen:
The personal refundable credit should only reduce taxes. Before, if the phaseout rate was positive, the credit could increase taxes. 

This does not affect baseline results. @martinholmer, could you please review?

### DIFF
--- a/taxcalc/functions.py
+++ b/taxcalc/functions.py
@@ -576,7 +576,7 @@ def Personal_Credit(c04500, MARS, II_credit, II_credit_ps, II_credit_prt):
     else:
         credit_phaseout = 0.0
 
-    _personal_credit = _personal_credit - credit_phaseout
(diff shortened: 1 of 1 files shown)


martinholmer:
@MattHJensen said:
> @martinholmer, could you please review?

Looks just fine to me because it brings the logic into agreement with the documentation in the ```current_law_policy.json``` file.

I've never figured out whether or not it is important in Pandas to pay attention to integer versus float, so I don't know if the following would make any difference:
```
_personal_credit = max(0.0, _personal_credit - credit_phaseout)
```
versus the integer version (which is what you have in the pull request)
```
_personal_credit = max(0, _personal_credit - credit_phaseout)
```

I'll be happy to merge or you could do it yourself.

MattHJensen:
Thanks @martinholmer for the review.

Issue #577, opened 2016-02-04
Wild Imputed Amount for Numextra


GoFroggyRun:
In `reocrds.py` we imputed `numextra` in this way:
```
        self._numextra = np.where(
            np.logical_or(self.AGERANGE >= 6,
                          np.logical_and(self.FDED != 2,
                                         np.logical_and(
                                             self.AGERANGE <= 0,
                                             self.e02400 > 0))),
            self._txpyers,
            np.where(np.logical_and(self.FDED == 2,
                                    self.p04470 >
                                    std_2009[self.MARS - 1]),
                     np.where(self.MARS != 2,
                              (self.p04470 -
                               std_2009[self.MARS - 1]) /
                              std_aged_2009[0],
                              (self.p04470 -
                               std_2009[self.MARS - 1]) /
                              std_aged_2009[1]), 0))
```
But It seems that there are 84 records in 08 puf will have `numextra` of amount more than 2 after the imputation process under 2014 law, where the largest number should be 2135. 

@feenberg: 
I'm not quite sure why this is happening. If I were not mistaken, `numextra` shall not exceed 2. And I was also wondering whether it's a good idea to cap `numextra` accordingly based on units' MARS during imputation phase.
Please advice, thanks!

cc @martinholmer @MattHJensen

martinholmer:
Sean said:
> In ```records.py``` we imputed ```_numextra``` in this way:
> ```
> COMPLICATED STATEMENT FROM the **impute_variables** function ...
> ```
> But It seems that there are 84 records in 08 puf will have numextra of amount more than 2 after the imputation process under 2014 law, where the largest number should be 2135.

Sean,  The above statement in ```impute_variables``` will **never** be executed because the ```impute_variables``` function will **never** be called for 2008 data.  So, unless I'm missing something, I don't see that there is anything to worry about.  You should never be trying to use the 2009 data "imputation process" on any data except data for 2009.  Can you provide us some background on what you're doing and why this issue came up?

cc @GoFroggyRun @feenberg @Amy-Xu @MattHJensen

feenberg:
Can you supply the amount of the standard deduction for one of the 
offending returns? The code below is difficult to follow, but I think it 
should work unless the file contains records with a standard deduction 
larger than can be achieved with 2 elderly taxpayers. If we had the data, 
we could tell for sure.

dan

MattHJensen:
[...]

PR #588, opened 2016-02-09
Modifying the Imputation for Numextra 

GoFroggyRun:
Previously, before the cps matched version of 09 puf being available, we partially rely on `p04470` for the imputation of numextra, which sometimes yields unexpected values (more than 2, or non-integer, see #577 for details). 
`AGERANGE` variable itself is quite powerful in deciding whether a unit's elder or not. When it's not available (i.e. 0, which indicates Not Present according to 09 puf booklet), we use the `age` variable obtained from cps to determine whether filers are considered as elder or not. 

@MattHJensen @martinholmer @feenberg 
Any comments, concerns or remarks would be appreciated.


martinholmer:
Sean,
How is it possible that this change in the ```_numextra``` imputation logic causes no changes in any of the tests?  If it doesn't cause a change in any test results, why are we making this change?

cc @GoFroggyRun @MattHJensen @feenberg

GoFroggyRun:
@martinholmer Thanks for the reminder. I forgot to update the diff files.

martinholmer:
Sean,  Adding the revisions in the ```requires_puf_csv``` tests makes the pull request look more sensible.

Are others comfortable with the new ```_numextra``` imputation logic?  If so, someone should merge this pull request into the master branch.   [We can't tell whether one member of a couple is elderly and the other is not, but that is because of a fundamental shortcoming of the data with which we are working.]

cc @GoFroggyRun @MattHJensen @feenberg

martinholmer:
I verified the pull request #588 logic change in imputing ```_numextra``` as follows:
```
$ python
Python 2.7.11 |Anaconda 2.3.0 (x86_64)| (default, Dec  6 2015, 18:57:58) 
[GCC 4.2.1 (Apple Inc. build 5577)] on darwin
>>> from taxcalc import *
>>> ppo = Policy()
>>> rec = Records()
>>> calc = Calculator(policy=ppo, records=rec)
You loaded data for 2009.
>>> import numpy as np
>>> nelist = list(calc.records._numextra)
>>> len(nelist)
219814
>>> nelist.count(0)
191018
>>> nelist.count(1)
13675
>>> nelist.count(2)
15121
>>> nelist.count(3)
0
>>> quit()
$ 
```
Notice that the sum of the number of filing units with 0, 1, or 2 ```_numextra``` values is equal to the total number of filing units (219814).  This shows the incorrect values greater than two that were imputed by the prior logic have been eliminated.

Thanks for this contribution, Sean.

cc @GoFroggyRun @MattHJensen @feenberg

Issue #589, opened 2016-02-10
Bug in handling large parameter values


martinholmer:
The following problem was encountered on the master branch using the 3a87f50 commit.
Consider the following test:
```
@pytest.mark.one
def test_high_mte():
    """
    Test different ways of eliminating the maximum taxable earnings (MTE)
    used in the calculation of the OASDI payroll tax.
    """
    # clarify start year and create Policy parameters object
    syr = 2013
    ppo = Policy(start_year=syr)
    # confirm that MTE (or '_SS_Earnings_c') has current-law value in 2016
    assert ppo._SS_Earnings_c[2016 - syr] == 118500
    # specify a "pop the cap" reform that 'eliminates' MTE cap in 2016
    # ... first way:
    reform = {2016: {'_SS_Earnings_c': [9.20E18]}}
    ppo.implement_reform(reform)
    assert ppo._SS_Earnings_c[2016 - syr] == 9200000000000000000
    # ... second way:
    reform = {2016: {'_SS_Earnings_c': [9.25E18]}}
    ppo.implement_reform(reform)
    assert ppo._SS_Earnings_c[2016 - syr] == 9250000000000000000
```

When I add this test to the test_policy.py file and run the test, I get these results:
```
$ py.test -m one
============================= test session starts ==============================
platform darwin -- Python 2.7.11, pytest-2.8.1, py-1.4.30, pluggy-0.3.1
rootdir: /Users/mrh/work/OSPC/tax-calculator, inifile: setup.cfg
plugins: pep8-1.0.6
collected 108 items 

tests/test_policy.py F

=================================== FAILURES ===================================
________________________________ test_high_mte _________________________________

    @pytest.mark.one
    def test_high_mte():
        """
        Test different ways of eliminating the maximum taxable earnings (MTE)
        used in the calculation of the OASDI payroll tax.
        """
        # clarify start year and create Policy parameters object
        syr = 2013
        ppo = Policy(start_year=syr)
        # confirm that MTE (or '_SS_Earnings_c') has current-law value in 2016
        assert ppo._SS_Earnings_c[2016 - syr] == 118500
        # specify a "pop the cap" reform that 'eliminates' MTE cap in 2016
        # ... first way:
        reform = {2016: {'_SS_Earnings_c': [9.20E18]}}
        ppo.implement_reform(reform)
        assert ppo._SS_Earnings_c[2016 - syr] == 9200000000000000000
        # ... second way:
[...]

PR #590, opened 2016-02-10
Fix bug in handling large parameter values

martinholmer:
This pull request switches numpy arrays of parameters from having datatype int64 to having datatype float64 in order to resolve the problems raised in issue #589.  These changes allow the intuitive setting of the maximum OASDI taxable earnings (MTE) level to infinity in reforms that eliminate the MTE.  These changes have only minor effects on the aggregate ```required_puf_csv``` test results and no effects on the marginal tax rate ```requires_puf_csv``` test results.  These changes require a shift in several py.test files from exact assert statements to the use of numpy.testing.assert_allclose with very small tolerances.  None of the tax-calculation logic in Tax-Calculator has been changed in this pull request.

@MattHJensen @feenberg @talumbau @GoFroggyRun


talumbau:
This PR may end up being the best way to handle https://github.com/open-source-economics/Tax-Calculator/issues/589 but I'd like a change to digest the issues raised in https://github.com/open-source-economics/Tax-Calculator/issues/589 and review the test case given there before we proceed with this PR.

taxcalc/parameters_base.py
```diff
@@ -395,7 +395,7 @@ def expand_1D(x, inflate, inflation_rates, num_years):
             if len(x) >= num_years:
                 return x
             else:
-                ans = np.zeros(num_years, dtype='f8')
+                ans = np.zeros(num_years, dtype=np.float64)
```
talumbau: The two statements here are equivalent.

MattHJensen:
Just chatted with @talumbau about this approach. We are all on the same page that this is the best solution. Merging.

PR #595, opened 2016-02-14
Updated test.py to Produce Extrapolated 'Inputs'

GoFroggyRun:
As discussed in #507, our calculator doesn't have a way of providing the real "inputs" that actually run through the tax-logic, which might pose problems when it comes to validation/debugging. Currently, how the calculator works can be roughly described as:
```
read puf -> imputation/extrapolation/weighting -> real inputs [missing] -> calculate.py/functions.py -> output variables
```
Note that the real inputs were hidden, and there's no automated way in getting them.

This PR has no impact on the output variables at all, but offers an option, using command `python test.py --realin`, to get the missing information when needed.

@martinholmer @MattHJensen Any comments, concerns or remarks would be appreciated.


martinholmer:
Sean,
There is at least one major problem with this pull request.  You should **never** do a copy and paste of VALID_READ_VARS into the test.py file; simply refer to it as ```Records.VALID_READ_VARS```.   If you need a list instead of a set, simply convert the ```Records.VALID_READ_VARS``` set into a list as follows:
```
valid_read_vars_list = list(Records.VALID_READ_VARS)
```
If you do the copy and paste, the test.py code will not work correctly after a change in the Records class alters the ```Records.VALID_READ_VARS``` set.

Fix this up and then we can focus on whether or not there any other issues with this pull request.

GoFroggyRun:
@martinholmer 
Thanks for your comments. I have updated the list by, instead of making a new one, referring to what we already have in `records.py`. How do you feel about this PR now?

martinholmer:
Sean (@GoFroggyRun),
The changes you made seem just fine, but I don't think the ```test.py``` file is the best place for this enhancement.  So, I've added the ability to write out post-blowup values for any taxyear to the ```inctax.py``` script by adding a ```--records``` option.  If you update your local computer to the Git Hub master branch and execute ```python inctax.py --help```, you'll see the documentation for the new option.

Your notion that this would be a useful capability has already been proven true.  I discovered a bug in the Records class logic that I'll fix in a separate pull request later this afternoon.

Meanwhile, I'm hoping you will test, and possibly fix, the new ```inctax.py``` ```--records``` logic.  Here is one way to start that testing.

It seems as the following command should produce a records output file that is essentially the same as the ```puf.csv``` input file.
```
$ python inctax.py puf.csv 2013 --records
```
The resulting output file is ```puf-13.records``` and they have not been blown up because the ```--blowup``` option was not used.  I started comparing the ```puf.csv``` file variable by variable with the ```puf-13.records``` output file using the following script:
```
import pandas as pd
import numpy as np
import sys
inp = pd.read_csv('puf.csv')
out = pd.read_csv('puf-13.records')
for varname in list(inp.columns):
  if varname == 'FLPDYR':
      continue
  inpdata = getattr(inp, varname)
  outdata = getattr(out, varname)
  if np.allclose(inpdata, outdata, rtol=0.0, atol=0.00001):
      pass
  else:
      sys.stdout.write('{} HAS DIFFERENCES\n'.format(varname))
      num = 0
      for i in range(0,len(inpdata)):
          if inpdata[i] != outdata[i]:
              num += 1
              if num > 5:
                  break
              else:
                  sys.stdout.write('{}= {:.4f} {:.4f}\n'.format(
                      'inp,out', inpdata[i], outdata[i]))
```
This script produces the following output:
```
iMac2:tax-calculator mrh$ python check.py
e15360 HAS DIFFERENCES
inp,out= 1040.0000 0.0000
inp,out= 320.0000 0.0000
inp,out= 5420.0000 0.0000
inp,out= 2080.0000 0.0000
inp,out= 3820.0000 0.0000
p87482 HAS DIFFERENCES
inp,out= nan nan
inp,out= nan nan
inp,out= nan nan
inp,out= nan nan
inp,out= nan nan
p87521 HAS DIFFERENCES
inp,out= nan nan
inp,out= nan nan
inp,out= nan nan
inp,out= nan nan
inp,out= nan nan
matched_weight HAS DIFFERENCES
inp,out= 9.0537 9.0540
inp,out= 11.0931 11.0930
inp,out= 11.0931 11.0930
inp,out= 9.0537 9.0540
inp,out= 9.0537 9.0540
```

The ```e15360``` differences are the bug I mentioned above.  But the other three sets of differences are puzzling to me.  Can you figure out what's going on with ```p87482```, p87521```, and ```matched_weight```?

Let me know if you have any questions.

GoFroggyRun:
@martinholmer:
Thanks for the prompt updates in `inctax.py`. Adding `--records` option makes it much easier to compare the inputs files. 
Regarding your concerns, `p87482` and `p87521` variables do have missing values in the original `puf` file, and thus `nan` output should correspond to those missing values. Also, it seems to me that the `matched_weight` differences are caused by some rounding errors.

martinholmer:
Sean (@GoFroggyRun)said:
> Thanks for the prompt updates in inctax.py.  Adding --records option makes it much easier to compare the inputs. 

Let me know if adding the use of the ```--blowup``` option produces the ```--records``` output you expect.

GoFroggyRun:
@martinholmer 
Yes. The outputs remain the same as the output on this PR, when using `--records --blowup 2013` option. Closed. Thanks @martinholmer .

Issue #596, opened 2016-02-15
Remove the `times_equal` helper function and rid the code base of `casting=unsafe`


talumbau:
See discussion on #381.

martinholmer:
T.J. (@talumbau),
As I think more about this issue, there seem to be some variables in the Records input file (e.g., ```puf.csv```) that are best left as integers after they are read into the Records class.  The main example of this kind of variable is MARS.  I guess any move away from the current int64 dtype for Records input would require some variables to remain int64 and others (that are involved in ```times_equal``` calls) to be converted to float64 dtype.

feenberg:
On Mon, 15 Feb 2016, Martin Holmer wrote:

> 
> T.J. (@talumbau),
> As I think more about this issue, there seem to be some variables in the Records input
> file (e.g., puf.csv) that are best left as integers after they are read into the
> Records class. The main example of this kind of variable is MARS. I guess any move
> away from the current int64 dtype for Records input would require some variables to
> remain int64 and others (that are involved in times_equal calls) to be converted to
> float64 dtype.
>

In taxsim we have always stored MARS as a real - and have not noticed any 
problems. If it  makes things easier to have everything in one type, I 
wouldn't worry about MARS. At the beginning of the program we create an 
integer out of the stored value, and use it as an array index, since 
Fortran doesn't allow real array indicies. It works fine.

dan

martinholmer:
Issue #596 has been resolved by the merging of pull request #603 into the master branch.

PR #603, opened 2016-02-18
Remove Records class use of casting='unsafe'

martinholmer:
This pull request resolves issue #596.  The new logic uses arrays with dtype=float64 for all the e-code-like variables (instead of using integer arrays), which means the results of many annual blow ups of the variables are slightly different than before.  This means that some intermediate tax results are slightly different than before, but the FICA and income tax liability aggregates are unchanged (at least when they are rounded to the nearest tenth of a billion dollars).

This pull request also introduces the Records.UNUSED_READ_VARS set that contains variables that are currently in the Records.VALID_READ_VARS set but are not used in the Tax-Calculator calculations.  The unused variables are not created in the Records class, which make the code easier to understand and execute faster.  The long-term objective is to remove all of these unused variables from the ```puf.csv``` file so that all these unused variable names can be removed from both the Records.VALID_READ_VARS and Records.UNUSED_READ_VARS sets.

This pull request also updates the ```var_labels.txt``` documentation file so that it includes only and all the variables that are in the Records.VALID_READ_VARS set and that are not in the Records.UNUSED_READ_VARS set.  This change means we are documenting only variables that are actually used in the Tax-Calculator and that we are documenting all the variables used in the Tax-Calculator (including variables that are not e-code variables).

cc @MattHJensen @feenberg @talumbau @GoFroggyRun


martinholmer:
The code-coverage report is, on net, a good thing.  But I find it strange that when I simplify the code (by reducing the number of statements by 15 and having tests that cover all my added code), it tells me that code coverage will be reduced if I merge the pull request.  This is technically true, but not very helpful.

taxcalc/records.py
```diff
@@ -290,253 +317,223 @@ def set_current_year(self, new_current_year):
...
-        times_equal(self.e11582, self.BF.ATXPY[year])
-        times_equal(self.e11583, self.BF.ATXPY[year])
-        times_equal(self.e10605, self.BF.ATXPY[year])
-        times_equal(self.e11900, 1.)
-        times_equal(self.e12000, 1.)
-        times_equal(self.e12200, 1.)
+        # pylint: disable=attribute-defined-outside-init
+        # pylint: disable=too-many-locals
+        # local copies of blowup factors for specified year:
+        AWAGE = self.BF.AWAGE[year]
```
talumbau: I'm surprised by a few things in this section. Here are a few general comments:

- the comment on line 325 is not correct. The variable assignments on lines 325-341 don't make copies of the values on the right hand side. Therefore, I find this section of code quite verbose, since I don't see what is gained by these variable assignments.

- my first comment is just a style issue (aside from the issue with the comment on line 325), and it seems reasonable that if someone would prefer to use `ADIVS` rather than `self.BF.ADIVS[year]`,  as long as he or she is willing to type out all of those variables, there is obviously no difference in how the code executes statements with either `ADIVS` or `self.BF.ADVIS[year]`. This comment is on the more serious performance implications of the changes here. The `time_equal` function used `np.multiply` to do in-place modification of the numpy arrays. Prior to using this `times_equal` function, all of the updates to variables here used the `*=` operator to do similar in-place operations. That is, they were all of the form:

```
    self.e00200 *= self.BF.AWAGE[year]
```

rather than 
```
  self.e00200  = self.e00200 * self.BF.AWAGE[year]
```
The reason to prefer the former is that it does not allocate new memory to hold the results of the multiplication and then assign this new object to the attribute `e00200` on the `self` object, whereas the latter choice does exactly that. So, in the code below, every assignment would involve a new memory allocation, which would then be filled with the results of the multiplication, and this object would be assigned to the attribute on the left, and the previous numpy array would no longer be referred to by anything, and so it would have to be garbage collected at some point.

- similarly, the `np.where` assignments previously used a similar technique to avoid new memory allocation. The old code did as follows:

```
  self.e00900p[:] = np.where(self.e00900p >= 0,
                                  self.e00900p *
                                  self.BF.ASCHCI[year],
                                  self.e00900p *
                                  self.BF.ASCHCL[year])
```

The left-hand side uses the `[:]` indexing style to essentially say "assign into all of the indices here", so that the values created on the right-hand side are place there. Without the `[:]`, a new numpy object is created on the right hand side, this is assigned to the attribute `e00900` on the object `self`, and then the previous numpy array referred to is no longer referred to by anyone, and so it must be garbage collected. So it is more efficient to use the `[:]` syntax on these assignments.

talumbau:
Agree that the code coverage report here is not particularly helpful in this case. Because it is a simple ratio, even an operation that simply replaces several lines of code with one equivalent line of code would lower the ratio and trigger a bad report.

martinholmer:
T.J., Thanks for the efficiency tips!  I've incorporated them in the Records._blowup method and have also applied them in other places in the Tax-Calculator code.

cc @MattHJensen @feenberg @talumbau @GoFroggyRun

talumbau:
sure no problem. This looks good to me now.

MattHJensen:
Thanks @martinholmer; this looks good. Merging.

Issue #572, opened 2016-02-02
Need to verify another 2015 EITC amount


martinholmer:
We added 2015 to the Internet-TAXSIM versus Tax-Calculator validation test suite in pull request #561.

There were no FICA differences between the two models using the c15.in sample of 100,000 randomly generated filing units.  However, there were 617 filing units with income tax differences of more than one cent.  The largest income tax difference was $77.93 for the unit with id 74480 and all that difference was caused by a difference in the two model's estimates of the EITC amount.

We need to compute this unit's EITC amount using the 2015 forms to verify that the Tax-Calculator current-law-policy parameters and logic are correct.

If we can confirm that the Tax-Calculator is computing the correct EITC for this filing unit, then we need to send a scanned image of the filled-out forms to Dan Feenberg at NBER.  That is the Internet-TAXSIM bug-reporting procedure.  Also, send Dan the 22-variable input line for the filing unit with id 74480, and remind him that we are using Internet-TAXSIM option ```56 1``` to generate results that we compare with Tax-Calculator results.

Sean, can you undertake this validation task?

cc @MattHJensen @feenberg @Amy-Xu @GoFroggyRun

GoFroggyRun:
@martinholmer Thanks for pointing out this issue. I might not be able to start looking in to this immediately, but please expect this issue to be verified no latter than this Thursday.

martinholmer:
Sean said:
> Thanks for pointing out this issue. I might not be able to start looking in to this immediately, but please expect the issue to be verified no later than this Thursday.

@GoFroggyRun,  That sounds great.  Thanks for looking into this matter.

GoFroggyRun:
@martinholmer:

I just checked record with id 74480. The unit has earned income $26,700, which yields a credit of $4,896; while his AGI is of $49,270, which yields a credit of $147. The smaller of which will be the EITC of that person, which is $147. Our calculator gives the EITC output of $148.22, while Internet Taxsim indicates $70.29 to be the EITC amount. 

cc @MattHJensen @feenberg

martinholmer:
Sean said:
> I just checked record with id 74480. The unit has earned income $26,700, which yields a credit of $4,896; while his AGI is of $49,270, which yields a credit of $147. The smaller of which will be the EITC of that person, which is $147. Our calculator gives the EITC output of $148.22, while Internet-TAXSIM indicates $70.29 to be the EITC amount.

Sean,  Thanks for the quick verification of this case.

[...]

PR #604, opened 2016-02-19
[WIP] Update EITC Threshold

GoFroggyRun:
This PR updates the EITC threshold for 1 kid and more.


martinholmer:
Sean,  This change in an EITC parameter reduces the number of differences by 16 (from 604 to 588), but the largest difference is now a little larger than before.  I'm wondering what source of information leads you to believe the parameter value needed to be changed.

Maybe we should list all 2015 EITC-related parameter values (from the current_law_policy.json file) so that Dan can compare what values TAXSIM is using for the same parameters.

Thanks for pursuing this issue.

cc @GoFroggyRun @MattHJensen @feenberg

GoFroggyRun:
@martinholmer :
So, as discussed in #572, record with id 74480 indicates different results for two calculators, where current master yields an EITC amount of $148.22, while TAXSIM approximates the amount to be $146.11. After looking at the table, I found that, since the EITC begins to phase out, it would make more sense to get an estimate lower than the actual value, $147.
And thus I modified the threshold a bit, and, currently, for that particular record, two calculators agree on the EITC amount, which is $146.11.But I'm surprised to see the largest difference gets larger.

martinholmer:
Sean said:
> So, as discussed in #572, record with id 74480 indicates different results for two calculators, where current master yields an EITC amount of $148.22, while TAXSIM approximates the amount to be $146.11. After looking at the table, I found that, since the EITC begins to phase out, it would make more sense to get an estimate lower than the actual value, $147.
And thus I modified the threshold a bit, and, currently, for that particular record, two calculators agree on the EITC amount, which is $146.11.

Sean,  Thanks for your continued work on this issue.  The more results you report, the more it looks to me as if Tax Calculator and Internet TAXSIM may have slightly different 2015 EITC parameters (particularly the phase-out-related parameters).

Do you think it makes sense for you to extract from the ```current_law_policy.json``` file all the 2015 EITC parameters and send them to Dan Feenberg at NBER, asking him to compare them with the 2015 EITC parameters that Internet TAXSIM is using.  If they are the same, then we can attribute the few-dollar cross-model-validation differences in EITC amounts to different rounding rules, I guess. 

What do you think is the most efficient way to resolve this issue?

cc @GoFroggyRun @feenberg @MattHJensen

GoFroggyRun:
@martinholmer 
Sure. I'll extract those parameters and get in touch with him.

martinholmer:
Sean said:
> Sure. I'll extract those parameters and get in touch with him.

Thanks.

feenberg:
Can you send me the thresholds you use for the EITC clawback?

dan

martinholmer:
Sean,  See pull request #605 for the correct EITC phaseout start values for 2015.  Just look at the source mentioned in the ```current_law_policy.json``` file to see what the 2015 IRS documents say are the parameter values.

martinholmer:
Closing pull request #604 because 2015 EITC validation problems have been resolved by pull requests #605 and #613.

PR #613, opened 2016-02-25
Update Internet-TAXSIM output and taxdiffs for c15.in sample

martinholmer:
New Internet-TAXSIM results for the c15.in sample of 100,000 randomly-generated filing units now agree (within one cent) on FICA and individual income tax liabilities, and therefore, on EITC amounts.  The taxcalc/validation/README.md file will be updated to reflect this milestone in a future pull request.


martinholmer:
The exact agreement between Tax-Calculator and Internet-TAXSIM on FICA and income tax liabilities has been achieved over the past few days by updating several 2015 EITC policy parameters to IRS published values.  After both models started using the correct EITC parameters, all the results for the c15.in sample came into agreement.  The EITC amounts are generated by equations in both models rather than by the use of a lookup table as the IRS forms provide.  As a result, there are many several dollar differences between the equation-generated EITC amounts from the two models (which agree except for one-cent rounding errors) and the table-generated EITC amounts from using the IRS instructions.  But given that neither group of model developers wants to adopt the lookup-table approach, these small differences between the model-generated EITC amounts and the lookup-table amounts would seem to be acceptable.

@MattHJensen @feenberg @GoFroggyRun @Amy-Xu

PR #618, opened 2016-02-26
[IGNORE] add DO_JIT flag to turn on/off jitting behavior

talumbau:

Issue #292, opened 2015-06-23
Should the Tax-Calculator be 'runnable' without numba?


talumbau:
The benefits might be:

- easier debugging
- able to run in environments that have pandas and numpy, but not numba

After a minute or two of searching, it seems like we could support testing this case using the `env` feature of Travis CI. We would basically just have an extra build environment without numba, so that our test matrix would be: [python 2.7, python 3.4] X [with numba, without numba]

http://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix

talumbau:
It turns out that this may be pretty simple. Looking at this section of code:

https://github.com/OpenSourcePolicyCenter/Tax-Calculator/blob/master/taxcalc/decorators.py#L223

There is a boolean flag `do_jit` in one of the helper functions for `iterate_jit`. It is always True. Presumably I did this to debug issues in the development of `iterate_jit`. It seems likely the only thing one would have to do to enable this behavior is:
 - come up with a reasonable way on how to turn this off/on
 - modify the TravisCI testing matrix to test w/o jitting.

MattHJensen:
@martinholmer, would the enhancement outlined in this issue have made things easier in #467?

martinholmer:
@MattHJensen said:
> would the enhancement outlined in this issue have made things easier in #467?

Honestly, I don't know.  My mistake was to assign a value to records.FLPDYR instead of incrementing the array like ```records.FLPDYR += 4``` in order to get from 2009 to 2013 without extrapolations, blowups, or reweighting.  My error was caused, I guess, by my lack of experience with numpy arrays.  But the disappointing thing about my experience is that the Python error traceback gave me no clue about what I was doing wrong.  The traceback info did lead me to the educated guess that some variable in the AMTI function was miss-specified.  But that wasn't much help because there look to be as many as sixty arguments to the AMTI function.  Everything is fine now; but it was a frustrating experience.

MattHJensen:
It looked to me like there might have been more useful python traceback info if numba hadn't kicked in. If you were willing to replicate the problem, you could try running things with `do_jit` = False and compare the tracebacks. Posting them here would provide good insight about how much we should prioritize this enhancement.

feenberg:
On Wed, 18 Nov 2015, Martin Holmer wrote:

> 
> @MattHJensen said:
>
>       would the enhancement outlined in this issue have made things easier in
>       #467?
> 
[...]

PR #619, opened 2016-02-26
Add jitting switch and bug fixes

talumbau:


talumbau:
This pull request enables the capability to turn off jitting, so the code in `functions.py` will run as straight Python code. See discussion in #292. When I enabled this capability, it became clear that some errors that occur in the jitted code are silently passing through. So, this PR also fixes those errors. I have also created PR #618 so that anyone who desires to look at the errors in depth can do so, without the fixes I included as the second commit in this PR. The fixes to get everything working again are as follows:

 - I added additional values to a few parameter arrays in `current_law_policy.json`. This is because I discovered that when running certain tests, we have `MARS` values that go up to 6, and yet we are indexing parameter arrays that only go up to 4. In the current jitted code base in master, this means that we are referencing garbage data. 

- In `functions.py`, we are returning values that are never declared in the function definition. This is because the functions are jitted and numba is doing the following: it reads all of the variables in the Python code, then creates a small C-like function that declares these variables as the necessary type (with the default initialization of that variable type following the rules of the C language) and then this function is compiled. In pure Python, we end up running the functions through code paths where certain variables are not declared prior to the return statement. Therefore, I added variable declarations for certain functions with default values of 0, which is what those variables default to now anyway. This just makes explicit what is happening in the compiled code anyway.

This PR should not be merged as is, but rather I think someone with appropriate domain knowledge should look at my changes to `current_law_policy.json` and both input the proper additional values and the proper names for the columns (which I just call `mars_5` and `mars_6`). Or someone could just comment on this PR and I will update my code to reflect the correct names and values.

martinholmer:
T.J. @talumbau,
Why are you undoing the work in pull request #606, which was merged into the master branch six days ago?  There was a reason for doing the rearrangement of all the EITC parameters, but I don't see a reason for undoing that rearrangement.  What's going on?

@MattHJensen

talumbau:
Yep, that's an error, although I don't understand at this moment how I messed it up. I will fix it and either create a new PR or merge the proper changes in to this PR. If you or someone else could comment on the changes to `current_law_policy.json` though, that would be very much appreciated. Thanks!

martinholmer:
T.J. @talumbau,
I'm confused by your assertion that the MARS values run from 1 to 6.  Here is the description of MARS from the 2009 IRS-SOI PUF documentation:
```
24. MARS Marital (Filing) Status:
Aggregated Return ..................................0
Single .............................................1
Married filing a joint return or Widow(er) with
   dependent child (surviving spouse)...............2
Married filing separately ..........................3
Head of household ..................................4

NOTE: For returns sampled above a 0.07 percent rate, 
returns filed with a marital status of head of household 
claiming no dependents have been converted to single.
```

Perhaps the Tax-Calculator code needs to fixed in a different way that in this pull request.

@MattHJensen @feenberg

talumbau:
@martinholmer yes I thought it would be surprising to all. Therefore, I opened #618 so that folks could verify that this is true for themselves. Please sync with that branch, set `DO_JIT` to `False`, and attempt to run the tests. You will see the test failure.

talumbau:
I conclude from your comment, then, that if I am seeing those values for `MARS` in our test suite, we are then testing with bad data. Is that correct?

feenberg:
Years ago married separate was either 3 or 6 depending on whether the 
couple lived together.

dan

taxcalc/decorators.py
```diff
@@ -8,6 +8,8 @@
 import ast
 import toolz
 
+DO_JIT=False
```
talumbau: Oops this should be `True`.

talumbau:
closing in favor of #620 which has changes based off latest master.

PR #620, opened 2016-02-26
Switch for turning off numba jitting

talumbau:
- defaults to DO_JIT=True
- add additional variable declarations in functions.py
- expand paramter arrays in current_law_policy.json

### DIFF
--- a/taxcalc/current_law_policy.json
+++ b/taxcalc/current_law_policy.json
@@ -1189,15 +1189,15 @@
         "irs_ref": "",
         "notes": "In this model, qualification for the Additional Additional Child Tax Credit is the same as qualificaiton rules for the Child Tax Credit under current law. ",
         "start_year": 2013,
-        "col_var": ["single", "joint", "separate", "head of household"], 
+        "col_var": ["single", "joint", "separate", "head of household", "mars_5", "mars_6"],
         "row_label": ["2013",
                       "2014",
                       "2015"],
         "cpi_inflated": false,
         "col_label": "",
-        "value": [[150000, 300000, 150000, 200000],
-		  [150000, 300000, 150000, 200000],
-                  [150000, 300000, 150000, 200000]]   
+        "value": [[150000, 300000, 150000, 200000, 200000, 200000],
+		          [150000, 300000, 150000, 200000, 200000, 200000],
+                  [150000, 300000, 150000, 200000, 200000, 200000]]
     },
 
     "_CTC_additional_prt":{
@@ -1528,8 +1528,8 @@
         "row_var": "FLPDYR",
         "row_label": ["2013"],
         "cpi_inflated": true,
-        "col_label": ["single", "joint", "separate", "head of household"],
-        "value": [ [0,           0,        0,             0]]
+        "col_label": ["single", "joint", "separate", "head of household", "mars_5", "mars_6"],
+        "value": [ [0,           0,        0,             0, 0, 0]]
     },
 
     "_II_credit_ps":{
--- a/taxcalc/decorators.py
+++ b/taxcalc/decorators.py
@@ -8,6 +8,8 @@ from six import StringIO
 import ast
 import toolz
 
+DO_JIT=True
+
 
 class GetReturnNode(ast.NodeVisitor):
     """
@@ -189,7 +191,7 @@ def create_toplevel_function_string(args_out, args_in, pm_or_pf,
     return s.getvalue()
 
 
-def make_apply_function(func, out_args, in_args, parameters, do_jit=True,
+def make_apply_function(func, out_args, in_args, parameters, do_jit=DO_JIT,
                         **kwargs):
     """
     Takes a '_calc' function and creates the necessary Python code for an
@@ -214,7 +216,10 @@ def make_apply_function(func, out_args, in_args, parameters, do_jit=True,
     '_apply' style function
 
     """
-    jitted_f = jit(**kwargs)(func)
+    if do_jit:
+        jitted_f = jit(**kwargs)(func)
+    else:
+        jitted_f = func
     apfunc = create_apply_function_string(out_args, in_args, parameters)
 
     func_code = compile(apfunc, "<string>", "exec")
@@ -226,7 +231,7 @@ def make_apply_function(func, out_args, in_args, parameters, do_jit=True,
         return fakeglobals['ap_func']
 
 
-def apply_jit(dtype_sig_out, dtype_sig_in, parameters=None, **kwargs):
+def apply_jit(dtype_sig_out, dtype_sig_in, parameters=None, do_jit=DO_JIT, **kwargs):
     """
     make a decorator that takes in a _calc-style function, handle
     the apply step
@@ -236,7 +241,11 @@ def apply_jit(dtype_sig_out, dtype_sig_in, parameters=None, **kwargs):
 
     def make_wrapper(func):
         theargs = inspect.getargspec(func).args
-        jitted_f = jit(**kwargs)(func)
+        if do_jit:
+            jitted_f = jit(**kwargs)(func)
+        else:
+            jitted_f = func
+
 
         jitted_apply = make_apply_function(func, dtype_sig_out,
                                            dtype_sig_in, parameters,
@@ -325,7 +334,7 @@ def iterate_jit(parameters=None, **kwargs):
                                                list(reversed(all_out_args)),
                                                in_args,
                                                parameters=all_parameters,
-                                               do_jit=True,
+                                               do_jit=DO_JIT,
                                                **kwargs_for_jit)
 
         def wrapper(*args, **kwargs):
--- a/taxcalc/functions.py
+++ b/taxcalc/functions.py
@@ -484,6 +484,7 @@ def StdDed(DSI, _earned, STD, p04470, e00100, e60000,
         c15100 = max(350 + _earned, STD[6])
         c04100 = min(STD[MARS - 1], c15100)
     else:
+        c15100 = 0.
         if MIDR == 1:
             c04100 = 0.
         else:
@@ -946,6 +947,7 @@ def AMTI(c60000, _exact, e60290, _posagi, e07300, c24517,
 
     if AGERANGE == 1 or (puf and (DSI == 1)):
         _ages = 0.
+        _agep = 0.
     else:
         if DOBYR >= 1 and DOBYR <= 99:
             _DOBYR = DOBYR + 1900.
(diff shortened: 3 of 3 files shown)


talumbau:
The codecov report is correct, in that now we have a code path that is not covered by TravisCI, which is the path of running the calculator with the jit code off. It seems valuable to run in this way, because doing so revealed problems in our test suite (undeclared variables returned in `functions.py`, and `MARS` values that exceeded size of parameter arrays).

martinholmer:
T.J. @talumbau,
It is great that you have added in pull request #620 the capability to execute the Tax-Calculator without using the Numba JIT speed-up capabilities.  We would never use this option in production, but as your recent experience shows, it is extremely helpful in debugging the Tax-Calculator code.

As far as I can tell, when DO_JIT is set to False in the decorators.py file and we run py.test, we get errors of two types.  The first type is that we are returning a calculated variable (for example, c15100) without it being initialized in a function in the functions.py file.  And the second type is that the code is trying to use MARS-based index values greater than 3 (that is, MARS values of 5 or 6).

In this pull request you have fixed these problems so that the py.test suite executes without any errors.

But I have my doubts about whether the fixes in this pull request are the most appropriate way to change the code.  I could be wrong about these doubts, but I think we should take our time to consider the best was to proceed now that we can run Tax-Calculator as a simple Python script.

I'm still thinking about the first type of error and I will post a comment here about that issue in the future.

You fixed the second type of error by expanding in the current_law_policy.json file, some policy parameters that are indexed by MARS, making the parameter arrays be indexed from 0 to 5 rather than from 0 to 3.  I doubt that is the correct approach because the puf.csv file we are using does not have any MARS values greater than 4, and therefore, the 0 to 3 indexing with MARS-1 values should be OK.  Here is a tabulation of the puf.csv file that shows we have no MARS values of 5 or 6.
```
$ awk -F, 'NR==1{for(i=1;i<=NF;i++)if($i~/MARS/)print i,$i}' puf.csv
27 MARS
$ awk -F, 'NR>1{t++;n[$27]++}END{print t;for(i in n)print i,n[i]}' puf.csv
219814
4 20495
1 85765
2 109265
3 4289
$ 
```

As you can see, MARS has values only in the 1-4 range with those counts add up to the 219,814 filing units in the puf.csv file.

My hunch is that we can eliminate the second type of errors by fixing the functions.py code to eliminate hard-coded access to MARS-indexed arrays that use indexes greater than 3.  Here is an example of that kind of problem in the functions.py file:
```
    # Calculate deduction for dependents
    if DSI == 1:
        c15100 = max(350 + _earned, STD[6])
        c04100 = min(STD[MARS - 1], c15100)
```

I have no idea what is going on in this code fragment, but the use of ```STD[6]``` seems wrong because the 2009 IRS PUF does not contain MARS values greater than 4.  So my hunch is that turing off the JIT business is revealing problems in the code that, for some reason, have been obscured by the JIT business.

I'm going to continue my investigations, but any comments or suggestion would be welcome.

@talumbau @MattHJensen @feenberg @GoFroggyRun @Amy-Xu

martinholmer:
Martin said:
> [T.J.] fixed the second type of error by expanding in the current_law_policy.json file, some policy parameters that are indexed by MARS, making the parameter arrays be indexed from 0 to 5 rather than from 0 to 3.  I doubt that is the correct approach because the puf.csv file we are using does not have any MARS values greater than 4, and therefore, the 0 to 3 indexing with MARS-1 values should be OK. Here is a tabulation of the puf.csv file that shows we have no MARS values of 5 or 6.
```
$ awk -F, 'NR==1{for(i=1;i<=NF;i++)if($i~/MARS/)print i,$i}' puf.csv
27 MARS
$ awk -F, 'NR>1{t++;n[$27]++}END{print t;for(i in n)print i,n[i]}' puf.csv
219814
4 20495
1 85765
2 109265
3 4289
$ 
```
>
> As you can see, MARS has values only in the 1-4 range with those counts add up to the 219,814 filing units in the puf.csv file.

But further inquiry shows that the ancient 1991 data used in the pytest suite does have MARS values greater than four.  Here are the tabulation results:
```
$ awk -F, 'NR==1{for(i=1;i<=NF;i++)if($i~/MARS/)print i,$i}' puf91taxdata.csv
14 MARS
$ awk -F, 'NR>1{t++;n[$14]++}END{print t;for(i in n){c+=n[i];print i,n[i]};print c}' puf91taxdata.csv
8799
4 299
5 3
6 87
1 3198
2 4705
3 507
8799
$
```

So, an alternative strategy to eliminating the second type of problem occurring when DO_JIT=False is to simply fix the ancient 1991 data to be consistent with more modern IRS PUF data.

MattHJensen:
@feenberg, do the latest years of the IRS full file still use all six values of MARS, or have they gone to four?

martinholmer:
Matt asked Dan:
> @feenberg, do the latest years of the IRS full file still use all six values of MARS, or have they gone to four?

From 2007 through 2010 (the years I checked the documentation), the IRS-SOI PUF has MARS values less than equal to four.  However, the full SOI sample (which as far as I know is not publicly available) for those year does have filing units with values of 5 or 6.

@MattHJensen @feenberg

MattHJensen:
@martinholmer said 

> From 2007 through 2010 (the years I checked the documentation), the IRS-SOI PUF has MARS values less than or equal to four. However, the full SOI sample (which as far as I know is not publicly available) for those year does have filing units with values of 5 or 6.

Thanks. 

>So, an alternative strategy to eliminating the second type of problem occurring when DO_JIT=False is to simply fix the ancient 1991 data to be consistent with more modern IRS PUF data.

@martinholmer, are you also suggesting that we stop maintaining the capability to handle MARS[0,5] as a general matter?

Several SOI economists have expressed interest in running the python calculator on the full file. When we ask JCT or CBO to cross-validate models, that will also give us an opportunity to use MARS[0,5]. I believe we also use MARS[0,5] when we test against taxcalc (SAS).  It seems worthwhile to maintain this capability for those purposes, so long as the full SOI sample still uses MARS[0,5] from 2013 on.

martinholmer:
@MattHJensen said:
> Several SOI economists have expressed interest in running the Python calculator on the full file. When we ask JCT or CBO to cross-validate models, that will also give us an opportunity to use MARS[0,5].  I believe we also use MARS[0,5] when we test against taxcalc (SAS).  It seems worthwhile to maintain this capability for those purposes, so long as the full SOI sample still uses MARS[0,5] from 2013 on.

I can see part of what you're saying (the part about SOI economists wanting capabilities beyond what their own PUF file supports), but where do we get the full SOI sample file to, say, "test against taxcalc (SAS)"?  Are you saying the Sean has "the full file" on his computer?  If not, how will we "cross-validate models" with JCT or CBO?  As you can see, I don't understand.

feenberg:
On Sat, 27 Feb 2016, Matt Jensen wrote:

> 
> @feenberg, do the latest years of the IRS full file still use all six values
> of MARS, or have they gone to four?

It has been 1-4 since 2000, at least. I don't think we have to consider 
other values.

dan

feenberg:
On Sat, 27 Feb 2016, Martin Holmer wrote:

> 
> Matt asked Dan:
>
>       @feenberg, do the latest years of the IRS full file still use
>       all six values of MARS, or have they gone to four?
> 
> From 2007 through 2010 (the years I checked the documentation), the IRS-SOI
> PUF has MARS values less than equal to four. However, the full SOI sample
> (which as far as I know is not publicly available) for those year does have
> filing units with values of 5 or 6.

Yes, I wasn't thinking about that. Matt is right, we should take into 
consideration the full file. I can check on the meaning of 5 on Tuesday, 
but 6 is the same as 3.

dan

feenberg:
On Sat, 27 Feb 2016, Martin Holmer wrote:

> 
> @MattHJensen said:
>
>       Several SOI economists have expressed interest in running the
>       Python calculator on the full file. When we ask JCT or CBO to
>       cross-validate models, that will also give us an opportunity to
>       use MARS[0,5]. I believe we also use MARS[0,5] when we test
>       against taxcalc (SAS). It seems worthwhile to maintain this
>       capability for those purposes, so long as the full SOI sample
>       still uses MARS[0,5] from 2013 on.
> 
> I can see part of what you're saying (the part about SOI economists wanting
> capabilities beyond what their own PUF file supports), but where do we get
> the full SOI sample file to, say, "test against taxcalc (SAS)"? Are you
> saying the Sean has "the full file" on his computer? If not, how will we
> "cross-validate models" with JCT or CBO? As you can see, I don't understand.

The only way I can see to test the Python calculator is to take a million 
records or so and compare with the SAS calculator. Since the SAS 
calculator has been tested against the INSOLE file, (by transitivity) the 
Python calculator could be confidently said to be tested also. But before 
that is done, let me send the most recent version of the SAS calculator, 
which I continue to work on. But it is very good now. For the first time 
it works for non-itemizing AMT taxpayers with itemized deductions.

dan

martinholmer:
@MattHJensen and Dan @feenberg,

OK, we'll go with MARS-indexed policy parameters having **six** array elements.  If that will please the SOI economists that is reason enough to have MARS values of 5 and 6.  MARS=5 is WIDOW and MARS=6 is SEPARATE, which (as Dan pointed out above) is equivalent to MARS=3.

In order to implement this decision, I need to know more about the Lee-Rubio reform.   Is the following (which is different from what T.J. put in pull request #620) correct about Lee-Rubio?
```
    "_CTC_additional_ps": {
        "long_name": "Additional additional child tax credit phaseout starting AGI (Lee-Rubio)",
        "description": "The amount of additional additional child tax credit is reduced for taxpayers with AGI higher than this level. ",
        "irs_ref": "",
        "notes": "In this model, qualification for the Additional Additional Child Tax Credit is the same as qualification rules for the Child Tax Credit under current law. ",
        "start_year": 2013,
        "col_var": ["single", "joint", "separate", "head of household", "widow", "separate"],
        "row_label": ["2013",
                      "2014",
                      "2015"],
        "cpi_inflated": false,
        "col_label": "",
        "value": [[150000, 300000, 150000, 200000, 150000, 150000],
                  [150000, 300000, 150000, 200000, 150000, 150000],
                  [150000, 300000, 150000, 200000, 150000, 150000]]
    },
```

In other words, in the Lee-Rubio reform proposal, is WIDOW given the same CTC_additional_ps parameter value as SINGLE?  Or should the reform value for WIDOW be equal to some other value?

MattHJensen:
> In other words, in the Lee-Rubio reform proposal, is WIDOW given the same CTC_additional_ps parameter value as SINGLE? Or should the reform value for WIDOW be equal to some other value?

Let's give WIDOW the same value as JOINT. That seems much more typical than giving WIDOW the same value as SINGLE.

martinholmer:
With only minor differences, the content of pull request #620 has been merged into the master branch via pull request #621.  So, pull request #620 is being closed.

Thanks again, T.J., for this valuable debugging option.

@talumbau @MattHJensen @feenberg @Amy-Xu

martinholmer:
Closing #620.

PR #621, opened 2016-02-28
Alternative follow-up to adding switch for Numba JIT logic

martinholmer:
This pull request is, for the most part, logically equivalent to pull request #620.  Many thanks to T.J. for adding in pull request #620 the ability to run the Tax-Calculator without the JIT speed-up logic.  The changes in this pull request to the decorators.py and functions.py files are the same as in #620, except that several unused local variables and function arguments in the decorators.py file have been removed.  Two MARS-indexed arrays in the current_law_policy.json file have been widened, just as in pull request #620, except that the values of one reform parameter are different for MARS=5 and MARS=6.

Also, this pull request includes a number of cosmetic changes in the current_law_policy.json file in order to make its format more consistent and to make the file more readable.  These changes mean that this json file meets the same PEP8 formatting standards (apart from long lines) that are imposed on the Python source code.  So, after these changes, the following command generates no PEP8 warnings:
```
pep8 --ignore=E501 current_law_policy.json
```

The changes in this pull request have been tested with DO_JIT=False and DO_JIT=True.  In both cases all the tests (including the ```requires_pufcsv``` tests) pass.  The only difference between the two DO_JIT values is the py.test execution time:
```
DO_JIT = True   --->    194 seconds
DO_JIT = False  --->   3348 seconds
```

So, the pytest suite (including the two ```requires_pufcsv``` tests) executes roughly 17 times faster when DO_JIT is true.

@talumbau @MattHJensen @GoFroggyRun @Amy-Xu @GoFroggyRun


martinholmer:
The slight drop in the code coverage statistic is caused by the same reasons that pull request #620 experienced a slight decline in code coverage.  This is not a problem because the all the tests in the py.test suite pass both when DO_JIT=True and when DO_JIT=False.

martinholmer:
Matt said:
> Let's give WIDOW the same value as JOINT [for Lee-Rubio reform parameter CTC_additional_ps]. That seems much more typical than giving WIDOW the same value as SINGLE.

I've made this change to pull request #621.  So, T.J.'s JIT-switch enhancement is now ready to be implemented by merging pull request #621 into the master branch.

@MattHJensen @feenberg @talumbau @Amy-Xu @GoFroggyRun

MattHJensen:
Thanks @martinholmer!

PR #626, opened 2016-03-01
User version info from taxcalc package for RTD version info

talumbau:


talumbau:
This makes it so that the latest tagged version in master will be displayed on the documentation. We grab the major/minor version from the latest version string (which might be something like `0.6-63-gbcc4564`) and use that for version and release info.

talumbau:
merging this will fix the current bad version info on taxcalc.readthedocs.org:

screen shot 2016-03-01 at 2 50 35 pm


cc @MattHJensen @martinholmer

talumbau:
merging so RTD docs get updated.

PR #628, opened 2016-03-02
Enhancement of `increment year`

GoFroggyRun:
This PR adds an optional way of doing increment year. Previously, increment year could be achieved by 
```
for i in range(4):
    calc.increment_year()
assert calc.current_year == 2017
```

Right now, alternatively, we could use
```
calc.year_increment(2017)
```
to specify the desired tax year. Iteration and assertion will be automatically made. 

Previous `increment_year()` option was still retained, as it was excessively used in testing suite.

@Amy-Xu @MattHJensen @martinholmer 
Please take a look, any comments, concerns or remarks would be appreciated.

### DIFF
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -160,6 +160,27 @@ class Calculator(object):
         self.policy.set_year(self.policy.current_year + 1)
         self.behavior.set_year(self.policy.current_year)
 
+    def year_increment(self, new_current_year):
+        iteration = new_current_year - self.policy.current_year
+        for i in range(iteration):
+            if self.growth.factor_adjustment != 0:
+                if not np.array_equal(self.growth._factor_target,
+                                      Growth.REAL_GDP_GROWTH_RATES):
+                    msg = "adjustment and target factor \
+                        cannot be non-zero at the same time"
+                    raise ValueError(msg)
+                else:
+                    adjustment(self, self.growth.factor_adjustment,
+                               self.policy.current_year + 1)
+            elif not np.array_equal(self.growth._factor_target,
+                                    Growth.REAL_GDP_GROWTH_RATES):
+                target(self, self.growth._factor_target,
+                       self.policy.current_year + 1)
+            self.records.increment_year()
+            self.policy.set_year(self.policy.current_year + 1)
+            self.behavior.set_year(self.policy.current_year)
+        assert self.policy.current_year == new_current_year
+
     @property
     def current_year(self):
         return self.policy.current_year
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -160,19 +160,19 @@ class Calculator(object):
         self.policy.set_year(self.policy.current_year + 1)
         self.behavior.set_year(self.policy.current_year)
 
-    def year_increment(self, advance_to_year):
+    def advance_to_year(self, year):
         '''
         The year_increment function gives an optional way of implementing
         increment year functionality by immediately specifying the year
         as input. New year must be at least the current year.
         '''
-        iteration = advance_to_year - self.records.current_year
+        iteration = year - self.records.current_year
         if iteration < 0:
             raise ValueError("New current year must be " +
                              "greater than current year!")
         for i in range(iteration):
             self.increment_year()
-        assert self.records.current_year == advance_to_year
+        assert self.records.current_year == year
 
     @property
     def current_year(self):
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -160,11 +160,35 @@ class Calculator(object):
         self.policy.set_year(self.policy.current_year + 1)
         self.behavior.set_year(self.policy.current_year)
 
-    def year_increment(self, new_current_year):
-        iteration = new_current_year - self.policy.current_year
-        for i in range(iteration):
-            self.increment_year()
-        assert self.policy.current_year == new_current_year
+    def year_increment(self, advance_to_year):
+        '''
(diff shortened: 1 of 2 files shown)


taxcalc/calculate.py
```diff
@@ -160,6 +160,12 @@ def increment_year(self):
         self.policy.set_year(self.policy.current_year + 1)
         self.behavior.set_year(self.policy.current_year)
 
+    def year_increment(self, new_current_year):
+        iteration = new_current_year - self.policy.current_year
```
talumbau: I think this function needs:
- a doc string on what it does
- a check that `iteration` is non-zero, with an exception if `new_current_year` is less than `self.policy.current_year`

The PR also needs:

 - a test on expected behavior in the regular case (properly advancing the year)
 - a test to make sure the function raises an exception when the year is less than the current year

I would also suggest a better name, such as `advance_to_year`, which indicates that the argument is a year.

GoFroggyRun:
@talumbau:
Thank you so much for your thoughtful comments. Please take a look at my recent improvement. 

Attached is an example of the output: 
screen shot 2016-03-02 at 5 21 33 pm

talumbau:
I think you were closer in the second commit.

```
def advance_to_year(self, new_current_year):
   ''' Provides an optional way of advancing to a new year.
      New year must be at least the current year.
   '''
    iteration = new_current_year - self.policy.current_year
    if iteration < 0:
        raise ValueError("new current year must be greater than current year!")
    for i in range(iteration):
        self.increment_year()
    assert self.records.current_year == advance_to_year
```

Then also add new tests in taxcalc/tests/test_calculate.py. Unless you add tests that cover both the exception case and the non-exception case, codecov will reject because you are adding new code without tests that test the code.

taxcalc/calculate.py
```diff
@@ -160,6 +160,36 @@ def increment_year(self):
         self.policy.set_year(self.policy.current_year + 1)
         self.behavior.set_year(self.policy.current_year)
 
+    def year_increment(self, advance_to_year):
+        '''
+        The year_increment function gives an optional way of implementing
+        increment year functionality by immediately specifying the year
+        as input. Robustness has been added to ensure input is of good shape.
+        '''
+        if isinstance(advance_to_year, int):
```
talumbau: In Python, we typically try to avoid type checking unless we really need to. Here I think it is not necessary.

taxcalc/calculate.py
```diff
@@ -160,6 +160,36 @@ def increment_year(self):
...
+                    for i in range(iteration):
+                        self.increment_year()
+                    assert self.records.current_year == advance_to_year
+                    print("Your data have been further extrapolated to " +
+                          str(self.records.current_year) + ".")
+            elif iteration == 0:
+                print("You are already in " +
+                      str(self.records.current_year) + ".")
+            else:
+                print(str(self.records.current_year) + " is the earliest " +
```
talumbau: We should raise (exceptions)[https://docs.python.org/2/tutorial/errors.html] on bad input. I would suggest a ValueError.

GoFroggyRun:
@talumbau:
Sorry for the multiple commits. What do you think of this PR now?

talumbau:
Your change to `test_calculate.py` actually removes an assert statement, so we definitely need that back in. A bigger point though is that you shouldn't feel the need to change an existing test for this new method. A good rule of thumb would be 'new feature, new test'. This feature is very small, so the test can be extremely simple. The new tests are really just an assertion of this new behavior you are adding. The idea behind the tests should be something like 'if someone were to remove or break this functionality, how could they get a nice failure to tell them what happened?'

```
def test_Calculator_year_increment():
    policy = Policy()
    puf = Records(data=TAXDATA, weights=WEIGHTS, start_year=2009)
    calc = Calculator(policy=policy, records=puf)
    calc.year_increment(2015)
    assert calc.current_year == 2015
    with pytest.raises(ValueError):
        calc.year_increment(2014)
```

Now if someone were to break either of the aspects of this method, this test would fail. I was curious as to how codecov could give a 'pass' to the PR when you had not tested the line that raises an exception. It turns out that it only reports coverage down to the hundredths of a percent, so your PR just barely kept the test coverage from going down enough to trigger the alarm:

Current test coverage: 1469 lines covered, 217 lines missed
>>> 1469./(1469 + 217)
0.8712930011862396

This PR (without the test above): 1475 lines covered, 218 lines missed
>>> 1475./(1475 + 218)
0.8712344949793266

If you add my test above, you won't increase the 'missed' lines at all.

martinholmer:
T.J. said:
> I would also suggest a better name, such as advance_to_year, which indicates that the argument is a year.

Sean,  You absolutely need to take T.J.'s advice here.  Naming the new method ```year_increment``` is unacceptable because the English word increment has a strong connotation of `adding one`.  Why don't you take T.J.'s advice and name the new method ```advance_to_year```?

cc @GoFroggyRun @talumbau @MattHJensen @Amy-Xu

GoFroggyRun:
@talumbau:
My apologies. I might have mis-interpreted your suggestion regarding `advance_to_year`. Sorry about that. 
I have adopted your comments regarding test suites, by keeping the original test for `increment_year` while adding an extra simple test for the new `advance_to_year` function. 
Please take a look again, and let me know if there's any improvements to make.

talumbau:
Looks good. Test coverage went up since every new line added is tested. +1 from me.

martinholmer:
T.J. said:
> Looks good. Test coverage went up since every new line added is tested. +1 from me.

I agree and will merge this pull request immediately.

PR #629, opened 2016-03-04
Step 10

ahunter1:
Added command line for Windows users in step 10

### DIFF
--- a/docs/source/contributor_guide.rst
+++ b/docs/source/contributor_guide.rst
@@ -80,7 +80,7 @@ Setup Git
 
        tax-calculator$ source activate taxcalc-dev
 
-If you are working on Windows, use the following the command line:
+    If you are working on Windows, use the following from the command line:
 
     .. code-block:: python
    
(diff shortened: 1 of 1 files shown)


talumbau:
yep, good catch. thanks!

docs/source/contributor_guide.rst
```diff
@@ -80,6 +80,12 @@ Setup Git
 
        tax-calculator$ source activate taxcalc-dev
 
+If you are working on Windows, use the following the command line:
```
MattHJensen: Is this missing a "from"? "from the command line"? 

Also, when I view the file [here](https://github.com/ahunter1/Tax-Calculator/blob/patch-1/docs/source/contributor_guide.rst), I see that the formatting is a bit off. Try indenting the line starting "if you are working on windows..." and see if that fixes it.

You will do all this by just adding another commit or two locally to the same branch and then pushing to your fork. The PR will update automatically.

MattHJensen:
@talumbau said

> yep, good catch. thanks!

Agreed! Thanks @ahunter1 and welcome.

MattHJensen:
Looks great. Merging. Thanks @ahunter1!

Issue #627, opened 2016-03-02
Enhance transparency and replicability for users w/o proprietary data


MattHJensen:
The fact that tax-calculator relies on semi-proprietary data for many use-cases, namely the IRS public use file, gives us a wonderful opportunity to explore and implement solutions to the following question:

“For modeling outfits that rely on proprietary data, what supplementary information can they publish to enable replication and deep peer review?” 

After settling on the answer, I envision us creating a directory of these supplementary information whenever we bump the version of tax-calculator or TaxData, and then we would make that directory available via a link from the TaxBrain results page and the readmes of the repos. 

One could argue that we don’t need to publish all of this since our code is open source; even where that might be true, it would be very helpful for others who don’t have open source models to produce these outputs, and part of our mission is to set an example. 

Different classes of information that I have thought to include so far:
1. Results of the model, using the proprietary data, for baseline and numerous example reforms. 
   - This allows for a quick assessment of how the model works. 
   - @amy-xu has taken a big step towards showing reform examples with her work [here](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/comparison/reform_results.txt).
   - @martinholmer has done the same for the baseline with his work [here.](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/tests/pufcsv_agg_expect.txt)
2. Dummy datasets and model results for those datasets. 
   - This allows for users to test our calculator
   - @martinholmer and @gofroggyrun have made considerable progress on a dummy dataset and baseline results for for the 22 internet TAXSIM input variables and 28 intermediate variables. 
   - We could also make the modified 1991 puf available and provide baseline results for it for _all_ intermediate and final variables. @GoFroggyRun, could you look into this? 
   - We don’t currently generate results for all years in the budget window, and we don’t demonstrate how results vary with different behavioral assumptions. 
3. Basic summary statistics for every input and calculated variable, using the proprietary data.
   - This allows users to intuitively understand the extrapolation as well as to try to match the key characteristics of our baseline data in their own dataset. 
   - @amy-xu is working on this. 
4. A correlation matrix for all input data. 
   - This would help users to impute any variables they are missing in their own data. 
   - @amy-xu is working on this. 
5. A list of extrapolation targets and blowup factors for each variable in the dataset. Equations relevant to the extrapolation. 
   - This would help users match our extrapolated dataset if they can construct a dataset that looks like ours in the base year.
     - @Amy-Xu is working on this. 
[...]

PR #631, opened 2016-03-04
Create summary statistics for variables

Amy-Xu:
In this PR, I'm creating a script that generates summary statistics (issue #627, task 3) for the PUF variables used in Tax-Calculator, and 16 calculated variables mostly referenced in analysis. In long-run, we may want to include all input and output variables, but in short run, we are only to include the ones mentioned above for simplicity reasons. 

Currently in this version, only the average of each variable has been included. @MattHJensen proposed to also include 1) total, 2) number_zero, and 3) value for each percentile, in addition to average. I'm not experienced enough to tell what information would be necessary for someone else who may want to replicate our results without PUF. (Correlation matrix would be provided separately.) Would appreciate any suggestions on this issue. @feenberg @martinholmer 

Regarding the format, currently the output file is in CSV format (I think the preview on Github looks better) and it will be subject to change. Would love to hear any feedback as well, as it might get a bit tricker to put together a few more stats for 10 years.

A side note: one input file `e_variable_info.csv` is currently [here](https://docs.google.com/spreadsheets/d/1WlgbgEAMwhjMI8s9eG117bBEKFioXUY0aUTfKwHwXdA/edit#gid=1029315862). We want to move it to our Tax-Calculator repo, and clean it up, which will be discussed in a separated issue.


martinholmer:
Amy,

I have two questions about pull request #631:

(1) Why are you creating the e_variable_info.csv file when we already have a var_label.txt file that is supposed to be the documentation on the input variables used by Tax-Calculator?  Matt and I have been discussing the var_list.txt file for months and he refused to follow my suggestion of deleting it.  So, now we are adding another file with similar information.  I don't understand our long-term strategy here.  Can you or Matt explain?

(2) Why are you not using Records.VALID_READ_VARS and Records.UNUSED_READ_VARS?  It is bad programming practice to copy and paste key data structures because as soon as they changes they need to be updated in two or more places.  That is a time consuming task that is very error prone (because people tend to forget where data structures were copied).

If these questions are off-base, it is because there is no clear statement of what you are trying to accomplish with all this work.

@Amy-Xu @MattHJensen

Amy-Xu:
Martin,

Sorry I wasn't clear in my original post. The point of this PR is to create `variable_stats_summary.csv`, instead of ` e_variable_info.csv`. Here's the reply to your questions. 

1) This file actually wasn't created by me (two interns: Xiyu and Maria) and it has been there for a few month as it includes more information than var_labels.txt. Currently it's saved in my google doc, as a [google spreadsheet](https://docs.google.com/spreadsheets/d/1WlgbgEAMwhjMI8s9eG117bBEKFioXUY0aUTfKwHwXdA/edit#gid=1029315862). If you take a look at this document, you can see we have already included two more columns, form reference and whether one variable is in PUF. These two pieces of information do not exist in var_labels.txt and are certainly useful for users who want to verify tax logic with IRS forms or to check non-PUF zero values or to use their own imputations. In future, if any contributor wants to update the variable descriptions to latest IRS forms, this spreadsheet would potentially save a lot time as well. So function wise, I won't consider this spreadsheet is completely redundant. However, I do think we need to integrate the first column of this spreadsheet and var_labels.txt, because the information is kind of repetitive. But it seems to me this integration is a separate issue from the stats added in this PR, so I opened issue #633 for further discussion. I know currently it doesn't look neat, but I don't think it's a super urgent issue either.

The reason why I want to move this spreadsheet to Tax-Calculator also partially because I need to import variable descriptions in accordance with 2014 form from somewhere. And this seems to be an easy and feasible way. If you have any other suggestion, I would love to know more about it for sure.

2) My bad on this, I'll change them in my next commit.


Last but not least, my understanding of this task is just to give our users, who don't have PUF, as much information as we're allowed, so they could replicate our results. It would also facilitate data wise comparisons with other agencies who own similar extrapolated datasets. I believe @MattHJensen is able to explain clearer on this.

Amy-Xu:
@martinholmer I think this PR is ready now. Could you review? Thanks!

Amy-Xu:
Ahh sorry I haven't replaced the variable list names yet. Will do in a sec.

MattHJensen:
This looks good to me. Obviously we need to think harder about ` e_variable_info.csv` and our other documentation of input variables, but I am hoping to have this functionality into master by tomorrow so I will merge later today unless anyone raises major objections. My thinking is that we can then go back and refactor to deal with the e-code documentation issue.

MattHJensen:
@Amy-Xu should there be some test coverage for the new script?

Amy-Xu:
I don't think so, but I'm not sure. 

But as you have already merged the correlation PR, we can close this one because that PR contains everything here. @MattHJensen

MattHJensen:
> But as you have already merged the correlation PR, we can close this one because that PR contains everything here. @MattHJensen

Ok, closing. 

> I don't think so, but I'm not sure.

Why not? How do we know if we are breaking these new capabilities?

PR #640, opened 2016-03-11
Correlation matrix for 2016 added

Amy-Xu:
In this PR, 

1) Refactored Stats_Summary.py so output type can be added as an argument
2) Added correlation matrix as an output type. Currently it would be a correlation matrix for 2016. Usage: python Stats_Summary.py --output correlation

This PR is based on #631. Once that PR gets merged, the diff for this PR will be clearer.

@martinholmer Sorry I haven't been able to wait for your feedback on #631 and open this PR already. In general, this is a part of our effort enhancing data transparency. If you have more questions on why we're doing this, @MattHJensen will have a much better explanation.

Could you review? @martinholmer @MattHJensen


martinholmer:
@Amy-Xu,  There is a very strong tradition in Python of naming scripts in lower case letters with underscores to connect words.  This is such an established convention that all the scripts in the Tax-Calculator repo are named this way.  And this naming scheme is widely followed in the international Python community, so much so that your new script --- Stats_Summary.py --- will trigger a pylint error.

Why don't you change its name to ```stats_summary.py```?  Is there any problem with the lower-case version?

@MattHJensen @talumbau @GoFroggyRun

Amy-Xu:
@martinholmer just changed the name to lower case.

martinholmer:
Amy said:
just changed the [script] name to lower case

Thanks.

MattHJensen:
My comment on #631 applies here as well:

> This looks good to me. Obviously we need to think harder about e_variable_info.csv and our other documentation of input variables, but I am hoping to have this functionality into master by tomorrow so I will merge later today unless anyone raises major objections. My thinking is that we can then go back and refactor to deal with the e-code documentation issue. cc @Amy-Xu and @martinholmer

PR #656, opened 2016-03-21
Updated WEIGHTS.csv for the new version of CPS-matched PUF

Amy-Xu:
This new dataset has more records than the previous one, so I updated the WEIGHTS.csv by running the extrapolation with the same set of factors and targets. I also updated relevant files in comparison and test folders.

All the reform results still look pretty much the same, but it seems to me that the average of capital gain and AMT variable changed quite a bit. Not sure whether those changes make sense or not.

@MattHJensen @martinholmer Any suggestions?


Amy-Xu:
This PR is done but should not be merged until the time deemed as proper. @MattHJensen

MattHJensen:
@Amy-Xu, could you remind us of the status of this PR?

Amy-Xu:
This was a update of CPS-matched 09 PUF, from March 14 CPS to March 15 CPS. As mentioned in the previous post, the reform results seems fine but the AMT and capital gain were low. When this PR's up, it was one week before WSJ editorial, so we decided to spend a bit more time digging into AMT and capital and postponed updating the dataset. 

Should we close this PR as we're already had 2010 PUF?

MattHJensen:
> Should we close this PR as we're already had 2010 PUF?

That makes sense to me. Closing.

PR #662, opened 2016-03-24
Enter reform parameters in sorted order

talumbau:
this keeps the page from jumping around frequently when selenium enters
values into TaxBrain. This change prevents the unintentional 'double
click' of the _AMT_CG_thd2_cpi flag

### DIFF
--- a/taxcalc/taxbrain/reforms.py
+++ b/taxcalc/taxbrain/reforms.py
@@ -252,7 +252,7 @@ def taxbrain_param_values_insert(driver, start_year, reform_spec):
     or more year:value pairs, and the specified start_year.
     Function returns nothing.
     """
-    for param in reform_spec:
+    for param in sorted(reform_spec.keys()):
         if TRACING:
             sys.stdout.write('==> PARAM {}\n'.format(param))
             sys.stdout.flush()


talumbau:
@martinholmer if you re-run `all-16-lvlnoi.json` with this change to `reforms.py`, I believe you will get a different result from taxbrain/2057. I think you will get a result like http://www.ospc.org/taxbrain/2172/ which is what I just got. My hypothesis is that there is something that occurs as selenium jumps the page around from section to section that clicks the CPI flag button a second time. You can see this is true if you take the current reforms.py and put a breakpoint just before the clicking of the submit button. If you do this for the current reforms.py, you will see that `_AMT_CG_thd2_cpi` is back to the default value just before the submit. With this change, the window doesn't jump around as much, which seems to preserve the desired CPI settings flags.

martinholmer:
T.J.,  Thanks very much for this fix, which I guess works around some kind of shortcoming in the selenium package.  I guess I would have been able to see this problem if the forthcoming enhancements to TaxBrain had been available when I first started.  Thanks again for the help on this.

I'll merge this now, and then use the updated master branch to record the improvement in the test results.

@talumbau @MattHJensen @feenberg @Amy-Xu @GoFroggyRun

PR #664, opened 2016-03-24
Edit current_law_policy.json to use 'fraction' instead of 'percent'

martinholmer:
This pull request contains non-substantive changes to the ```current_law_policy.json``` file that are intended to clarify that Tax-Calculator uses decimal fraction parameter values (not percentage parameter values).


Amy-Xu:
This looks good to me!

Issue #671, opened 2016-03-29
Use a single site for documentation browsing


zrisher:
This is a minor issue, but I figured it would still be helpful to mention it.

As a starting contributor, it's a little confusing to go through the Contributor Guide on `readthedocs.org` and when you're finally done and ready to start using it, be redirected to `nbviewer.jupyter.org` for the usage guide. It makes you wonder how navigable the documentation is going to be and if you'll be able to find what you need.

As far as I can tell, the only reason for this is that most of the docs are in `.rst`, whereas a couple files are in `.ipynb`. What would be the argument against converting the files in `docs/notebooks` to `.rst` and moving them into `docs/source`?

MattHJensen:
Thanks @zrisher. I'd say that the biggest problem here is that the ipython notebook is way out of date. I am going to open a PR to update the notebook, and then we can talk about whether it would be better as a .rst. I suspect we'll want to keep it as a notebook since someone can download it and run all of the cells with their local version of tax-calculator. [Here is a gallery](https://github.com/ipython/ipython/wiki/A-gallery-of-interesting-IPython-Notebooks) of how some other projects use notebooks.

zrisher:
@MattHJensen Ahhhh ok, ipython notebook content can be directly loaded into code. Clearly I did not realize that, very cool! So yes, if it was updated to current use and perhaps if we amended the note referencing it to suggest that a bit more directly for users unfamiliar with ipython notebooks (I can handle that part), this would be fixed for me.

MattHJensen:
@zrisher ^^

zrisher:
I'm not sure the best way to say "you're going to a ipython notebook that's part of the repo. This file format can be easily imported into running code" succinctly, and I don't have a great feel for how necessary it is to be that specific for most of the python community. 

So I'm going to drop the requirement of "amended the note referencing it to suggest that a bit more directly for users unfamiliar with ipython notebooks". 

Will pick it back up at a later time when I'm more familiar with these if anyone else deems it useful.

PR #672, opened 2016-03-30
10 min

MattHJensen:
Addresses #671. 

The 10-min notebook is working now but extremely bare boned. It will need to be improved from here. 

cc @martinholmer, @zrisher, @Amy-Xu.


MattHJensen:
Merging as this fixes something that is broke and doesn't change calculator results.

PR #695, opened 2016-04-19
Add taxcalc/taxcalcsas/compare.sas script and a13 results

martinholmer:
This pull request adds a SAS script used to execute taxcalc.sas as part of the comparison of taxcalc.sas and Tax-Calculator results.  As described in the revised README.md file, using the 100,000 tax filing units in the a13.in sample as input produces from the two programs the same income tax results (apart from one-cent rounding-error differences) except for the fact the Tax-Calculator does not compute Schedule R tax credits.

This pull request does not change any tax-calculation logic or any py.test results or any validation test results.


martinholmer:
@MattHJensen and Dan @feenberg,
Was there a strategic decision early on in the project to not include in Tax-Calculator the Schedule R calculation of the non-refundable credit for the elderly?

The Schedule R calculations are done in the taxcalc.sas program, but not in Tax-Calculator.  Instead Tax-Calculator reads E07200 from the `puf.csv` file.  This means that Tax-Calculator does not produce accurate results when E07200 is not included among the input variables.

Issue #659, opened 2016-03-24
Numba should not be a dependency of `taxcalc`


talumbau:
With the merge of #621, it would be helpful if the conda package for taxcalc did not have in its set of dependencies the numba package. Most users don't install taxcalc via a conda package, so it likely makes little difference to users. Typically, they would create a conda environment based on the environment.yaml in this repo and then work directly with the source repository, locally installing via the `setup.py` if necessary. This workflow would not change. This change would be very helpful in automated deployments, especially on Heroku, where the size of the deployed code is limited. Without requiring numba, the `taxcalc` package is much smaller. This should be done in such a way that if taxcalc is run without numba, it will succeed, but give warnings that the run time will be long because numba is not installed. If it runs and numba is present, it will behave as it does today.

martinholmer:
T.J. said among other things:
> It would be helpful if the conda package for taxcalc did not have in its set of dependencies the numba package. [... explanation for why this would be very helpful ...]

So what would have to be done to the Tax-Calculator code in order to reach this desirable goal?

@talumbau

talumbau:
#713 is the PR that will resolve this issue. Needs at least another commit which I hope to do this week.

martinholmer:
T.J. said:
> #713 is the PR that will resolve this [#659] issue. Needs at least another commit which I hope to do this week.

But this seems to be a waste of your valuable time because all of the code that you're changing in #713 is soon moving out of `records.py` to the taxdata repository.  See discussion in issue #688.

@talumbau @MattHJensen @Amy-Xu

martinholmer:
Merged pull request #754 seems to have resolved this issue #659.
@talumbau and @MattHJensen, are there other aspects to this issue, or should issue #659 be closed?

martinholmer:
Martin Holmer said on May 27:
> Merged pull request #754 seems to have resolved this issue #659.
> @talumbau and @MattHJensen, are there other aspects to this issue, or should issue #659 be closed?

T.J.'s work in merged pull request #799 --- plus no comments on this issue over the past two months --- suggests that this issue #659 has been resolved.

PR #713, opened 2016-04-26
[WIP] Jit itemizer

talumbau:
This PR is a work in progress.

The first step has been completed, which is to replace the `@vectorize` implementation with an implementation that works with `@jit`. The next step is to provide a dummy implementation with `@jit` when it's not available (that is, numba is not installed). This is not too difficult. When this is done, `taxcalc` will have no hard dependency on numba, which is the goal of #659


martinholmer:
T.J. (@talumbau),
Now that pull request #754 has been merged into the master branch, I think the Tax-Calculator code does not depend on the numba package.  That is because the cmbtp_itemizer code has been moved out of the Records class into the taxdata repository.  This move of cmbtp-calc logic seems to have two implications:

(1) This pull request #713 is no longer necessary.  If I'm right about that, you should close #713.

(2) Some cleanup work needs to be done.  When I grep for numba (after #754 is merged), I get this:
```
tax-calculator$ grep -r --include="*py" numba .
./docs/source/conf.py:MOCK_MODULES = ['numba', 'numba.jit', 'numba.vectorize', 'numba.guvectorize']
./taxcalc/decorators.py:from numba import jit, vectorize, guvectorize
./taxcalc/decorators.py:    to guvectorized numba functions
./taxcalc/decorators.py:    to vectorized numba functions
./taxcalc/decorators.py:            # print ("This should only be seen in RTD, if not install numba!")
./taxcalc/tests/test_decorators.py:from numba import jit, vectorize, guvectorize
./taxcalc/utils.py:    A sanity check decorator. When combined with numba.vectorize
```
Is there a need for any of this now?  If not, do you want to do the cleanup?  If you want me to do the cleanup, please give me some guidance so I don't do anything stupid.

@MattHJensen

MattHJensen:
@martinholmer said:

> Now that pull request #754 has been merged into the master branch, I think the Tax-Calculator code does not depend on the numba package. 

@talumbau, is that your take as well? If so, does it mean this PR should be closed?

talumbau:
yes. this PR is no longer needed. Closing.

talumbau:
@martinholmer said:
> Is there a need for any of this now? If not, do you want to do the cleanup? If you want me to do the cleanup, please give me some guidance so I don't do anything stupid.

that code will still be needed. When there is no numba dependency, the code will just do a check for its existence and not use numba features if it's not there. Out of curiosity, do you know of a use case where someone is unable to use the Tax-Calculator right now because of its dependency on `numba`?

MattHJensen:
I have wondered whether it could be useful to use Tax-Calculator with `numba` turned off to get nicer error messages.

talumbau:
oh yes, that is a nice use case.

PR #687, opened 2016-04-11
Refining EIC Calculation Using Age Variable

GoFroggyRun:
This PR deals with EIC calculation. Previously, 'numextra' is used to determine elder units and exclude them from claiming EIC. Thanks to the 'age' variable in the CPS matched puf file, we are able to further refine the criteria that is used to determine whether a unit is eligible for EIC or not. To be specific, units under age 25 or over 65 are being excluded from EITC, where the age thresholds have also been parametrized to incorporate [related reform](https://www.whitehouse.gov/sites/default/files/docs/eitc_report.pdf).

Comparing our prediction against SOI 2013 data, we have:

1. EIC earned income amount of 497.9B against SOI’s 501.6B and
2. EIC amount of 67.5B  against SOI’s 68.1B. 

And here is the diagnostic table:
base

Moreover, the modification of EIC logic also improves the estimation of revenue changes resulted from EITC reform:
```
reform0 = {2016: {'_EITC_rt':[[0.153, 0.3400, 0.4000, 0.4500]]
                 ,'_EITC_prt':[[0.153, 0.1598, 0.2106, 0.2106]]
                 ,'_EITC_c':[[1005, 3359, 5548, 6242]]
                 ,'_EITC_ps':[[11500, 18150, 18150, 18150],]
                 }
          }
```
Previously, implementing this reform would result in revenue change of amount -$101 B (http://www.ospc.org/taxbrain/2355/), while in [C. Expand the EITC for Workers without Qualifying
Children](https://www.jct.gov/publications.html?func=startdown&id=4881), JCT estimates a revenue change of -$71.6 B. (Notice that JCT has, additionally, taken the change of age threshold into consideration, which means that if we further implement reform of age threshold, the difference would be even larger.) This patch narrows the gap: 
screen shot 2016-04-11 at 6 10 52 pm
Implementing the same reform now gives us a revenue change of - $55.4 B, which seems to be much more appropriate. 

Lastly, when we also incorporate the change of age threshold for EIC eligibility: 
```
reform1 = {2016: {'_EITC_rt':[[0.153, 0.3400, 0.4000, 0.4500]]
                 ,'_EITC_prt':[[0.153, 0.1598, 0.2106, 0.2106]]
                 ,'_EITC_c':[[1005, 3359, 5548, 6242]]
                 ,'_EITC_ps':[[11500, 18150, 18150, 18150],]
                  ,'_EITC_Age_flr':[21]
                  ,'_EITC_Age_cel':[67]
                 }
          }
``` 
[...]

PR #714, opened 2016-05-01
Added DSI Variable to Disqualify Certain Units from EITC

GoFroggyRun:
Per issues identified in #687 and #707 regarding EITC, this PR continues to refine the tax logic involves EITC. 

According to [09 Public Use Booklet](http://users.nber.org/~taxsim/gdb/gdb09.pdf), `DSI` variable indicates that whether a unit is claimed a dependent on another taxpayer (`DSI == 1`) or not (`DSI == 0`). Meanwhile, the [tax form](https://www.irs.gov/pub/irs-pdf/i1040gi.pdf) says that, with slight simplification of the real tax logic, if 'you are claimed as a dependent on someone else's 2015', then 'you' will be ineligible for EITC. 

In light of this, `DSI` variable has been added to the current tax logic to filter those units that are disqualify for EITC. There are some minor changes to the puf output (w.r.t both agg and mtr) and reform results. The consistency between tax-calc and TAXSIM, however, remain the same. 

@martinholmer @MattHJensen Please take a look. Comments, concerns or remarks would be appreciated.


martinholmer:
Sean,

The PEP8 tests for #714 failed because of a formatting problem I introduced in a recent change to the `simpletaxio.py` code.  Reduce the indention of line 536 and then you will be able to pass the PEP8 tests.  I'm not sure why Travis-CI did not flag this when I checked in this code change.  Here is the error message you are getting (and the one I should have gotten when I check in that code):
```
/home/travis/build/open-source-economics/Tax-Calculator/taxcalc/simpletaxio.py:536:32: E127 continuation line over-indented for visual indent
                               'illegal value')
                               ^
```

@GoFroggyRun

GoFroggyRun:
@martinholmer：

Thanks  for your instruction on resolving the PEP8 error. After getting rid of the indention, however, the Travis-DI still reports the same issue, and I'm not  quite sure why this is the case. Please advice, thanks!

martinholmer:
Sean said:
> Thanks for your instruction on resolving the PEP8 error. After getting rid of the [excess] indention, however, the Travis-CI still reports the same issue [...]

No,  it is not the same issue.  You did fix the PEP8 warning, but somehow the CodeCov program got confused about your simple change.  I think we can safely ignore its confusion.  I'll merge your pull request in a few minutes.

PR #719, opened 2016-05-03
Fix to better support policy reforms with 'None'

talumbau:
- For multi-valued policy parameters, it is sometimes desirable to only
  specify a few of the values of the reform for a given year. A user may
  specify 'None' for the other values, which indicates that current
  policy would be followed for the unspecified value for that year.

### DIFF
--- a/taxcalc/parameters_base.py
+++ b/taxcalc/parameters_base.py
@@ -432,9 +432,9 @@ class ParametersBase(object):
             for row in x:
                 keep_user_data_mask.append([1 if i != -1 else 0 for i in row])
                 keep_calc_data_mask.append([0 if i != -1 else 1 for i in row])
-                if not np.any(row == -1):
+                if not np.all(row == -1):
                     last_good_row += 1
-                else:
+                if np.any(row == -1):
                     has_nones = True
             if x.shape[0] >= num_years and not has_nones:
                 return x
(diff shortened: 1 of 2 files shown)


talumbau:
This use case is most often seen with TaxBrain users, who are filling out text fields with comma-separated values for future years for a particular filing status of a policy parameter.

martinholmer:
T.J. said:
> For multi-valued policy parameters, it is sometimes desirable to only specify a few of the values of the reform for a given year. A user may specify 'None' for the other values, which indicates that current policy would be followed for the unspecified value for that year.

Can you give an example of how this new capability would be used?  I looked at the new tests and they didn't answer my question, but maybe I didn't understand the new tests.  Can you help me understand this enhancement?

@talumbau

talumbau:
Sure. One example would be from TaxBrain. Consider a user who wants to modify the standard deduction on TaxBrain for the years 2016, 2017, and 2018 (you can't see the 2018 entries in the screenshot):

screen shot 2016-05-03 at 11 16 39 am

From `current_law_policy.json`, the latest specified values are for 2016:

```
[6300, 12600, 6300, 9300, 12600, 6300, 1050]
```

for the seven filing statuses. On TaxBrain, you can only specify values for the first four filing statuses. So, for the rest of the 2016 values, we can fill in current law. For the 2017, 2018, and 2019 values, the user has said nothing (in this case, because it's not possible in the interface). The translated reform dictionary looks like this:

```
{2016: {_STD':  [[12200, 24400, 12200, 17900, 12600.0, 6300.0, 1050.0],
                 [12400, 24800, 12400, 18000, None, None, None],
                 [12600, 25200, 12600, 18500, None, None, None]]}}
```

These values can then be filled in based on the CPI flag (potentially modified by the user) to create a full specification of the reform.

The tests in this PR are another example. In this case, the user is specifying a reform starting in 2016 with values specified fully up to 2018, but only specifying the value of `_II_brk2` for `single` filing status for 2019, leaving the rest of the values unspecified. This parameter is CPI-inflated, so for 2019 the parameter should be the combination of what the user specified (here, only the single filing status) and the rest of the parameter should be the CPI-inflated 2018 values.

So, this PR would allow both of these to work. (Note that the TaxBrain reform I specify above won't work right now due to this issue).

martinholmer:
T.J. said:
> This use case is most often seen with TaxBrain users, who are filling out text fields with comma-separated values for future years for a particular filing status of a policy parameter.

But `None` does not mean (to me) what you seem to be suggesting as the typical use of this enhancement.  To me, `None` means the parameter does not exist, but (as best I can tell) you mean leave that parameter at its current-law-policy values.  Unless I'm confused here, I think we need a word that better communicates the intended meaning than does `None`.

@talumbau

talumbau:
The placement of `None` is an expression of user intent. Just as the specification of a value for a reform parameter is an expression of the intent of the user, stating `None` says that there is no intent to change the value.

martinholmer:
T.J. said:
> The **translated reform dictionary** looks like this:
> ```
>    {2016: {_STD':  [[12200, 24400, 12200, 17900, 12600.0, 6300.0, 1050.0],
>                     [12400, 24800, 12400, 18000, None, None, None],
>                     [12600, 25200, 12600, 18500, None, None, None]]}}
> ```
> These values can then be filled in based on the CPI flag (potentially modified by the user) to create a full specification of the reform.

I'm fine with the above example because the TaxBrain user is not being asked to enter `None` in a text box.  But rather the `None` is something that TaxBrain is writing in the reform dictionary and your pull request enhances `parameters_base.py` to understand what to do when `None` is in the reform dictionary.  (Actually, I thought you had done that weeks ago.)

Then T.J. said:
> The tests in this PR are another example. In this case, the user is specifying a reform starting in 2016 with values specified fully up to 2018, but only specifying the value of _II_brk2 for single filing status for 2019, leaving the rest of the values unspecified. This parameter is CPI-inflated, so for 2019 the parameter should be the combination of what the user specified (here, only the single filing status) and the rest of the parameter should be the CPI-inflated 2018 values.
>
> So, this PR would allow both of these to work.

But you have not explained what the TaxBrain user is supposed to type into the text box in order to specify this second reform.  My main interest is in what you are proposing for that.  Certainly, I hope you are not proposing `None` because we already have a symbol for TaxBrain users to specify that the parameter should be indexed as usual (namely, the `*` character).  In your second example, what would a TaxBrain user enter to specify this reform?

@talumbau @MattHJensen

talumbau:
Ah I see. So the question is, how would a TaxBrain user specify this reform:

```
_II_brk2 = [[36000, 72250, 36500, 48600, 72500, 36250],
            [38000, 74000, 36900, 49400, 73800, 36900],
            [40000, 74900, 37450, 50200, 74900, 37450],
            [41000, None, None, None, None, None]]

```

For that case, it turns out that a TaxBrain user could not specify this exact reform, because a TaxBrain user can only enter data for the first four filing status categories. However, they can come close, and no new syntax is required. First, let's take the 2016 values from current law for `_II_brk2` for the six filing statuses:

```
[37650, 75300, 37650, 50400, 75300, 37650]
```

We'll have to use those values for the last two filing status categories that a TaxBrain user can't access. In order to get as close as possible to the above, the TaxBrain user would do as follows:

Enter: `36000, 38000, 40000, 41000` in the `Single` box
Enter: `72250, 74000, 74900` in the `Married filing jointly` box
Enter: `36500, 36900, 37450` in the `Married filing separately` box
Enter: `48600, 49400, 50200` in the `Head of Household` box


This would result in a reform that looks like this:

```
_II_brk2 = [[36000, 72250, 36500, 48600, 75300, 37650],
            [38000, 74000, 36900, 49400, None, None],
            [40000, 74900, 37450, 50200, None, None],
            [41000, None, None, None, None, None]]
```

Essentially, the situation only comes up when a TaxBrain user declines to enter the same number of additional years for each part of a multi-valued parameter. Does all of the above make sense?

martinholmer:
T.J. said after explaining how TaxBrain user would specify a "uneven" reform:
> Essentially, the situation only comes up when a TaxBrain user declines to enter the same number of additional years for each part of a multi-valued parameter. Does all of the above make sense?

Yes, this all makes sense.  Thanks for the explanation.  Should I merge this pull request #719 into the master branch?  Or, do others want to review it further?

@talumbau @MattHJensen

MattHJensen:
> Thanks for the explanation. Should I merge this pull request #719 into the master branch? 

LGTM too. Merging.

Issue #688, opened 2016-04-13
Can we move imputation logic from records.py to taxdata repo?


martinholmer:
Recent discussion regarding issue #292 raises again the basic question Matt Jensen posed in issue #436: can imputations of variables derived from `puf.csv` data be moved out of `records.py` into the `taxdata` repository's `puf-cps-processing.py` script?

Can't we simply move the following functions from `records.py` to `puf-cps-processing.py`?

```
    def _impute_variables(self):
        """
        Impute variables in 2009 PUF Records data
        """
        self._cmbtp_itemizer = self._imputed_cmbtp_itemizer()
        self._cmbtp_standard = self.e62100 - self.e00100 + self.e00700
        # impute number of taxpayers
        self._txpyers[:] = np.where(
            np.logical_or(self.MARS == 2,
                          np.logical_or(self.MARS == 3, self.MARS == 6)),
            2., 1.)
        # impute number of extra standard deductions for aged
        self.numextra[:] = np.where(
            np.logical_or(self.AGERANGE >= 6,
                          np.logical_and(self.age >= 65, self.AGERANGE == 0)),
            self._txpyers, 0.)
        # impute the ratio of household head in total household income
        total = np.where(self.MARS == 2,
                         self.wage_head + self.wage_spouse, 0)
        earnings_split = np.where(total != 0,
                                  self.wage_head / total, 1.)
        one_minus_earnings_split = 1.0 - earnings_split
        self.e00200p[:] = earnings_split * self.e00200
        self.e00200s[:] = one_minus_earnings_split * self.e00200
        self.e00900p[:] = earnings_split * self.e00900
        self.e00900s[:] = one_minus_earnings_split * self.e00900
        self.e02100p[:] = earnings_split * self.e02100
        self.e02100s[:] = one_minus_earnings_split * self.e02100

    def _imputed_cmbtp_itemizer(self):
        """
        Private class method calls global function defined below.
        """
        return imputed_cmbtp_itemizer(self.e17500, self.e00100, self.e18400,
                                      self.e62100, self.e00700,
                                      self.p04470, self.e21040,
                                      self.e18500, self.e20800)

    # DO NOT MOVE def _extrapolate_2009_puf(self):

@vectorize([float64(float64, float64, float64,
                    float64, float64,
                    float64, float64,
                    float64, float64)])
def imputed_cmbtp_itemizer(e17500, e00100, e18400,
                           e62100, e00700,
                           p04470, e21040,
[...]

PR #721, opened 2016-05-04
[WIP] Move _cmbtp_standard logic from records.py to functions.py file

martinholmer:
This pull request continues making changes that move toward the goal of eliminating the `Records.impute_variables()` method as described in issue #688.  But instead of moving the calculation of `_cmbtp_standard` into the taxdata final-prep script, the logic is moved into the functions.py file, which used the `_cmbtp_standard` variable just once in the `AMTI()` function.

The expectation was that making this logic change would have no effect on unit test results, but it turned out to produce very minor changes in the results of the two requires_pufcsv unit tests.  The reason for these changes is that moving the `_cmbtp_standard` logic involved fixing a problem in the master branch code: the `e00100` variable was not being "aged" in the `Records._blowup()` method.


MattHJensen:
Thanks @martinholmer. I'll take a look at this later today.

GoFroggyRun:
Thanks @martinholme, these changes look good to me.

MattHJensen:
Still considering whether this works. Hopefully I'll figure it out soon. _cmbtp is a min imputation of amt preferences.

martinholmer:
Will transform pull request #721 into two: (1) blowup-bug-fix and (2) impute-logic-move.  That will demonstrate that the impute-logic-move causes no changes in test results after the blowup-bug-fix has been implemented.

taxcalc/functions.py
```diff
@@ -824,7 +824,7 @@ def AMTI(c60000, _exact, e60290, _posagi, e07300, c24517,
         c62100 += _cmbtp
     if puf and _standard > 0:
         if f6251 == 1:
-            _cmbtp = _cmbtp_standard
+            _cmbtp = e62100 - e00100 + e00700
```
MattHJensen: @martinholmer, I believe AMT preferences (`_cmbtp`) ought to be imputed with the base year's data and then blown up separately. Therefore, moving the imputation for standard deduction filers functions.py here is not correct. I think we ought to treat _cmbtp like all of our other imputations: do the imputation in TaxData and then include the starting year values in puf.csv.

martinholmer:
Retract pull request #721 in favor of future pull request that moves imputation of the _cmbtp_standard (and _cmbtp_itemizer) variable from the Records class to the taxdata repository.

Issue #710, opened 2016-04-26
Possible Issues When Implementing Reform 


GoFroggyRun:
I happened to notice that, when implementing reforms, different expressions could lead to different results. 

As an example, if we specify two reform as follow:

```
reform1 = {2014: {'_EITC_rt':[[0, 0, 0, 0]]
                 ,'_EITC_prt':[[0, 0, 0, 0]]}
          }
policy1.implement_reform(reform1)

reform2 = {2014: {'_EITC_rt':[[0, 0, 0, 0]]},
           2014: {'_EITC_prt':[[0, 0, 0, 0]]
                 }
          }
policy2.implement_reform(reform2)
```
, which are essentially identical. 

The results, however, will be different:

Reform 1:
![screen shot 2016-04-26 at 3 43 12 pm](https://cloud.githubusercontent.com/assets/13324931/14831906/9edf799a-0bc5-11e6-917b-dc68e1abe9a5.png)

Reform 2:
![screen shot 2016-04-26 at 3 43 18 pm](https://cloud.githubusercontent.com/assets/13324931/14831911/a614bd6a-0bc5-11e6-85c0-90e35184d16f.png)

The issue is that, while trying to implement some reforms, each year entry can only be modified once, and thus all reforms for that specific year shall be included. I thought it would be helpful to have a warning message when users are trying to specify duplicated entries of the same year.

martinholmer:
Sean said:
> I happened to notice that, when implementing reforms, different expressions could lead to different results.

Thanks for reporting this issue, but your presentation of the results from the two reforms (`reform1` and `reform2`) is incomprehensible.  You need to show us the **difference** between the two diagnostic tables, line by line.  If you don't know how to produce the differences in the two tables, then look at the code in the `test_agg()` function in the `taxcalc/tests/test_pufcsv.py` file for one way.  Perhaps you can develop an even better way.  In any case, your current approach is not helpful to others on the development team because it imposes too high a time cost in understanding what you have discovered.

@GoFroggyRun

MattHJensen:
@GoFroggyRun, what is the status of this issue?

GoFroggyRun:
@martinholmer @MattHJensen:

[...]

PR #722, opened 2016-05-04
Added Reform Test Based on Calculated Results 

GoFroggyRun:
This PR adds a test unit to verify an issue that has been identified in #710, where reform expressed in two different ways might yield different results. 

Although #711, which added a test unit that examines the consistency of implementing the same reform in two ways, has passed, this PR that carries out the verification based on the weighted sum of calculated individual income tax results of two calculators with essentially the same reform fails to pass. The test failure suggests that we might be not properly implementing the two different expressions of the same reform. 

It is however somewhat confusing that #711 should report no error. Another guess is that there might be something wrong with the calculation part, instead of implementation of the reforms. 

@martinholmer @MattHJensen @Amy-Xu 
Comments, concerns or remarks would be appreciated.

### DIFF
--- a/taxcalc/tests/test_calculate.py
+++ b/taxcalc/tests/test_calculate.py
@@ -357,6 +357,8 @@ def test_same_reform_expressed_two_ways_result():
     reform2 = {2017: {'_STD': [[0, 0, 0, 0, 0, 0, 0]]},
                2017: {'_EITC_MinEligAge': [26]}
                }
+    print reform1
+    print reform2
     # specify two separate policies
     policy_1 = Policy()
     policy_2 = Policy()
--- a/taxcalc/tests/test_calculate.py
+++ b/taxcalc/tests/test_calculate.py
@@ -345,6 +345,40 @@ def test_Calculator_using_nonstd_input(rawinputfile):
     assert_allclose(mtr_fica, exp_mtr_fica)
 
 
+def test_same_reform_expressed_two_ways_result():
+    """
+    Test that two ways of specifying the same reform produce the same results.
+    """
+    # specify the same reform in two different ways
+    reform1 = {2014: {'_EITC_rt': [[0, 0, 0, 0]], '_EITC_prt': [[0, 0, 0, 0]]}}
+    reform2 = {2014: {'_EITC_rt': [[0, 0, 0, 0]]},
+               2014: {'_EITC_prt': [[0, 0, 0, 0]]}}
+    # specify two separate policies
+    policy_1 = Policy()
+    policy_2 = Policy()
+    # load the input data separately as well
+    TAXDATA_1 = Records(data=TAXDATA, weights=WEIGHTS, start_year=2009)
+    TAXDATA_2 = Records(data=TAXDATA, weights=WEIGHTS, start_year=2009)
+
+    # set up two parallel calculators
+    calcX = Calculator(policy=policy_1, records=TAXDATA_1)
+    calcY = Calculator(policy=policy_2, records=TAXDATA_2)
+
+    # implement the same reforms that are expressed in different ways
+    policy_1.implement_reform(reform1)
+    policy_2.implement_reform(reform2)
+
+    calcX.advance_to_year(2014)
+    calcY.advance_to_year(2014)
+
+    # calculate the results for the two calculators
+    calcX.calc_all()
+    calcY.calc_all()
+    # confirm that the two calculators yield the same iitax liability
+    assert (calcX.records._iitax * calcX.records.s006).sum() == \
+        (calcY.records._iitax * calcY.records.s006).sum()
+
+
 class TaxCalcError(Exception):
     '''I've stripped this down to a simple extension of the basic Exception for
     now. We can add functionality later as we see fit.
--- a/taxcalc/tests/test_calculate.py
+++ b/taxcalc/tests/test_calculate.py
@@ -14,6 +14,8 @@ from taxcalc import create_distribution_table, create_difference_table
 # use 1991 PUF-like data to emulate current puf.csv, which is private
 TAXDATA_PATH = os.path.join(CUR_PATH, '..', 'altdata', 'puf91taxdata.csv.gz')
 TAXDATA = pd.read_csv(TAXDATA_PATH, compression='gzip')
+PUF_PATH = os.path.join(CUR_PATH, '..', '..', 'puf.csv')
+PUFDATA = pd.read_csv(PUF_PATH,)
 WEIGHTS_PATH = os.path.join(CUR_PATH, '..', 'altdata', 'puf91weights.csv.gz')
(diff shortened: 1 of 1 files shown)


martinholmer:
Sean,  Thanks for your pull request #722 that adds a new test to see if two ways of specifying the same reform produce exactly the same weighted aggregate income tax revenue.  Your test produces this assertion error message:
```
assert 989949173284.3723 == 989937010502.1674
```
In other words, the two ways to specifying the same reform produce income tax revenue amounts that have a ratio of about `1.000012`.  That is a relative difference of about 0.0012 percent or about one-thousandth of a percent.   The two revenue amounts, when expressed in billions of dollars rounded to the nearest tenth of a billion (which is the practice on TaxBrain), are `989.9` and `989.9`.

If you think it is important to figure out the cause of this difference and fix the code to make it go away, then proceed with the work.

@GoFroggyRun @MattHJensen @feenberg @Amy-Xu

GoFroggyRun:
@martinholmer:

Thanks for looking into this. My guess is that, since the test units are replying on 1991 data, while we are looking at reforms in 2014, it might be the case that the actually difference resulting from the way we implement the reforms are being significantly diminished due to the extrapolation process, or the test data itself that we are using. 

I tried to replicate pretty much the same thing using `puf` data [here](https://github.com/GoFroggyRun/Notebook/blob/master/Diff_Policy.ipynb), the ratio using the same reform options but fed by different data input is quite different, where the notebook suggests a ratio of 8.7%. This number is not acceptable at all. And I have not yet found a reasonable explanation to such discrepancy.

martinholmer:
Sean said:
> I tried to replicate pretty much the same thing using puf data here, the ratio using the same reform options but fed by different data input is quite different, where the notebook suggests a ratio of 8.7%. This number is not acceptable at all. And I have not yet found a reasonable explanation to such discrepancy.

Then change your new test in pull request #722 to use the `puf.csv` file as Records input.  Then after you get the bigger difference, proceed in pull request #722 to change the Tax-Calculator code in a way that eliminates the bigger difference in the new test.  Then your new test will pass and you will have made a valuable contribution.

@GoFroggyRun @MattHJensen @feenberg @Amy-Xu

feenberg:
On Wed, 4 May 2016, Martin Holmer wrote:

> 
> Sean, Thanks for your pull request #722 that adds a new test to see if two
> ways of specifying the same reform produce exactly the same weighted aggregate
> income tax revenue. Your test produces this assertion error message:
> 
> assert 989949173284.3723 == 989937010502.1674
> 
> In other words, the two ways to specifying the same reform produce income tax
> revenue amounts that have a ratio of about 1.000012. That is a relative
> difference of about 0.0012 percent or about one-thousandth of a percent. The
> two revenue amounts, when expressed in billions of dollars rounded to the
> nearest tenth of a billion (which is the practice on TaxBrain), are 989.9 and
> 989.9.
> 
> If you think it is important to figure out the cause of this difference and
> fix the code to make it go away, then proceed with the work.

This is not due to drop-Q? If not, the error is similar in magnitude to 
what one might expect from rounding a character value to single-precision. 
That is, the reforms might not be exactly the same expressed as binary 
single precision numbers.

dan

GoFroggyRun:
Updates on this issue: 

1. I am, temporarily, using `puf.csv` on my local computer, and it turns out that the assertion difference is now `2736191846302.0684 == 2628870375603.8765`, which shall call more attention on this issue, given the difference is about 4%.

2. Another interesting finding is that, if we choose to print out those reforms immediately after initializing them, we have: 
```
{2017: {'_STD': [[0, 0, 0, 0, 0, 0, 0]], '_EITC_MinEligAge': [26]}}
{2017: {'_EITC_MinEligAge': [26]}}
```
while the initialization and printing commands are given
```
reform1 = {2017: {'_STD': [[0, 0, 0, 0, 0, 0, 0]],
                  '_EITC_MinEligAge': [26]}}
reform2 = {2017: {'_STD': [[0, 0, 0, 0, 0, 0, 0]]},
                  2017: {'_EITC_MinEligAge': [26]}}
print reform1
print reform2
```
This suggests that, if I were not mistaken, `reform2` is not being correctly recorded in the first place. And the reason being is that our reforms are assumed to be in `JSON` format, which is also considered to be a dictionary. A dictionary, however, won't allow duplicated keys (in our case the keys are just different years), and thus specifying duplicated year entries will over-write each others, and only the very last one will be recorded by the system. Given that this seems to be the characteristic (or assumption) of JSON files, the system won't give you any warning message. 

I'm thus not sure what's the best way to proceed and to avoid such possible mis-specification of reforms. All of our existing reforms/ tests assume a `JSON` format and we might not want to change this format. But it also seems hard to raise some warning messages to avoid such mis-specifications of reforms, given the characteristic of JSON file described above. 

@martinholmer @MattHJensen @talumbau 
Any ideas or possbile solutions would be more than appreciated on this issue.

martinholmer:
@GoFroggyRun, what is the status of your pull request #722?  Am I confused, or has #722 been superseded by your pull request #749?  If #722 is no longer relevant, you might consider closing #722.

GoFroggyRun:
@martinholmer thanks for your attention on #722. 

Sorry that the two PRs, #722 and #749, might have confused you. 

#749 is intended to add an explicit test that examines the consistency of addressing ID surtax reform and ID haircut reform.

While this PR, #722, tries to verify the consistency of expressing any same reform in different ways, which is note quite relevant to #749.

MattHJensen:
@GoFroggyRun, this test and your documentation seem very helpful for spurring a conversation about how to deal with situations where a user might mis-specify a reform by listing the year key twice.

@martinholmer said:  

> I don't see how the Tax-Calculator code can check for such a programming error, but I may be wrong on this and would be eager to consider ways of checking for such unintended errors in dictionary construction.

I am in the same boat. Unless @talumbau or others have any suggestions here, it seems like our only option may be to add documentation in some appropriate place for our users. 

@GoFroggyRun do you have any ideas for what that documentation would be and where it should go?

GoFroggyRun:
@MattHJensen:
>
do you have any ideas for what that documentation would be and where it should go?

An obvious option I could think of is to always raise some warning message when users are using the function `implement_reform` in `policy.py`. The content of the message could be something general like:

```Please be advised that Python dictionaries only allow **unique** key values!```,

or something more specific like:

```Please cluster reforms of the same year inside one key!```.

What do you think @MattHJensen?

cc @martinholmer

MattHJensen:
> An obvious option I could think of is to always raise some warning message when users are using the function implement_reform in policy.py. 

I suppose another option would be to add the warning to the doc string for `implement_reform`. @martinholmer, what do you think?

GoFroggyRun:
I'm closing this PR since proper documentation has been added in PR #795 to deal with this issue.

MattHJensen:
Thanks @GoFroggyRun for walking everyone through this potential source of error.

PR #725, opened 2016-05-04
Updated stats_summary.py for current master

Amy-Xu:
This PR fixed two issues with the script:
1) The unused variable list doesn't exist anymore
2) New age variables

The actual summary file doesn't seem different from previous version in terms of dollar amounts. The sequence of variables is not the same.

@MattHJensen Could you review and merge it?


MattHJensen:
This looks good except for the messy git diff for e-variables. Merging. 

@Amy-Xu is going to followup with a comment on how that happened and how to prevent it in the future. 

@martinholmer and @GoFroggyRun, I recommend reviewing the diff for the calculated variables to make sure everything has changed as you expected.

Amy-Xu:
The e-variables summary remains the same but c-varaibles changed due to EITC and Child Tax Credit related logic modification.

All files in this folder haven't been updated during the past few weeks. Right after UNUSED_VARS got removed from records.py, the sequence of PUF e-variables was different at that point. I'm thinking it might be more efficient to include summary statistics updates as a part of test suite, inside the tests using PUF, so the updates would be a part each PR. But reforms update still stays as a standalone script, as it takes about 20~30 min. Will open a PR for this soon.

Issue #750, opened 2016-05-17
Move Earning Split Calculation out of Imputation?


GoFroggyRun:
The `_impute_variables` function in `recorsd.py` includes not only the imputation of `_cmbtp_itemizer` and `_cmbtp_standard`, but also a few earning-splits that are extensively used in the `functions.py`. 

If we were, however, to switch off the imputation by using `consider_imputations=False`, the earning-split parts will be skipped as well, since they are inside the `_impute_variables` function. But this is not what we want, as skipping the earning split process will lead to malfunction of the calculator: 
If the process were being skipped, all earning split variables like `e00200p`, `e00200s`, `e00900p` will be zero. And functions like `EI_FICA` and `F2441` won't work in the desired way. 

I'd thus suggest move the earning split calculation out of the `_impute_variables` function.

feenberg:
On Tue, 17 May 2016, Sean.Wang wrote:

> 
> The _impute_variables function in recorsd.py includes not only the imputation of
> _cmbtp_itemizer and _cmbtp_standard, but also a few earning-splits that are
> extensively used in the functions.py.
> 
> If we were, however, to switch off the imputation by using consider_imputations=False,
> the earning-split parts will be skipped as well, since they are inside the
> _impute_variables function. But this is not what we want, as skipping the earning
> split process will lead to malfunction of the calculator:
> If the process were being skipped, all earning split variables like e00200p, e00200s,
> e00900p will be zero. And functions like EI_FICA and F2441 won't work in the desired
> way.
> 
> I'd thus suggest move the earning split calculation out of the _impute_variables
> function.
> 
Remember that the imputation of earning split must be done on the raw 
data, not after any aging or growth parameters are applied. That is why I 
did it with the other imputations in the SAS program. Also, we may be 
getting a better earnsplit variable after 2010.

dan

GoFroggyRun:
Thanks for the minder @martinholmer.

PR #751, opened 2016-05-19
Moved Earning Split Calculation Out of `_impute_variables` Function   

GoFroggyRun:
As discussed in #750, the `_impute_variables` in `records.py` currently includes the calculation of earning splits, based on the data obtained from the cps-matched `puf` data. Meanwhile, our calculator offers an option to skip the imputation process, namely `consider_imputations`. When `consider_imputations` is set to **False**, whole `_impute_variables` function will be skipped. 

This, however, is not what we want, as skipping the earning split process will damage the tax-logic: in function `EI_FICA` and `F2441`, variable `e00200p`,` e00200s`, `e00900p`, `e00900s`, `e02100p` and `e02100s` will be definitely used to have the right tax-logic, and cannot be submitted. 

This PR moved the calculation of earning split out of the `_impute_variables` function, into a new function `_split_earnings ` that won't be skipped whatsoever, which ensures our calculation for `EI_FICA` and `F2441` won't get wrong. 

@martinholmer @Amy-Xu Could you review?

### DIFF
--- a/taxcalc/records.py
+++ b/taxcalc/records.py
@@ -264,9 +264,11 @@ class Records(object):
             msg = ('Records.constructor start_year is neither None nor '
                    'an integer')
             raise ValueError(msg)
-        if consider_imputations and self.current_year == Records.PUF_YEAR:
-            self._impute_variables()
-            self._extrapolate_2009_puf()
+        if self.current_year == Records.PUF_YEAR:
+            self._split_earnings()
+            if consider_imputations:
+                self._impute_variables()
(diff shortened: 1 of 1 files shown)


martinholmer:
DON'T DO THIS

GoFroggyRun:
@martinholmer , yes you are right. I'm closing this PR.

martinholmer:
Sean @GoFroggyRun,  If you are closing pull request #751, wouldn't it be appropriate to close the related issue #750?

PR #754, opened 2016-05-24
Remove cmbtp-calc and earnings-split logic from Records class

martinholmer:
The removed logic is now in the taxdata repository's final-preparation script (see [taxdata pull request 28](https://github.com/open-source-economics/taxdata/pull/28)), so it is being executed once, when the `puf.csv` file is created, instead of every time the Tax-Calculator is executed.  These two coordinated pull requests are meant to resolve issue #688.


martinholmer:
** AFTER 1st COMMIT**

All the unit tests pass after the first commit in this pull request (using the new `puf.csv` file) except for minor changes in the marginal tax rates histograms and the failure of `test_percentage_change_gdp()` in the `test_macro_elasticity.py` file.  This later failure is probably caused by the use of old 1991 data that has not been prepared in the new manner.

After the first commit in this pull request, all the validation tests passed and all the comparison/reform_results were unchanged.

martinholmer:
**AFTER 2nd COMMIT**

All the unit tests now pass (using the new `puf.csv` created by [taxdata pull request 28](https://github.com/open-source-economics/taxdata/pull/28)) including `test_percentage_change_gdp()`, which has been moved from `test_macro_elasticity.py` to `test_pufcsv.py` and now uses the new `puf.csv` file instead of the old 1991 input data to create the Records objects.

martinholmer:
**THE 3rd COMMIT**

Removes logic in Records class for six variables that are no longer in the `puf.csv` file because they are used only in the cmbtp-calc and earnings-split logic, which has been moved to the taxdata final-prep script in [taxdata pull request 28](https://github.com/open-source-economics/taxdata/pull/28).

martinholmer:
**THE 4th COMMIT**

Removes eleven variables from the VALID_READ_VARS set because those variables are unused in Tax-Calculator logic, and therefore, have been dropped from the `puf.csv` file created by [taxdata pull request 28](https://github.com/open-source-economics/taxdata/pull/28).

These changes have no effect on the unit or validation test results and do not change the comparison reform_results.

martinholmer:
**THE 5th COMMIT**

Adds `DATAPREP.md` with Tax-Calculator Data-Preparation Guidelines.

martinholmer:
@talumbau needs to look at this [Travis-CI crash (not a coding error) message](https://travis-ci.org/open-source-economics/Tax-Calculator/jobs/133241591), which is reproduced here:
```
The command "conda create -n taxcalcdev python=$TRAVIS_PYTHON_VERSION pytest setuptools pandas=0.16.2 numba toolz six" failed and exited with 1 during .
```
I hope this can be fixed quickly.

@MattHJensen

MattHJensen:
@martinholmer, the new DATAPREP.md is *really* helpful. Thanks!

DATAPREP.md
```diff
@@ -0,0 +1,47 @@
...
+e00900 = e00900p + e00900s
+
+e02100 = e02100p + e02100s
+```
+Obviously, when `MARS` is not equal to 2 (married filing jointly), the
+values of the three `s` variables are zero and the values of the three
+`p` variables are equal to the value of their corresponding
+filing-unit variable.  And obviously, the data file can omit any one,
+or all, of these variables.
+
```
zrisher: So since the tax calculator expects:
* `e00200s = e00900s = e02100s = 0 if MARS != 2`
* `e00200 = e00200p + e00200s`, `e00900 = e00900p + e00900s`, and `ce02100 = e02100p + e02100s` always

and if these conditions aren't met we're using an erroneous input set, perhaps we should check for them when the records are loaded and raise an error if they fail?

martinholmer:
Zack (@zrisher) said:
> So since the tax calculator expects:
>```
>e00200s = e00900s = e02100s = 0 if MARS != 2
> e00200 = e00200p + e00200s, e00900 = e00900p + e00900s, and e02100 = e02100p + e02100s always
>```
> and if these conditions aren't met we're using an erroneous input set, perhaps we should check for them when the records are loaded and raise an error if they fail?

Zack,  Thanks for your thoughtful comment on the earnings-split input data expected by Tax-Calculator.

In my own projects, I would have included the checking logic and error message you suggest, but I was warned off about that style of programming when I started working on Tax-Calculator.  I would be happy to add what you are suggesting if others on the development team agree with you that it would be a good idea.

@MattHJensen @feenberg @talumbau @Amy-Xu @GoFroggyRun

zrisher:
Sure, so to summarize why I think this would be a positive change: catching an incorrect state as soon as it's apparent within the code best suited to understanding why it's an issue:
* decreases the amount of time a user must spend debugging unexpected output, if the state where to go otherwise unhandled
* improves understanding of the issue by defining the exact erroneous condition and combining it with a descriptive message - i.e. self-documenting code
* increases modularity / strengthens the API wall by centralizing handling - e.g. we wouldn't need to check for this in Taxbrain, inctax, etc but could instead pass along the message
* code maintenance impact is small as there's usually similar checks that can be grouped together, and placing it within this context can make it quickly comprehensible without reading

Just my two cents, obviously happy to work with whatever best fits the team.

martinholmer:
Argument, made by @zrisher in (closed) pull request #754, in favor of adding checking for earnings-splitting errors when the input file is read by the `Records` class seems persuasive to me.  The only thing I didn't agree with is his characterization of the argument being worth only two cents.

If nobody raises concerns over the next two days about adding such error-checking code, I'll go ahead and do that towards the end of the week.

@MattHJensen @feenberg @talumbau @Amy-Xu @GoFroggyRun

MattHJensen:
@martinholmer, do you remember where the previous discussion about the error checking is? I don't remember the tradeoffs. If you don't remember, could you give another instance of a case where this sort of error checking logic would be useful and what we do (or don't do) instead?

martinholmer:
Matt asked:
> Do you remember where the previous discussion about the error checking is? I don't remember the tradeoffs.

I don't remember.  It was early last year.

And Matt also asked:
> If you don't remember, could you give another instance of a case where this sort of error checking logic would be useful and what we do (or don't do) instead?

There is extensive error-checking of the values read by `simtax.py` from an Internet-TAXSIM-formatted input file (see [`simpletaxio.py`](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/simpletaxio.py)).  I think the case @zrisher makes (in the conversation about closed pull request #754) is quite similar to the reasons we error-check the values in the Internet-TAXSIM-formatted file.

PR #755, opened 2016-05-26
Remove c62100_everyone and associated logic

martinholmer:
This pull request implements the following suggestion made by Dan Feenberg:
>> ```
>> c62100_everyone = c62100
>> if c09600 == 0 and e60000 == 0:
>>     c60000 = 0.
>>     c62100 = 0.
>> ```
>> Why is c62100 zeroed out when c09600 and e60000 are both zero?
> 
> This is a decision to accept the taxpayers decision to file the 6251 or not. If e60000 is zero then the taxpayer did not file a 6251 or pay any AMT. This follows SOI practice. It isn't necessarily better than using our own calculation, indeed I think for purposes of simulating reforms which might affect the decision to file we should not do this.

This pull request simply removes the above code block and changes `c62100_everyone` to `c62100`.  This change means the `e60000` variable is no longer needed in the `VALID_READ_VARS` set or in the `puf.csv` file (changes that will be made when a new `puf.csv` file is distributed).  This change also means that the Tax-Calculator diagnostic table now produces larger values for AMT taxable income; all other values in the diagnostic table are the same as before.

@feenberg @MattHJensen @Amy-Xu @GoFroggyRun


feenberg:
I did look at the taxcalc.sas file and the code to zero c62100 and c60000 
is behind an "if _exact" guard. The _exact means "sacrifice flexibility to 
reproduce soi thinking" and "sacrifice marginal rate calculation to 
taxpayer liability calculation where there is an discrepency"

dan

martinholmer:
Dan @feenberg said:
> I did look at the `taxcalc.sas` file and the code to zero c62100 and c60000 is behind an "if _exact" guard. The _exact means "sacrifice flexibility to reproduce soi thinking" and "sacrifice marginal rate calculation to taxpayer liability calculation where there is a discrepency".

Thanks for the clarification.  I guess the `if _exact` guard got lost in translation.  I'll go ahead and merge pull request #755.

taxcalc/functions.py
```diff
@@ -865,18 +863,14 @@ def AMTI(c60000, _exact, e60290, _posagi, e07300, c24517,
     c63100 = max(0., c63100)
     c63200 = max(0., c63000 - c63100)
     c09600 = c63200
-    c62100_everyone = c62100
```
GoFroggyRun: Sorry for the late comment, but the change here doesn't make much sense to me now. In the diagnostic table, we are reporting AMTI, number of AMT filers and AMT liability. For AMT filers and AMT liability, if I were not mistaken, we are only reporting filers who actually **owe** positive AMT liability. 

After the code line changes here, however, for AMTI, we are reporting the total amount that includes everyone. What we are doing here for AMTI doesn't seem to accord with our following calculation for # of filers and AMT liability. 

I thus have a few suggestions regarding how to deal with AMTI:


- Still adopt the code that zeros out AMTI, however, with slight modifications. In particular, 
```
 if c09600 == 0.:
       c60000 = 0.
       c62100 = 0.
```

- Change the name of AMTI column in our current diagnostic table to a new name: effective AMTI.

martinholmer: @GoFroggyRun said:
> Sorry for the late comment, but the change here doesn't make much sense to me now. In the diagnostic table, we are reporting AMTI, number of AMT filers and AMT liability. For AMT filers and AMT liability, if I were not mistaken, we are only reporting filers who actually owe positive AMT liability.
>
> After the code line changes here, however, for AMTI, we are reporting the total amount that includes everyone. What we are doing here for AMTI doesn't seem to accord with our following calculation for # of filers and AMT liability.
>
> I thus have a few suggestions regarding how to deal with AMTI:
>
> Still adopt the code that zeros out AMTI, however, with slight modifications. In particular,
>```
> if c09600 == 0.:
>       c60000 = 0.
>       c62100 = 0.
>```
> Change the name of AMTI column in our current diagnostic table to a new name: effective AMTI.

Changing the way things are now (after pull request #755 was merged in the master branch on 26-May-2016) would break our Internet-TAXSIM validation tests.  Dan said he was OK with the way AMTI is now defined in Tax-Calculator.  I don't understand why things should be changed.

GoFroggyRun:
@martinholmer said:

>
Dan said he was OK with the way AMTI is now defined in Tax-Calculator. I don't understand why things should be changed.

Firstly, putting the validation results aside, is the current definition of AMTI **helpful** and **informative**? Or, in other words, why are we reporting the portion of AMTI that does not contribute to the AMT liability?

Now, regarding the validation:
>
Changing the way things are now would break our Internet-TAXSIM validation tests.

Having validation consistency against other calculators is definitely something we'd like to see, as this makes us feel much more confident about what we are doing. Braking the validation, in the meantime, shall not be something to be afraid of, nor a good reason for the Tax-Calculator to compromise.

As mentioned in the `README.md` in the validation folder, 

>
The premise behind cross-model validation work is that independently developed tax simulation models are unlikely to contain the same bug.

As far as I am concerned, validation comes after enhancements and developments of the Tax-Calculator, not the other way around. I won't be surprised to see discrepancies between two **independent** models, as every model has its own assumptions. If the discrepancy is something we can explain, I won't be too worried. If the discrepancy is something we can explain, and we have a compelling reason to do so, we do the right thing. Although the AMTI issue here is minor, the essence of validation still applies.

If you agree with my understanding of validation, we can deal with the first question, which is "is the current definition of AMTI **helpful** and **informative**". If not, let me know what validation means in your opinion.

MattHJensen:
@GoFroggyRun said: 

> Putting the validation results aside, is the current definition of AMTI helpful and informative? 

I'm in full agreement that this is the question we should be asking ourselves about everything we report and appreciate that Sean is bringing it up explicitly in this conversation.  

> Or, in other words, why are we reporting the portion of AMTI that does not contribute to the AMT liability?

In this case, I'm not sure there is a clear answer. 

An advantage of reporting AMTI (alternative minimum taxable income) for everyone is that it makes it easier to identify whether a policy reform is affecting AMT liability by increasing AMTI or increasing AMT rates. Furthermore, it allows us to see how AMTI is distributed across the income distribution. If we zero AMTI for non AMT payers, then we lose that distributional information. 

On the other hand, reporting AMTI for everyone is somewhat at odds with the fact that we zero deductible expenses for non itemizers (and in fact data limitations force us to do so). 

With only these considerations in mind, it seems to make more sense to report AMTI for everyone instead of just AMTI that contributes to AMT liability. But there could certainly be other considerations. 

In either case, we need to be very careful about properly documenting what is being reported on TaxBrain -- I will look into that once we make a decision here.

feenberg:
On Sat, 4 Jun 2016, Matt Jensen wrote:

> 
> @GoFroggyRun said:
>
>       Putting the validation results aside, is the current definition of AMTI helpful and
>       informative?
> 
> I'm in full agreement that this is the question we should be asking ourselves about everything
> we report and appreciate that Sean is bringing it up explicitly in this conversation.
>
>       Or, in other words, why are we reporting the portion of AMTI that does not
>       contribute to the AMT liability?
> 
> In this case, I'm not sure there is a clear answer.
>


Yes, this is a problem with no good answer. Surely some people will be 
interested in average tax rate with AMTI as the denominator, and more 
people will be interested in amount of AMT liability before the credit for 
regular tax, in which case we should report AMTI for all taxpayers. But 
most will want to know what effect a change in the rules has on total tax 
liability, and for them only the effective AMTI is relevant. Perhaps we 
should offer both.

I note that we do not offer a way to subset results by filing status, but 
that if we did, it might be nice to offer separate results for AMT/non-AMT 
taxpayers and itemizers/non-itemizers on request.

dan

martinholmer:
Thank you Sean, Matt, and Dan for the helpful discussion of AMT taxable income as part of the conversation of (merged) pull request #755.

My view on this matter is that we should not be zeroing out anybody's AMT taxable income (AMTI).

I think this is the most useful approach for several reasons.  The main reason is the one Matt brings up: some reforms effect that tax base while others effect tax rates applied to the base.  If we were to zero out AMTI for those with zero AMT liability, a reform in AMT tax rates will cause a spurious change in AMTI.  This is the reason we do not zero out regular taxable income for those with a zero (or negative) regular tax liability.

I guess it can be true that sometimes an analyst wants to know the AMTI total for only those with positive AMT liability, but that is no problem because the analyst can use Tax-Calculator to generate all AMTI amounts and then total those amounts for any subset of the sample they want.  This is the whole point of open-source software.  But if we were to censor the AMTI amounts, an analyst who wants all AMTI amounts would be forced to reprogram Tax-Calculator to get the answer they want.  This would be a very undesirable situation.

@GoFroggyRun @MattHJensen @feenberg

PR #760, opened 2016-05-30
Add delta_wages to e00200p in behavior.py update_ordinary_income()

martinholmer:
While I was not involved in the original development of the `Behavior` class and `behavior` function (which implement the behavioral-response logic in the TaxBrain partial-equilibrium dynamic analysis), and therefore, could be missing something, it seems as if the change in wage-and-salary income induced by the behavioral responses to a tax reform should be added to the `e00200p` variable as well as being added to the `e00200` variable.  Neglecting to add the change to the `e00200p` variable means that the change in earnings caused by the behavioral response to a tax reform is not being used to calculate a change in payroll taxes.

We can assess the implications of this single-line change in the code by using the following script:
```
taxcalc/taxbrain$ python behavior.py --help
usage: python behavior.py [-h] REFORM_YEAR CALC_YEAR SUB_ELAST INC_ELAST

Writes to stdout behavioral-response results produced by Tax-Calculator via
the taxcalc package running on this computer.  These results can be compared
with hand-generated results produced by TaxBrain running in the cloud.

positional arguments:
  REFORM_YEAR  calendar year of pop-the-cap reform
  CALC_YEAR    calendar year of calculation results
  SUB_ELAST    substitution elasticity of behavior
  INC_ELAST    income elasticity of behavior

optional arguments:
  -h, --help   show this help message and exit
```
Before the one-line code change, we have these results:
```
taxcalc/taxbrain$ python behavior.py 2020 2020 0.25 0.0
REFORM: pop-the-cap in 2020
SUB_ELASTICITY,INC_ELASTICITY= 0.25 0.0
2020,ITAX,REV_STATIC(S),REV_DYNAMIC(D),D-S= 1737.5 1703.9 -33.6
2020,FICA,REV_STATIC(S),REV_DYNAMIC(D),D-S= 1416.7 1416.3 -0.5
```
After the one-line code change in this pull request, we have these results:
```
taxcalc/taxbrain$ python behavior.py 2020 2020 0.25 0.0
REFORM: pop-the-cap in 2020
SUB_ELASTICITY,INC_ELASTICITY= 0.25 0.0
2020,ITAX,REV_STATIC(S),REV_DYNAMIC(D),D-S= 1737.5 1703.9 -33.6
2020,FICA,REV_STATIC(S),REV_DYNAMIC(D),D-S= 1416.7 1405.7 -11.0
```
Notice that the behavioral responses to this reform induce a decline in income tax revenue that is unchanged by the code revision: -33.6 billion 2020 dollars.  But the decline in payroll taxes is now much larger and more reasonable for this policy reform: -11.0 billion versus -0.5 billion 2020 dollars.

Any thoughts or comments?

@MattHJensen @talumbau @Amy-Xu

### DIFF
--- a/taxcalc/behavior.py
+++ b/taxcalc/behavior.py
@@ -32,6 +32,7 @@ def update_ordinary_income(behavioral_effect, calc_y):
                               0.)
 
     calc_y.records.e00200 = calc_y.records.e00200 + delta_wages
+    calc_y.records.e00200p = calc_y.records.e00200p + delta_wages
 
     calc_y.records.e00300 = calc_y.records.e00300 + delta_other_inc
 


MattHJensen:
@martinholmer, you are correct. Thank you for finding this! I forgot to update the behavior logic after we implemented the earning split.

MattHJensen:
> I forgot to update the behavior logic after we implemented the earning split.

If anyone has a suggestion for how I could avoid this kind of mistake in the future it would be very much appreciated!

martinholmer:
Matt said:
> I forgot to update the behavior logic after we implemented the earning split.
> If anyone has a suggestion for how I could avoid this kind of mistake in the future it would be very much appreciated!

In my experience, there is no silver-bullet method for avoiding errors.  We are all human, and therefore, make mistakes from time to time.

I stumbled across this error after I added a new test --- located in the [`taxcalc/taxbrain/behavior.py` script](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/taxbrain/behavior.py) --- that is stronger than any of the unit tests in the [`taxcalc/tests/test_behavior.py` file](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/tests/test_behavior.py).

The only answer to your question that comes to my mind is "develop more strong tests."

@MattHJensen @feenberg @Amy-Xu @talumbau @GoFroggyRun

feenberg:
On Tue, 31 May 2016, Martin Holmer wrote:

> 
> Matt said:
>
>       I forgot to update the behavior logic after we implemented the earning
>       split.
>       If anyone has a suggestion for how I could avoid this kind of mistake in
>       the future it would be very much appreciated!

When I was building taxsim I thought about this kind of error quite a bit, 
because I saw lots of instances where models had multiple names for 
essentially the same concept, because they were on different forms and 
users forgot to increment all the names when calculating a marginal tax 
rate. In particular it was typical to omit the AMT from marginal tax 
calculations.

For example, capital gains appears on Schedule D and on Form 6251 with 
different names in the SOI database and most tax calculators. The 
justification for this is that they are sometimes different. On the 6251 
they are described as "as adjusted for the AMT" to reflect this 
possibility. In taxsim I ignored the difference (since I didn't have the 
data) and used the same variable for both. In taxcalc.sas I calculate the 
amount of adjustment in the init.sas, and for the AMT calculation I use 
the sum of the Schedule D amount and the adjustment. That way a user can 
come along and vary the Schedule D amount, and still get results that 
reflect the AMT calculation. This works because init.sas is just called 
once while comp.sas is called before and after the increment to capital 
gains. The result can't be perfect since we never know the reason for the 
adjustment, but I believe it is close.

Possibly something similar could be done for earnings. There are not 
actually that many variables with this problem that will affect us, since 
most variables with multiple names (such as AGI, taxable income...) are 
not data for us. I can only think of capital gains and earnings. There are 
some itemized deductions that have multiple names, but I don't think they 
are supposed to be adjusted.

dan

PR #765, opened 2016-06-01
Streamline Behavior class logic, add elasticity validity checking, and add unit tests

martinholmer:
Also, streamline logic of `Behavior._update_ordinary_income()` method.

There is no change in Behavior class logic (other than the fix of the income effect calculation), and therefore, there is no change in the results generated by the unit tests, validation tests, reform comparisons, and by the taxbrain/behavior.py script (which assumes a zero income elasticity, and therefore, its results are unchanged).


martinholmer:
The behavior-response income-effect elasticity calculation (in the `behavior.py` file) is changed in pull request request #765.  The nature of this change is described first and then the rationale for the change is presented.

Before pull request #765, the income-effect elasticity was defined as the the proportional change in taxable income divided by the dollar change in combined tax liability.  Notice that the denominator of this elasticity is not a proportional change.  Also, note that this earlier definition of the income-effect elasticity is logically different from the definition of the substitution-effect elasticity: the proportional change in taxable income divided by the proportional change in marginal net-of-tax rate.  In general, elasticities are defined in the economics literature as the ratio of two proportional (or percentage) changes.

In pull request #765, the definition of the income-effect elasticity is changed to be the proportional change in taxable income divided by the proportional change in after-tax income.  This change make the elasticity conceptually comparable with the substitution-effect elasticity by being the ratio of two proportional changes.  And also, this change brings the income-effect elasticity into line with that used in the economics literature as can be seen in equation 2 on page 10 of Gruber and Saez, "The elasticity of taxable income: evidence and implications", _Journal of Public Economics_ 84:1-32 (2002), which is available at [this link](http://piketty.pse.ens.fr/files/GruberSaez2002.pdf).

feenberg:
On Fri, 3 Jun 2016, Martin Holmer wrote:

> 
> The behavior-response income-effect elasticity calculation (in the behavior.py file) is changed
> in pull request request #765. The nature of this change is described first and then the
> rationale for the change is presented.
> 
> Before pull request #765, the income-effect elasticity was defined as the the proportional
> change in taxable income divided by the dollar change in combined tax liability. Notice that
> the denominator of this elasticity is not a proportional change. Also, note that this earlier
> definition of the income-effect elasticity is logically different from the definition of the
> substitution-effect elasticity: the proportional change in taxable income divided by the
> proportional change in marginal net-of-tax rate. In general, elasticities are defined in the
> economics literature as the ratio of two proportional (or percentage) changes.
> 
> In pull request #765, the definition of the income-effect elasticity is changed to be the
> proportional change in taxable income divided by the proportional change in after-tax income.
> This change make the elasticity conceptually comparable with the substitution-effect elasticity
> by being the ratio of two proportional changes. And also, this change brings the income-effect
> elasticity into line with that used in the economics literature as can be seen in equation 2 on
> page 10 of Gruber and Saez, "The elasticity of taxable income: evidence and implications",
> Journal of Public Economics 84:1-32 (2002), which is available at this link.

Income effects are not usually specified as elasticities. In Feldstein and 
Feenberg ( http://www.nber.org/w5155 ) we specified the income effect as
the dollar change in earnings for each dollar of change in tax. No 
proportions were used. The actual amount we used was .15. As far as I 
recall no one ever raised an objection.

If I look at section 2.2.1 of the Gruber and Saez paper, I don't see an 
elasticity formulation for the income effect. Look at the inline equation
after "income effect parameter". The term (dz/dR) is the .15. dR is the 
change in tax.

dan
617-863-0343

martinholmer:
Dan @feenberg said:
>  Income effects are not usually specified as elasticities. In [Feldstein and Feenberg (1995 NBER working paper)](http://www.nber.org/w5155) we specified the income effect as the dollar change in earnings for each dollar of change in tax. No proportions were used. The actual amount we used was .15. As far as I recall no one ever raised an objection.

Thanks for the link to your 1995 paper.  Which equation were you referring to when you said the income-effect elasticity was defined in a way that "no proportions were used"?

Dan also said:
> If I look at section 2.2.1 of the Gruber and Saez paper, I don't see an elasticity formulation for the income effect. Look at the inline equation after "income effect parameter". The term (dz/dR) is the .15. dR is the change in tax.

I'm not sure what you're referring to in section 2.2.1, but what what I was talking about was equation 2 in section 2.2.2 (on page 10), which is the equation they actually estimate.  It clearly has the income-effect elasticity as the proportional change in taxable income divided by the **proportional** change in after tax income.

Do you think Gruber and Saez estimated the wrong equation?  This paper was published in the _Journal of Public Economics_, so I would assume it was thoroughly reviewed.

@MattHJensen

feenberg:
On Fri, 3 Jun 2016, Martin Holmer wrote:

> 
> Dan @feenberg said:
>
>       Income effects are not usually specified as elasticities. In Feldstein and Feenberg
>       (1995 NBER working paper) we specified the income effect as the dollar change in
>       earnings for each dollar of change in tax. No proportions were used. The actual
>       amount we used was .15. As far as I recall no one ever raised an objection.
> 
> Thanks for the link to your 1995 paper. Which equation were you referring to when you said the
> income-effect elasticity was defined in a way that "no proportions were used"?

Equation 10 on page 14 and then a few lines down where "dwH/dy =.15". dwH 
in wage times hours (which is first order aproximation to change in tax 
liability) divided by the change in income. It isn't the percentage change 
in either numerator or denominator.

dan

MattHJensen:
@feenberg, Gruber and Saez (confusingly) don't use the elasticity specification for the income effect in equation 1 (page 9) but they do in equation 2 (page 10). Could you explain the intuition behind why it might be misguided to use the elasticity specification?

martinholmer:
Dan @feenberg said several days ago:
> Income effects are not usually specified as elasticities. In [Feldstein and Feenberg]( http://www.nber.org/w5155) we specified the income effect as the dollar change in earnings for each dollar of change in tax. No proportions were used. 

I'm not knowledgeable enough to know about the validity of that first sentence, but I have changed the Behavior code back to the way it was: where the income effect is expressed as the _dollar_ change in taxable income divided by the _dollar_ change in after-tax income.

Now the only remaining issue is to address the concerns of @MattHJensen about how best to change itemized deductions by the itemized-deduction share of the total dollar change produced by the sum of the income and substitution effects.  Right now the code is this:
```
calc.records.e19570 = np.where(ided > 0.,
                               calc.records.e19570 + delta_ided,
                               0.)
# TODO, we should create a behavioral modification variable instead of using e19570
``` 
The problem with adding `delta_ided` to the `e19570` variable is that it seems to not always be used in the tax-calculation code.  For example, in the `ItemDed` function in the functions.py file, we have [these lines of code](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/functions.py#L309-L313) that by-pass the `e19570` variable when `puf` is True (which I'm assuming is the case when we are using the `puf.csv` file as Tax-Calculator input).

I'm not sure we should be adding a new "behavioral modification variable" as the TODO comment suggests.  But we should think about which one of the existing `e` variables should have the `delta_ided` amount added to it.  Any thoughts or suggestions?

taxcalc/behavior.py
```diff
@@ -113,105 +94,144 @@ def has_response(self):
...
-        calc_y.policy.
+          from the policy reform that involves moving from calc_x.policy to
+          calc_y.policy.
         Returns new Calculator object --- a deepcopy of calc_y --- that
-        incorporates behavioral responses to the reform.
+          incorporates behavioral responses to the reform.
+        Note: the use here of a dollar-change income elasticity (rather than
+          a proportional-change elasticity) is consistent with Feldstein and
+          Feenberg, "The Taxation of Two Earner Families" NBER Working Paper
+          No. 5155 (June 1995).  A proportional-change elasticity has become
```
MattHJensen: Would the following be a more accurate statement than "has become more common since..."?

> Gruber and Saez used a proportional change elasticity in  "The elasticity of taxable income: evidence and implications", Journal of Public Economics 84:1-32 (2002) [see equation 2 on page 10].

MattHJensen:
I said: 

> @feenberg, Gruber and Saez (confusingly) don't use the elasticity specification for the income effect in equation 1 (page 9) but they do in equation 2 (page 10). Could you explain the intuition behind why it might be misguided to use the elasticity specification?

@feenberg, I'd still appreciate any feedback you have on why the elasticity specification doesn't make sense for the income effect. @martinholmer has switched things back to the dollar change in taxable income divided by the dollar change in after-tax income specification, but this question is likely to come up in the future.

MattHJensen:
@martinholmer said: 

> I'm not sure we should be adding a new "behavioral modification variable" as the TODO comment suggests. But we should think about which one of the existing e variables should have the delta_ided amount added to it. Any thoughts or suggestions?

@martinholmer, I am not sure about the right thing to do. 

At first I was thinking that we should add the feedback to a new variable that feeds into itemized deduction because a new variable would not be subject to potential policy reforms (any existing e variable is subject to reforms). But then I started second guessing whether that is actually the right objective. What about the case where someone eliminates all itemized deductions with haircuts -- in that case, we would want the new deductible expenses resulting from the feedback to be subject to those reforms, right? And if so, does that mean the best option would be to spread the feedback across all of the deductible expense e variables? Alternatively, do you think we should be considering this issue in relation to your work in #788?

taxcalc/behavior.py
martinholmer: @MattHJensen said:
> Would the following be a more accurate statement than "has become more common since..."?
>
>> Gruber and Saez used a proportional change elasticity in "The elasticity of taxable income: evidence and implications", Journal of Public Economics 84:1-32 (2002) [see equation 2 on page 10].

It has been several weeks since I looked at [Emmanuel Saez, Joel Slemrod, and Seth H. Giertz, "The Elasticity of Taxable Income with Respect to Marginal Tax Rates: A Critical Review" _Journal of Economic Literature_ 2012](http://eml.berkeley.edu/~saez/saez-slemrod-giertzJEL12.pdf).  The way I remembered it was that many (most?) studies since 2002 have used the Gruber-Saez income effect specification.  But now that I look back at the 2012 JEL review, I can't seem to find a definitive statement to that effect.  Given that, I'll change the comment to read as you suggest.

martinholmer:
@MattHJensen said:
> I am not sure about the right thing to do.
>
> At first I was thinking that we should add the feedback to a new variable that feeds into itemized deduction because a new variable would not be subject to potential policy reforms (any existing e variable is subject to reforms). But then I started second guessing whether that is actually the right objective. What about the case where someone eliminates all itemized deductions with haircuts -- in that case, we would want the new deductible expenses resulting from the feedback to be subject to those reforms, right? And if so, does that mean the best option would be to spread the feedback across all of the deductible expense e variables? Alternatively, do you think we should be considering this issue in relation to your work in #788?

You're right, this is a complicated issue.  All the more reason to defer this issue.  It has nothing to do with the other changes in pull request #765.

MattHJensen:
@martinholmer said:

> You're right, this is a complicated issue. All the more reason to defer this issue. It has nothing to do with the other changes in pull request #765.

That makes sense. The changes here lgtm. Why don't you merge whenever you are ready.

martinholmer:
@MattHJensen said:
>> You're right, this is a complicated issue. All the more reason to defer this issue. It has nothing to do with the other changes in pull request #765.
>
> That makes sense. The changes here lgtm. Why don't you merge whenever you are ready.

OK.

PR #768, opened 2016-06-03
Add error checks for earnings-split variables in Records constructor

martinholmer:
This pull request implements the suggestion made by @zrisher in the aftermath of pull request #754.  It also streamlines the handling of non-puf.csv input data files and adds new Records tests in an attempt to increase code coverage.  There are no tax-calculation logic changes in this pull request, and thus, there are no changes to the results produced by the unit tests, validation tests, or reform comparisons.


martinholmer:
T.J. (@talumbau),
In a (not completely successful) attempt to achieve complete code coverage of the Records class, I've made some changes in the reading of the blowup factors and sample weights (see [these lines](https://github.com/martinholmer/Tax-Calculator/blob/1c0330670edc867768d8e9d9ec0bef6ca5c3bc1d/taxcalc/records.py#L458-L533) in pull request #768).

I'm hoping you can give me feedback on two issues:

(1) Does the revised code (particularly the code that reads the blowup factors and sample weights from the `taxcalc` package via EGG) look like it will work?  I wasn't sure if passing all the unit tests on my local computer and in the Travis-CI process was conclusive proof that TaxBrain would work smoothly with the revised code.

(2) Is there anyway you can suggest to write a unit test in the `test_records.py` file that would exercise the `except` code in those lines?  Don't spend any time researching this, but if you can make a suggestion from your vast Python knowledge, I'd love to hear about it.

Thanks in advance.

talumbau:
I took a quick look. For the two items:

1. As far as I can tell, the revised reading of the blowup factors looks reasonable, as it is fairly similar to the existing code. The main thing to know is that if new files are added they must be added to the MANIFEST.in at the top of the repo. You are right to wonder about these lines of code, because Travis CI does not currently test them (more elaboration in item 2)

2. The last time I looked at this issue, I did not have a good way to make Travis CI test this situation, as it is happy to read the files directly from the source repository it is testing in, instead of reading from the package.

One manual way to test is to install the package with `python setup.py install` locally, then `rm` the various weights files from your local directory, then run the tests and make sure everything passes.

It would be good to hit those lines of code via a test on Travis CI. I never came up with a clever way to do that and then moved on to the other items. Let me know if you find something!

martinholmer:
T.J. (@talumbau),
Thanks so much for the helpful information in response to my questions in pull request #768.

PR #783, opened 2016-06-12
Add utils tests to increase code coverage

martinholmer:
Non-substantive changes to code in the `utils.py` file plus new tests in the `test_utils.py` file to increase code coverage.  No change in tax-calculation logic.

### DIFF
--- a/taxcalc/utils.py
+++ b/taxcalc/utils.py
@@ -2,6 +2,8 @@ import numpy as np
 import pandas as pd
 from pandas import DataFrame
 from collections import defaultdict
+import pytest
+
 
 STATS_COLUMNS = ['_expanded_income', 'c00100', '_standard', 'c04470', 'c04600',
                  'c04800', 'c05200', 'c62100', 'c09600', 'c05800', 'c09200',
@@ -31,7 +33,6 @@ DIFF_TABLE_LABELS = ["Tax Units with Tax Cut", "Tax Units with Tax Increase",
                      "Percent with Tax Increase", "Percent with Tax Decrease",
                      "Share of Overall Change"]
 
-
 LARGE_INCOME_BINS = [-1e14, 0, 9999, 19999, 29999, 39999, 49999, 74999, 99999,
                      200000, 1e14]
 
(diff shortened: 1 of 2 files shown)


taxcalc/utils.py
```diff
@@ -2,6 +2,8 @@
 import pandas as pd
 from pandas import DataFrame
 from collections import defaultdict
+import pytest
```
talumbau: I didn't have a chance to comment before this got merged. What is the reason that `pytest` is imported here? I don't think it should be imported into library code, only test code.

taxcalc/utils.py
```diff
@@ -225,38 +214,33 @@ def get_sums(df, na=False):
...
     outputs = []
     for col in STATS_COLUMNS:
-        if hasattr(c, 'records') and hasattr(c, 'policy'):
-            if hasattr(c.policy, col):
-                outputs.append(getattr(c.policy, col))
-            else:
-                outputs.append(getattr(c.records, col))
-        else:
-            outputs.append(getattr(c, col))
+        outputs.append(getattr(calc.records, col, np.nan))
```
talumbau: This is a change in behavior. Previously, this function accepted raw DataFrames as well as Calculator objects. The idea is that the intent of this function is to collect up a certain set of attributes from the passed in object and return a collection of them. The caller of the function did not have to care if the object was a Calculator object or raw DataFrame. I would be in favor or reverting this function to its previous implementation.

taxcalc/utils.py
martinholmer: @talumbau said:
> I didn't have a chance to comment before this got merged. What is the reason that pytest is imported here? I don't think it should be imported into library code, only test code.

Thanks for pointing out this oversight.  I'll remove that unused import statement.

taxcalc/utils.py
martinholmer: @talumbau said:
> This is a change in behavior. Previously, this function accepted raw DataFrames as well as Calculator objects. The idea is that the intent of this function is to collect up a certain set of attributes from the passed in object and return a collection of them. The caller of the function did not have to care if the object was a Calculator object or raw DataFrame. I would be in favor or reverting this function to its previous implementation.

Well then I don't understand why the old documentation said this:
```
Parameters
----------
c : Calculator object
```
There was nothing there saying the results function "accepted raw DataFrames as well as Calculator objects".
In fact. the old documentation was quite explicit about the results argument being a Calculator object.

If you want it to work in the undocumented manner that you describe above, then I suggest you change it to be the way you want.  You can do that as you add more tests to `test_utils.py` so that the utils code coverage is complete.

PR #786, opened 2016-06-13
Revert results function to accept multiple types

talumbau:
- `results` collects attributes from an object and creates a DataFrame.
   This object could be a Calculator, in which case it must look for the
   attributes in other objects. Otherwise, just collect them, create the
   DataFrame and return.

### DIFF
--- a/taxcalc/tests/test_utils.py
+++ b/taxcalc/tests/test_utils.py
@@ -137,9 +137,6 @@ def test_create_tables():
     records2 = Records(data=TAXDATA, weights=WEIGHTS, start_year=2009)
     calc2 = Calculator(policy=policy2, records=records2)
     calc2.calc_all()
-    # test incorrect call of results() function
-    with pytest.raises(ValueError):
-        res = results(dict())
     # test creating various distribution tables
     dt1 = create_difference_table(calc1, calc2, groupby='large_income_bins')
     dt2 = create_difference_table(calc1, calc2, groupby='webapp_income_bins')
@@ -552,3 +549,16 @@ def test_ascii_output_function(csvfile, asciifile):
     assert filecmp.cmp(output_test.name, asciifile.name)
     output_test.close()
     os.remove(output_test.name)
+
+
+def test_results():
+    # create a current-law Policy object and Calculator object calc1
+    policy1 = Policy()
+    records1 = Records(data=TAXDATA, weights=WEIGHTS, start_year=2009)
+    calc1 = Calculator(policy=policy1, records=records1)
+    calc1.calc_all()
+    res_df = results(calc1)
+    assert all(res_df.columns == STATS_COLUMNS)
+    # results function works on DataFrames, too
+    res_df2 = results(res_df)
+    assert all(res_df2.columns == STATS_COLUMNS)
--- a/taxcalc/utils.py
+++ b/taxcalc/utils.py
@@ -238,11 +238,8 @@ def results(calc):
     """
     outputs = []
     for col in STATS_COLUMNS:
-        if hasattr(calc, 'records') and hasattr(calc, 'policy'):
-            if hasattr(calc.policy, col):
-                outputs.append(getattr(calc.policy, col))
-            else:
-                outputs.append(getattr(calc.records, col))
+        if hasattr(calc, 'records'):
+            outputs.append(getattr(calc.records, col))
         else:
             outputs.append(getattr(calc, col))
     return DataFrame(data=np.column_stack(outputs), columns=STATS_COLUMNS)
--- a/taxcalc/utils.py
+++ b/taxcalc/utils.py
@@ -224,22 +224,27 @@ def get_sums(df, na=False):
 
 def results(calc):
     """
-    Get results from Calculator object and organize them into a table.
+    Get all STATS_COLUMNS attributes from a Calculator-like object
+    and organize them into a DataFrame.
 
     Parameters
     ----------
-    calc : Calculator object
+    calc : a Calculator object or some other object with attribute names
+           matching the list in STATS_COLUMNS
 
     Returns
     -------
     DataFrame object
     """
-    if not hasattr(calc, 'records'):
-        msg = 'calc does not have records attribute in results(calc) utility'
-        raise ValueError(msg)
     outputs = []
     for col in STATS_COLUMNS:
-        outputs.append(getattr(calc.records, col, np.nan))
+        if hasattr(calc, 'records') and hasattr(calc, 'policy'):
+            if hasattr(calc.policy, col):
+                outputs.append(getattr(calc.policy, col))
+            else:
+                outputs.append(getattr(calc.records, col))
+        else:
+            outputs.append(getattr(calc, col))
     return DataFrame(data=np.column_stack(outputs), columns=STATS_COLUMNS)
 
 


martinholmer:
@talumbau proposed revising the utility function `results` in #786 to have this code:
```
    outputs = []
    for col in STATS_COLUMNS:
        if hasattr(calc, 'records'):
            outputs.append(getattr(calc.records, col))
        else:
            outputs.append(getattr(calc, col))
    return DataFrame(data=np.column_stack(outputs), columns=STATS_COLUMNS)
```
When calc has a records attribute, another test I added yesterday has ensured that `calc.records` has all the variable names in the `STATS_COLUMNS` list.

But what if the calc object does not have a records attribute and the calc object does not contain a variable name in the `STATS_COLUMNS` list?  The Python documentation says this about `getattr`:
```
If the named attribute does not exist, default is returned if provided,
otherwise AttributeError is raised.
```
So, in this untested case, it seems as if your proposed code (which does not provide a default value to `getattr`) will raise an AttributeError and the application using the `results` utility function will crash.  Is that what we want?

Also, is there any real-world use of the `results` utility function when it is called with an argument that is not a Calculator object?

talumbau:
The calling code will not crash. It will receive an `AttributeError` exception, which it will either handle or pass back up to its caller, and so on up the call stack. The alternative would be to put an additional try/except block inside of this function, which would catch one type of exception and either re-raise the same type of exception or some other type of exception, and then all behavior after that would be exactly the same as what is in this PR. The goal of exception handling should be to add value of some kind and provide a helpful message to calling code about what went wrong. This function is about seven lines long and it's not doing very much; it's hard to think of a more helpful error message than something like `AttributeError: XXX not found`, which is what we would get.

The use case is in one of the repos that make up the TaxBrain application, although I can't find it right now. A lot of downstream processing of Tax-Calculator results are done in DataFrames, so it seems useful to have a function that works on both.

martinholmer:
@talumbau said in response to this question:
>> Also, is there any real-world use of the results utility function when it is called with an argument that is not a Calculator object?
>
> The use case is in one of the repos that make up the TaxBrain application, although I can't find it right now. A lot of downstream processing of Tax-Calculator results are done in DataFrames, so it seems useful to have a function that works on both.

I've looked at the webapp-public and drop-q repositories and I don't see any use of the *Tax-Calculator* results utility function.  Have I missed a use?

What I do see is that you have a [results function in the drop-q repository](https://github.com/open-source-economics/drop-q/blob/master/dropq/dropq.py#L259-L267).  Why the duplication?  And the same question applies to this duplicate results function: why is its logic made more complex in order to handle non-Calculator objects when this results function is always called with a Calculator object as its argument.  Why all the (apparently) unnecessary complication?

So far, I don't see any need for pull request #786.  Have I missed something?

@MattHJensen

talumbau:
ah yes, I see. I believe at some point in the past I needed both ways but this has been refactored. Since `STATS_COLUMNS` is defined in Tax-Calculator, I think it would make the most sense to get rid of the `results` function in dropq and just use this one.

Feel free to close if you wish, but I don't understand the following things:

 - the proposed PR is the exact same number of lines of code as what is in master? How is the logic "more complex"?

 -  The task of collecting attributes and giving back a DataFrame seems generally useful. Why the insistence on having a `records` attribute to achieve this goal? It seems like an artificial restriction on a utility function.

- Why is the current test of the capabilities of the `results` function inside a test called `test_create_tables`? Usually a good idiom for testing a function `foo` is a test named `test_foo`.

I'm fine abandoning this PR. I can add additional utility functions to drop-q if necessary.

martinholmer:
@talumbau said:
> ah yes, I see. I believe at some point in the past I needed both ways but this has been refactored. Since STATS_COLUMNS is defined in Tax-Calculator, I think it would make the most sense to get rid of the results function in drop-q and just use this one.

I think that is very sensible.

> Feel free to close if you wish, but I don't understand the following things:
>
> the proposed PR is the exact same number of lines of code as what is in master? How is the logic "more complex"?

Same number of lines only because in this pull request you don't check to see if calc has a records attribute; you just rely on the Python interpreter to raise an error when you try to access calc.records.  I know you put a low value on such explicit checking, but that is why the number of statements is the same.  If the master branch code dropped the explicit checking, it would have three fewer statements.

> The task of collecting attributes and giving back a DataFrame seems generally useful. Why the insistence on having a records attribute to achieve this goal? It seems like an artificial restriction on a utility function.

My view is that there is a cost (in terms of testing and code maintenance) of adding features.  If the feature is needed by other parts of the project, then the cost of adding the feature is worth incurring.  But if the feature is never used, then a benefit-cost analysis of adding the feature always produces a negative net benefit (that is, zero minus the cost).  You could say (correctly) that even though the feature is not now used, it could be used sometime in the future.  But that is an uncertain benefit, which still leads to an unfavorable benefit-cost result.  If the need for the feature actually does materialize in the future, then the feature can be added then.  I guess I have a JIT (just-in-time) philosophy about adding features to an already complex project.

> Why is the current test of the capabilities of the results function inside a test called test_create_tables? Usually a good idiom for testing a function foo is a test named test_foo.

Well, the location of the test is a little arbitrary.  As I remember, I put it in the `test_create_tables` function because the `results` utility function appeared to be called only from the functions that create tables (difference tables and distribution tables).

> I'm fine abandoning this PR. I can add additional utility functions to drop-q if necessary.

OK.  I'll close pull request #786.

martinholmer:
Closing pull request #786 with permission from T.J. (@talumbau).

zrisher:
Sorry to be a little late to the party here. I think this discussion brought up some interesting points though, so I just wanted to add my thoughts:

> if the feature is never used, then a benefit-cost analysis of adding the feature always produces a negative net benefit

I agree entirely, and I would even say that if it's *rarely* used, the net benefit is probably still negative.

In this particular case, we have a function `results()` that returns the calculated data from a `Calculator` object. Correct me if I'm wrong, but I believe most uses of the calculator are to calculate these stats_columns, and most uses of the stats_columns would like to deal with them as a DataFrame (or dict) instead of as attributes on the records object. So this function is not just valuable, but intrinsic to the Calculator object.

Because this function is so closely linked to the main use case of the Calculator, and because its only input is a Calculator, it seems to me that it should actually be a function on the Calculator object itself, instead of a general utility function. This would obviate the need for checking that the records object has the expected stats_columns as well, because when we're in the context of an instance method we can generally assume that our internal attributes are correctly prepared (and a general error is acceptible if they aren't). 

This may bring up the question of "well, if we're going that route, shouldn't we go all the way down and make it a function on the Records object instead?" I personally would say "no", because it actually seems to me that the data a Calculator calculates should be in a seperate Results object linked to the Records by shared record_ids. I think that would more clearly represent the relationships between data, but I understand why the current route was taken given how TAXSIM works. 

Now, getting back to the point, if we see this strong relationship between `results()` and a `Calculator` object, it makes less sense to generalize this function to take some unknown dictionary which is *supposed* to be a records-like object. If we really needed this functionality (which it sounds like we may not), I would add the general stats attributes collecting as an instance method on a `Records` object, and then have the `calculator.results()` function proxy `calculator.records.results()`.

Just my opinion, based on a more limited understanding of Tax-Calculator. Very curious to hear your thoughts @talumbau @martinholmer.

talumbau:
I now know why this behavior is needed. The calls "in the wild" happen when we go to make results tables in the `dropq` package:

https://github.com/open-source-economics/drop-q/blob/master/dropq/dropq.py#L240

(and similar lines). The situation is as follows:

`create_difference_table` and `create_distribution_table` just need some columns to operate on. They grab those columns from the passed in object using the `results` function. In the case of `dropq`, we do a significant amount of upstream work and end up with a DataFrame that has the columns that we need to make tables out of, instead of the raw Calculator. So, the way the code was prior to #783, you could pass in a Calculator object or a DataFrame into those functions and it would do the right thing. Conceptually, this polymorphic behavior seems pretty straightforward.

Handling as a method seems reasonable in the abstract as well, @zrisher. At the very least, it would avoid the "look before you leap" style that exists in current master.

So, the situation was as follows: two useful functions were written to create tables based on certain data inside Calculator objects. At about the same time, `dropq` was then written which does unique computational work resulting in DataFrames. Similar tables were needed out of the data in the DataFrames, so the code prior to #783 handled both ways.

martinholmer:
@zrisher,  Thanks for your thoughtful comments.  I particularly like the idea of making the results utility function into a Calculator method.  That will serve all the uses in the Tax-Calculator.

@talumbau @MattHJensen

martinholmer:
@talumbau, Thank you for your reaction to the comments by @zrisher on closed pull request #786.

In particular, you say:
> Handling as a method seems reasonable in the abstract as well, @zrisher.

Does that mean you favor converting the results utility function into a Calculator method (as suggested by @zrisher)?  Or am I confused about what you're saying here?  Then you could simply use a drop-q results function that works with DataFrames, right?  I guess I'm not crystal clear about what you are suggesting as the best way forward.

Alos, I still don't understand the role of the drop-q results function.
Do you ever call that function in drop-q?

@MattHJensen

talumbau:
> Does that mean you favor converting the results utility function into a Calculator method (as suggested by @zrisher)? Or am I confused about what you're saying here? Then you could simply use a drop-q results function that works with DataFrames, right? I guess I'm not crystal clear about what you are suggesting as the best way forward.

My suggestion for the best way forward is to merge this PR. 

@martinholmer said
> If the need for the feature actually does materialize in the future, then the feature can be added then. I guess I have a JIT (just-in-time) philosophy about adding features to an already complex project.

The need for the feature exists today. It's used here:

https://github.com/OpenSourcePolicyCenter/dropQ/blame/master/dropq/dropq.py#L240

and here:

https://github.com/OpenSourcePolicyCenter/dropQ/blame/master/dropq/dropq.py#L246

and has been since April of 2015. The benefits of merging the PR would be:

 - reverting to the previous functionality of the code prior to an undiscussed commit to master
 - allowing existing dropq code to work again with master
 - allowing the `create_*_table` functions to take either DataFrames or Calculator objects, or really any object that has the proper attributes.

Some alternatives would be:

 - do nothing and leave the results of #783. The downside is that `dropq` will have to be changed because it passes DataFrames to the `create_*_table` functions, which are then passed as arguments to `results`

 - refactor the `results` function as a method on Calculator, and then modify the `create_*_table` functions in `utils.py` to not call `results` (either as a free standing function or a method) and then make sure that any call to `create_*_table` functions passes `calc.results()` instead of `calc`.

A bad outcome would be:

- refactor the `results` function as a method on Calculator, and then modify the `create_*_table` functions in `utils.py` to call `calc.results()` internal to the function. This would mean that `dropq` would be unable to use the `create_*_table` functions, which would be bad, and such a change would offer very little benefit to this package. 

All of the alternatives are more effort than merging this PR which essentially just enhances code coverage of an existing use case.

martinholmer:
@talumbau, Thanks for your clarification of your earlier comments on the comments by @zrisher.

I'll think about what you are proposing, but I still don't have an answer from you to my question:
> Also, I still don't understand the role of the drop-q results function.
> Do you ever call that function in drop-q?

talumbau:
> Also, I still don't understand the role of the drop-q results function.
> Do you ever call that function in drop-q?

@talumbau said:
> ah yes, I see. I believe at some point in the past I needed both ways but this has been refactored. Since STATS_COLUMNS is defined in Tax-Calculator, I think it would make the most sense to get rid of the results function in drop-q and just use this one.

I think the specific sequence of events was: created a `dropq` version of `create_distribution_table` (along with its own version of the `results` function), then realized the usage was identical to that in Tax-Calculator, deleted `dropq` version and switched to Tax-Calculator version, left in the `results` function which now has no callers. Does that answer the question?

martinholmer:
@talumbau said:
> I think the specific sequence of events was: created a dropq version of create_distribution_table (along with its own version of the results function), then realized the usage was identical to that in Tax-Calculator, deleted dropq version and switched to Tax-Calculator version, left in the results function which now has no callers. Does that answer the question?

Yes, thank you.

martinholmer:
@talumbau said:
> I think the specific sequence of events was: created a dropq version of create_distribution_table (along with its own version of the results function), then realized the usage was identical to that in Tax-Calculator, deleted dropq version and switched to Tax-Calculator version, left in the results function which now has no callers. Does that answer the question?

Yes, this answers my question, but I'm still confused about your dropq.py code.  It looks to me as if you are calling the `dropQ results` function four times in the `run_nth_year_mtr_calc` function: twice in [these lines](https://github.com/open-source-economics/drop-q/blob/master/dropq/dropq.py#L403-L404) and twice in [these lines](https://github.com/open-source-economics/drop-q/blob/master/dropq/dropq.py#L444-L445).

Of course, you can continue to do that.  But if you want to remove the `dropQ results` function and use the new Tax-Calculator `results` function, you will have to pass the new `results` utility function a Records object (not a Calculator object).

Issue #789, opened 2016-06-14
Behavior class handling of itemized-deduction response


martinholmer:
After pull request #765 was merged into the master branch, the main issue with the Behavior class is the concern of @MattHJensen about how best to change itemized deductions by the itemized-deduction share of the total dollar change produced by the sum of the income and substitution effects. Right now the code is this:
```
calc.records.e19570 = np.where(ided > 0.,
                               calc.records.e19570 + delta_ided,
                               0.)
# TODO, we should create a behavioral modification variable instead of using e19570
```
The problem with adding `delta_ided` to the `e19570` variable is that `e19570` seems to not always be used in the tax-calculation code. For example, in the `ItemDed` function in the `functions.py` file, we have [these lines of code](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/functions.py#L309-L313) that by-pass the `e19570` variable when `puf` is True (which I'm assuming is the case when we are using the `puf.csv` file as Tax-Calculator input).

I'm not sure we should be adding a new "behavioral modification variable" as the TODO comment suggests. But we should think about which one of the existing `e` variables should have the `delta_ided` amount added to it.

Then in response @MattHJensen said:
> I am not sure about the right thing to do.
>
> At first I was thinking that we should add the feedback to a new variable that feeds into itemized deduction because a new variable would not be subject to potential policy reforms (any existing e variable is subject to reforms). But then I started second guessing whether that is actually the right objective. What about the case where someone eliminates all itemized deductions with haircuts -- in that case, we would want the new deductible expenses resulting from the feedback to be subject to those reforms, right? And if so, does that mean the best option would be to spread the feedback across all of the deductible expense e variables?

feenberg:
On Tue, 14 Jun 2016, Martin Holmer wrote:

> 
> After pull request #765 was merged into the master branch, the main issue with the
> Behavior class is the concern of @MattHJensen about how best to change itemized
> deductions by the itemized-deduction share of the total dollar change produced by the
> sum of the income and substitution effects. Right now the code is this:
> 
> calc.records.e19570 = np.where(ided > 0.,
>                                calc.records.e19570 + delta_ided,
>                                0.)
> # TODO, we should create a behavioral modification variable instead of using e19570
> 
> The problem with adding delta_ided to the e19570 variable is that e19570 seems to not
> always be used in the tax-calculation code. For example, in the ItemDed function in
[...]

PR #791, opened 2016-06-15
Remove itemized-deduction logic from Behavior response method

martinholmer:
This pull request was prepared at the request of Matt Jensen.  The idea is to remove the itemized-deduction logic in the Behavior class response method until the problems identified in issue #789 can be resolved.


MattHJensen:
Thanks @martinholmer. This LGTM. Merging.

PR #792, opened 2016-06-16
Revise create_distribution_table utility function to support dropQ usage

martinholmer:
The extended discussion of closed pull request #786 belatedly revealed that code in the dropQ repository requires the Tax-Calculator `create_distribution_table` utility function to accept a DataFrame object as its first argument.  This pull request revises the `create_distribution_table` utility function accordingly, and therefore, it now supports the dropQ usage.

@MattHJensen @talumbau @zrisher @Amy-Xu @GoFroggyRun

### DIFF
--- a/taxcalc/tests/test_calculate.py
+++ b/taxcalc/tests/test_calculate.py
@@ -288,7 +288,8 @@ def test_Calculator_create_difference_table():
     puf2 = Records(data=TAXDATA, weights=WEIGHTS, start_year=2009)
     calc2 = Calculator(policy=policy2, records=puf2)
     # create difference table and check that it is a Pandas DataFrame
-    dtable = create_difference_table(calc1, calc2, groupby="weighted_deciles")
+    dtable = create_difference_table(calc1.records, calc2.records,
+                                     groupby="weighted_deciles")
     assert isinstance(dtable, pd.DataFrame)
 
 
--- a/taxcalc/tests/test_calculate.py
+++ b/taxcalc/tests/test_calculate.py
@@ -248,10 +248,12 @@ def test_Calculator_create_distribution_table():
                    'Individual Income Tax Liabilities',
                    'Payroll Tax Liablities',
                    'Combined Payroll and Individual Income Tax Liabilities']
-    dt1 = create_distribution_table(calc, groupby="weighted_deciles",
+    dt1 = create_distribution_table(calc.records,
+                                    groupby="weighted_deciles",
                                     result_type="weighted_sum")
     dt1.columns = dist_labels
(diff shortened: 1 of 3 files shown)


talumbau:
I think this will handle the existing `dropq` use case, but I am wondering about the interface change. I thought we had some nice examples of creating different tables in our [example](https://github.com/open-source-economics/Tax-Calculator/blob/master/docs/notebooks/10_Minutes_to_Tax-Calculator.ipynb) [notebooks](https://github.com/open-source-economics/Tax-Calculator/blob/master/docs/notebooks/Behavioral_example.ipynb), but I was mistaken. Maybe we had examples of creating tables in those notebooks in the past and they were removed? In any case, my own view is that it's a bit odd to have the interface be different between `create_distribution_table` and `create_difference_table`. So now we would have folks do:

```
 dt1 = create_distribution_table(calc1.records,
                                    groupby="weighted_deciles",
                                    result_type="weighted_sum")

```
And then after that have:
```
difftable = create_difference_table(calc1, calc2, groupby="weighted_deciles")
```

which is a bit unintuitive (one might think it would be `create_difference_table( calc1.records, calc2.records...)`. But, `dropq` is a caller of only one of the functions, so my use case is handled by this PR. I would suggest making the interface consistent for the two functions (so, omitting the changes to lines 405 and 406 from `utils.py` in this PR), but I am not a regular user of the calculator so it may be of little to no importance to actual users of the calculator.

martinholmer:
The arguments of the `create_difference_table` have been changed at the suggestion of @talumbau.  Now both the `create_distribution_table` and `create_difference_table` utility functions take the same kind of beginning arguments.

talumbau:
This looks good to me. Thanks for making this change so that `dropq` will not have to change right away. I hope this does not present an undue burden for users of the `create_*_table` functions. Now that the [dropq repo](https://github.com/open-source-economics/drop-q/tree/master/dropq) is completely open, I had wondered about the potential for creating tests in Tax-Calculator that actually depend on a certain tagged version of `dropq`. That would allow for future changes to these functions to happen in lockstep with dropq. That may or may not be a good idea. I'll open an issue for discussion.

martinholmer:
T.J. (@talumbau) said:
> This [pull request #792] looks good to me. Thanks for making this change so that dropq will not have to change right away. I hope this does not present an undue burden for users of the create_*_table functions.

I don't think there will be any burden at all.  Shall I wait for other comments until the end of the day before merging?

@MattHJensen

MattHJensen:
This LGTM.  @Amy-Xu and @GoFroggyRun note the table API change.

MattHJensen:
@martinholmer, I'm going to go ahead and merge so we can release the new version of tax-calculator.

PR #794, opened 2016-06-17
Propagate read_csv args into read_egg_csv

talumbau:


talumbau:
This PR changes `_read_egg_csv` so that it can take additional keyword arguments for the Pandas `read_csv` function. In particular, we need to pass through the `index_col="YEAR"` argument when reading the blowup factors. 

Discovered this issue when doing integrating testing with the webapp and dropq.

martinholmer:
I don't understand these changes.
In particular, I don't understand the vague and mysterious use of `**kwargs`?  Why not just make the following change in the `Records._read_egg_csv` method?
```
-            vname_dict = pd.read_csv(vname_fname)
+            vname_dict = pd.read_csv(vname_fname, index_col='YEAR')
```
Won't that do what you want?

talumbau:
`_read_egg_csv` is used in two places. In the call at this line:

https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/records.py#L514

we need to read a CSV file with `pandas.read_csv` and not pass in the `index_col="YEAR"` keyword argument. In the call at this line:

https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/records.py#L491

we do need the call. This occurred when you did refactoring with this commit:

https://github.com/open-source-economics/Tax-Calculator/commit/86c5b6c429bbbb27bef4ed481021e024eee73edd

I think the refactoring was nice. Essentially, we had two very similar chunks of code, so you created one function out of it with two calls. As you can see in the commit https://github.com/open-source-economics/Tax-Calculator/commit/86c5b6c429bbbb27bef4ed481021e024eee73edd
, the calls to `pd.read_csv` were not identical. So, this PR changes `_read_egg_csv` to allow "pass-through" arguments to the pandas function. That way, the caller of `_read_egg_csv` determines what arguments to give to pandas `read_csv` instead of it being hard-coded in the function.

talumbau:
So to be explicit:

> Why not just make the following change in the Records._read_egg_csv method?
> 
> -            vname_dict = pd.read_csv(vname_fname)
> +            vname_dict = pd.read_csv(vname_fname, index_col='YEAR')
> Won't that do what you want?

No, it won't because of the two calls to `_read_egg_csv`.

martinholmer:
T.J. (@talumbau),  Thanks for the gentle and thorough explanation of my mistake in commit 86c5b6c.

martinholmer:
@talumbau,  Thanks for fixing my mistake.

MattHJensen:
Lgtm. Merging.

PR #795, opened 2016-06-20
Add to documentation of Policy.implement_reform method

martinholmer:
This pull request adds to the Policy.implement_reform method docstring a warning about avoiding more than one YEAR:MODS pair that have the same YEAR value.  The need for such a documentation enhancement was the conclusion of the pull request #722 conversation.

Being only a comment change, there is no change in tax-calculation logic.


MattHJensen:
These instructions look very helpful to me. What do you think @GoFroggyRun?

GoFroggyRun:
@MattHJensen Agreed. LGTM as well.

PR #796, opened 2016-06-21
Add new itemized-deduction logic to Behavior class

martinholmer:
This pull request attempts to resolve issue #789.

@MattHJensen @feenberg @Amy-Xu @GoFroggyRun @zrisher @talumbau


MattHJensen:
@martinholmer, I don't understand the intuition behind deductible expenses falling when there is a reduction in marginal tax rates. I would have thought the opposite: your incentive to work goes up, you work more, buy a bigger houses, pay more taxes, and give more to charity. I suppose I can comprehend an argument that charity and housing are complements to leisure, but it feels tenuous. Do you know if there is evidence for this? 

An alternative would be to subtract ided from agi for the denominator instead of summing them, and thereby allow an increase for each other_income, agi, and ided that still nets out to the predicted change in taxable income. 

Here are a couple of examples for anyone who isn't following along with the code. 

Existing:

image

Proposed:

image

cc @feenberg

martinholmer:
Thanks, Matt, for you insightful comment on pull request #796, which is an attempt to fix the logic error identified in issue #789.  Your comment asks why #796 subtracts from, rather than adds to, itemized deductions a share of the change in taxable income.  I haven't seen in the economics literature cited on TaxBrain any example of how researchers allocate the total change in taxable income to different components of taxable income.  Have I missed part of that literature?  How do other researchers do this allocation to the components of taxable income?  Given that I've not seen anything in the literature about how to do the allocation, pull request #796 does a subtraction because that was how I interpreted Dan Feenberg's following comment (made as part of issue #789):
```
In the past I have always subtracted the behavioral change in itemized
deductions from "other deductions" and allowed that to go negative. It is
an arbitrary allocation, but no more so than any other.
```
I interpreted this to mean "subtracting from" was what researchers did.  Did I misunderstand this?

@MattHJensen @feenberg @Amy-Xu

martinholmer:
Revise logic of the `Behavior.response` method to reflect approach suggested by @MattHJensen in the conversation about pull request #796.

MattHJensen:
Thanks @martinholmer for identifying the bug in #789 and fixing it in this PR. After looking for and not finding any literature to the contrary, I think we are right to have deductible expenses move in the same direction as taxable income. Merging this now, but will probably keep stewing on how best to communicate to the user -- and perhaps give the user some control over -- how the taxable income effect is split up amongst the components of taxable income.

Issue #780, opened 2016-06-11
Need more decorators tests


martinholmer:
The new test in the merged pull request #779 reduced the number of statements in the `decorators.py` file that are untested from 8 to 6.  You can see the current master branch `decorators.py` code and the six untested statements at [this link](https://codecov.io/gh/open-source-economics/Tax-Calculator/src/825f442f34fd6c97e95bdcdf4bcf3aee5b0f6a82/taxcalc/decorators.py).  I do not understand the code in the `decorators.py` file well enough to write other tests that would reduce the number of untested statement from six to zero.

@talumbau, as the original author of the `decorators.py` functions, could you add the extra tests to the `test_decorators.py` file that are needed to get to complete code coverage?

@MattHJensen

martinholmer:
T.J. (@talumbau),  If you don't have time for working on the request in issue #780, can you give me some guidance on how to add the needed unit tests?  I'd be happy to do the work, but I'm not that familiar with the decorators.py logic.

@MattHJensen

martinholmer:
Closing issue #780 because T.J. added more decorator tests in merged pull request #799.
Thanks, @talumbau.

PR #799, opened 2016-06-21
No numba dependency

talumbau:


talumbau:
I need another commit here to deal with Codecov issues, but I wanted to point out that running the tests w/o `numba` makes the test cycle with Travis CI about 13 minutes along, instead of about 3 with `numba`. I don't consider this a big deal but I thought I should point it out as something to consider for this PR.

cc @Amy-Xu @GoFroggyRun @MattHJensen @martinholmer

martinholmer:
T.J. said:
> I need another commit here to deal with Codecov issues, but I wanted to point out that running the tests w/o numba makes the test cycle with Travis CI about 13 minutes along, instead of about 3 with numba. I don't consider this a big deal but I thought I should point it out as something to consider for this PR.

This is a VERY HIGH COST to pay.  Can you explain why all six versions of the tests need to be run after every commit to a pull request?  Maybe the BENEFITS from this enhancement are greater than this high cost.  Can you provide us a benefit-cost analysis that shows why this should be the way things should be in the future?

@MattHJensen @talumbau @GoFroggyRun @Amy-Xu @zrisher

martinholmer:
@talumbau, Pull request #799 also has a noticeable decline in code coverage.  What caused this?

@MattHJensen

talumbau:
Not all combinations of Python {3.4, 3.5, 2.7} X {with numba, without numba} have to be run. It's just the default. It can be changed to whatever we want. However, all the versions typically start within a few seconds of each other, so even if we just picked one to run with each PR, say "Python 2.7 w/o numba", that one would take about 12-13 minutes, so the overall Travis CI run time would not be reduced by much, if any. 

I thought this was a desired feature - that is, to make the code runnable without numba. Is that incorrect? The alternative would be to not test the code without numba, but that seems like a bad idea. Consider the case of debugging some issue that is giving a numba error. A developer would like to understand if the issue is really a problem in numba or in their own code. So, the developer runs their latest code without numba. If we are not testing without numba, you could not have the expectation of what should happen at that point. In other words, you might hit an error, but you wouldn't know if the error was already sitting in master, or if it is because of your code changes.

talumbau:
@martinholmer Code coverage is good to go. There were initially 4 untested lines in utils.py, but that function is no longer used so I removed it. The two package-level .py files (utils.py and decorators.py) both remain at 100% coverage, as you can see in the updated report:


https://codecov.io/gh/open-source-economics/Tax-Calculator/compare/8314d09bb989c105b0da6a33fc5f597b71ff52b9...ea5f4b39b939a134178ec71e1920d330dfe1e6c0

(the Codecov report at the top of this PR is based on the initial commits)

martinholmer:
@talumbau said:
> Not all combinations of Python {3.4, 3.5, 2.7} X {with numba, without numba} have to be run. It's just the default. It can be changed to whatever we want. However, all the versions typically start within a few seconds of each other, so even if we just picked one to run with each PR, say "Python 2.7 w/o numba", that one would take about 12-13 minutes, so the overall Travis CI run time would not be reduced by much, if any.

So, you're saying the six tests run in parallel, right?  And the 12-13 minutes is just the Tax-Calculator running at normal Python-interpreted speed, right?

If my understanding above is correct, then I would hope the without-numba tests could be made optional.  Your point about needing the without-numba tests when debugging is valid, but most of the pull-request commits are not doing debugging.  I don't know enough about how Travis-CI works to know if the without-numba tests can be run only when requested.

@MattHJensen @GoFroggyRun @Amy-Xu @zrisher

talumbau:
> So, you're saying the six tests run in parallel, right? And the 12-13 minutes is just the Tax-Calculator running at normal Python-interpreted speed, right?

Yes and yes.

> If my understanding above is correct, then I would hope the without-numba tests could be made optional. Your point about needing the without-numba tests when debugging is valid, but most of the pull-request commits are not doing debugging. I don't know enough about how Travis-CI works to know if the without-numba tests can be run only when requested.

The point about debugging is more about how one would reason about the code if a problem occurred at some point in the future. So for example, when debugging https://github.com/open-source-economics/Tax-Calculator/issues/405, a first step would have been to run the code w/o numba (if such a feature had been available at the time), hoping to get a clean test report. This would prove that the real trouble in that issue was numba behavior. However, unless we test w/o numba in each merge to master, we would not know what to expect when running the tests w/o numba in the future.

But, there is always a balance to be struck on these issues. This is a good problem to have (deciding time per testing cycle vs. amount of testing per PR). 

I have a suggestion on how to proceed based on a brand new Travis CI feature - [cron jobs](https://docs.travis-ci.com/user/cron-jobs/)! This was just launched last month and it seems like the right answer for our issue. My suggestion is that we run the "without numba" tests every night and keep the current tests on all PRs the same. So, if something got in to master that makes the test fail w/o numba, we would know by the next morning that something bad happened the previous day. How does that sound?

martinholmer:
@talumbau said:
> But, there is always a balance to be struck on these issues. This is a good problem to have (deciding time per testing cycle vs. amount of testing per PR).
>
> I have a suggestion on how to proceed based on a brand new Travis CI feature - cron jobs! This was just launched last month and it seems like the right answer for our issue. My suggestion is that we run the "without numba" tests every night and keep the current tests on all PRs the same. So, if something got in to master that makes the test fail w/o numba, we would know by the next morning that something bad happened the previous day. How does that sound?

Your suggestion is brilliant!   I'm all in favor of that approach.

@MattHJensen

Amy-Xu:
Sounds great! 👍

MattHJensen:
@martinholmer said: 

> Your suggestion is brilliant!

@Amy-Xu said:

> Sounds great! 👍

Agreed!

martinholmer:
T.J., While you're working in the `test_decorators.py` file, don't forgot issue #780.

@talumbau  @MattHJensen

talumbau:
OK, this PR does as follows now:

 - removes Numba as a hard dependency from Tax-Calculator
 - tests w/o Numba in Travis CI cron jobs which run daily and not connected to a PR so we will get a notification if they fail but PRs still test fast
 - adds test coverage to `decorators.py` (not 100% but a marked improvement)

MattHJensen:
Thanks @talumbau! This looks good to me. 

cc @martinholmer @GoFroggyRun

martinholmer:
@talumbau,  Thanks for finding a solution to the long no-numba test run times.
Shall I go ahead and merge this pull request #799 into the master branch?

@MattHJensen

talumbau:
I'll merge now. It appears that cron jobs are running at 10am Central time daily, so I'm curious to see if this will run as expected.

martinholmer:
@talumbau said:
> I'll merge [#799] now. It appears that cron jobs are running at 10am Central time daily, so I'm curious to see if this will run as expected.

What?  I thought your idea was to run the cron jobs in the middle of the night?  Why run them during the business day?

@MattHJensen

talumbau:
The interface for this new Travis CI feature is minimal. the options for how to run the cron jobs are just "monthly, weekly, daily". When I turned the feature on, I picked "daily". I assumed there would be an additional config to set the time. Apparently, it uses the current time when the job is created as the start time for the daily job. There is no interface element to pick the time of day. It's been running this way for the last week, although obviously the cron job was running exactly the same way as our normal PRs. This will be the first time it is doing something differently.

martinholmer:
@talumbau said:
> The interface for this new Travis CI feature is minimal. the options for how to run the cron jobs are just "monthly, weekly, daily". When I turned the feature on, I picked "daily". I assumed there would be an additional config to set the time. Apparently, it uses the current time when the job is created as the start time for the daily job. There is no interface element to pick the time of day. It's been running this way for the last week, although obviously the cron job was running exactly the same way as our normal PRs. This will be the first time it is doing something differently.

Thanks for the explanation.  Maybe they will add more flexibility sometime in the future.

PR #800, opened 2016-06-22
Remove inappropriate _exact code

martinholmer:
The value of `_exact` is always zero in the Tax-Calculator.  The removed code uses three variables that are not in Records.VALID_READ_VARS or in the 2009 or 2010 IRS-SOI PUF file (and therefore, never in the `puf.csv` input file).  The removed code executes only when `_exact` is nonzero.  Given all this, it would seem that the removed code could never produce a meaningful result.  So, it seems prudent to reduce the complexity of the code in the `functions.py` file by removing this unused code.  Given that `_exact` is always zero in Tax-Calculator, this pull request has no effect on the unit or validation test results.


feenberg:
On Wed, 22 Jun 2016, Martin Holmer wrote:

> 
> The value of _exact is always zero in the Tax-Calculator. The removed code uses
> variables that are not in Records.VALID_READ_VARS or in the 2009 or 2010 IRS-SOI PUF
> file (and therefore, never in the puf.csv input file). The removed code executes only
> when _exact is nonzero. Given all this, it would seem that the removed code could
> never produce a meaningful result. So, it seems prudent to reduce the complexity of
> the code in the functions.py file by removing this unused code. Given that _exact is
> always zero in Tax-Calculator, this pull request has no effect on the unit or
> validation test results.

The function of _exact in taxcalc.sas is to make debugging easier. It uses 
the entries in form 6251 for deductions and capital gains, instead of 
copying the values from Schedules A and D. It uses E59560 (EIC earnings) 
instead of wages plus self-employment. That reduces the number of 
unbalanced returns and means that remaining unbalanced returns are 
probably bugs, rather than returns with hidden adjustments. Given the 
nature of the OSPC use of the python calculator, _exact is probably not 
useful.

dan

martinholmer:
Dan @feenberg said:
> The function of _exact in taxcalc.sas is to make debugging easier. It uses
the entries in form 6251 for deductions and capital gains, instead of
copying the values from Schedules A and D. It uses E59560 (EIC earnings)
instead of wages plus self-employment. That reduces the number of
unbalanced returns and means that remaining unbalanced returns are
probably bugs, rather than returns with hidden adjustments. Given the
nature of the OSPC use of the python calculator, _exact is probably not
useful.

Dan,  Thanks for the useful description of the role of the `_exact` variable in your taxcalc.sas program.

@MattHJensen @Amy-Xu @GoFroggyRun @zrisher

martinholmer:
Commit 636af7f in pull request #800 removes two uses of `_exact`  in the functions.py file.  In the AMTI function, there was a redundant use of `_exact`.  In the NumDep function, when `_exact == 1` (which in the default operation of Tax-Calculator, was never true), the `c59560` variable was set to the `e59560` variable instead of the `_earned` variable.  This logic is clearly wrong because the largest components of `_earned` are wage indexed in the blowup, while `e59560` was blowup by a different factor.  So, commit 636af7f cleans-up and simplifies the code without any changes in Tax-Calculator unit or validation test results, all of which assume `_exact` is zero.

martinholmer:
Commit 8799d15 removes the unused `_fixeic` variable, which makes the `e59560` variable unused.  See the discussion above about why `e59560` was not a desirable variable given that it was blownup with an inappropriate factor.  The `e59560` variable will remain in the Records.VALID_READ_VARS set until a new `puf.csv` file, which does not contain `e59560`, can be distributed.

martinholmer:
Pull request #800 resolves the ambiguity in the meaning of the unused `_exact` variable in the Tax-Calculator code.  Before pull request #800, `_exact` had two meanings: (1) use alternative variables, and (2) suppress "stair-step-smoothing" calculations (in order to produce exact income tax liabilities at the cost of potentially unrealistic marginal tax rates).  After pull request #800, the first meaning has been eliminated from the code, leaving only the second meaning in which exact means no smoothing approximations.  This pull requests allows Tax-Calculator to generate exact income tax liabilities when `_exact` is one, which will be useful in cross-model validation and other uses of Tax-Calculator.

@MattHJensen @feenberg @zrisher @Amy-Xu @GoFroggyRun

PR #777, opened 2016-06-10
Charity Contribution Parameter Issue

GoFroggyRun:
This PR resolves parameter naming and description issue discussed in #372. 

In particular, both [Pub 526](http://www.irs.gov/pub/irs-pdf/p526.pdf) and [IRS Charitable Contribution Deductions](http://www.irs.gov/Charities-&-Non-Profits/Charitable-Organizations/Charitable-Contribution-Deductions) indicate that there's no particular cap for charity cash contribution, while there is a general 50% cap for total charity contribution. Thus I suggest we remove the cap for cash contribution and parametrize the general 50% cap. This PR makes no change toward any tests, validations or reforms result. 

In the webapp-public page, it seems that we are using the old parameter name [here](https://github.com/OpenSourcePolicyCenter/webapp-public/blob/master/webapp/apps/taxbrain/models.py#L166). We might want to rename in the webapp code as well to prevent any error.

@MattHJensen

PR #804, opened 2016-06-28
Simplify ItemDed function logic other than charity

martinholmer:
This pull request attempts to simplify ItemDed function logic.  One goal is to eliminate the use of the `puf` variable in the ItemDed function and the use of E-variables not documented in 2009 IRS-SOI PUF documents.  It might be true that future development will add some of these undocumented variables.

Note that this pull request does not change any charity itemized-deduction parameters or logic, which are known to be incorrect (see the conversation about pull request #777).

This pull request does not change any unit test results (apart from one case of minor rounding error) and does not change an validation test results.

This pull request #804 does not need to be merged if pull request #805 is merged.


martinholmer:
Pull request #804 has been closed because all the changes in #804 are included in pull request #805.  The only role of #804 is to show the first five commits of #805 cause no substantive changes in unit test results, validation test result, or reform comparison result.

PR #805, opened 2016-06-29
Correct ItemDed function logic for charity

martinholmer:
This pull request attempts to simplify ItemDed function logic. One goal is to eliminate the use of the puf variable in the ItemDed function and the use of E-variables not documented in 2009 IRS-SOI PUF documents. It might be true that future development will add some of these undocumented variables.

Note that this pull request changes charity itemized-deduction parameter names (as suggested in Sean Wang's pull request #777) and corrects charity logic, which is known to be incorrect as discussed in the #777 conversation.

This pull request #805 incorporates all the changes in closed pull request #804 as well as the charity parameter name changes and charity itemized-deduction logic changes.  The closed pull request #804 shows that the first five commits in this pull request #805 cause no substantive changes in unit test or validation test results.

The changes in tax-calculation results in the `pufcsv_agg_expect.txt` unit test results file reflect the fact that the prior erroneous logic allowed charitable deductions in excess of either the 30% noncash cap and/or the 50% all cap, neither of which were active in the prior code.


martinholmer:
The final commit 04622ac in pull request #805 removes the redundant usage of the `puf` variable from the AMTI function.  This changes in this commit have no effect on unit test, validation test, or reform comparison results.

MattHJensen:
@GoFroggyRun, could you take a look at this?

taxcalc/functions.py
```diff
@@ -207,146 +207,107 @@ def AGI(_ymod1, c02500, c02700, e02615, c02900, XTOT,
...
+    lim30 = min(ID_Charity_crt_noncash * _posagi, e20100)
+    c19700 = min(ID_Charity_crt_all * _posagi, lim30 + e19800)
+    charity_floor = ID_Charity_frt * _posagi  # floor is zero in present law
+    c19700 = max(0., c19700 - charity_floor) * (1. - ID_Charity_HC)
     # Casualty
-    if e20500 > 0:
-        c37703 = e20500 + ID_Casualty_frt * _posagi
-    else:
+    if e20500 > 0.0:  # assume e20500 was subject to a 10% disregard
+        c37703 = e20500 + 0.10 * _posagi  # add back disregarded amount
```
GoFroggyRun: Everything else look good to me, except this part. 

I notice that this "10% assumption" is probably stemmed from the fact that parameter `ID_Casualty_frt ` being `0.1`. That is, if this parameter were changed from `0.1` to say `0.2`, and everything else remain the same. 
Then this will become a "20% assumption", instead of "10% assumption", right? 
If so, should we replace the hardcoded value `0.10` with parameter `ID_Casualty_frt `?

cc @MattHJensen

martinholmer:
Sean said:
> Everything else look good to me, except this part:
```
    if e20500 > 0.0:  # assume e20500 was subject to a 10% disregard
        c37703 = e20500 + 0.10 * _posagi  # add back disregarded amount
```
> I notice that this "10% assumption" is probably stemmed from the fact that parameter `ID_Casualty_frt` being 0.1. That is, if this parameter were changed from 0.1 to say 0.2, and everything else remain the same.  Then this will become a "20% assumption", instead of "10% assumption", right? 

Wrong.

> If so, should we replace the hardcoded value 0.10 with parameter ID_Casualty_frt?

No.  The add back is to (blown-up) 2009 values which had a 10% disregard.  This was one of the bugs that is being fixed in this pull request #805.  If you don't understand this, look closely at the 2009 IRS-SOI PUF documentation for the `E20500` variable.

feenberg:
On Thu, 30 Jun 2016, Martin Holmer wrote:

> 
> Sean said:
>
>       Everything else look good to me, except this part:
>
>     if e20500 > 0.0:  # assume e20500 was subject to a 10% disregard
>         c37703 = e20500 + 0.10 * _posagi  # add back disregarded amount
>
>       I notice that this "10% assumption" is probably stemmed from the fact that parameter
>       ID_Casualty_frt being 0.1. That is, if this parameter were changed from 0.1 to say 0.2, and
>       everything else remain the same. Then this will become a "20% assumption", instead of "10%
>       assumption", right?
> 
> Wrong.
>
>       If so, should we replace the hardcoded value 0.10 with parameter ID_Casualty_frt?
> 
> No. The add back is to (blown-up) 2009 values which had a 10% disregard. This was one of the bugs that is being
> fixed in this pull request #805. If you don't understand this, look closely at the 2009 IRS-SOI PUF
> documentation for the E20500 variable.

Matt is right. E20500 is after the 10% disregard. We back calculate the 
original amount before the 10% floor and put the before amount in the 
database. Then taxbrain calculates the "after" amount using the user 
supplied floor. That way the user can increase the floor and get valid 
results. (Decreasing the floor won't change any calculation). If the 
taxbrain user wants to increase the percentage that wouldn't affect the 
taxpayer data, only the tax on that data.

dan

GoFroggyRun:
@martinholmer @feenberg:

Thanks for your detailed explanations. I now understand that this 10% assumption is associated with the dataset being used, or, more precisely, the tax year of the dataset. 

I have, however, another stupid concern:
If this "10% rule" were changed in 2017, or anytime in the future, to say 20%, then the current tax-calculator won't be compatible with input file of that year, is this quite right? 

If such assumption is associated with the dataset, would it be more appropriate to deal with it in the data repo?

cc @MattHJensen

martinholmer:
Sean said:
> Thanks for your detailed explanations. I now understand that this 10% assumption is associated with the dataset being used, or, more precisely, the tax year of the dataset.

Yes.  Well said.

> I have, however, another concern:
> If this "10% rule" were changed in 2017, or anytime in the future, to say 20%, then the current tax-calculator won't be compatible with input file of that year, is this quite right?

Yes, you are absolutely correct.  That won't happen for at least a decade but it is something that will have to be done, if there is a reform like that, given the current approach we are using in Tax-Calculator.

> If such assumption is associated with the dataset, would it be more appropriate to deal with it in the data repo?

As you can see from his comment in this conversation, Dan thinks the 2009 ten percent disregard has already been added back into the `e20500` variable in the preparation of the `puf.csv` file.  But when I look through the contents of the taxdata repository, I see nothing that suggests the `e20500` variable is different from what it is in the 2009 IRS-SOI PUF.  And furthermore, the `e_variable_info.csv` file calls the `e20500` variable `Sch A: Net casualty or theft loss`.  What does the `Net` mean if it isn't net of the ten percent disregard?  And a third bit of information that suggests Dan's recollection is not right is the nature of the code before this pull request:
```
    # Casualty
    if e20500 > 0:
        c37703 = e20500 + ID_Casualty_frt * _posagi
    else:
        c37703 = 0.
    c20500 = max(0., c37703 - ID_Casualty_frt * _posagi)
```
If the `e20500` variable had been gross (rather than net) with the ten percent disregard being already added back, then the `if` statement above is clearly incorrect: it is adding back the disregard **twice** (although in an incorrect manner).  So, who ever wrote the existing code did not think the disregard had already been added back.  Notice that the bug in the old code means that `c37703` is always equal to `e20500` no matter what the value of `ID_Casualty_frt` is in a reform.  That is the bug that is being fixed in these lines.

The problem with dealing with a possible future reform in the taxdata repo is that then we open the door to the `puf.csv` variables having different values than the ones in the IRS-SOI PUF, which have the exact same names.  To my mind that is chaos.  If there was a decision to use the IRS-SOI PUF variable names, then those variables had better contain the same data as they do in the IRS-SOI PUF data.

martinholmer:
Commit 58cae87 in pull request #805 changes how the value of `ID_Casualy_frt` in the `PUFCSV_YEAR` is specified.  This enhancement, which causes no change in any results, is an attempt to respond to the legitimate concerns expressed by Sean Wang earlier in this pull-request conversation.

GoFroggyRun:
@martinholmer,
LGTM. Thanks for adding the additional parameter `ID_Casualty_frt_in_pufcsv_year`. 

cc @MattHJensen

PR #806, opened 2016-06-29
Remove puf variable from LLC function

martinholmer:
This pull request simplifies LLC function logic, so that the puf variable in the LLC function and the use of E-variables not documented in 2009 IRS-SOI PUF documents can be eliminated. This pull request is part of a series that intends to eliminate the use of the puf variable in the functions.py file.

This pull request does not change any unit test results or any validation test results.


MattHJensen:
LGTM. Thanks @martinholmer!

PR #807, opened 2016-06-29
Handle non-existence of numba package

talumbau:
- `test_force_no_numba` removes the numba module from sys.modules, but
  does so in a way that assumes it exists in the first place. This fix
  handles the case of non-existence of the numba package

### DIFF
--- a/taxcalc/tests/test_decorators.py
+++ b/taxcalc/tests/test_decorators.py
@@ -337,7 +337,7 @@ def test_force_no_numba():
     hasattr(mck, 'jit')
     del mck.jit
     import taxcalc
-    nmba = sys.modules['numba']
+    nmba = sys.modules.get('numba', None)
     sys.modules.update([('numba', mck)])
     # Reload the decorators with faked out numba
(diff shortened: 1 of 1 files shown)


talumbau:
handles the following error from the cron job:

screen shot 2016-06-29 at 10 22 01 am

talumbau:
The good news is that the cron job is working correctly. The bad news is that it did not generate an email on failure of the cron. The docs don't say anything about what should happen in the case of a failure on cron, but I was hoping for some kind of notification.

martinholmer:
@talumbau said:
> The good news is that the cron job is working correctly. The bad news is that it did not generate an email on failure of the cron. The docs don't say anything about what should happen in the case of a failure on cron, but I was hoping for some kind of notification.

Sounds like your on the bleeding edge.

talumbau:
Unless there are any remaining comments/concerns, I'd like to merge this before 10am Central time today.

MattHJensen:
>  I'd like to merge this before 10am Central time today.

sounds good to me.

PR #809, opened 2016-06-29
Correct error in F2441 function logic

martinholmer:
This pull request fixes a logic error in the F2441 function.  The old code included these lines:
```
if MARS == 2 and puf:
    c32880 = e00200p
    c32890 = e00200s
else:
    c32880 = _earned
    c32890 = _earned
```
But this logic is clearly wrong because the `_earned` variable correctly includes self-employment income along with wage-and-salary income, while the `e00200p` and `e00200s` variables include only wage-and-salary income.  This pull request computes `_earned_p` and `_earned_s` in the EI_FICA function and then uses them in the F2441 function to set the values of the `c32880` and `c32890` variables without using the `puf` variable.

Fixing this bug causes very small changes in the two pufcsv unit test results, but no changes in the validation test or the reform comparison results.

This pull request also simplifies ExpEarnedInc function logic, which causes no additional changes in unit test results, validation test results, or reform comparison results.


MattHJensen:
@GoFroggyRun, could you also look at this one?

GoFroggyRun:
LGTM @MattHJensen.

PR #808, opened 2016-06-29
Remove puf variable from AmOppCr function

martinholmer:
This pull request simplifies AmOppCr function logic, so that the puf variable in that function can be eliminated. This pull request is part of a series that intends to eliminate the use of the puf variable in the functions.py file.  The logic changes in this pull request allow users to input either the expenses of each of at most four students or the credit amount.

This pull request does not change any unit test results (apart from a few mtr rounding differences) or any validation test results.

@MattHJensen @feenberg @Amy-Xu @GoFroggyRun @zrisher

PR #813, opened 2016-07-06
Remove puf variable from decorators logic and tests

martinholmer:
Merged pull requests #805, #806, #808, and #809, taken together, remove the usage of the `puf` variable in the `iterate_jit` decorator in the `functions.py` file.  As a result, logic in the `decorators.py` file and tests in the `test_decorators.py` file can be simplified.  This pull request contains some simplifications, but it may be possible to simplify the logic and/or the tests even more.  Suggestions for further simplifications are welcome.  If no concerns are raised or addition simplifications are suggested, it is anticipated that this pull request will be merged into the master branch sometime on Friday, July 8th.

### DIFF
--- a/taxcalc/decorators.py
+++ b/taxcalc/decorators.py
@@ -234,13 +234,12 @@ def iterate_jit(parameters=None, **kwargs):
         kwargs_for_func = toolz.keyfilter(in_args.__contains__, kwargs)
         kwargs_for_jit = toolz.keyfilter(jit_args.__contains__, kwargs)
 
-        # Any name that is a parameter (or the special case 'puf')
+        # Any name that is a parameter
         # Boolean flag is given special treatment.
         # Identify those names here
         dd_key_list = list(Policy.default_data(metadata=True).keys())
         allowed_parameters = dd_key_list
         allowed_parameters += list(arg[1:] for arg in dd_key_list)
-        allowed_parameters.append("puf")
         additional_parameters = [arg for arg in in_args if
                                  arg in allowed_parameters]
         additional_parameters += parameters
--- a/taxcalc/decorators.py
+++ b/taxcalc/decorators.py
@@ -19,7 +19,7 @@ try:
     jit = numba.jit  # pylint: disable=invalid-name
     DO_JIT = True
 except (ImportError, AttributeError):
-    def id_wrapper(*dec_args, **dec_kwargs):
+    def id_wrapper(*dec_args, **dec_kwargs):  # pylint: disable=unused-argument
         """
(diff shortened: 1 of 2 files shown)


martinholmer:
@talumbau,  I'm interested in your views about the status of the `decorators.py` function called `create_toplevel_function_string(args_out, args_in, pm_or_pf, kwargs_for_func={})`, and in particular, the role of the last argument: the `kwargs_for_func` dictionary.  Is this `kwargs_for_func` argument being used in Tax-Calculator?  It doesn't seem to be used in the `functions.py` code, and it is (no longer) being used in the `test_decorators.py` tests.  Does it make sense to remove this `kwargs_for_func` argument?  If not, where is it being used by the Tax-Calculator?

talumbau:
I believe it is correct to say that `kwargs_for_func` is no longer needed. It was kind of a pain to implement in the first place. The typical use case was like this (from `test_decorator.py`):

```
@iterate_jit(parameters=['w'], nopython=True, puf=True)
def Magic_calc5(w, x, y, z, puf):
    a = x + y
    if (puf):
        b = w[0] + x + y + z
    else:
        b = 42
    return (a, b)
```

Traditionally, in `functions.py`, we needed `puf` to have a default value of `True` if the caller did not pass `puf` to the decorated function. Because of the way the jitting behavior happens, the default value of the argument had to be passed in to the decorator instead of the function definition itself, so we ended up with a decorator that had keyword arguments meant for numba (e.g. `nopython=True`) *and* keyword arguments that were meant for the underlying wrapped function. So, it appears that we now only have the case of keyword arguments passed to `iterate_jit` that are meant for numba itself - the main use case being `nopython=True`. So, as long as we don't want to have default values passed to any function in `functions.py`, it should be fine to remove this functionality.

martinholmer:
@talumbau,  Thanks for your quick and comprehensive response to my question about the `kwargs_for_func` argument in the decorators.py file.  I'll work on removing it tomorrow morning.

martinholmer:
Pull request #813 commit edfa18d removes (with T.J.'s approval) the unused `kwargs_for_func` variable from the `decorators.py` file.  The five subsequent commits contain non-substantive changes that bring the `decorators.py` file into pylint compliance.

martinholmer:
Pull request #813 commit de3e8f0 attains complete `decorators.py` code coverage by removing a `try ... except` block that seems to be unneeded now that `jit` is defined even if the `numba` package is not present.

martinholmer:
@talumbau, as the original author of the `decorators.py` file, I would value your review of the proposed changes in pull request #813.

PR #816, opened 2016-07-07
Add the consumption.json file to the manifest

talumbau:
- Adding consumption.json to the manifest so that upon install, this
   JSON file gets installed as well.


martinholmer:
@talumbau,  Thanks for fix this.  I'll try to remember this going forward.

PR #817, opened 2016-07-07
add c28*00 variable declarations

talumbau:
- When functions in functions.py are jitted, there is an implicit
   variable declaration in the generated LLVM. When running in pure
   Python, there are code paths in SchR that can execute such that
   these variables do not exist. This causes tests to fail when running
   w/o numba. We declare the necessary variables at the top of the
   function


talumbau:
The lastest master was failing the cron jobs due to the issue fixed in this PR. Test logs for failed cron jobs here:

https://travis-ci.org/open-source-economics/Tax-Calculator/jobs/143072190

martinholmer:
I have a general question about pull request #817:  why did PR#810 work with Numba, but fail without Numba?  All the c28?00 variables are declared in the Records.CALCULATED_VARS set.  Is it a general rule that a function in the `functions.py` file must pass in as arguments all calculated (c?????) variables used in and returned by the function? If so, please explain to a novice like me exactly why that is so?

Also, unless you answer my last question no, will it not be more efficient to pass in as arguments the already created c28?00 variables (instead of declaring them locally in a redundant manner)?

@talumbau @MattHJensen @GoFroggyRun @Amy-Xu @zrisher

talumbau:
> why did PR#810 work with Numba, but fail without Numba?

When running in "pure Python mode" (so, `numba` is not available), the functions in `functions.py` behave (mostly) just like regular Python functions. In the function below, all code paths lead to the statement `return c`:

```
def foo(a):
    if a > 2:
        c = a + 3
    return c
```

If a > 2, then we reach a statement where `c` is defined, so the statement `return c` can succeed. If we don't hit the line of code declaring `c`, then it's not clear what the meaning of `return c` should be. 

```
>>> foo(4)
7
>>> foo(1)
  File "<stdin>", line 1, in <module>
  File "script.py", line 7, in foo
    return c
UnboundLocalError: local variable 'c' referenced before assignment
```

When we "jit" the function, meaning we compile it with numba, we can get slightly different behavior, because numba is translating the function into similar (but different) language primitives and then compiling those primitives down to machine code. So, for example, the equivalent C function for the foo function above would look something like this:

```
int foo(int a){
    int c;
    if a > 2:
        c = a + 3;
    return c;    
}
```

This is because the C language (and many other compiled languages) insist that we declare locally scoped variables so that the compiler will know the type, and can thus allocate sufficient memory on the call stack for this variable.

```
from numba import jit

@jit(nopython=True)
def foo(a):
    if a > 2:
        c = a + 3
    return c
```

After jitting the function we get this behavior:

```
>>> foo(1)
0
```

which is what one would expect from the C function I describe above.

> Is it a general rule that a function in the functions.py file must pass in as arguments all calculated (c?????) variables used in and returned by the function?

If you wish to *read* the value of any variable, you must pass it as an argument to the function. If you wish to *overwrite* the value of an existing variable by returning it from a function in `functions.py`, you may declare the variable inside the function, assign some value to it and return. If there are code paths in the function where the variable is not declared but is still returned, this will likely be OK because Numba will allocate a default value for the variable before compiling the function.  If there are code paths in the function where the variable is not declared but is still returned *and* you are running in pure Python mode, the function will raise an exception when it hits that code path  (because of the rules of the Python language).

> Also, unless you answer my last question no, will it not be more efficient to pass in as arguments the already created c28?00 variables (instead of declaring them locally in a redundant manner)?

I'm not sure if my answer above answers your question, so let me know if it did. Since you have now told me that the c28?00 variables already exist, it would be most efficient to pass them as arguments, because you are correct that we are allocating memory with my added lines of code. I can make that change, but I'll wait to hear from you to see if that is your desired way to move forward.

martinholmer:
@talumbau,  Thanks for your detailed explanation of this situation regarding the new SchR function.

My take-away from your explanation is that, in order for a function in the `functions.py` file to work correctly, that following must be true: **if a variable is used in the function's calculations or if it is returned by the function, then it must appear in that function's argument list**.  Is that a correct way to summarize what you said?

If so, then just add the c07200 and the c28?00 variables to the SchR argument list.  They are already allocated and set to zero in the Records class because they are in the CALCULATED_VARS set.

martinholmer:
@talumbau,  After looking at other functions in the `functions.py` file, I see my summary of your explanation is not correct.  But I would suggest that in pull request #817 you move the setting of the `c07200` and six `c28?00` variables to zero into a new `else:` clause at the bottom of the SchR function.  That approach avoids the duplication of effort inherent in the current version of pull request #817.

talumbau:
Before I make the change, I think I need to understand what the logic of the function should be. For each record where we perform the operations in `SchR`, do we care about the value of the `c28?00` variables as we enter the function? My guess is no since they are not passed in now. If I modify the function to just take `c28?00` variables as arguments and do nothing else, then for a given record, if we don't hit the code that modifies the `c28?00` variables, the values will stay the same. However, in the current code, those records are set to 0 (the default value of the variable created inside the function). So, to be specific, my question is, for a given record, if `_calc_schR and (age_head >= 65 or (MARS == 2 and age_spouse >= 65))` is False, do you want the value of the `c28?00` variables to be set to `0` or do you want them to be whatever value they had prior to the execution of the function (that is, not modified in any way)?

martinholmer:
@talumbau said:
> Before I make the change, I think I need to understand what the logic of the function should be. For each record where we perform the operations in SchR, do we care about the value of the c28?00 variables as we enter the function? My guess is no since they are not passed in now.

That is correct: we don't care at the beginning what the c28?00 values are.

> If I modify the function to just take c28?00 variables as arguments and do nothing else, then for a given record, if we don't hit the code that modifies the c28?00 variables, the values will stay the same. However, in the current code, those records are set to 0 (the default value of the variable created inside the function). So, to be specific, my question is, for a given record, if _calc_schR and (age_head >= 65 or (MARS == 2 and age_spouse >= 65)) is False, do you want the value of the c28?00 variables to be set to 0 or do you want them to be whatever value they had prior to the execution of the function (that is, not modified in any way)?

Set all the c28?00 variables to zero and set the c07200 variable to zero whenever the if clause if False.

Thanks for your patience.  We could streamline this whole process if the AUTHOR of the commit that fails the NO-NUMBA unit tests could get an email about the cron job test failure.  It is the author of the failing commit that is likely to know best how to fix the NO-NUMBA failure.

Is there a way to have the Travis-CI cron job email sent to the commit author?

@MattHJensen

talumbau:
@martinholmer I believe I have added changes that align with the discussion above. Please have a look.

cc @MattHJensen

martinholmer:
@talumbau said:
> I believe I have added changes that align with the discussion above. Please have a look.

Looks good.  Thanks for making these changes.  Let me know one way or another whether this code passes the Travis-CI chron test without numba.

PR #827, opened 2016-07-15
Updated .gitignore to ignore puf.csv file.

rickecon:
Unless this creates problems with the running of Tax-Calculator from TaxBrain, I thought it would be nice for running it on my local machine if git knew to ignore the PUF file in my Tax-Calculator directory structure.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,6 @@ docs/_build/
 
 # Sublime
 *.sublime-*
+
+# ignore puf data files
+puf.csv


martinholmer:
I'm not so convinced this is a good change.  To run the pufcsv unit tests (and others) I temporarily move a copy of `puf.csv` into the top-level directory of the Tax-Calculator source tree.  So, I routinely rely on `git status` to remind me that I've, for some reason, left a copy of `puf.csv` in the source tree.  If I understand the proposed change, the `git status` command would no longer show the stray copy of the `puf.csv` file.  I know we all have different working styles, but for me the proposed change would lower my productivity.  Is there a benefit to this change I'm overlooking?

@rickecon @MattHJensen

talumbau:
The advantage would be that someone would not accidentally add the PUF to their repo via `git add .` and then unintentionally put the PUF in a pull request. That seems like a nice safety feature to prevent the accidental disclosure of the PUF. I am +1 on this change.

PR #830, opened 2016-07-18
Add new test to test_functions.py

martinholmer:
This pull request adds a new test that ensures each variable in the Records.CALCULATED_VARS set is actually calculated in the functions.py code, rather than being used at its initialized value of zero.  This test simply enforces, going forward, the situation achieved after the merge of pull request #826.  There is no change in tax-calculating logic in this pull request.


martinholmer:
@talumbau,  I need some advice from someone who knows more about the Python `ast` package than I do.  As you can see, pull request #830 passes all the unit tests under Python 2.7, but fails in the new test under Python 3.4 and 3.5.  My basic problem is that I can't find much (useful) documentation on the `ast` package, so I'm in the dark about how this package has (apparently) changed in incompatible ways between Python 2.7 and Python 3.4+.  Can you point me to some useful documentation?  Even after I figure out the differences between the 2 and 3 versions of the `ast` package, how do I write test code to do one thing under Python 2.7 and another thing under Python 3.4+?  Any help on this matter would be greatly appreciated.

@MattHJensen

martinholmer:
@talumbau,  I've figured out one way to handle the differences in the `ast` package between Python 2 and Python 3 in pull request #830.  Now the new test in `test_functions.py` passes under all three versions of Python used in the Tax-Calculator Travis-CI process.

If you can suggest a better approach to handling this issue, I'd be interested in your thoughts.

@MattHJensen

taxcalc/tests/test_functions.py
```diff
@@ -9,9 +9,89 @@
...
 import re
-from taxcalc import IncomeTaxIO  # pylint: disable=import-error
+from taxcalc import IncomeTaxIO, Records  # pylint: disable=import-error
 from io import StringIO
 import pandas as pd
+import ast
+
+CUR_PATH = os.path.abspath(os.path.dirname(__file__))
+FUNCTIONS_PY_PATH = os.path.join(CUR_PATH, '..', 'functions.py')
+PYTHON3 = sys.version_info >= (3, 0)
```
talumbau: suggest using `six.PY2` as we already depend on `six`:

https://pythonhosted.org/six/

taxcalc/tests/test_functions.py
```diff
@@ -9,9 +9,89 @@
...
+        self.generic_visit(node)
+        return (self.fnames, self.fargs, self.cvars, self.rvars)
+
+    def visit_FunctionDef(self, node):  # pylint: disable=invalid-name
+        """visit FunctionDef node"""
+        self.fname = node.name
+        self.fnames.append(self.fname)
+        self.fargs[self.fname] = list()
+        for anode in ast.iter_child_nodes(node.args):
+            if PYTHON3:
```
talumbau: Here you could just do `if six.PY3`

martinholmer:
@talumbau, Thanks for the helpful suggestions about how to make pull request #830 better.  I've used the `six` package to determine which version of Python is running the new test.  And I've incorporated your suggested addition to the new test's docstring to make the purpose of the test more clear.  And finally, I've revised the test's error-message handling to be what you suggest in pull request #831 in order to make the error message standout visually.  In a subsequent pull request I plan to switch over to your suggested ValueError approach in other tests.  Thanks again for these useful suggestions.

@MattHJensen @Amy-Xu @GoFroggyRun @zrisher

PR #832, opened 2016-07-19
Add arguments and remove unused returned variables from functions.py

martinholmer:
This pull request was developed using a script that will soon be added as a new test in the test_functions.py file.  This script checks that each variable in Records.CALCULATED_VARS that is calculated in and returned from a function is an argument of that function.  There were a large number of variables that failed this test on the master branch.  The revisions in this pull request eliminate all those failures in one of two ways: (1) if the calculated-and-returned variable was used in subsequent tax calculations, then that variable was added to the list of that function's arguments; or (2) if the calculated-and-returned variable was not used in any subsequent tax calculations, then that variable was converted into a local function variable (that is, not placed in the function's argument list and not placed among the function's returned variables).

These changes have **no effect on tax results** generated by the unit tests, validation tests, or reform comparisons (apart from one tax filing unit moving from one marginal-tax-rate histogram-bin to an adjacent bin, which was probably caused by rounding error).

While the tax results are unchanged, the **time it takes to obtain tax results is noticeably less**.  This statement is based on before and after timings of the two `requires_pufcsv` unit tests, which involve going through the tax calculations many times for different years and for different kinds of marginal tax rates.  Before the changes in this pull request, the average execution time was about 99.5 seconds.  After the changes in this pull request, the average execution time is about 56.5 seconds, which is a reduction in execution time of about 43 percent.

The speed-up seems to be caused by the elimination of many calculated variables that were never used in subsequent tax calculations.  It would appear that these variables were temporary variables used in multiple-step calculations within a function.  It would appear that they were being saved for debugging purposes.  This approach to inspecting intermediate variables is very expensive in execution time (as the timing results above indicate), but they were necessary until very recently.  Now the Tax-Calculator can be run using one or a couple of filing units without the @iterate_jit and @jit decorators being used, which means the Python debugger can be used for debugging (rather than dumping out all the intermediate variables).


martinholmer:
Because pull request #833 contains all the changes in pull request #832 plus the new test promised in #832, pull request #832 is being closed.  The work on #833 is complete and unless there are concerns, objections, or questions concerning these changes, it is anticipated that #833 will be merged into the master branch at the end of the day on Friday, July 22.

PR #835, opened 2016-07-21
Add comments on using debugger to trace tax calculations

martinholmer:
This pull request does what the title says.  The new DEBUGGING variable in decorators.py is False, so that normally there is no change in the operation of Tax-Calculator.  When DEBUGGING is temporarily set to True, the debugger will stop execution and begin an interactive debugging session at the start of the Calculator.calc_one_year() method.  The most effective way to use this option is to use an input file (via the inctax.py interface to Tax-Calculator) that contains only one or just a few filing units.

### DIFF
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -13,7 +13,6 @@ from .records import Records
 from .behavior import Behavior
 from .growth import Growth
 from .consumption import Consumption
-from .decorators import DEBUGGING
 # import pdb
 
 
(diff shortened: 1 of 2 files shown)


talumbau:
I am -1 on this change. There are legitimate reasons to put additional logging information in the package, but adding code to the package to trigger the debugger is different from that use case and very unusual. Typically a user's debug session code would not be part of the package code.

martinholmer:
@talumbau said:
> There are legitimate reasons to put additional logging information in the package, but adding code to the package to trigger the debugger is different from that use case and very unusual. Typically a user's debug session code would not be part of the package code.

Is the latest version better?  If not acceptable, how would you suggest doing this?

martinholmer:
Pull request #835 now changes no Tax-Calculator logic; it only adds a few comments to suggest one way of invoking the Python debugger.

Issue #171, opened 2015-03-09
Medical floor rate for medical deduction in itemized deduction


Amy-Xu:
`
c17750 = .075*_posagi;             
c17000 =  max(0,e17500-c17750);
`

In the original snippet for medical deduction, the floor rate was hard coded as 7.5%. (Later on I parameterized it as ID_medical_frt.) However starting from the beginning of 2013, this rate increases to 10% for most taxpayers and remains as 7.5% for taxpayers or their spouses no younger than 65.  http://www.irs.gov/taxtopics/tc502.html

I changed this rate from 7.5% to 10% in this pull request #169 , but have not incorporated the exceptions for elderly taxpayers yet. 

Need to find a way to separate taxpayers by the age of 65.

MattHJensen:
We can solve this with the CPS match

martinholmer:
Issue #171 has been resolved by the merger into the master branch of pull request #839.

PR #169, opened 2015-03-09
Added long names and descriptions for new parameters

Amy-Xu:
Here are the main revisions on this branch:

1. added descriptions for the hair cut parameters of several adjustments
2. added descriptions for itemized deduction related parameters
3. changed the parameter name of itemized deduction phase out start from ID_pe to ID_ps; changed the description of ID_crt to be more accurate according to the current law.

ready to merge upon review by @MattHJensen

PR #839, opened 2016-07-25
Attempt to resolve bug identified in issue 171

martinholmer:
@Amy-Xu said during March 2015 in issue #171:
> In the original [code] for medical deduction, the floor rate was hard coded as 7.5%. (Later on I parameterized it as ID_medical_frt.) However starting from the beginning of 2013, this rate increases to 10% for most taxpayers and remains as 7.5% for taxpayers or their spouses no younger than 65. http://www.irs.gov/taxtopics/tc502.html
>
> I changed this rate from 7.5% to 10% in this pull request #169 , but have not incorporated the exceptions for elderly taxpayers yet.
>
> Need to find a way to separate taxpayers by the age of 65.

This pull request attempts to fix this bug in a way that requires the least amount of change in Tax-Calculator and TaxBrain.  Fixing this bug, as Amy indicated above, increases the number of elderly itemizers, and therefore, slightly reduces aggregate income tax liability in the 2013-2016 period; there is no change for 2017+.  The decline in 2016 income tax liability is about $1.1 billion, which is less than 0.08 percent of aggregate income tax liability.  This small bug fix also causes changes in some marginal tax rates and in some reform comparison results.

@MattHJensen @Amy-Xu @feenberg @GoFroggyRun @zrisher


taxcalc/current_law_policy.json
```diff
@@ -902,6 +902,21 @@
...
+        "description": "Elderly taxpayers have this fraction added to the value of _ID_Medical_frt.",
+        "irs_ref": "Form 1040 Schedule A, line 3, in-line. ",
+        "notes": "This addon cannot be decreased to lower than -0.025 due to limited data on elderly non-itemizers.",
+        "start_year": 2013,
+        "col_var": "",
+        "row_var": "FLPDYR",
+        "row_label": ["2013", "2014", "2015", "2016", "2017"],
+        "cpi_inflated": false,
+        "col_label": "",
+        "value": [-0.025, -0.025, -0.025, -0.025, 0.0],
```
Amy-Xu: I guess I'm a bit confused here. So after 2016, there's no rate add-on anymore?

martinholmer:
@Amy-Xu said:
> I guess I'm a bit confused here. So after 2016, there's no rate add-on anymore?

My understanding of the [IRS document you cited](http://www.irs.gov/taxtopics/tc502.html) is that the add-on becomes zero beginning in 2017.  That is, the floor becomes 10% of AGI for everybody beginning in 2017.  Did I misunderstand current-law policy or did I specify it incorrectly in the `current_law_policy.json` file?

martinholmer:
Unless there are other questions, comments, or concerns regarding Tax-Calculator pull request #839, it will be merged into the master branch sometime after the end of the day on Wednesday, July 27th.

@MattHJensen @Amy-Xu @feenberg @GoFroggyRun @zrisher

Amy-Xu:
@martinholmer You're right on that parameter. This PR looks good to me!

Issue #277, opened 2015-05-27
params in params.json should be ordered by appearance in functions.py


MattHJensen:
The parameters would be much easier to find in `params.json` if they were in the same order as they appear in `functions.py`.

martinholmer:
Issue #277 has been resolved by merged pull request #845.

PR #845, opened 2016-07-27
Reorder parameters in current_law_policy.json file

martinholmer:
This pull request resolves issue #277 by putting the policy parameters in the order they are first used in the functions.py file.  This is a cosmetic change that has no effect on tax-calculating logic, and therefore, causes no change in results from the unit tests, validation tests, or reform comparisons.


MattHJensen:
Thanks Martin!

PR #847, opened 2016-07-28
Move Behavior.response call out of Calculator.diagnostic_table

martinholmer:
This pull request removes the call of the Behavior.response(base_calc, reform_calc) function from the Calculator.diagnostic_table method.  This is being done for several reasons.  First, computing the behavioral response to a policy reform has no logical connection to preparing a diagnostic table of summary statistics from a single Calculator object.  And second, this removal will facilitate the eventual movement of diagnostic_table from being a Calculator method to being a utility function, as envisioned in issue #213.


martinholmer:
This pull request #847 is now complete and will soon be followed up by another pull request that converts the Calculator.diagnostic_table method into a utility function.

I've looked briefly at the other OSPC repositories and did not see any calls to the Calculator.diagnostic_table method, so I'm assuming the changes to that method in this pull request #847 will not cause any problems for other repositories.  I'm I correct about that?

@MattHJensen @talumbau @feenberg @Amy-Xu @GoFroggyRun @zrisher

PR #848, opened 2016-07-28
Move diagnostic table logic from calculate.py to utils.py

martinholmer:
This pull request moves diagnostic table logic from the Calculator class to utility functions as envisioned in issue #213.  The new create_diagnostic_table utility function does not change the Calculator object passed to it (per #213), but extracts aggregate, weighted information for the current_year and organizes that summary information in a Pandas DataFrame.

This pull request does not implement a second utility function envisioned in issue #213, which was raised in April 2015 before the addition of the Behavior class.  The main reason this is so is that it is not clear how to handle in the same multi-year function two very different cases: one where the Calculator object has no behavioral responses and the other where the Calculator object does have behavioral responses, and therefore, needs a baseline (usually current-law policy) calculator with which to compute the behavioral responses to a reform.  This second utility function will be developed in a future pull request.

Note that this pull request #848 is build on top of, and therefore, include all pull request #847.
The actual move of the diagnostic table logic is in the final commit a90475c.


martinholmer:
Unless there are questions, concerns or objections, pull requests #847 and #848 will be merged into the master branch at the end of the day on Monday, August 1.

Issue #132, opened 2015-02-20
Webapp users should be able to change the number of tax brackets


MattHJensen:
Many tax proposals increase or decrease the number the number of tax brackets. Decreasing the number of tax brackets is most common, and the webapp currently allows users to decrease the number of brackets by zeroing out thresholds. Some users will want to increase the number of brackets, and zeroing out thresholds to decrease the number of brackets is not particularly elegant. We should extend taxcalc and TaxBrain to accommodate this.

MattHJensen:
This has not proven to be high-priority. @GoFroggyRun will need to add a tax bracket as part of his work on https://github.com/open-source-economics/plot_concepts/issues/3, but there is not a need right now for user-defined numbers of taxbrackets in Tax-Calculator and TaxBrain right now.

feenberg:
On Thu, 4 Aug 2016, Matt Jensen wrote:

> 
> This has not proven to be high-priority. @GoFroggyRun will need to add a tax bracket
> as part of his work on open-source-economics/plot_concepts#3, but there is not a need
> right now for user-defined numbers of taxbrackets in Tax-Calculator and TaxBrain right
> now.

It is certainly easy enough to use fewer brackets - just repeat the last 
bracket info. But additional brackets are not possible. We could just add 
an extra bracket that repeats the currentl last bracket to the standard 
web page if we wanted to give the opportunity to add an additional bracket 
without spending a lot of time developing a mechanism for an arbitraty 
number of brackets.

One additional bracket is a characteristic of many proposed reforms in the 
past.

dan

MattHJensen:
> We could just add an extra bracket that repeats the currentl last bracket to the standard web page if we wanted to give the opportunity to add an additional bracket without spending a lot of time developing a mechanism for an arbitrary number of brackets.

@gofroggyrun, could you keep this in mind when you are adding an extra bracket to tax-calculator fpr the plot work--that the default should be set to duplicate the existing top rate? Also, once the functionality is in tax-calculator, could you open an issue on webapp-public requesting that it be added to TaxBrain as well?

GoFroggyRun:
@feenberg:
>
We could just add an extra bracket that repeats the currentl last bracket to the standard web page if we wanted to give the opportunity to add an additional bracket without spending a lot of time developing a mechanism for an arbitrary number of brackets.

Thanks for your thoughtful suggestion on dealing with extra tax brackets. I'll add such feature to tax-calculator.

@MattHJensen:
>
[...]

PR #858, opened 2016-08-07
Added extra tax bracket for II and CG

GoFroggyRun:
Per discussions in issue #132, this PR adds extra tax bracket on top of the current last bracket for individual income (II) and capital gain (CG), including AMT CG, respectively. 

The default value of these added parameters duplicates the existing top rates and thresholds, and thus there are no tax logic change. 

Test suites have been modified as well to incorporate these added parameters, and there're no change in any current diff files. I plan to add more tests regarding these extra parameters in another PR soon. 

@MattHJensen @martinholmer @feenberg Comments, concerns or remarks would be appreciated.


GoFroggyRun:
@talumbau T.J., I'm confused by the Travis CI test failure, since I wasn't able to track the error message. Could you please advice what's going on?

talumbau:
@GoFroggyRun there is something wrong with Travis CI this morning - at least, as far as I can tell. When I click on the link to look at the build logs, here is what I see:

screen shot 2016-08-08 at 9 41 37 am

taxcalc/functions.py
```diff
@@ -540,9 +540,17 @@ def TaxGains(e00650, c01000, c04800, c23650, p23250, e01100, e58990,
         else:
             addtax = 0.
 
+        if c24540 > CG_thd_xtr[MARS - 1]:
+            addtax = (CG_rt_xtr - CG_rt3) * c24517 + addtax
+        elif c24540 <= CG_thd_xtr[MARS - 1] and _taxinc > CG_thd_xtr[MARS - 1]:
+            addtax = addtax + (CG_rt_xtr - CG_rt3) * \
+             min(_dwks21, _taxinc - CG_thd_xtr[MARS - 1])
```
talumbau: I think this is the offending line. One way to make this PEP8 compliant would be to format this line in the same way that lines 538-539 are formatted.

GoFroggyRun: Thanks for helping me locating the issue T.J. @talumbau!

MattHJensen:
Thanks @GoFroggyRun! I wonder if Xtra is the right thing to append to the name, given that a user will need to modify both this rate and the rate below it in order to raise the top rate. Perhaps it should just be given the next number. E.g., `_AMT_CG_rt_4` instead of `_AMT_CG_rt_xtr`

MattHJensen:
Something else I've been thinking about is whether we should add a function somewhere that automatically increases this rate if the rate below it is increased. 

That would maintain backward compatibility. Otherwise someone whose script raises the top rate will be broken by this PR. 

I'm thinking we should wait until @martinholmer is back from vacation before merging this PR, i trust you can still use it for your widget work in the meantime, @GoFroggyRun, right?

feenberg:
On Tue, 9 Aug 2016, Matt Jensen wrote:

> 
> Something else I've been thinking about is whether we should add a function somewhere
> that automatically increases this rate if the rate below it is increased.
> 
> That would maintain backward compatibility. Otherwise someone whose script raises the
> top rate will be broken by this PR.

Doesn't this depend on the default threshold for the optional bracket? If 
the threshold is set at infinity, then rate 5 can be changed without 
changing rate 6.

MattHJensen:
> If the threshold is set at infinity, then rate 5 can be changed without changing rate 6.

That is much more clever. Thanks @feenberg. cc @GoFroggyRun

GoFroggyRun:
@MattHJensen said:
>
 I wonder if Xtra is the right thing to append to the name, given that a user will need to modify both this rate and the rate below it in order to raise the top rate. Perhaps it should just be given the next number.

Thanks for your comments. I have modified names of these extra parameters.

@feenberg said:
>
If the threshold is set at infinity, then rate 5 can be changed without changing rate 6.

Thank you so much for your suggestion Dan! This is a much more clever strategy. I have updated these thresholds, as well as reverted modifications on all tests.

MattHJensen:
@GoFroggyRun, do you think the rates should be set to 1 or infinity, in parallel to the brackets being set to infinity, so that raising rt7 doesn't bump into rt8? 

@martinholmer, I think @GoFroggyRun and I have this covered now. I'll merge once we get these last changes. Thanks for the tip about the rates on the phone.

GoFroggyRun:
@MattHJensen said:
>
do you think the rates should be set to 1 or infinity

Setting these two rates to one makes more sense to me.

MattHJensen:
Thanks @GoFroggyRun. Merging.

PR #867, opened 2016-08-18
Cap benefits from deductions at a percent of deductible expenses 

MattHJensen:
This is still work in progress as I am generating results to compare against TPC's estimate from their [analysis](http://www.taxpolicycenter.org/publications/analysis-hillary-clintons-tax-proposals) of Clinton's tax plan. 

In the meantime, I'll welcome anyone's feedback on the implementation.


MattHJensen:
28% cap, exempting charitable contributions, in 2020:

This PR: $7.8 B
TPC: $38 B

I think the reason for the difference is that TPC includes many exclusions in the cap.

Here is their description:

> The 28 percent limit would apply to: all itemized deductions (except for charitable contributions), tax-exempt state and local bond interest, employer-sponsored health insurance paid for by employers or with before-tax employee dollars, health insurance costs of self-employed individuals, employee contributions to defined contribution retirement plans and IRAs, the deduction for income attributable to domestic production activities, certain trade or business deductions of employees, moving expenses, contributions to health savings accounts and Archer MSAs, and interest on education loans.

Here is the only [news article](http://bigstory.ap.org/article/34af7a6fce404c0089290d4cd3cfb7fc/clinton-propose-350-billion-college-affordability-plan) I can find that doesn't just reference the TPC study

> The cost of Clinton's plan would be offset by capping itemized tax deductions for wealthy families at 28 percent, like those taken by high-income taxpayers for charitable contributions and mortgage interest. That proposal, which has long been included in President Barack Obama's annual budget, would raise more than $600 billion in the next decade, according to the Treasury Department.

The Treasury Department [Green Book](https://www.treasury.gov/resource-center/tax-policy/Documents/General-Explanations-FY2017.pdf) (153-154) contains a proposal that is more similar to what TPC analyzed than what the journalist described:

> The income exclusions and deductions limited by this provision would include any tax-exempt
State and local bond interest, employer-sponsored health insurance paid for by employers or with
before-tax employee dollars, health insurance costs of self-employed individuals, employee
contributions to defined contribution retirement plans and IRAs, the deduction for income
attributable to domestic production activities, certain trade or business deductions of employees,
moving expenses, contributions to health savings accounts and Archer MSAs, and interest on
education loans. 

It looks like I need to go back and add some more options to be included under the cap in this PR. 

I plan to also make more options available for the ID_BenefitSurtax, and I will try to do so without breaking existing functionality.

MattHJensen:
This isn't working right yet. When I add charitable contributions back to the cap, the score goes down.

MattHJensen:
From the CBO Budget Options, Option 53, page 33:

image

MattHJensen:
I'd appreciate feedback on this PR now. 

In 2020, a 28% cap on all itemized deductions raises $5.3 billion. Interestingly, if we _exclude_ state and local taxes from the cap, it raises $9.04 billion. More on this below.

The best comparison is CBO, since they analyze this provision alone. (see https://github.com/open-source-economics/Tax-Calculator/pull/867#issuecomment-241563423). They get $15B in 2020. 

As an interesting side note, this relationship is reversed when we compare TC and CBO for limiting the value of itemized deductions to 6% of adjusted gross income, which I model with an itemized deduction benefit surtax with a 6% credit rate and 100% rate. For this analysis, TC gets $27 billion in 2020 and CBO gets only $7 billion. 

Now let's return to the odd behavior where we exclude an itemized deduction from the cap to find that the revenue raised from the cap goes up. After some experimentation, I have identified at least two reasons for this: 

First, some taxpayers have fully reduced their tax liability through itemized deductions. When we add more itemized deductions to the cap, the deductible expenses go up (meaning a higher cap) but the value of the benefit remains unchanged (meaning fewer excess benefits.)

Second, some taxpayers are in the itemized deduction phaseout range, and again, adding more itemized deductions to the cap means that the cap increases faster than the benefit. When I only apply the cap to taxpayers with positive `_combined` and set the itemized deduction maximum phaseout to zero (`_ID_crt`), then adding itemized deductions to the cap uniformly increases revenue. 

Given all of the odd behavior with this feature, I would greatly appreciate others' review, feedback, and suggestions. 

cc @martinholmer, @feenberg, @gofroggyrun, @andersonfrailey

martinholmer:
@MattHJensen, I'm not in the office on Friday morning, but will look at PR#867 this afternoon.

feenberg:
On Fri, 23 Sep 2016, Matt Jensen wrote:

> 
> I'd appreciate feedback on this PR now.
> 
> In 2020, a 28% cap on all itemized deductions raises $5.3 billion. Interestingly, if
> we exclude state and local taxes from the cap, it raises $9.04 billion. More on this
> below.

I don't see way to specify a cap on the taxbrain web page. Also, I am not 
sure what the cap is. 28% of what? Or do you mean that each dollar of 
itemized deductions reduces tax by no more than 28 cents? I don't see a 
straightforward way of doing that calculation, so I would want to know how 
it was being done before commenting.


> 
> The best comparison is CBO, since they analyze this provision alone. (see #867
> (comment)). They get $15B in 2020.
> 
> As an interesting side note, this relationship is reversed when we compare TC and CBO
> for limiting the value of itemized deductions to 6% of adjusted gross income, which I
> model with an itemized deduction benefit surtax with a 6% credit rate and 100% rate.
> For this analysis, TC gets $27 billion in 2020 and CBO gets only $7 billion.
>

The first place I would look would be at the treatment of AMT. Is this 
stacked before or after the AMT? The child credit is also a stacking 
issue.

Also, I do not understand why the default surtax rate is shown as 1.0 (and 
the credit also as 1.0) in the "Surtax on Itemized Deduction Beefit" 
section. Surely the surtax rate should be .0. How is the benefit 
calculated?

> Now let's return to the odd behavior where we exclude an itemized deduction from the
> cap to find that the revenue raised from the cap goes up. After some experimentation,
> I have identified at least two reasons for this:
> 
> First, some taxpayers have fully reduced their tax liability through itemized
> deductions. When we add more itemized deductions to the cap, the deductible expenses
> go up (meaning a higher cap) but the value of the benefit remains unchanged (meaning
> fewer excess benefits.)
>

I can't comment on this till I know what the definition of the cap is.

martinholmer:
@MattHJensen,  I think the questions raised by Dan @feenberg are all relevant.  Also, I would like to see the reform dictionary that you use to simulate such a reform.  Dan's question about where in the sequence of the tax-calculating logic this reform comes in is particularly relevant.  Here is some relevant code you've added:
```
        # Calculate cap value for itemized deductions
        capped_benefit = deductible_expenses * calc.policy.ID_BenefitCap_rt
        # Add the difference between the actual benefit and capped benefit
        # to income tax and combined tax liabilities.
        excess_benefit = np.maximum(benefit - capped_benefit, 0)
        calc.records._iitax += excess_benefit
        calc.records._surtax += excess_benefit
        calc.records._combined += excess_benefit
```
The first statement goes right to one of Dan's questions: "28% of what?"
This statement (so it seems to me) says allowed itemized deductions are only 28% of total itemized deductions under current law.  Is that the way these reform proposals work?

Another issue is that any constraint on allowed itemized deductions in your formulation of the logic does **not** feed back into AMT calculations.  Is that really how these reform proposals are specified?  Or do you need to do this constraining of allowed itemized deductions earlier in the tax-calculating logic sequence.  Why not just apply the 28% limitation in the ItemDed function?

@Amy-Xu @GoFroggyRun @andersonfrailey

martinholmer:
The reform being analyzed in pull request #867 is described by [CBO (Option 53 on page 33)](https://www.cbo.gov/sites/default/files/cbofiles/attachments/49638-BudgetOptions.pdf) as:
```
Limit the tax benefits of itemized deductions to 28 percent of their total value.
```
**A limitation rate of 28% is identical to a hair-cut of 0.72.**

If this statement is correct, then a couple of things logically follow:
(1) pull request #867 is unneeded and should be closed without merging into master; and
(2) the reform CBO simulates should be added to `reforms.json` for comparison.

Am I missing something here?  If not, Tax-Calculator can already simulate these kind of itemized-deduction limitation reforms, and do it in a way that correctly simulates the AMT interactions.

MattHJensen:
@martinholmer and @feenberg, let me explain what I am trying to accomplish:

1) Calculate how much a set of itemized deductions reduce each taxpayers `_iitax` liability using 'ComputeBenefit()'
- The way that 'ComputeBenefit()' works is that we calculate '_iitax' for two seperate calculators. One calculate has the haircut parameters set to 1 for the set of itemized deductions, thereby zeroing out the deductible expenses;  the other calculator has the haircut parameters set to 0 as they are under current law. We call the difference in `_iitax` between these two calculators the "benefit" from that set of itemized deductions. 
- This approach was recommended by @feenberg for use in our itemized deduction surtax function that allows us to limit the value of itemized deductions to x% of AGI. 
- The approach should properly account for all interactions with AMT, standard deduction, and every other provision of the individual income tax code in our model when calculating the value that a taxpayer gets from an itemized deduction. 
- Note that in the snippet below, 'ID_switch' defines which itemized deductions will be included under the cap. 
```
def ComputeBenefit(calc, ID_switch):
    """
    Calculates the value of the benefits accrued from itemizing.
    """
    # compute income tax liability with no itemized deductions allowed for
    # the types of itemized deductions covered under the BenefitSurtax
    no_ID_calc = copy.deepcopy(calc)
    if ID_switch[0]:
        no_ID_calc.policy.ID_Medical_HC = 1.
    if ID_switch[1]:
        no_ID_calc.policy.ID_StateLocalTax_HC = 1.
    if ID_switch[2]:
        no_ID_calc.policy.ID_RealEstate_HC = 1.
    if ID_switch[3]:
        no_ID_calc.policy.ID_Casualty_HC = 1.
    if ID_switch[4]:
        no_ID_calc.policy.ID_Miscellaneous_HC = 1.
    if ID_switch[5]:
        no_ID_calc.policy.ID_InterestPaid_HC = 1.
    if ID_switch[6]:
        no_ID_calc.policy.ID_Charity_HC = 1.
    no_ID_calc.calc_one_year()
    # compute surtax amount and add to income and combined taxes
    # pylint: disable=protected-access
    benefit = np.where(
        no_ID_calc.records._iitax - calc.records._iitax > 0.,
        no_ID_calc.records._iitax - calc.records._iitax, 0.)
    return benefit
```
2) Calculate total deductible expenses for the subset of itemized deductions under the cap. 
- I do this by adding up the relevant c variables for each deduction that we calculate earlier in the calculator. 
```
        deductible_expenses = 0.
        if calc.policy.ID_BenefitCap_Switch[0]:  # Medical
            deductible_expenses += calc.records.c17000
        if calc.policy.ID_BenefitCap_Switch[1]:  # StateLocal
            deductible_expenses += ((1. - calc.policy.ID_StateLocalTax_HC) *
                                    np.maximum(calc.records.e18400, 0.))
        if calc.policy.ID_BenefitCap_Switch[2]:
            deductible_expenses += ((1. - calc.policy.ID_RealEstate_HC) *
                                    calc.records.e18500)
        if calc.policy.ID_BenefitCap_Switch[3]:  # Casualty
            deductible_expenses += calc.records.c20500
        if calc.policy.ID_BenefitCap_Switch[4]:  # Miscellaneous
            deductible_expenses += calc.records.c20800
        if calc.policy.ID_BenefitCap_Switch[5]:   # Mortgage and interest paid
            deductible_expenses += calc.records.c19200
        if calc.policy.ID_BenefitCap_Switch[6]:  # Charity
            deductible_expenses += calc.records.c19700
```
3) Define the `capped_benefit` as "X% of the `deductible_expenses`. 
- In this sentence describing the reform, "Limit the tax benefits of itemized deductions to 28 percent of their total value," I interpret "total value" to refer to the total amount of of deductible expenses. 

4) Calculate the 'excess_benefit' as difference between the amount of benefit that the taxpayer actually received from the itemized deductions from step 1, and the `capped_benefit` from step 3, not allowing `excess_benefit` to be negative. 
```
 excess_benefit = np.maximum(benefit - capped_benefit, 0)
```
5) Add excess_benefit to our various tax variables.
```
        calc.records._iitax += excess_benefit
        calc.records._surtax += excess_benefit
        calc.records._combined += excess_benefit
```

@martinholmer, based on the method in step 1, I believe this approach does properly take into account interactions with AMT. Also, these reforms are designed to make it so that taxpayers in high tax brackets do not get more value from itemized deductions than taxpayers in lower tax brackets. Therefore, I do not believe that modeling a limitation rate of 28% as a haircut of 0.72 is correct. That would reduce itemized deductions for taxpayers in every tax bracket. It would also not take into account differences in the value that taxpayers can actually derive from itemized deductions.

MattHJensen:
@feenberg and @martinholmer, here's a summary: 

Goal: Cap the value to the taxpayer of itemized deductions at X% of total deductible expenses. 

Method for calculating "value to the taxpayer": Run one calculator all the way through with itemized deductions turned off via haircuts. Run another calculator all the way through with itemized deductions unchanged. Compare the difference in individual income tax liability. 

Comments: 

- I believe this takes into account all interactions with AMT, standard deduction, and other provisions of the code. 
- This is not available on TaxBrain yet because we are just working on it.

martinholmer:
@MattHJensen said:
> @martinholmer, based on the method in step 1, I believe this approach does properly take into account interactions with AMT. Also, these reforms are designed to make it so that taxpayers in high tax brackets do not get more value from itemized deductions than taxpayers in lower tax brackets. **Therefore, I do not believe that modeling a limitation rate of 28% as a haircut of 0.72 is correct. That would reduce itemized deductions for taxpayers in every tax bracket.** It would also not take into account differences in the value that taxpayers can actually derive from itemized deductions.

But, as far as I can see, the code in pull request #867 applies the 28% limitation to **all** taxpayers.  Am I wrong?  If so, where is the code that applies the 28% limitation to only those with high incomes?

So, I continue to think #867 is completely unnecessary to simulate the current 28% limitation reform proposals.  Is the following description from 2013 incorrect?
```
Previously published estimates from Citizens for Tax Justice of this proposal
assumed that it would only affect taxpayers with adjusted gross income (AGI)
above $250,000 for married couples and above $200,000 for singles.  Such 
an income threshold was included in the version of the proposal that appeared
in the President’s jobs bill in 2011, the only actual legislative language for
the proposal. However, documents recently released by the administration
make clear that there is no longer such an income threshold. 
```
Above quote is from [this 2013 document](http://ctj.org/pdf/28percentlimit.pdf).

MattHJensen:
@martinholmer, the 28% limitation, as I have intended to model it, would *primarily* apply to individuals in the 33%, 35%, and 39.6% brackets -- in particular, it would apply when the effective tax rate on deductible expenses is greater than 28%. It might affect a few taxpayers who have effective rates on itemized deductions that are higher than their statutory marginal rate. 

As the snippet from CTJ indicates, if we want to allow users to limit who is affected, we could also add an AGI threshold, but I figure we can ignore that for not. 

TPC has the same interpretation of the provision as what I have described above ([page 12](http://www.taxpolicycenter.org/sites/default/files/alfresco/publication-pdfs/2000638-an-analysis-of-hillary-clintons-tax-proposals.pdf)): 

> The 28 percent limit on the value of certain deductions and exclusions would be another form of minimum tax on selected tax preferences. Specifically, higher-income filers would need to compute their income tax liability with and without various excluded forms of income—including the value of employer contributions to health insurance, contributions to retirement plans, and tax-exempt bond interest—plus itemized deductions other than charitable contributions and certain other deductions added to taxable income. If the difference is greater than 28 percent of the value of those items, they would add the difference to tax liability. 

Similarly, this WSJ [says this](http://www.wsj.com/articles/SB10001424127887323916304578402952175617808) about the proposal:

> For example, $100,000 in deductions would be worth up to $39,600 for a very high-income household now, because the top tax rate is 39.6%; the proposal would limit that tax break to $28,000.

I don't think we can model such a new "minimum tax on selected preferences" with existing code (although it does share some characteristics with the surtax on itemized deduction benefits); in particular, it isn't equivalent to cutting 28% off the top of all deductible expenses.

martinholmer:
@MattHJensen said:
> TPC has the same interpretation of the provision as what I have described above (page 12):
>
>> The 28 percent limit on the value of certain deductions and exclusions would be another form of minimum tax on selected tax preferences. Specifically, higher-income filers would need to compute their income tax liability with and without various excluded forms of income—including the value of employer contributions to health insurance, contributions to retirement plans, and tax-exempt bond interest—plus itemized deductions other than charitable contributions and certain other deductions added to taxable income. If the difference is greater than 28 percent of the value of those items, they would add the difference to tax liability.

OK, when it is explained that way, what you are doing makes more sense.

So, we're back to the question of why you can't get a revenue estimate similar to what they get.  Is that because you're not subjecting the income exclusions to the 28% limitation?  In the quote from the TPC above, it seem clear that they are computing "benefits" to be limited as itemized-deduction benefits and excluded-income benefits.  But pull request #867 seems to limit only itemized deduction expenses.  Is that true?  Or have I missed something else?

Also, now that I've read what several other organizations say about this kind of reform proposal, I think your new `BenefitCap` function would be better named `BenefitLimitation`.  To me, cap implies a dollar limit.

@feenberg @Amy-Xu @GoFroggyRun @andersonfrailey

martinholmer:
@MattHJensen, I doubt my observation in this comment on pull request #867 has anything to do with your questions about the size of the revenue effect of a 28% limitation reform, but they seem important enough to make.

Aren't there some missing haircut parameters in the following code?
```
    if calc.policy.ID_BenefitCap_rt != 1.:
        benefit = ComputeBenefit(calc, calc.policy.ID_BenefitCap_Switch)
    # Calculate total deductible expenses under the cap.
        deductible_expenses = 0.
        if calc.policy.ID_BenefitCap_Switch[0]:  # Medical
            deductible_expenses += calc.records.c17000
        if calc.policy.ID_BenefitCap_Switch[1]:  # StateLocal
            deductible_expenses += ((1. - calc.policy.ID_StateLocalTax_HC) *
                                    np.maximum(calc.records.e18400, 0.))
        if calc.policy.ID_BenefitCap_Switch[2]:
            deductible_expenses += ((1. - calc.policy.ID_RealEstate_HC) *
                                    calc.records.e18500)
        if calc.policy.ID_BenefitCap_Switch[3]:  # Casualty
            deductible_expenses += calc.records.c20500
        if calc.policy.ID_BenefitCap_Switch[4]:  # Miscellaneous
            deductible_expenses += calc.records.c20800
        if calc.policy.ID_BenefitCap_Switch[5]:   # Mortgage and interest paid
            deductible_expenses += calc.records.c19200
        if calc.policy.ID_BenefitCap_Switch[6]:  # Charity
            deductible_expenses += calc.records.c19700
        # Calculate cap value for itemized deductions
        capped_benefit = deductible_expenses * calc.policy.ID_BenefitCap_rt
        # Add the difference between the actual benefit and capped benefit
        # to income tax and combined tax liabilities.
        excess_benefit = np.maximum(benefit - capped_benefit, 0)
        calc.records._iitax += excess_benefit
        calc.records._surtax += excess_benefit
        calc.records._combined += excess_benefit
```
Only two of the seven haircut parameters are used; why not all seven?  That would be the correct way to simulate the unlikely case in which some haircut reform provisions were included in the same reform as a 28% limitation provision.

martinholmer:
@MattHJensen,  When computing `deductible_expenses` in the new `BenefitCap` function, shouldn't the itemized-deduction phase-out calculations be taken into account?

MattHJensen:
@martinholmer said:

> So, we're back to the question of why you can't get a revenue estimate similar to what they get. Is that because you're not subjecting the income exclusions to the 28% limitation? In the quote from the TPC above, it seem clear that they are computing "benefits" to be limited as itemized-deduction benefits plus excluded-income benefits. But pull request #867 seems to limit only itemized deduction expenses. Is that true? 

Yes, that is true. We should not be matching TPC's estimate right now for that reason. I am planning on adding a few excluded income benefits in a follow on PR, but many of the benefits that TPC includes we don't currently model, such as the exclusion for employer provided health insurance. The ones I will add are non-taxable social security benefits and tax-exempt bond interest. 

While we shouldn't get close to TPC's number, I would expect us to be closer to CBO's. The difference w.r.t CBO is my main source of concern for this PR. 

@martinholmer said:

>  now that I've read what several other organizations say about this kind of reform proposal, I think your new BenefitCap function would be better named BenefitLimitation. To me, cap implies a dollar limit.

Yes, that is a good suggestion. I will add a commit with that change. 

@martinholmer said:

> When computing deductible_expenses in the new BenefitCap function, shouldn't the itemized-deduction phase-out calculations be taken into account?

I don't think so. The effect of the itemized deduction phaseout is captured by `ComputeBenefit()`. If the phaseout reduces the benefit a taxpayer receives from itemized deductions, that taxpayer will have a lower `excess_benefit` as a result.

feenberg:
This is the correct general approach. I have no idea if other modelers use 
it - there are shortcuts that would mostly work, but not for people who 
straddle tax brackets, or have excess credits, or move on or off the AMT. 
It wouldn't surprise me if some modelers use the shortcuts. However, I 
would not expect the shortcuts to explain nearly the deviation that Matt 
reported. such as the negative revenue. That is something beyond my 
understanding.

dan

feenberg:
On Sat, 24 Sep 2016, Martin Holmer wrote:

> 
> @MattHJensen said:
>
>       @martinholmer, based on the method in step 1, I believe this approach does
>       properly take into account interactions with AMT. Also, these reforms are
>       designed to make it so that taxpayers in high tax brackets do not get more
>       value from itemized deductions than taxpayers in lower tax brackets.
>       Therefore, I do not believe that modeling a limitation rate of 28% as a
>       haircut of 0.72 is correct. That would reduce itemized deductions for
>       taxpayers in every tax bracket. It would also not take into account
>       differences in the value that taxpayers can actually derive from itemized
>       deductions.
> 
> But, as far as I can see, the code in pull request #867 applies the 28% limitation to all
> taxpayers. Am I wrong? If so, where is the code that applies the 28% limitation to only
> those with high incomes?
>

There is a thread in the tax literature about the injustice of high 
marginal rate taxpayers getting a bigger subsidy from a deduction than low 
marginal rate taxpayers. The proposed solution is always to limit the 
benefit the the amount that a lower rate taxpayer would have received, in 
this case 28%. This might well be described in a press release as high 
income and low income, but my guess is that the actual implementation 
would be based on tax bracket, not AGI. This would probably mean using tax 
before AMT and credits to calculate the benefit. Press releases are not to 
be taken literally.

dan

martinholmer:
> @martinholmer asked:
> When computing deductible_expenses in the new BenefitCap function, shouldn't the itemized-deduction phase-out calculations be taken into account?

In response, @MattHJensen said:
> I don't think so. The effect of the itemized deduction phaseout is captured by ComputeBenefit().

I definitely agree that the `ComputeBenefit` function correctly calculates the `benefit` variable, which is income tax revenue benefit including any effects of the itemized-deduction phase out.

But what I'm asking about is whether or not the `capped_benefit` variable is correctly computed.  In your code `deductible_expenses` is the sum of component itemized-deductions **before** the phase-out of total itemized deductions is applied.

> And @MattHJensen continued:
> If the phaseout reduces the benefit a taxpayer receives from itemized deductions, that taxpayer will have a lower excess_benefit as a result.

```
capped_benefit = deductible_expenses * calc.policy.ID_BenefitCap_rt
excess_benefit = np.maximum(benefit - capped_benefit, 0)
```
Yes, in the second statement above, when `benefit` goes down, `excess_benefit` also goes down.  So, I agree with your statement above.  But there's more.  If `deductible_expenses` are too high (because the itemized-deduction phase out was not accounted for in the calculation of `deductible_expenses` in the `BenefitCap` function), then `capped_expenses` are also too high, and therefore, `excess_benefits` are too low.

Isn't that the problem you're having with the comparative revenue estimates: that Tax-Calculator estimate is lower than CBO's?  Isn't this what is worrying you:
```
In 2020, a 28% cap on all itemized deductions raises $5.3 billion [using #867 code]. ...
The best comparison is CBO, since they analyze this provision alone; they get $15B in 2020.
```

MattHJensen:
@martinholmer said:

> 
> If deductible_expenses are too high (because the itemized-deduction phase out was not accounted for in the calculation of deductible_expenses in the BenefitCap function), then capped_expenses are also too high, and therefore, excess_benefits are too low.
> 
> Isn't that the problem you're having with the comparative revenue estimates? The Tax-Calculator estimate is lower than CBO's as follows:

Thanks for the clarification, @martinholmer. I understand what you are recommending now, but I find it very difficult to interpret "Limit the value of itemized deductions to 28% of their total value" to mean "Limit the value of itemized deductions to 28% of their value after that value has already been reduced by the Pease limitation". It would be difficult, I think, to explain the latter interpretation to users; it is difficult to justify on policy grounds, and it would also require substantial surgery in `functions.py` to generate a separate Pease-capped value for each deduction (currently we apply the Pease limitation to a sum variable several itemized deductions.)

MattHJensen:
@martinholmer inspired me to read more closely how CBO is treating the Pease limitation in their 28% limitation experiment. [Here](https://www.cbo.gov/sites/default/files/cbofiles/attachments/49638-BudgetOptions.pdf) is what they say (pg 33): 

> This option considers three alternative approaches **that would replace the Pease limitation** with broader restrictions on the total amount of itemized deductions that taxpayers are allowed to take. 

This indicates to me that CBO includes Pease repeal in the baseline when they estimate the value of the 28% limitation. 

With Pease repeal in the baseline, TC calculates that a 28% limitation on itemized deductions would raise $15.4 billion in 2020. CBO estimates it would raise $15 billion in 2020. 

My recommendation is that we merge this PR as is, unless others have additional stylistic feedback. 

@martinholmer, what do you think?

martinholmer:
@MattHJensen said:
> This [quote] indicates to me that CBO includes Pease repeal in the baseline when they estimate the value of the 28% limitation.
>
> With Pease repeal in the baseline, TC calculates that a 28% limitation on itemized deductions would raise $15.4 billion in 2020.  CBO estimates it would raise $15 billion in 2020.

That's much better.  I guess this shows that one of the hardest things involved in comparing reform estimates is understanding exactly what reform is being estimated by the other organization.

MattHJensen:
Thanks a lot @martinholmer and @feenberg for your review and suggestions!

feenberg:
On Mon, 3 Oct 2016, Matt Jensen wrote:

> 
> @martinholmer said:
>
>       If deductible_expenses are too high (because the itemized-deduction phase
>       out was not accounted for in the calculation of deductible_expenses in the
>       BenefitCap function), then capped_expenses are also too high, and
>       therefore, excess_benefits are too low.
>
>       Isn't that the problem you're having with the comparative revenue
>       estimates? The Tax-Calculator estimate is lower than CBO's as follows:
> 
> Thanks for the clarification, @martinholmer. I understand what you are recommending
> now, but I find it very difficult to interpret "Limit the value of itemized deductions
> to 28% of their total value" to mean "Limit the value of itemized deductions to 28% of
> their value after that value has already been reduced by the Pease limitation". It
> would be difficult, I think, to explain the latter interpretation to users; it is
> difficult to justify on policy grounds, and it would also require substantial surgery
> in functions.py to generate a separate Pease-capped value for each deduction
> (currently we apply the Pease limitation to a sum variable several itemized
> deductions.)

What is your interpretation of what the 28% rule means? Is there a 
(possibly ambiguous) source for the idea?

dan

feenberg:
On Mon, 3 Oct 2016, Matt Jensen wrote:

> 
> @martinholmer inspired me to read more closely how CBO is treating the Pease
> limitation in their 28% limitation experiment. Here is what they say (pg 33):
>
>       This option considers three alternative approaches that would replace the
>       Pease limitation with broader restrictions on the total amount of itemized
>       deductions that taxpayers are allowed to take.
> 
> This indicates to me that CBO includes Pease repeal in the baseline when they estimate
> the value of the 28% limitation.
> 
> With Pease repeal in the baseline, TC calculates that a 28% limitation on itemized
> deductions would raise $15.4 billion in 2020. CBO estimates it would raise $15 billion
> in 2020.
>

Page 33 does imagine alternatives to Pease, not additions. The sentence:

    The first alternative would
    limit the tax benefits of itemized deductions to 28 percent
    of the deductions’ total value.

implies that it is the actual tax benefit that is limited to 28 percent of 
sum of itemized deductions. This requires calculating tax with and without 
itemized deductions, then the tax is the larger of the tax with no 
deductions less 28% of itemized deductions and the tax with all 
deductions. There is ambiguity as to the treatment of the standard 
deduction. Is the standard deduction part of this calculation? The 
benefits of itemization would seem to include only the excess over the 
standard deduction, but I don't know if that is intended.

dan

MattHJensen:
@feenberg, thanks for confirming the interpretation of the CBO report. I think we have this feature right now. 

The standard deduction is incorporated: when we calculate tax without itemized deductions, then taxpayers switch to the standard deduction. Therefore the itemized deduction benefit only captures the value that the taxpayer gets from itemizing that is in excess of what they would get from taking the standard deduction.

Issue #880, opened 2016-08-29
Calculate MTR w.r.t currently "invalid" income types


MattHJensen:
Over in B-Tax, @jdebacker asked:

> This line yields the following error:
> `[mtr_fica_mtg, mtr_iit_mtg, mtr_combined_mtg] = calc1.mtr('e19200')`
> `ValueError: mtr income_type_str="e19200" is not valid`
> 
> I see this variable in the PUF. Am I not able to calculate the MTR w.r.t. this item with the Tax-Calculator? Something else causing the error?

To which @Amy-Xu replied: 

> It seems from [this line](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/calculate.py#L229) of code in Tax-Calculator that we only allow a certain set of variables to be passed into MTR function. This set currently does not include interest paid. We should be able to add this variable very easily. What do you think? @MattHJensen

To which @jdebacker responded:

> Thanks for finding that @Amy-Xu. I recall something like this, but couldn't find the lines of code to point out. Does this limitation of the valid strings have to do with some TaxCalc logic or for another reason (e.g. to save computational time it does not compute MTRs for every line item on the 1040)?
> 
> Relatedly, there are some variables such as tax exempt interest that is (obviously) zero-taxed under current law. But one could imagine a reform where the tax-exempt status of this income source were to change. How would one handle that? In particular, would the Tax Calc output then put all interest income in e00300? Or would the formerly tax exempt still show up under e00400, but now face the non-zero tax? I'm just trying to think about how I might compute a MTR on interest income as a whole (taking into account that some might be taxable and some not, with the taxable status changing between the baseline and reform).
>

martinholmer:
@jdebacker said:
> Thanks for finding that @Amy-Xu. I recall something like this, but couldn't find the lines of code to point out. Does this limitation of the valid strings have to do with some TaxCalc logic or for another reason (e.g. to save computational time it does not compute MTRs for every line item on the 1040)?

The limitation has everything to do with the fact that different income types have different MTR logic associated with them as the Calculator.mtr() method shows.  We will never be able to allow any income type to be passed to the mtr() method.

So, do you what to request another valid income type for the Calculator.mtr() method?  If so, which one?

@jdebacker @Amy-Xu @MattHJensen @feenberg @GoFroggyRun @zrisher

jdebacker:
@martinholmer thanks for the quick follow up.

[...]

PR #881, opened 2016-08-29
Add mtr calculation for two itemized-deduction-expense variables

martinholmer:
This pull request resolves issue #880, which requests the ability to compute marginal tax rates with respect to two itemized-deduction-expense variables (rather than with respect to a particular income type.)  Adding this capability made the `Calculator.VALID_MTR_INCOME_TYPES` a misnomer, so it has been changed to the more general `Calculator.VALID_MTR_VARIABLES`.  In a similar fashion, the first parameter of the `Calculator.mtr()` method has been changed from `income_type_str` to `variable_str`.  Those calling the `Calculator.mtr()` method in their own package or notebook may need to revise the call if they are using the named parameter.

There is no change in Tax-Calculator logic (other than the two new variables for which marginal tax rates can be computed) or results (other than the two new sets of results in the `pufcsv_mtr_expected.txt` file).

@jdebacker @Amy-Xu @MattHJensen @feenberg @GoFroggyRun @zrisher @codykallen

### DIFF
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -226,14 +226,14 @@ class Calculator(object):
     def current_year(self):
         return self.policy.current_year
 
-    MTR_VALID_INCOME_TYPES = ['e00200p', 'e00900p',
-                              'e00300', 'e00400',
-                              'e00600', 'e00650',
-                              'e01400', 'e01700',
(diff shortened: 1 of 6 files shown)


jdebacker:
thanks @martinholmer !

Issue #862, opened 2016-08-12
Parameterize a "Buffet Rule"


MattHJensen:
Clinton's tax plan includes a Buffet Rule.

All I can find from Clinton's [own documents](https://www.hillaryclinton.com/briefing/factsheets/2016/01/12/investing-in-america-by-restoring-basic-fairness-to-our-tax-code/) is this:

> Hillary will reiterate her call for the “Buffett Rule,” which ensures that those making more than $1 million per year pay at least an effective tax rate of 30 percent. 

TPC implemented something much more complicated in [their analysis](https://www.hillaryclinton.com/briefing/factsheets/2016/01/12/investing-in-america-by-restoring-basic-fairness-to-our-tax-code/) of her proposal: 

 [From TPC:](http://www.taxpolicycenter.org/publications/analysis-hillary-clintons-tax-proposals): 

> Q2. The documentation states that Secretary Clinton would propose the “Buffett Rule” to ensure
> that “millionaires must pay at least a 30 percent effective rate.” Does Secretary Clinton’s
> version of the Buffett Rule differ in any way from the version that was proposed in the
> Administration’s FY2016 budget, described as follows in the Treasury Green Book (available
> at https://www.treasury.gov/resource-center/tax-policy/Documents/GeneralExplanations-FY2016.pdf):
> 
> The proposal would impose a new minimum tax, called the Fair Share Tax (FST), on high income
> taxpayers. The tentative FST would equal 30 percent of AGI less a credit for charitable
> contributions. The charitable credit would equal 28 percent of itemized charitable contributions
> allowed after the overall limitation on itemized deductions (so called Pease limitation). The final
> FST would be the excess, if any, of the tentative FST over the sum of the taxpayer’s (1) regular
> income tax (after certain credits) including the 3.8-percent net investment income tax, (2) the
> alternative minimum tax, and (3) the employee portion of payroll taxes. The set of certain credits
> subtracted from regular income tax would exclude the foreign tax credit, the credit for tax
> withheld on wages, and the credit for certain uses of gasoline and special fuels. The amount of
> FST payable (i.e., the excess of tentative FST over regular tax) would be phased in linearly
> starting at $1 million of AGI ($500,000 in the case of a married individual filing separate
> return). The FST would be fully phased in at $2 million of AGI ($1 million in the case of a married
> individual filing a separate return). 
> 
> 
> A2. We assume the Secretary’s version of the Buffett Rule would be the same as the
[...]

PR #890, opened 2016-08-31
Adds a function to calculate the Fair Share Tax aka the Buffett Rule

andersonfrailey:


martinholmer:
The following PEP8 code-formatting rules are violated in pull request #890.
```
__________________________ PEP8-check(ignoring E402) ___________________________
taxcalc/functions.py:1234:80: E501 line too long (90 > 79 characters)
def BuffettRule(c00100, c05200, Buffett_tentRate, Buffett_minAGI, Buffett_phaseRate, fst):
                                                                               ^
taxcalc/functions.py:1243:80: E501 line too long (86 > 79 characters)
    Buffett_minAGI: The minimum AGI needed to be subject to the FST. Default = 1000000
                                                                               ^
taxcalc/functions.py:1255:48: E226 missing whitespace around arithmetic operator
        rate = (float(c00100 - Buffett_minAGI))/Buffett_minAGI * Buffett_phaseRate
                                               ^
taxcalc/functions.py:1255:80: E501 line too long (82 > 79 characters)
        rate = (float(c00100 - Buffett_minAGI))/Buffett_minAGI * Buffett_phaseRate
                                                                               ^
```
Each of these four errors need to be fixed before #890 can be considered on its merits.

You should be running all these PEP8 test on your computer so that you don't have to mess with these errors when you submit a pull request to GitHub.

@andersonfrailey @MattHJensen

martinholmer:
@andersonfrailey, Thanks for the fix of several PEP8 errors in pull request #890, but you still have one more.  You need to ask @MattHJensen to show you how to check for these kinds of errors on your own computer.  Nobody is going to spend their time reviewing your work until it passes the PEP8 tests.

MattHJensen:
@andersonfrailey, when you open a PR, could you include a reference to the issue you are trying to resolve? Also, when you are adding a new reform option to TC it is helpful to include a revenue estimate. Ideally, you will estimate a version of the reform that has been analyzed by others and then try to understand and explain why the revenue effects are similar or different.

andersonfrailey:
This is to address issue #862. 

Estimated revenue effects of implementing the FST from the OSPC model and other sources are below (billions of dollars):

| Model                                          |  2016  |  2017  |  2018  |  2019  |  2020  |  2021  | Total |
| :-----------------------------: | :------: | :------: | :------: | :------: | :------: | :------: | :-----: |
|  [TPC](http://www.taxpolicycenter.org/publications/analysis-hillary-clintons-tax-proposals/full)                                            |    2.2     |     2.4    |     4.7    |    10.2   |    12.5    |    12.8   | 44.8 |
| [Treasury (2016)](https://www.treasury.gov/resource-center/tax-policy/Documents/General-Explanations-FY2016.pdf)                         |    6.7     |     -.093   |     1.2     |     2.8    |     3.7    |    3.9     |   17.37 |
| [Treasury (2017)](https://www.treasury.gov/resource-center/tax-policy/Documents/General-Explanations-FY2017.pdf)                         |     0.0    |      7.8    |    -.062    |      1.3    |    3.1     |    4.0     |  15.58 |
| [Joint Committee on Taxation](http://www.finance.senate.gov/imo/media/doc/03-20-12%20Updated%20JCT%20Analysis%20of%20Buffett%20Rule%20Legislation.pdf)   |    4.8     |     5.3    |   5.7     |     6.1    |     6.4    |    6.7     | 35 |
| Avg. of other Estimates | 3.4 | 3.6 | 2.7 | 5.1 | 6.4 | 6.8| 28.18 |
| **OSPC** |   **4.0**     |     **4.3**     |    **4.7**     |      **5.2**    |    **5.7**     |    **6.3**     | **30.2**|

I stopped at 2021 because that is the last year TPC estimated the revenues for the individual year.
 
The estimates vary quite a bit. The most obvious reason for the difference between OSPC and others is this what I've put together is a more simplified version of the rule. While the other models factor in regular income tax, AMT, and a 4% surcharge on AGI when determining the FST owed, this only uses regular income tax.

TPC claims to follow the general methodology used by the Joint Committee on Taxation and Treasury for estimation, so the other differences are probably due to differences in data and the parameters used in their respective models. 

I have yet to find an explanation for large reduction in revenue from the Treasury estimates after the first year the tax is imposed and jump in revenue between 2018 and 2019 in TPC's.

feenberg:
On Fri, 2 Sep 2016, andersonfrailey wrote:

> 
> This is to address issue #862.
> 
> 
> The estimates vary quite a bit. The most obvious reason for the difference between
> OSPC and others is this what I've put together is a more simplified version of the
> rule. While the other models factor in regular income tax, AMT, and a 4% surcharge on
> AGI when determining the FST owed, this only uses regular income tax.


Aren't the other estimates for fiscal year, and OSPC for calander year?

The AMT and surcharge may be a big part of the explanation - since they 
address precisely the same taxpayers.

The negative income is bizarre, but must reflect some behavioral 
considerations. Perhaps there is a lag before behavior sets in? Perhaps 
withhold is also lagged?


dan

MattHJensen:
@andersonfrailey, I think this implementation would be closer to what Clinton and others have proposed if it keyed off of the sum of individual income tax and employee payroll rather than regular income tax. Does that seem right to you as well? 

You'll need to be cautious to ensure that the itemized deduction surtax is accounted for. I.e., If the surtax is applied, then the FST liability ought to be lower.

andersonfrailey:
@MattHJensen I believe it would be closer. I'll rerun the program with those changes and post the results.

MattHJensen:
Sounds good @andersonfrailey. 

As a note on nomenclature, we normally refer to the results from this model as "Tax-Calculator results" or "TC Results" rather than "OSPC results". The reason for this is that TC is an open source model (in the public domain, even) with contributors from many organizations. It could be even more accurate to refer to the results as Tax-Calculator + TaxData  (TC+TD) results, since the data we rely on come from the TaxData repo, but normally that's too wordy, and we just write "TC".

andersonfrailey:
@MattHJensen, when I used income tax, payroll, and 4% surtax on AGI above $5 million I ended up with pretty odd result. Estimated revenues jumped from $4-6 billion a year to $18-$21 billion. For comparison I switched income tax back to regular tax and the result was closer to the other estimates.

|           | 2016 | 2017 | 2018 | 2019 | 2020 | 2021 |
| ----- | ----- | ----- | ----- | ----- | ----- | ----- |
| Income Tax, FICA, Surtax | 18.32 | 18.87 | 19.55 | 20.03 | 20.75 | 21.68 | 119.20 |
| Regular Tax, FICA, Surtax | 2.67 | 2.85 | 3.11 | 3.44 |3.73 | 4.07 | 19.88|

MattHJensen:
Try dropping the 4% surtax. Also, if you push another commit when you post the new results, then others can see exactly what is modeled and it is easily reversible.

andersonfrailey:
@MattHJensen, here are the results using just the individual income tax and FICA:

| | 2016 | 2017 | 2018 | 2019 | 2020 | 2021 |
|----| ----- | ----- | ----- | ----- | ----- | ----- |
|       | 27.8 | 28.55 | 29.43 | 30.05 | 31.13 | 32.47 |

andersonfrailey:
Re-checking that last push to fix the errors that came up.

MattHJensen:
@andersonfrailey, Sees like we have more thinking to do on why the numbers are higher than other estimates. Here are a few other comments in the meantime:

First. 

Our codebase tends to use 

`fst = max((tentFST - _iitax - _fica) * rate, 0)` 

instead of
```
        fst = (tentFST - _iitax - _fica) * rate
        if fst < 0.0:
            fst = 0.0
```

and similar usage of mins. 

Second. 

Based on where `BuffetRule()` is called in calculate.py, I don't expect it to interact properly with `BenefitSurtax()`. 

Third. 

Since `BuffetRule()` comes after _iitax and _fica are defined, you'll need to add fst to `_combined` and `_iitax` like in `SurtaxBenefit()`. 

When you generate reform results, the proper way to do it would be to set up one calculator with `Buffet_tentRate = 0` (current law) and then another with `Buffet_tentRate = .3` (reform) and compare the differences in `_combined`. 

Fourth.

A recent PR from Martin renamed _fica to something else. 

Fifth. 

I believe the standard buffet rule proposals only include the employee payroll tax payments not both employee and employer.

martinholmer:
@andersonfrailey,  I don't understand why the new function and parameters are called BuffetRule when the new function returns the value of the "fair share tax" `fst`.   Is there a good reason to call these things by separate names?  If not, it seem potentially confusing.  So, for example, why isn't the new function called `FairShareTax(...)`?

@MattHJensen

andersonfrailey:
@martinholmer, it's because the "fair share tax" is the name the Obama administration/Clinton team gave the tax their Buffett Rule proposals imposed. I'll change it in the next push for clarity.

PR #893, opened 2016-09-01
add TaxForm and child classes to support Luca integration

zrisher:
This PR introduces the `TaxForm` base class and a child class for each tax form from which our input `evars` are derived. Please provide your feedback!

**Purpose**
These classes are designed to help convert raw tax form field data into `evars` to support reading `luca` data (and eventually user-supplied tax form pdfs) as `taxcalc` input. 

Eventually, they will also support converting `taxcalc` output to tax form data in `luca` and pdf format. They could also hold form-specific tax calculation logic, i.e. `taxcalc.functions.TaxGains()`, if deemed useful.

**Status**
This is only part of the overall work I'm doing to support Luca input. However, it's the foundation for everything else and I believe it's ready for contributor suggestions and feedback (@martinholmer @talumbau @MattHJensen @feenberg @Amy-Xu @GoFroggyRun @codykallen). It's probably a little rough around the edges still, but it's complete with working tests and passes PEP8.

**Effects on Taxcalc**
This does not touch `taxcalc` logic or even existing files besides including the new classes into the taxcalc module and a small helper function added to `utils.py`.

**Notes on Architecture**

**Class per form**
It seemed to me that one class per form made significantly more sense than one big mapping class for a few reasons:
* The field to evar mappings are not always 1-1, can depend on the filing year, and take a fair number of lines (see `to_evars_direct()` and `to_evars_indirect()`).
* Some forms have significantly less revisions than others and will be much easier to code mappings for. I wanted to easily indicate what years we were ready to work with (see `check_year_support`).
* Some forms can be submitted multiple times for a single filing and computing their evars requires knowing about all the forms (e.g. P87482, E87487, E87492, E87497). Working with a collection of forms for a given filing will be handled by a `Filing` class to be introduced later (see https://github.com/zrisher/Tax-Calculator/commit/eb4a40b2c07fbd41599b48f7e01f36bcb35c4bb7). Converting to a `Record` set will then be handled by a `FilingCollection` (which can support multiple filings from multiple taxpayers).
* Some tax forms have different fields for taxpayer id. 
* There's other logic we may want to segment per form in the future (e.g. pdf handling, form-specific calculations).
* I'm a big fan of Domain Driven Design for maximizing maintainability and flexibility.

**Field names**
Luca provides user tax form data in JSON format with fields keys very similar to the format proposed here (the only differences are for some categorical things like filing status, where Luca has a variable similar to MARS but implemented differently). So passing data from Luca into these forms is nice and trivial, and can be handled by a `luca_io` class similar to `incometaxio` (see https://github.com/zrisher/Tax-Calculator/commit/eb4a40b2c07fbd41599b48f7e01f36bcb35c4bb7).   However, Luca cannot read from tax forms in pdf format, and its ability to write to them is limited to 2013 and earlier (2012 in some circumstances).

The field labels for IRS-provided PDFs are not particularly usable, so I believe Luca's "line5a" format that we've adopted is better. Here's a selection of the form metadata from the 1040 available [here](https://www.irs.gov/pub/irs-pdf/f1040.pdf) for lines 7-9b:

```
<IncomeSection xmlns:xfa="http://www.xfa.org/schema/xfa-data/1.0/" xfa:dataNode="dataGroup"/>
<f1_35/>
<f1_36/>
<f1_37/>
<f1_38/>
<f1_39/>
<f1_40/>
<f1_41/>
<f1_42/>
<f1_43/>
<f1_44/>
```


martinholmer:
@zrisher, there is a large volume of impressive work in your pull request #893 that would add a TaxForm class and many child classes to support Luca integration with the Tax-Calculator.  After a quick review of pull request #893, I think I have a rough (and, no doubt, partial) understanding of what this code does.

But I have a basic question about the Luca project.  There are many lines on the real-world tax forms that contain a filing unit's tax attributes (that is, input variables that are used in tax calculations).  But Tax-Calculator can handle only a subset of those real-world tax attributes.  This is because it is a tax simulation model (which abstracts from many details), not commercial tax-preparation software (which handles all the details).  My question is this: How is the Luca project going to handle tax attributes that are not recognized by Tax-Calculator?

@MattHJensen @feenberg

feenberg:
On Wed, 7 Sep 2016, Martin Holmer wrote:

> 
> @zrisher, there is a large volume of impressive work in your pull request
> #893 that would add a TaxForm class and many child classes to support Luca
> integration with the Tax-Calculator. After a quick review of pull request
> #893, I think I have a rough (and, no doubt, partial) understanding of what
> this code does.
> 
> But I have a basic question about the Luca project. There are many lines on
> the real-world tax forms that contain a filing unit's tax attributes (that
> is, input variables that are used in tax calculations). But Tax-Calculator
> can handle only a subset of those real-world tax attributes. This is because
> it is a tax simulation model (which abstracts from many details), not
> commercial tax-preparation software (which handles all the details). My
> question is this: How is the Luca project going to handle tax attributes
> that are not recognized by Tax-Calculator?

In my experience, it is always valid to use "0" for a tax attribute and 
produces the expected result. We should keep an eye out for a contrary 
case, but starting with "0" is my suggestion. In the IRS world, missing is 
always zero.

dan

martinholmer:
@martinholmer asked:
> How is the Luca project going to handle tax attributes that are not recognized by Tax-Calculator?

Dan @feenberg replied:
> In my experience, it is always valid to use "0" for a tax attribute and produces the expected result.  We should keep an eye out for a contrary case, but starting with "0" is my suggestion.  In the IRS world, missing is always zero.

I agree, Dan, but my question was: How will the Luca-add-on-to-Tax-Calculator handle nonzero values for tax attributes that are not recognized by Tax-Calculator?   By "not recognized" I mean the attribute is not in the Records.VALID_READ_VARS set.

@MattHJensen @zrisher

taxcalc/filings/forms/tax_form.py
```diff
@@ -0,0 +1,179 @@
...
+        return results
+
+    def to_evars_indirect(self):
+        """
+        Get any evars that can't be determined with a simple 1-1 mapping,
+        e.g. if an evar depends on multiple lines.
+        To be extended by child classes.
+        Should convert values to their expected types.
+        """
+
```
talumbau: Should there be a `pass` here? or a `raise NotImplementedError`?

taxcalc/tests/filings/forms/test_tax_form.py
```diff
@@ -0,0 +1,163 @@
...
+
+def test_tax_form_year_support():
+    child_class = type("TestTaxForm", (TaxForm,), {
+        '_SUPPORTED_YEARS': [2014, 2015],
+    })
+    with pytest.raises(UnsupportedFormYearError):
+        form = child_class(2013)
+    with pytest.raises(UnsupportedFormYearError):
+        form = child_class(2016)
+    form = child_class(2015)
```
talumbau: cool test!

talumbau:
@zrisher, I like the code organization in this PR. I made a couple of comments on the source directly. I noticed some repetition in the test code, so I wondered if you might want to make use of `pytest.mark.parametrize` to reduce duplication of code:

http://docs.pytest.org/en/latest/parametrize.html

I was thinking of `test_year_support` especially. Nice work!

martinholmer:
@zrisher,  I'm finding the information in your pull request #893 about where each Tax-Calculator input variable comes from on the IRS forms very useful in my cross-model validation work.  Thanks for all this work.  So far, I do have one question: How are you thinking about using IRS form information to specify the `age_head` and `age_spouse` input variables?

@MattHJensen @feenberg

martinholmer:
@talumbau,  Why have you assigned to yourself pull request #893 that was developed by and submitted by @zrisher?

If you are going to be in charge, then it would seem that you need to answer my questions (see above in #893) before this is merged?  There is very useful information in this pull request, but the overall utility of using Luca-like or PDF-form input for Tax-Calculator seem unclear to me.  We need some kind of discussion of these matters before adding this massive amount of code to the taxcalc package.  

And now that I write that last sentence, I wonder why isn't all this code in a separate filings (ro some other name) package, which **uses** the taxcalc package (rather than being part of the taxcalc package)?  We need a discussion of the pros and cons of including all this extra code in the taxcalc package.  It certainly is not going to be used by TaxBrain, is it?  The `simtax.py` and `inctax.py` scripts are part of the taxcalc package because they are needed for the taxcalc validation tests.  But I don't see any possible use of the Luca-like or PDF-form input in the cross-model validation work.

@MattHJensen @feenberg

talumbau:
Hi @martinholmer, I put myself on the "Assignees" list as a way to indicate that I was going to review this PR. I frequently use this "Assignees" list on PRs this way. The idea is that assignees on issues would be the person implementing the feature/bug fix/whatever, but assignees for PRs are on the reviewer list. It was just a way to keep the task top of mind. I gave my feedback so I'm fine being removed from the list.

martinholmer:
@zrisher,  In addition to my earlier question about how pull request #893 sets the `age_head` and `age_spouse` variables, I have two other questions: one about the setting of the `EIC` variable and the other about the setting of the `n24` variable.

First, the number of EIC qualified children, `EIC`, does not seem to be set correctly given the information supplied on Schedule EIC.  Suppose a parent has three children living at home and the parent enters the information about each child into Schedule EIC.  But one of the children is too old to be EIC qualified, so the value of the `EIC` variable is 2.  But the following logic in the `filings/forms/us1040seic.py` file would set the value of the `EIC` variable to 3, if I'm not mistaken.
```
# EIC - Earned Income Credit code: categorical variable
child_ssns = ['line2_child1', 'line2_child2', 'line2_child3']
results['EIC'] = sum(1 if fields.get(key) else 0 for key in child_ssns)
```
It would seem that the `EIC`-setting logic needs to be a little more complicated than the way it is now in pull request #893.

Second, the number of children qualified for the Child Tax Credit, Tax-Calculator input variable `n24`, does not seem to be set anywhere in pull request #893.  Did I miss something?

@MattHJensen @feenberg

martinholmer:
@zrisher,  Note that Tax-Calculator input from Form 8863 has been simplified in pull request #917.  These changes should make it much easier for you to specify input variables off of Form 8863.  Once #917 is merged, there will be only two Form 8863 lines used by Tax-Calculator as input:

| Var-name | IRS-form | Form-line | Var-description
| :------- | :------- | :-------- | :--------------
| e87530 | 8863 | 10 | total LLC adjusted qualified expenses for all students
| p87521 | 8863 | 1 | total tentative AmOppCredit amount for all students

taxcalc/filings/forms/tax_form.py
zrisher: We don't need a `raise NotImplementedError`, because we expect this function to produce `None` if a child TaxForm class has no indirect evar mappings.

I believe we don't need a `pass` because the function has a comment block, thus avoiding the `IndentationError` that a function without any body content would produce. If we added a `pass` here, might we encounter pylint message W0107 - "Unneccesary pass statement"? If not and you believe the pass statement adds clarity I'm happy to add it.

taxcalc/tests/filings/forms/test_tax_form.py
zrisher: Yeah I had to look that one up. : )

zrisher:
@martinholmer asked:
> How is the Luca project going to handle tax attributes that are not recognized by Tax-Calculator?

@feenberg replied:
> In my experience, it is always valid to use "0" for a tax attribute and produces the expected result. We should keep an eye out for a contrary case, but starting with "0" is my suggestion. In the IRS world, missing is always zero.

@martinholmer replied:
> I agree, Dan, but my question was: How will the Luca-add-on-to-Tax-Calculator handle nonzero values for tax attributes that are not recognized by Tax-Calculator? By "not recognized" I mean the attribute is not in the Records.VALID_READ_VARS set.

Currently this implementation will only generate input vars that are specifically returned by the direct or indirect evar mapping functions - form fields that aren't mapped to anything are simply ignored. The info is retained inside the TaxForm's field list, however, in case we want to use it later.

@talumbau said:
>  I noticed some repetition in the test code, so I wondered if you might want to make use of pytest.mark.parametrize to reduce duplication of code: http://docs.pytest.org/en/latest/parametrize.html
I was thinking of test_year_support especially.

Yep, looks good. I will implement this.

@martinholmer asked:
> How are you thinking about using IRS form information to specify the age_head and age_spouse input variables? Where on the IRS forms do filers enter their age (or birthday)

Nowhere, unfortunately. This will need to be added to the input set via other means, as it is not listed on any tax form.

@martinholmer I will respond to your other questions later tonight or tomorrow, sorry for the delay.

martinholmer:
@zrisher said:
> Currently this implementation will only generate input vars that are specifically returned by the direct or indirect evar mapping functions --- form fields that aren't mapped to anything are simply ignored. The info is retained inside the TaxForm's field list, however, in case we want to use it later.

This approach works at a certain level, but I'm concerned about one thing.  If someone using Luca input specifies a non-zero value for a variable that is not recognized by Tax-Calculator, then that person is likely to very confused about the Tax-Calculator output (because the output does not change in the way the person expects).  This is likely to cause chaos with people using Luca input to Tax-Calculator.  All this could be avoided if the code in this pull request #893 was to raise an error saying that a variable (which is not in the `Records.VALID_READ_VARS set`) cannot be specified to be any value other than zero.

Here is an example of the confusion I'm concerned about.  A Luca-input user specifies some wage income and no other income for a single person with no kids.  They feed that input into Tax-Calculator that generates results that make sense to the user.  Then the user adds some income that Tax-Calculator does not recognize, feeds in the new input, and gets back exactly the same results from Tax-Calculator.  At that stage, this user is likely to be confused and wondering why the extra income did not change tax liability.

Does this make sense to you?  Is there a better way to avoid the confusion I think will happen?

@MattHJensen

martinholmer:
@martinholmer asked:
> How are you thinking about using IRS form information to specify the age_head and age_spouse input variables? Where on the IRS forms do filers enter their age (or birthday)

@zrisher said:
> Nowhere, unfortunately. This will need to be added to the input set via other means, as it is not listed on any tax form.

TurboTax handles this by starting the "tax interview" with a Personal Information form that includes birthdays of everybody in the filing unit.  This Personal Information form is not an IRS form, but it does collect information that is necessary to complete the IRS forms.

Maybe the Luca-input scheme needs to be broadened beyond IRS forms to include something like the TurboTax Personal Information form.  Does this make sense?  Is there a better approach to gathering essential data that do not appear on IRS forms?

@MattHJensen

MattHJensen:
> Maybe the Luca-input scheme needs to be broadened beyond IRS forms to include something like the TurboTax Personal Information form. Does this make sense? Is there a better approach to gathering essential data that do not appear on IRS forms?

That makes sense to me, too.

MattHJensen:
> This approach works at a certain level, but I'm concerned about one thing. If someone using Luca input specifies a non-zero value for a variable that is not recognized by Tax-Calculator, then that person is likely to very confused about the Tax-Calculator output (because the output does not change in the way the person expects). This is likely to cause chaos with people using Luca input to Tax-Calculator. All this could be avoided if the code in this pull request #893 was to raise an error saying that a variable (which is not in the Records.VALID_READ_VARS set) cannot be specified to be any value other than zero.

Another approach to solving this problem would be to allow the user to include non-zero values for the unused variables in their input file, but then tell the user both before and after the calculation which values are not used in the calculations. A note could also explain that while TC does not currently accomodate those variables, pull requests are appreciated. 

The primary advantage of this approach is that the user could keep one input file that corresponds to his actual tax return and not need to update the file when he gets the error the first time and again  each time TC is expanded. That would be more friendly to the user, I think, and also encourage contributions.

zrisher:
@martinholmer said:
> Maybe the Luca-input scheme needs to be broadened beyond IRS forms to include something like the TurboTax Personal Information form. Does this make sense? Is there a better approach to gathering essential data that do not appear on IRS forms?

@MattHJensen said:
> That makes sense to me, too.

The `TaxForm`s simply exist to facilitate whatever overarching process is taking data in tax form format and wants to send it through `taxcalc`. 

So a `lucatax.py` command line tool could take in this data via command line arguments or a file and add it into the variables derived from some`TaxForm`s before sending them to `taxcalc`. Similarly `webapp-public` could take field input and merge it with the variables extracted from uploaded tax forms. 

So the architecture proposed in this PR could work with any method of retrieving that data, but yes I could definitely define an additional `TaxForm` specifically to hold useful data the IRS *knows* about but doesn't collect on any actual tax form. Then both processes could lean on this form to simplify their code. I'll do this unless there are any objections.

zrisher:
@martinholmer said:

> This approach works at a certain level, but I'm concerned about one thing. If someone using Luca input specifies a non-zero value for a variable that is not recognized by Tax-Calculator, then that person is likely to very confused about the Tax-Calculator output (because the output does not change in the way the person expects). This is likely to cause chaos with people using Luca input to Tax-Calculator. All this could be avoided if the code in this pull request #893 was to raise an error saying that a variable (which is not in the Records.VALID_READ_VARS set) cannot be specified to be any value other than zero.

@MattHJensen said:

> Another approach to solving this problem would be to allow the user to include non-zero values for the unused variables in their input file, but then tell the user both before and after the calculation which values are not used in the calculations. A note could also explain that while TC does not currently accommodate those variables, pull requests are appreciated.

> The primary advantage of this approach is that the user could keep one input file that corresponds to his actual tax return and not need to update the file when he gets the error the first time and again each time TC is expanded. That would be more friendly to the user, I think, and also encourage contributions.

100% agree with Matt for exactly the reasons he mentions. It's not really an error that someone included data we don't use since it won't negatively affect the program. It might be an error, though, that our calculator doesn't make use of that data. One of the big benefits of making it easy to run real tax returns through `taxcalc` is expanding the possible contributor base and exposing cases like this for them to work on.

This:

> tell the user both before and after the calculation which values are not used in the calculations. A note could also explain that while TC does not currently accommodate those variables, pull requests are appreciated.

would be handled by the higher level process, but would require each TaxForm to know which form fields exist for it and which it uses. I'll get to work on this unless there's any objection.

zrisher:
@martinholmer said:
> I wonder why isn't all this code in a separate filings (ro some other name) package, which uses the taxcalc package (rather than being part of the taxcalc package)? We need a discussion of the pros and cons of including all this extra code in the taxcalc package. It certainly is not going to be used by TaxBrain, is it? The simtax.py and inctax.py scripts are part of the taxcalc package because they are needed for the taxcalc validation tests. But I don't see any possible use of the Luca-like or PDF-form input in the cross-model validation work.

I agree that (by design) there aren't any dependencies between `taxcalc` and `TaxForm`, and it would thus make sense to provide this as a separate package within the Tax-Calculator repo. My decision to include in directly into `taxcalc` for this PR was motiviated by a previous discussion I remember reading (but now can't find) where someone basically said "this is a fair amount of code that doesn't necessarily belong in taxcalc" and the response was "yes, but most users aren't familiar with the entire package anyway and it makes things simpler." I'm happy to do whatever makes the most sense to this contributor base.

zrisher:
@martinholmer said:

> the number of EIC qualified children, EIC, does not seem to be set correctly given the information supplied on Schedule EIC. Suppose a parent has three children living at home and the parent enters the information about each child into Schedule EIC. But one of the children is too old to be EIC qualified, so the value of the EIC variable is 2. But the following logic in the filings/forms/us1040seic.py file would set the value of the EIC variable to 3, if I'm not mistaken... It would seem that the EIC-setting logic needs to be a little more complicated than the way it is now

You're right, if someone listed a non-qualifying child on the form, the proposed logic would be incorrect. I wasn't entirely familiar with the tax law here and assumed someone would only list a child on the form if they qualified.

Looking at the second page of the form, it seems all the data required to confirm their elibigility is listed on this form except one point -  is the child filing a joint return, and if so is it "only to claim a refund of withheld income tax or estimated tax paid"? I suppose that's another point of data for the "things the IRS knows" form?

zrisher:
@martinholmer said:  
> the number of children qualified for the Child Tax Credit, Tax-Calculator input variable n24, does not seem to be set anywhere

If a filer has 4 or less children, we could derive this from column 4 of field 6c on Form 1040. However, if they have more, this must be derived at the `TaxFormCollection` level, where we know how many Schedule 8812's have been submitted and can count how many of fields A, B, C, and D have at least one box checked. To avoid code duplication, I was leaving this variable to that class. Does that sound correct?

martinholmer:
@martinholmer said:
> I wonder why isn't all this code in a separate filings (ro some other name) package, which uses the taxcalc package (rather than being part of the taxcalc package)? We need a discussion of the pros and cons of including all this extra code in the taxcalc package. It certainly is not going to be used by TaxBrain, is it? The simtax.py and inctax.py scripts are part of the taxcalc package because they are needed for the taxcalc validation tests. But I don't see any possible use of the Luca-like or PDF-form input in the cross-model validation work.

@zrisher said in response:
> I agree that (by design) there aren't any dependencies between taxcalc and TaxForm, and it would thus make sense to provide this as a separate package within the Tax-Calculator repo. My decision to include in directly into taxcalc for this PR was motiviated by a previous discussion I remember reading (but now can't find) where someone basically said "this is a fair amount of code that doesn't necessarily belong in taxcalc" and the response was "yes, but most users aren't familiar with the entire package anyway and it makes things simpler." I'm happy to do whatever makes the most sense to this contributor base.

As I have worked with pull request #893, I've come around to your decision to make the filings/forms code tree part of the taxcalc package.  I'm just fine with the current approach.

@MattHJensen @talumbau

zrisher:
Per #918 I should remove E11070, E11550, and E11580 from this PR (and probably inspect changes to `VALID_READ_VARS` since this was opened to catch any others).

zrisher:
I will also need to update this per https://github.com/open-source-economics/Tax-Calculator/pull/917 (and probably inspect changes to VAR_LABELS since this was opened, too).

martinholmer:
@zrisher,  Do you want to make the test and code changes you mentioned above on pull request #893?  Or would it be easier for you to make these changes after #893 is merged into master?

zrisher:
@martinholmer Yep, I'm working on:

* make use of [pytest.mark.parametrize](http://docs.pytest.org/en/latest/parametrize.html) to reduce duplication of code, test_year_support especially.
* define an additional TaxForm specifically to hold useful data the IRS knows about but doesn't collect on any actual tax form - e.g. `age_head` and `age_spouse`
* support "tell the user both before and after the calculation which values are not used in the calculations" by enabling each TaxForm to know which form fields exist and which it can use
* make the EIC logic more sophisticated and introduce required variables to the non-IRS form
* review changes to VALID_READ_VARS and VAR_LABELS since this was opened and adjust mappings accordingly

If that's a satisfactory list to everyone, and I implement those things correctly, yes I would love to get this merged.

martinholmer:
@zrisher said:
> Yep, I'm working on:
>
> 1. make use of pytest.mark.parametrize to reduce duplication of code, test_year_support especially.
>
> 2. define an additional TaxForm specifically to hold useful data the IRS knows about but doesn't collect on any actual tax form - e.g. age_head and age_spouse.
>
> 3. support "tell the user both before and after the calculation which values are not used in the calculations" by enabling each TaxForm to know which form fields exist and which it can use.
>
> 4. make the EIC logic more sophisticated and introduce required variables to the non-IRS form.
>
> 5. review changes to VALID_READ_VARS and VAR_LABELS since this was opened and adjust mappings accordingly.
>
> 6. add notes on _cmbtp_itemizer and _cmbtp_standard to Form 6251.
>
> If that's a satisfactory list to everyone, and I implement those things correctly, yes I would love to get this merged.

This list looks good to me.  To me items (2) and (4) could be solved using the same approach as used by TurboTax: a non-IRS form called Personal Information (or some such name) that includes ages of everyone (adults and children) in the filing unit along with other info needed to fill out the IRS forms.

So, just one question.  Does your last sentence indicate that you plan to work on items (1)  - (6) while these enhancements are on pull request #893, or are you suggesting the remaining work would be easier if we merged #893 into the master branch now?  If it is the latter, I'm happy to do the merge now.

zrisher:
@martinholmer I'm happy to do either. Merging these commits into master would make it slightly easier to continue my work. If you don't mind doing so with the listed issues outstanding, that would be great!

martinholmer:
@zrisher said:
> Merging these commits [pull request #893] into master would make it slightly easier to continue my work.  If you don't mind doing so with the listed issues outstanding, that would be great!

OK,  I'll do the merge and open a new ISSUE that starts out with the last few contributions to this discussion.

PR #900, opened 2016-09-03
Set e01500=e01700 in SimpleTaxIO._specify_input method

martinholmer:
This pull request sets total (taxable plus nontaxable) pension benefits, `e01500`, equal to taxable pension benefits, `e01700`, when Internet-TAXSIM-like input data is read by the SimpleTaxIO class.  See the discussion associated with merged pull request #898 for the reasons for making this change, other from the fact the whole-includes-the-parts logic dictates this change.


martinholmer:
The one-line change in pull request #900 was incorrect and has been undone as part of pull request #901 for the reasons provided there.

PR #930, opened 2016-09-19
Change call order of FairShareTax function

martinholmer:
This pull request started out with the goal of improving `current_law_policy.json` syntax (so as to eliminate several pylint warnings) and of streamlining FairShareTax calculations (so as to slightly speed them up).  In addition to these cosmetic changes, the `calculate.py` call order of the FairShareTax function has been changed so that it is the final income tax calculation.  Before this change, the FairShareTax was incorrectly called before the income tax calculation in the BenefitSurtax function.

@MattHJensen @feenberg @andersonfrailey @Amy-Xu @GoFroggyRun

### DIFF
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -189,13 +189,13 @@ class Calculator(object):
         C1040(self.policy, self.records)
         DecomposeEITC(self.policy, self.records)
         IITAX(self.policy, self.records)
-        ExpandIncome(self.policy, self.records)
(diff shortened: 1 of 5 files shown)


martinholmer:
Unless I hear today any concerns about pull request #930, which changes the call order of the new FairShareTax function, I'm going to merge #930 into the master branch at the end of the day.

@andersonfrailey @MattHJensen @feenberg @Amy-Xu @GoFroggyRun

PR #941, opened 2016-09-22
Include two new items in _surtax subtotal

martinholmer:
This pull request adds into the `_surtax` total both the Fair Share Tax and the high-AGI surtax amounts.  Previously, the only item included in the `_surtax` total was the itemized-deduction-benefits surtax, which continues to be included in the `_surtax` total.  Because the Fair Share Tax and the high-AGI surtax are not part of current-law policy, these changes have no effect on Tax-Calculator "Misc. Surtax" results.


MattHJensen:
@martinholmer, I had originally used "_surtax" as shorthand for "_benefitsurtax". I don't have a strong opinion on whether to rename `_surtax` as `_benefitsurtax` or alternatively to lump other surtaxes into the ` _surtax` variable. One complication with following the latter approach is that it is not always easy to identify what is a surtax. For instance, should `NIIT` be added to `_surtax` rather than `othertaxes`?

martinholmer:
@MattHJensen,  Per our conversation, I've changed the diagnostic table row label from "Misc. Surtax ($b)" to "Reform Surtaxes ($b)" to suggest that all surtax-like reforms are added together as an income tax subtotal on this line.

PR #945, opened 2016-09-24
_ID_BenefitSurtax_trt 1.0 -> 0.0

### LINKED ISSUE #943: Should the BenefitSurtax() rate be set at 0 or 1 under current law?
Over at #867, @feenberg asked:

> Also, I do not understand why the default surtax rate is shown as 1.0 (and
the credit also as 1.0) in the "Surtax on Itemized Deduction Beefit"
section. Surely the surtax rate should be .0. 

The reason why I set the surtax rate to 1 is that many (all) reforms in the news that use this feature have a surtax rate of 1. I'm talking about reforms along the lines of Feldstein, Feenberg, and Macguineas where the value of itemized deductions is limited to X% of AGI. If the surtax rate were set at 0, then the user would have to change both the surtax rate and the credit rate. With the surtax rate at 1, then the user will likely only have to change one parameter. 

The advantage of setting the surtax rate at 0 is that the user would have to think through exactly what the rate and credit parameters mean in order to specify their intended reform . Perhaps that would be better. What do you think @feenberg? 

cc @martinholmer, @andersonfrailey, @GoFroggyRun

MattHJensen:
Resolves #943.

### DIFF
--- a/taxcalc/current_law_policy.json
+++ b/taxcalc/current_law_policy.json
@@ -1727,7 +1727,7 @@
         "row_label": ["2013"],
         "cpi_inflated": false,
         "col_label": "",
-        "value": [1.0]
+        "value": [0.0]
     },
 
     "_PT_rt1": {


MattHJensen:
I'll merge this tomorrow if no one objects.

PR #948, opened 2016-09-26
Add MTR Plotting

GoFroggyRun:
This PR (mainly) adds two functions to `utils.py` file, namely the `get_mtr_data()` function and the `mtr_plot` function. Some tiny modifications have been applied to certain current functions. (mostly generalizations, and their tests have been updated accordingly) 
Instructions, descriptions as well as test suites are available for the two aded functions. Also, the added `styles.py` file defines a few layout formats for the plot, which can be used for all further plotting. 
Comments, concerns or remarks on these functions are more than welcomed. 
@MattHJensen @martinholmer

It seems that the Travis CI have failed since we don't really have Bokeh package installed. T.J. @talumbau, I was wondering would it be sufficient if I just add Bokeh to the `environment.yml` file? Please advice, thanks!


MattHJensen:
@talumbau, do you have any feedback on this PR? I'll be interested to hear from @martinholmer once he is back in town next week as well. Is there anything @GoFroggyRun could add to help you assess this? 

The goal is to start adding some out-of-the-box plotting capabilities to TC. I expect we'll have several other default plots we'd like to add as well.

MattHJensen:
@GoFroggyRun, do you know what's going on w the test failures?

GoFroggyRun:
@MattHJensen , it seems that the test failures are coming from the fact that Bokeh is not installed on TravisCI, which I am not so sure how to install. There's no test error on my local machine.

talumbau:
Sorry for the delay in reply. This is an interesting PR and definitely adds new capabilities to Tax-Calculator. With this PR as-is, it would mean that `bokeh` is now a hard dependency of the Tax-Calculator. So, if someone were to install `taxcalc` and not install `bokeh`, they would get error messages on the lack of `bokeh`. `bokeh` is a pretty hefty package, so I don't know if the right answer is to insist that bokeh be installed when Tax-Calculator is installed. I would prefer to make `bokeh` a "soft" dependency, in the same way that `numba` is a soft dependency. The idea would be that if bokeh is not installed, if the user attempts to call the functions that need bokeh, one would get an error message saying "bokeh must be installed. Please execute `conda install bokeh` to install bokeh."

talumbau:
To answer @GoFroggyRun's question about how to install `bokeh` in the Travis CI test environment, you need to modify the `.travis.yml` file. In this line:

https://github.com/open-source-economics/Tax-Calculator/blob/master/.travis.yml#L16

you should change it to include `bokeh` as a package that is installed in to the test environment, so the line would look like this:

```
  - conda create -n taxcalcdev python=$TRAVIS_PYTHON_VERSION pytest setuptools pandas=0.18 toolz six mock bokeh

```

martinholmer:
@talumbau suggested:
> `bokeh` is a pretty hefty package, so I don't know if the right answer is to insist that `bokeh` be installed when Tax-Calculator is installed. I would prefer to make `bokeh` a "soft" dependency, in the same way that `numba` is a soft dependency. The idea would be that if `bokeh` is not installed, if the user attempts to call the functions that need `bokeh`, one would get an error message saying "bokeh must be installed. Please execute 'conda install bokeh' to install bokeh."

I like this suggestion, but if this suggestion is followed how do the unit tests work?  Can `test_mtr_plot` in the `test_utils.py` file be written so that it is automatically skipped when `bokeh` is not installed?

@GoFroggyRun @MattHJensen

martinholmer:
@GoFroggyRun, I don't understand your proposed new file called `taxcalc/styles.py`.  For example, what does microseconds have to do with plotting marginal tax rates?  If there is anything in the `styles.py` file that is used by your mtr plotting capability, please put what is needed in your new utility function.  I don't see the need for a new `taxcalc/styles.py` file.

GoFroggyRun:
@martinholmer:
```
 I don't see the need for a new taxcalc/styles.py file.
```
Thanks for your comments. I added the `taxcalc/styles.py` file that includes some other layout which might be helpful for other future plotting functions. And yes, you are right, the file is not necessary for the moment, and thus I removed it.

```
bokeh is a pretty hefty package, so I don't know if the right answer is to insist that bokeh be installed when Tax-Calculator is installed. I would prefer to make bokeh a "soft" dependency, in the same way that numba is a soft dependency. The idea would be that if bokeh is not installed, if the user attempts to call the functions that need bokeh, one would get an error message saying "bokeh must be installed. Please execute 'conda install bokeh' to install bokeh."
```
I agreed as well. But I'm having similar concerns, which is how do the unit tests work if we were not forcing `bokeh` dependency?

After adding `bokeh` to the environment (for the moment, and before the previous question is dealt with), I'm still having one error message reads 
```
NameError: name 'TAXDATA' is not defined
```
I have no idea why I'm having such error message, as this `TAXDATA` file has been used many times in `test_utils.py` file.

martinholmer:
@GoFroggyRun,  You need to add more tests of your new code in pull request #948.

@MattHJensen

GoFroggyRun:
@martinholmer:
I just added one more test to ensure codecov tests will pass. Please take a look.

martinholmer:
@GoFroggyRun said:
> I just added one more test to ensure codecov tests will pass. 

Good.

I'll try to do some plotting with this pull request soon and give you some feedback.

But meanwhile, what is the answer to my question about what is going to happen on computers where bokeh is not installed.  What is going to happen with the unit tests?  It is not acceptable that they fail given all the reasons advanced by @talumbau.

@GoFroggyRun @MattHJensen

talumbau:
Sorry for the delay in reply. If this PR is functionally complete, I would like to suggest some additional commits, which I can do as a pull request to Sean's branch. The commits that I will add will make the following changes:

 - for this branch as-is, `taxcalc` will fail to import if `bokeh` is not installed. In my proposed commits, `taxcalc` will succeed at importing even in `bokeh` is not installed

 - the function `mtr_plot` will raise an exception when called if `bokeh` is not installed. The exception will contain a message telling the user to install `bokeh` in order to access this functionality. If `bokeh` is installed, the `mtr_plot` function will execute as normal.

 - In this pull request, we test the functions in the presence of `bokeh`. I will add some way of testing in the absence of `bokeh` (similar to what we do with Numba right now).

martinholmer:
@talumbau said:
> Sorry for the delay in reply. If this PR is functionally complete, I would like to suggest some additional commits, which I can do as a pull request to Sean's branch. The commits that I will add will make the following changes:
>
> * for this branch as-is, taxcalc will fail to import if bokeh is not installed. In my proposed commits, taxcalc will succeed at importing even in bokeh is not installed
>
> * the function mtr_plot will raise an exception when called if bokeh is not installed. The exception will contain a message telling the user to install bokeh in order to access this functionality. If bokeh is installed, the mtr_plot function will execute as normal.
>
> * In this pull request, we test the functions in the presence of bokeh. I will add some way of testing in the absence of bokeh (similar to what we do with Numba right now).

@talumbau,  These contributions would be very much appreciated, especially the last one.

@GoFroggyRun @MattHJensen

talumbau:
@GoFroggyRun I set up a PR to your branch here:

https://github.com/GoFroggyRun/Tax-Calculator/pull/1

It looks like my code is not completely in sync with your branch because I had a more recent version of master when I branched off and pulled your changes. The relevant commit for the items above is this one:

https://github.com/GoFroggyRun/Tax-Calculator/pull/1/commits/ab3c5f5ea842adf4bbced6a4d6a35ea6decea90f

talumbau:
I think the correct course of action would be either of the following:

 - @GoFroggyRun can merge master into his local `mtr_plot` branch and then push to `GoFroggyRun/Tax-Calculator`, which would then mean that my PR is only commit GoFroggyRun@ab3c5f5

 - @GoFroggyRun can merge my PR into his branch directly, which would get the branch up to date with master, and then trigger Travis CI to run everything and see if test coverage is good with my proposed changes.

GoFroggyRun:
Thanks @talumbau for your PR. 
I followed your second suggestion and merged your PR into my branch directly. I much appreciate your contributions.

talumbau:
The code coverage indicates that these lines are not executed during testing:

```
except ImportError:
    BOKEH_AVAILABLE = False
```

this is because Travis CI is set up to install `bokeh` for these tests. In `test_mtr_plot_force_no_bokeh` I force the execution of the non-bokeh code path by manually setting the BOKEH_AVAILABLE variable, but this does not force the `ImportError` exception to occur, which is the only way to get the above lines to run. The only way to get these lines to run would be to the run the whole test suite w/o bokeh, which seems duplicative.

martinholmer:
@GoFroggyRun, I downloaded pull request #948 and all 255 unit-tests passed on that branch because I have `bokeh` installed on my computer.

Then I tried to use the new utility functions you've added in this pull request.  What happened is below.  I don't understand the error message.  Did I do something wrong?
```
iMac2:tax-calculator mrh$ cp ../tax-calculator-data/puf.csv .
iMac2:tax-calculator mrh$ python
Python 2.7.11 |Anaconda 2.3.0 (x86_64)| (default, Dec  6 2015, 18:57:58) 
[GCC 4.2.1 (Apple Inc. build 5577)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
Anaconda is brought to you by Continuum Analytics.
Please check out: http://continuum.io/thanks and https://anaconda.org
>>> from taxcalc import *
>>> calcx = Calculator(policy=Policy(), records=Records())
You loaded data for 2009.
>>> calcy = Calculator(policy=Policy(), records=Records())
You loaded data for 2009.
>>> calcy.policy.implement_reform({2013:{'_SS_Earnings_c':[1e99]}})
>>> df = get_mtr_data(calcx,calcy,'weighted_count')
taxcalc/utils.py:122: FutureWarning: sort(columns=....) is deprecated, use sort_values(by=.....)
  df.sort(income_measure, inplace=True)
  File "<stdin>", line 1, in <module>
  File "taxcalc/utils.py", line 709, in get_mtr_data
    wgtpct_x = gp_x.apply(weights, 'mtr_iit')
  File "/Users/mrh/anaconda/lib/python2.7/site-packages/pandas/core/groupby.py", line 634, in apply
    @wraps(func)
  File "/Users/mrh/anaconda/lib/python2.7/functools.py", line 33, in update_wrapper
    setattr(wrapper, attr, getattr(wrapped, attr))
AttributeError: 'str' object has no attribute '__module__'
>>> 
```

@MattHJensen @talumbau

GoFroggyRun:
@martinholmer,
Sorry about the mistakes in the documentation of `get_mtr_data`, to choose from different groupby objects, the quotes are not required. So, for example,
```
df = get_mtr_data(calcx,calcy,weighted_perc_dec) 
```
would go through. 

I also realized that `weighted_count` and `weighted_share_of_total` cannot be supported by the function, and thus removed them from the documentation. Please take a look and let me know if you find anything else looks suspicious.

martinholmer:
@GoFroggyRun, After your recent fix to the `get_mtr_data` weights documentation, I downloaded the pr-948 branch again.  Here is what I tried and below that I ask a question about why I can't see the plot.
```
iMac2:tax-calculator mrh$ cp ../tax-calculator-data/puf.csv .
iMac2:tax-calculator mrh$ python
Python 2.7.11 |Anaconda 2.3.0 (x86_64)| (default, Dec  6 2015, 18:57:58) 
[GCC 4.2.1 (Apple Inc. build 5577)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
Anaconda is brought to you by Continuum Analytics.
Please check out: http://continuum.io/thanks and https://anaconda.org
>>> from taxcalc import *
>>> calcx = Calculator(policy=Policy(), records=Records())
You loaded data for 2009.
>>> calcy = Calculator(policy=Policy(), records=Records())
You loaded data for 2009.
>>> calcy.policy.implement_reform({2013:{'_SS_Earnings_c':[1e99]}})
>>> df = get_mtr_data(calcx,calcy,weighted_mean)
>>> fig = mtr_plot(df)
>>> fig.show()
  File "<stdin>", line 1, in <module>
AttributeError: 'Figure' object has no attribute 'show'
>>> show(fig)
  File "<stdin>", line 1, in <module>
NameError: name 'show' is not defined
>>> exit()
iMac2:tax-calculator mrh$ 
```
What do you mean by the following in the `plot_mtr` docstring?
```
Returns
 +    -------
 +    Figure Object (Use show() option to visualize)
```
Each of these two new utility functions should have a docstring that explains what the function arguments mean and what to do with the returned objects.  I know I'm not the most clever person around, but the current documentation is not clear enough for me to produce a graph that I can see.  To me the words `Use the show() option to visualize` means that the Figure object returned by the `plot_mtr` functions has a method called `show`, but that doesn't seem to be the case.
What am I doing wrong?

GoFroggyRun:
@martinholmer:

Thanks for looking into this, I never tried such function in command line environment and thus was aware of this issue. Thanks for bringing this up. 

I used the following code:
```
(taxcalc-dev) Sean-Wangs-MacBook-Pro:tax-calculator seanwang$ python
Python 2.7.11 |Continuum Analytics, Inc.| (default, Dec  6 2015, 18:57:58) 
[GCC 4.2.1 (Apple Inc. build 5577)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
Anaconda is brought to you by Continuum Analytics.
Please check out: http://continuum.io/thanks and https://anaconda.org
>>> from taxcalc import *
>>> calcx = Calculator(policy=Policy(), records=Records())
You loaded data for 2009.
>>> calcy = Calculator(policy=Policy(), records=Records())
You loaded data for 2009.
>>> calcy.policy.implement_reform({2013:{'_SS_Earnings_c':[1e99]}})
>>> df = get_mtr_data(calcx,calcy,weighted_mean)
>>> fig = mtr_plot(df)
>>> output_file("test_plot.html")
>>> show(fig)
>>> 
```
and was able to get the output in a pop-up window. 

The issue here is that, when using a command line, an output file has to be defined, here we use `output_file("test_plot.html")`, to have the plot shown up. 

I'll add this to documentation.

martinholmer:
@GoFroggyRun, Two things: a question and then a request.

You said:
> I never tried such function in command line environment and thus was aware of this issue.

How do you use these new plotting functions?

The request:  Please fix the PEP8 errors generated by you latest commit.  And also, please correct/improve the docstring documentation for the `get_mtr_data` function.

GoFroggyRun:
@martinholmer:
>
How do you use these new plotting functions?

I used a tool called [jupyter notebook](https://ipython.org/notebook.html), which is similar to the command line tool. 

>
Please fix the PEP8 errors generated by you latest commit. And also, please correct/improve the docstring documentation for the get_mtr_data function.

Done. Please take a look again. Thanks!

martinholmer:
@GoFroggyRun said:
> The issue here is that, when using a command line [Python interpreter], an output file has to be defined, here we use output_file("test_plot.html") to have the plot shown up.
I'll add this to documentation.

Still no luck.  What am I doing wrong?
```
iMac2:tax-calculator mrh$ cp ../tax-calculator-data/puf.csv .
iMac2:tax-calculator mrh$ python
Python 2.7.11 |Anaconda 2.3.0 (x86_64)| (default, Dec  6 2015, 18:57:58) 
[GCC 4.2.1 (Apple Inc. build 5577)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
Anaconda is brought to you by Continuum Analytics.
Please check out: http://continuum.io/thanks and https://anaconda.org
>>> from taxcalc import *
>>> calcx = Calculator(policy=Policy(), records=Records())
You loaded data for 2009.
>>> calcy = Calculator(policy=Policy(), records=Records())
You loaded data for 2009.
>>> calcy.policy.implement_reform({2013:{'_SS_Earnings_c':[1e99]}})
>>> df = get_mtr_data(calcx, calcy, weighted_mean)
taxcalc/utils.py:122: FutureWarning: sort(columns=....) is deprecated, use sort_values(by=.....)
  df.sort(income_measure, inplace=True)
>>> fig = mtr_plot(df)
>>> output_file('pr-948-fig.html')
  File "<stdin>", line 1, in <module>
NameError: name 'output_file' is not defined
>>> fig.output_file('pr-948-fig.html')
  File "<stdin>", line 1, in <module>
AttributeError: 'Figure' object has no attribute 'output_file'
>>> exit()
iMac2:tax-calculator mrh$ 
```

GoFroggyRun:
@martinholmer:

Not sure why this is happening.

Could you try to run
```
source activate taxcalc-dev
```
before running `python`?

martinholmer:
@GoFroggyRun said:
>Not sure why this is happening.
> Could you try to run
> `source activate taxcalc-dev`
> before running python?

Why do I need to do that?  Shouldn't your pull request #948 work in all environments?

MattHJensen:
@martinholmer said:

> Why do I need to do that? Shouldn't your pull request #948 work in all environments?

I don't think so. Currently tax-calculator is only guaranteed to work in environments described by its [environment.yml](https://github.com/open-source-economics/Tax-Calculator/blob/master/environment.yml). That environment is easily activated by running  `source activate taxcalc-dev` as described in our getting started guide. It could be that those requirements should be relaxed, but isn't that a separate issue?  I don't think we should expect new PRs to work in environments that TC isn't guaranteed to handle.

martinholmer:
@MattHJensen said:
> Currently Tax-Calculator is only guaranteed to work in environments described by its `environment.yml`. An environment that meets those criterium is easily activated by running `source activate taxcalc-dev` as described in our getting started guide.

Actually, it requires `conda env create` before `source activate taxcalc-dev`; then I did `conda install bokeh`.

But PR#948 still doesn't work for me.  I'm still getting the same `output_file is not defined` error I got before I changed environments.  So, it would seem as if the conda environment issue has nothing to do with this problem in PR#948.

What I did is below.  What am I doing wrong?
```
(taxcalc-dev) iMac2:tax-calculator mrh$ cat pr-948-test.py 
from taxcalc import *
calcx = Calculator(policy=Policy(), records=Records())
calcy = Calculator(policy=Policy(), records=Records())
calcy.policy.implement_reform({2013: {'_SS_Earnings_c': [1e99]}})
df = get_mtr_data(calcx, calcy, weighted_mean)
fig = mtr_plot(df)
output_file('pr-948-fig.html')
show(fig)

(taxcalc-dev) iMac2:tax-calculator mrh$ python pr-948-test.py 
You loaded data for 2009.
You loaded data for 2009.
/Users/mrh/work/OSPC/tax-calculator/taxcalc/utils.py:122: FutureWarning: sort(columns=....) is deprecated, use sort_values(by=.....)
  df.sort(income_measure, inplace=True)
/Users/mrh/anaconda/envs/taxcalc-dev/lib/python3.5/site-packages/bokeh/util/deprecation.py:33: BokehDeprecationWarning: legend_spacing was deprecated in Bokeh 0.12.3 and will be removed, use Legend.spacing instead.
  warn(message)
/Users/mrh/anaconda/envs/taxcalc-dev/lib/python3.5/site-packages/bokeh/util/deprecation.py:33: BokehDeprecationWarning: legend_padding was deprecated in Bokeh 0.12.3 and will be removed, use Legend.padding instead.
  warn(message)
  File "pr-948-test.py", line 7, in <module>
    output_file('pr-948-fig.html')
NameError: name 'output_file' is not defined
(taxcalc-dev) iMac2:tax-calculator mrh$ 
```

GoFroggyRun:
@martinholmer:

Sorry that I just realized `output_file` and `show` were being removed from `utilis.py` after T.J. helped me with the test suite, and thus the error message reads that they are not defined. 

After the most recent commit, everything should be good to go.  Please re-run all commands.

martinholmer:
@GoFroggyRun said:
> Sorry that I just realized output_file and show were being removed from utilis.py after T.J. helped me with the test suite, and thus the error message reads that they are not defined.
> After the most recent commit, everything should be good to go. Please re-run all commands.

OK, I downloaded the newest version of pull request #948 and did this (trying to follow the documentation):
```
iMac2:tax-calculator mrh$ cat pr-948-test.py
from taxcalc import *
calcx = Calculator(policy=Policy(), records=Records())
calcy = Calculator(policy=Policy(), records=Records())
calcy.policy.implement_reform({2013: {'_SS_Earnings_c': [1e99]}})
df = get_mtr_data(calcx, calcy, weighted_mean, mtr_measure='_combined')
fig = mtr_plot(df)
output_file('pr-948-fig.html')
show(fig)
iMac2:tax-calculator mrh$ python pr-948-test.py
You loaded data for 2009.
You loaded data for 2009.
/Users/mrh/work/OSPC/tax-calculator/taxcalc/utils.py:122: FutureWarning: sort(columns=....) is deprecated, use sort_values(by=.....)
  df.sort(income_measure, inplace=True)
  File "pr-948-test.py", line 5, in <module>
    df = get_mtr_data(calcx, calcy, weighted_mean, mtr_measure='_combined')
  File "/Users/mrh/work/OSPC/tax-calculator/taxcalc/utils.py", line 714, in get_mtr_data
    wpct_x = DataFrame(data=wgtpct_x, columns=['w_mtr'])
UnboundLocalError: local variable 'wgtpct_x' referenced before assignment
iMac2:tax-calculator mrh$ 
```
What am I doing wrong?

Also, please fix your code so that you avoid using obsolete library methods and eliminate the `FutureWarning: sort(columns=....) is deprecated, use sort_values(by=.....)` message.

GoFroggyRun:
@martinholmer thanks for these feedback. Fixed more documentation, and deprecation issue as well.

martinholmer:
@GoFroggyRun said:
> [T]hanks for these feedback.  Fixed more documentation, and deprecation issue as well.

Using the latest version of PR #948, I still cannot produce a plot.  What am I doing wrong?
```
iMac2:tax-calculator mrh$ cat pr-948-test.py
from taxcalc import *
calcx = Calculator(policy=Policy(), records=Records())
calcy = Calculator(policy=Policy(), records=Records())
calcy.policy.implement_reform({2013: {'_SS_Earnings_c': [1e99]}})
df = get_mtr_data(calcx, calcy, weighted_mean, mtr_measure='_combined')
fig = mtr_plot(df)
output_file('pr-948-fig.html')
show(fig)

iMac2:tax-calculator mrh$ cp ../tax-calculator-data/puf.csv .

iMac2:tax-calculator mrh$ python pr-948-test.py 
You loaded data for 2009.
You loaded data for 2009.
  File "pr-948-test.py", line 6, in <module>
    fig = mtr_plot(df)
  File "/Users/mrh/work/OSPC/tax-calculator/taxcalc/utils.py", line 753, in wrapped_f
    return fn(*args, **kwargs)
  File "/Users/mrh/work/OSPC/tax-calculator/taxcalc/utils.py", line 818, in mtr_plot
    PP.legend.spacing = 5
  File "/Users/mrh/anaconda/lib/python2.7/site-packages/bokeh/plotting/helpers.py", line 401, in __setattr__
    setattr(x, attr, value)
  File "/Users/mrh/anaconda/lib/python2.7/site-packages/bokeh/core/properties.py", line 722, in __setattr__
    (name, self.__class__.__name__, text, nice_join(matches)))
AttributeError: unexpected attribute 'spacing' to Legend, similar attributes are legend_spacing
iMac2:tax-calculator mrh$ 
```

talumbau:
@martinholmer I was able to reproduce the problem you show above. I fixed it by updating bokeh as follows:

```
conda update bokeh
```

So it seems clear this PR depends on the latest version of bokeh, version 0.12.3. This PR should reflect that, probably in a version specification in the environment.yaml.

GoFroggyRun:
@martinholmer: try execute `source activate taxcalc-dev` before going into python environment. 
If it still doesn't work, make sure you have the latest bokeh package:
```
bokeh                     0.12.3                   py35_0 
```
via command `conda list`.

martinholmer:
@GoFroggyRun said:
> Try execute `source activate taxcalc-dev` before going into python environment. 

I don't understand why you keep saying this; we have established that this is **not** the problem.

> If it still doesn't work, make sure you have the latest bokeh package:
> `bokeh               0.12.3                   py35_0`
> via command `conda list`.

This is the problem as @talumbau pointed out.  Your pull request #948 does not specify which version of `bokeh` is required.  You need to fix that in your pull request.

Also, if you are going to add `bokeh` to the `environment.yml` file, which makes it a **mandatory** part of the Tax-Calculator environment, why do we need all the BOKEH wrapper code in `utils.py`?

GoFroggyRun:
@martinholmer:
>
This is the problem as @talumbau pointed out. Your pull request #948 does not specify which version of bokeh is required. You need to fix that in your pull request.

Instead of requiring bokeh version, I revert part of the changes in [db5cad0](https://github.com/open-source-economics/Tax-Calculator/pull/948/commits/db5cad04f4750b6aecd4e98e81d74634ca7573fb) so that the code will also support pervious bokeh versions. (a deprecation error in the latest version though)

>
Also, if you are going to add bokeh to the environment.yml file, which makes it a mandatory part of the Tax-Calculator environment, why do we need all the BOKEH wrapper code in utils.py?

We still need these wrapper code, otherwise there might be an error.

martinholmer:
After upgrading to the current version of `bokeh` --- which appears to be a requirement for pull request #948 to work even though this was not mentioned anywhere in #948 and not specified in the `environment.yml` file --- I can get a graph of mtr values by setting the `get_mtr_data` function parameter `weights` to `weighted_mean`.  However, the documentation for the `weights` parameter is inaccurate: if I follow those directions and set `weights` to a string, things don't work.

The documentation of the `weights` parameter of the `get_mtr_data` function docstring needs to be corrected.  And in the process of doing that correction the documentation needs to be enhanced to say what all seven of the allowed values of `weights` mean and in what circumstances they would be useful.

@GoFroggyRun @MattHJensen @talumbau

martinholmer:
@GoFroggyRun said in response to a question:
>> Also, if you are going to add bokeh to the environment.yml file, which makes it a mandatory part of the Tax-Calculator environment, why do we need all the BOKEH wrapper code in utils.py?
>
> We still need these wrapper code, otherwise there might be an error.

I don't understand what you mean by "otherwise there might be an error".  With the complicated wrapper code in `utils.py` there will be an error when `bokeh` is not installed.  So, if the `environment.yml` file is going to include `bokeh` as a required part of the Tax-Calculator environment, why is there a need to add all the code complexity to check for the availability of `bokeh`?

@MattHJensen  @talumbau

talumbau:
> I don't understand what you mean by "otherwise there might be an error". With the complicated wrapper code in utils.py there will be an error when bokeh is not installed. So, if the environment.yml file is going to include bokeh as a required part of the Tax-Calculator environment, why is there a need to add all the code complexity to check for the availability of bokeh?

The `environment.yml` file is a helpful way to specify the packages (and versions of those packages) that are needed in order to make use of the Tax-Calculator. The term "needed" here is a bit fuzzy: it can mean anything from "if you don't have the specified version of pandas, the code will break" to "you can't access the full functionality of Tax-Calculator if you don't have this package installed."

A _dependency_ is a more formal relationship specified in whatever tool is used to package up the Tax-Calculator. If a formal dependency is specified, this is saying that the installation of the package (in this case, `taxcalc`) _requires_ the installation of the packaged specified as a dependency.

When we perform the release procedure for `taxcalc`, we create two separate "builds" of taxcalc, one for PyPI, hosted here:

https://pypi.python.org/pypi/taxcalc

and one with the equivalent conda package, hosted here:

https://anaconda.org/ospc/taxcalc

These releases have formal dependencies specified within the metadata particular to the given format. In both cases, when one installs the package, the tool attempts to install each dependent package listed in the metadata. This is the place where we wish to avoid the "hard dependency" of bokeh. If bokeh is listed as a hard dependency for the PyPI distribution or the conda distribution, it will place an undue burden on anyone attempting to use the Tax-Calculator by installing these packages. One of the biggest issues is that it would bloat the installation size of the web application.

martinholmer:
@talumbau, Thanks for your explanation of the complexities of "needed" versus "required" Python packages for Tax-Calculator.  So, if I understand correctly, `bokeh` is not in either the PyPI distribution or the conda distribution because those distributions do not need `bokeh`.  But pull request #948 would always install `bokeh` on the local computer of a user who follows the [Contributor Guide](http://taxcalc.readthedocs.io/en/latest/contributor_guide.html) installation instructions, right?  So, if all the local users --- the people who might want to use plotting utilities in #948 --- have `bokeh`, why is all the BOKEH_AVAILABLE and wrapper business in the `utils.py` file needed?

You explanation about "needed" versus "required" was useful, but after reading your explanation, the presence of all the `bokeh` wrapping code seems even more puzzling.

talumbau:
The word choice is not particularly important ("needed" vs. "required"), but rather the meaning behind the addition of libraries to the `environment.yml` vs. adding libraries in the `setup.py` or the `meta.yaml`, for PyPI packaging or conda packaging, respectively.

For your question, "why have the helper code in the utils.py that checks for bokeh?", perhaps it will help to explain the consequences of not having that code. If this branch is merged without this code, here is what would happen in a couple scenarios:

In the case of no `bokeh` wrapper helper:
1. A developer of Tax-Calculator (who regularly works with the code in the repo) updates to master. The developer does not have bokeh installed. The code will raise an exception upon import of `taxcalc` and so no work can be accomplished until `bokeh` is installed. The developer can execute `conda env update` and, because the `environment.yml` has been updated to include the line `bokeh`, `bokeh` will be installed and the user can go back to using `taxcalc`, which may or may not involve `bokeh` at all.

2. A released version of `taxcalc` is installed on a user's machine via `pip install` or `conda install`, or `taxcalc` is installed on one of the machines that is used to run the TaxBrain web appplication. The code will raise an exception on the import of `taxcalc` and won't be able to do anything unless `bokeh` is installed, even though`bokeh` is not required for any of the calculations involving `taxcalc` on the web app.

In the case of having the `bokeh` wrapper helper:
1. A developer of Tax-Calculator (who regularly works with the code in the repo) updates to master. The developer does not have bokeh installed. The developer can continue to use the Tax-Calculator with no issues until code is called that requires `bokeh`. The developer can then execute `conda env update` and, because the `environment.yml` has been updated to include the line `bokeh`, `bokeh` will be installed and the user can proceed with using the parts of the Tax-Calculator that require `bokeh`.

2.  A released version of `taxcalc` is installed on a user's machine via `pip install` or `conda install`, or `taxcalc` is installed on one of the machines that is used to run the TaxBrain web appplication. The user's code will work as expected as long as the user does not call code that requires `bokeh`. The web application will continue to work as expected because it does not call this `bokeh`-specific code.

GoFroggyRun:
@martinholmer, I have modified the input type for weighting option to `String that coincides with tax-calculator defined function`, but still not so sure whether this is the most accurate way. I have also got rid of weighting methods that do not seem to be helpful, while provided more documentations for the remaining ones.

martinholmer:
@GoFroggyRun said:
> I have modified the input type for weighting [parameter of the `get_mtr_data` utility function] to String that coincides with tax-calculator defined function, but still not so sure whether this is the most accurate way. I have also got rid of weighting methods that do not seem to be helpful, while provided more documentations for the remaining ones.

There are at least two sets of problems with the new `weighting` parameter: (1) the documentation of the `weighting` parameter in the function docstring is incorrect in saying the parameter is of type `String`, and (2) there is no error trapping when incorrect values for the `weighting` parameter.  Here are the details:

(1)  Incorrect documentation:
```
$ cat pr-948-test.py
from taxcalc import *
calcx = Calculator(policy=Policy(), records=Records())
calcy = Calculator(policy=Policy(), records=Records())
calcy.policy.implement_reform({2013: {'_SS_Earnings_c': [1e99]}})
df = get_mtr_data(calcx, calcy, 'wage_weighted',
                  mtr_measure='_combined')
fig = mtr_plot(df, plot_width=850, plot_height=500)
output_file('pr-948-fig.html')
show(fig)

$ python pr-948-test.py
You loaded data for 2009.
You loaded data for 2009.
  File "pr-948-test.py", line 6, in <module>
    mtr_measure='_combined')
  File "/Users/mrh/work/OSPC/tax-calculator/taxcalc/utils.py", line 716, in get_mtr_data
    wgtpct_x = gp_x.apply(weighting, 'mtr_combined')
  File "/Users/mrh/anaconda/lib/python2.7/site-packages/pandas/core/groupby.py", line 634, in apply
    @wraps(func)
  File "/Users/mrh/anaconda/lib/python2.7/functools.py", line 33, in update_wrapper
    setattr(wrapper, attr, getattr(wrapped, attr))
AttributeError: 'str' object has no attribute '__module__'
$
```

(2) No error trapping:
```
$ cat pr-948-test.py
from taxcalc import *
calcx = Calculator(policy=Policy(), records=Records())
calcy = Calculator(policy=Policy(), records=Records())
calcy.policy.implement_reform({2013: {'_SS_Earnings_c': [1e99]}})
df = get_mtr_data(calcx, calcy, weighted_count_lt_zero,
                  mtr_measure='_combined')
fig = mtr_plot(df, plot_width=850, plot_height=500)
output_file('pr-948-fig.html')
show(fig)

$ python pr-948-test.py
You loaded data for 2009.
You loaded data for 2009.
$
```
In case (2), there is a plot produced, but it doesn't correspond to anything in the documentation of the `get_mtr_data` utility function.

GoFroggyRun:
@martinholmer, thanks for your feedback. 
I am interested in your tips on addressing the two issues here. In particular, what do you think is the most accurate type for `weighting` parameter. 

One quick fix that would also deal with the two issues above would be making the `weighting` parameter of string type, and adding an if loop. Not sure which approach (change `weighting` to string or not) is better, please advice. Thanks!

martinholmer:
@GoFroggyRun said:
> One quick fix that would also deal with the two issues above would be making the weighting parameter of string type, and adding an if loop. Not sure which approach (change weighting to string or not) is better, please advice.

That would make sense: the series of if statements can trap undefined weighting string values.

martinholmer:
@GoFroggyRun,  The latest version looks much better with respect to the `weighting` parameter.

Before going any further in your development, you need to merge the most recent version of master into `mtr_plot` development branch.  You haven't done this for some time, so there is are a lot of code changes on master that merging your branch would wipe out.  This would require an undoing of the merger of your pull request.  So, to avoid this sort of time wasting for everybody involved, please update your master branch on your local computer and then merge it into this development branch before working on the following list of issues.

You need to augment/correct the documentation (in the docstring) for the remaining parameters of the `get_mtr_data` utility function.  The issues are below:
```
   MARS : Integer
        options for input: 1, 2, 3, 4
        Choose different filling status
```
Integer is incorrect (or incomplete type) and you need to document the 'ALL' option.
```
    income_measure : String object
        options for input: '_expanded_income', 'c00100', 'e00200'
        classifier of income bins/deciles
```
Please define for novice users what these three options mean (e.g., `c00100` means AGI).
```
    mtr_measure : String object
        options for input: '_iitax', '_combined'
        Choose different marginal tax rate measure
```
Explain in words what the two options mean and please change the default value of this parameter to '_combined'.
```
    complex_weight : Boolean
        The cumulated sum will be carried out based on weighted income measure
        if this option is true
```
I have no idea what this parameter does.  Can you give more explanation?

GoFroggyRun:
@martinholmer sorry about the delay. Just addressed all your comments in the latest commit, please take a look.

martinholmer:
@GoFroggyRun said:
> Just addressed all your comments in the latest commit, please take a look.

This is much better documentation of the `get_mtr_data` function, but I'm still confused about the complex_weight parameter, which has a default value of False.  Looking at the code, which is not something anybody using the `get_mtr_data` function should have to do, it seems as if "complex" simply means that the income measure percentile bins are constructed using sample weights.  Is that correct?  If so, why the use of "complex"?  Why not something a little more informative, like "weighted_income_measure"?

If my interpretation is correct, I have two questions (one minor and one major) about the effect of changing complex_weight from False to True.

(1) Why do the percentile bins run from 0 to 99 when complex_weight=False, but run from 1 to 100 when complex_weight=True?  This anomaly does nothing but create doubts in the minds of users.

(2) Why are the graphs so different?  [At this stage I should be able to show you the two very different looking graphs, but I don't know how to do that and there isn't any explanation about how to do that in the documentation of the `mtr_plot` function.]   Here is the script is used:
```
$ cat pr-948-test.py
from taxcalc import *
calcx = Calculator(policy=Policy(), records=Records())
calcy = Calculator(policy=Policy(), records=Records())
calcy.policy.implement_reform({2013: {'_SS_Earnings_c': [1e99]}})
df = get_mtr_data(calcx, calcy, complex_weight=False) # change False to True
fig = mtr_plot(df, plot_width=850, plot_height=500)
output_file('pr-948-fig-complexweight-F.html') # change F to T
show(fig)
```
I ran it twice: once the way it is above and once with the two changes described in the comments. [I will send you a private email with the two HTML plots attached.]

The plot of data generated with complex_weight=False, shows that the top 10% of (unweighted) filing units would have higher taxes.  But the plot of data generated with complex_weight=True, shows that the top 45% of (weighted) filing units would have higher taxes.  I simply do not believe that nearly half of the population will have higher taxes when the Social Security maximum taxable earnings is raised to infinity.  Have I done something wrong?

GoFroggyRun:
@martinholmer, regarding your concerns:
> Why not something a little more informative, like "weighted_income_measure"?

I changed the option name to `weight_by_income_measure`. (See reasons below)

>(1) Why do the percentile bins run from 0 to 99 when complex_weight=False, but run from 1 to 100 when complex_weight=True? This anomaly does nothing but create doubts in the minds of users.

Fixed. Thanks for noticing this. 

>(2) Why are the graphs so different? [At this stage I should be able to show you the two very different looking graphs, but I don't know how to do that and there isn't any explanation about how to do that in the documentation of the mtr_plot function.] 

I just realized my wording wasn't quite right. The option actually allows `s006` not only to be sorted according to certain income measure, but also to be weighted by this type of income measure. So it's actually not a matter of weighted or not, we are looking at completely different measure when `weight_by_income_measure` is True. Does this make sense to you?

martinholmer:
@GoFroggyRun said in response to this question:
>> Why are the graphs so different? [At this stage I should be able to show you the two very different looking graphs, but I don't know how to do that and there isn't any explanation about how to do that in the documentation of the mtr_plot function.]
>>
>> The plot of data generated with complex_weight=False, shows that the top 10% of (unweighted) filing units would have higher taxes. But the plot of data generated with complex_weight=True, shows that the top 45% of (weighted) filing units would have higher taxes. I simply do not believe that nearly half of the population will have higher taxes when the Social Security maximum taxable earnings is raised to infinity.
>
> I just realized my wording wasn't quite right. The option actually allows s006 not only to be sorted according to certain income measure, but also to be weighted by this type of income measure. So it's not a matter of being weighted or not, but, rather, we are looking at completely different measure when weight_by_income_measure is True. Namely s006 (sorted by income_measure), and s006*income_measure (sorted by income_measure). Does this make sense to you?

No, this does not make any sense to me.  All I know is that the graph with the weight_by_income_measure=True does not make any sense.

And, by the way, what's the answer to my question about how to embed these graphs in our conversation?

GoFroggyRun:
@martinholmer said:
>
No, this does not make any sense to me. All I know is that the graph with the weight_by_income_measure=True does not make any sense.

So the second plot doesn't any longer, after setting `weight_by_income_measure=True`, mean that 
45% of (weighted) **filing units** would have higher taxes. It means that 45% of the **aggregate income** (sum of s006*e00200) is subject to higher taxes. This option is not about weighted vs unweighted, but rather about what to weight s006 by. This option allows us to investigate different activity (wage/agi/etc..) of interests, instead of filling units. 

>
And, by the way, what's the answer to my question about how to embed these graphs in our conversation?

Not sure what do you mean by embedding these graphs. Are you suggesting a different way of presenting the plots?

martinholmer:
@GoFroggyRun said:
> So the second plot doesn't any longer, after setting weight_by_income_measure=True, mean that 45% of (weighted) filing units would have higher taxes. It means that 45% of the aggregate income (sum of s006*e00200) is subject to higher taxes. This option is not about weighted vs unweighted, but rather about what to weight s006 by. This option allows us to investigate different activity (wage/agi/etc..) of interests, instead of filling units.

The above explanation make sense (except for the confusing use of "activity".  But the explanation above is not what is in the documentation of the `weight_by_income_measure` parameter in the `get_mtr_data` docstring.  You need to revise that docstring to explain what you are saying above, but avoid the confusing "activity" word and say what you mean directly.

@GoFroggyRun also said:
> Not sure what do you mean by embedding these graphs.

Here is a concrete example: how do I put a generated graph into a LaTeX document?  I know how to display an EPS graph as a LaTeX figure, but how do I include your graphs in my document?

Here is a second example: how do I show a generated graph in a GitHub (issue or pull request) conversation?

martinholmer:
@GoFroggyRun,  I was using the MTR plotting capability as currently structured in pending pull request #948 and got a result that was puzzling to me.  Here is the script I ran:
```
$ cat pr-948-test.py
from taxcalc import *
calcx = Calculator(policy=Policy(), records=Records())
calcy = Calculator(policy=Policy(), records=Records())
calcy.policy.implement_reform({2013: {'_SS_Earnings_c': [1e99]}})
df = get_mtr_data(calcx, calcy, MARS='ALL', weight_by_income_measure=False)
# print df
fig = mtr_plot(df, plot_width=850, plot_height=500)
output_file('pr-948-fig-complexweight-F-MARS-ALL.html')
show(fig)

$ python pr-948-test.py
You loaded data for 2009.
You loaded data for 2009.
```
Here is a screenshot of what the `show(fig)` statement displayed in my browser:

image

I expected the red reform line to rise above the blue base line at high percentiles of filing unit earnings (e00200), but I did not expect to see any visible reform effect (that is, red higher than blue) at lower percentiles, certainly not between the 10th and 20th earnings percentile.

When I uncommented the `print df` statement and processed the printed data, I get this:
```
$ awk 'BEGIN{tol=1e-3}NR>1{d=$3-$2;if(d>tol||d<-tol)print $0,d}' mtr-plot-dump
%tile   base    reform     diff=reform-base
13  0.355000  0.361953 0.006953        <--------- ?
14  0.368135  0.370225 0.00209         <--------- ?
17  0.272222  0.277141 0.004919        <--------- ?
87  0.323556  0.325307 0.001751
89  0.340054  0.341650 0.001596
90  0.354713  0.356973 0.00226
91  0.370610  0.378307 0.007697
92  0.347936  0.386784 0.038848
93  0.352334  0.391306 0.038972
94  0.354834  0.388781 0.033947
95  0.355733  0.390448 0.034715
96  0.362743  0.405254 0.042511
97  0.363556  0.412745 0.049189
98  0.383581  0.448407 0.064826
99  0.406233  0.482432 0.076199
```
I expect the reform MTR to be higher than the base MTR at the high earnings percentiles, but I can't think of a reason why there should be a reform effect at the three percentiles below the 20th percentile.

Perhaps these results are showing me something about this tax reform that I don't understand and need to learn about.  Can you think of what that might be?  If not, is there some chance that your new `get_mtr_data` function is not working as expected?

@MattHJensen @feenberg @Amy-Xu @andersonfrailey

martinholmer:
@GoFroggyRun,  This comment follows up my prior comment on puzzling MTR-plotting results for a "pop-the-cap" reform.  The plot below is generated using exactly the same script as used in the prior comment except for one change, the `get_mtr_data` parameter `income_measure` is now `'c00100'` (AGI) instead of `'e00200'` (wages and salaries):

image

Why is there no noticeable reform effect except in the 0-10th and 30-40th percentile ranges?  This graph seems to be suggesting that nobody affected by the reform (which affects only those individuals with earnings above $113,700 in 2013) is in the top half of the AGI distribution.  Can that really be true?

@MattHJensen @feenberg @Amy-Xu @andersonfrailey

GoFroggyRun:
@martinholmer:
I just updated docstring that addresses your concerns with regard to `weight_by_income_measure` parameter and embedding. Please take a look. 

Regarding your following concerns, I believe they have gone beyond the scope of this PR for the following reasons:

1. The main purpose of adding the `get_mtr_data` and `mtr_plot` functions, especially the latter one, are not for debugging purpose. It would be easier, in my opinion, to start with suspicious records themselves. 

2. The `get_mtr_data` function simply collects some existed taxcalc utilities and puts them together to prepare for plotting data, where the calculation of mtrs and the classification of bins are based upon `mtr()` and `add_weighted_decile_bins` respectively. If there's some case where the `get_mtr_data` function would generate unexpected data (comparing with calling `mtr()` and `add_weighted_decile_bins` separately to generate data of similar format), I'd be more than happy to investigate `get_mtr_data` function. Otherwise, even if there are indeed some issues, they are more likely to come from `mtr()`, `add_weighted_decile_bins`, or both functions.

3. I'd like to note that (although this probably won't fully address your latest concern about plots against AGI) `c00100` are calculated values and thus would be variant under reforms. Meaning that, when implementing different reforms, filers' AGI will be different and thus the bins for each line will be different as well. 

That said, if there are records with odd MTRs and you'd like me to investigate, I'd be more than happy to take a look at them in a separate issue. Or, if there are cases where `get_mtr_data` function won't coincide with results generated by calling `mtr()` and `add_weighted_decile_bins` separately, I'd be more than happy to deal with them within this PR as well.

MattHJensen:
@GoFroggyRun, could you run @martinholmer's plot and see if you get the same thing. Once you have, could you try to identify whether the problem is with `get_mtr_data`, `_mtr()`, or `add_weighted_decile_bins`? If there is a problem in any of those areas, then it will mean that the new features in this PR are not helpful until the problem is fixed.

martinholmer:
@GoFroggyRun,  I have learned something on my own from the plot of average combined MTR by wage-and-salary (e00200) percentile for a pop-the-cap reform, which is shown again here:

image

The increase in MTR for those just below the 20th wage-and-salary percentile are most likely filing units with little or no wages-and-salaries, but considerable self-employment income.  So that graph is no longer puzzling to me.

However, I'm still puzzled by the plot of average combined MTR by AGI (c00100) percentile for a pop-the-cap reform, which is shown again here:

image

My expectation is that AGI is unchanged by a pop-the-cap reform, whose reform dictionary is as follows:
```
{2013: {'_SS_Earnings_c': [1e99]}}
```
So, your argument that  "AGI (c00100) is a calculated value and thus would be variant under reforms" is true in general, but not in this case, unless I'm missing something.  Am I missing something about the effects of this particular reform?

GoFroggyRun:
All right. I have some very mysterious findings.

First of all, @martinholmer is absolutely right about the plot. This is not how the plot should be looking like. 

The plot should be the following:
bokeh_plot 38
 
I was able to generate the same plot in two, somewhat independent way. The first method is based on @martinholmer's script, but with modifications:

```
from taxcalc import *
calcx = Calculator(policy=Policy(), records=Records())
calcy = Calculator(policy=Policy(), records=Records())
calcy.policy.implement_reform({2013: {'_SS_Earnings_c': [1e99]}})
df = get_mtr_data(calcx, calcy, MARS='ALL', income_measure='e00200', weight_by_income_measure=False)
df = get_mtr_data(calcx, calcy, MARS='ALL', income_measure='c00100', weight_by_income_measure=False)
# print df
fig = mtr_plot(df, plot_width=850, plot_height=500)
output_file('pr-948-fig-complexweight-F-MARS-ALL.html')
show(fig)
```

I'll prepare the other method in another post.

GoFroggyRun:
Here's the other script that yields the same plot:

```
from taxcalc import *
calcx = Calculator(policy=Policy(), records=Records())
calcy = Calculator(policy=Policy(), records=Records())
calcy.policy.implement_reform({2013: {'_SS_Earnings_c': [1e99]}})
a, mtr_iit_x, mtr_combined_x = calcx.mtr()
a, mtr_iit_y, mtr_combined_y = calcy.mtr()
df_x = exp_results(calcx)
df_y = exp_results(calcy)
df_x['combine'] = mtr_combined_x
df_y['combine'] = mtr_combined_y
df_y['c00100'] = df_x['c00100']
df_x.sort_values(by='c00100', inplace=True)
df_y.sort_values(by='c00100', inplace=True)
df_x = add_weighted_decile_bins(df_x, income_measure='c00100',
                             num_bins=100)
df_y = add_weighted_decile_bins(df_y, income_measure='c00100',
                             num_bins=100)
weighted_effect_x={}
weighted_effect_y={}

for i in range(1,101):
    temp = df_x[df_x['bins']==i]
    weighted_effect_x[i] = sum(temp['s006']*temp['combine'])/sum(temp['s006'])

for i in range(1,101):
    temp = df_y[df_y['bins']==i]
    weighted_effect_y[i] = sum(temp['s006']*temp['combine'])/sum(temp['s006'])

weighted_effect_x = pd.DataFrame.from_dict(weighted_effect_x, orient="index")
weighted_effect_y = pd.DataFrame.from_dict(weighted_effect_y, orient="index")

merged = pd.concat([weighted_effect_x, weighted_effect_y], axis=1,
                       ignore_index=True)
merged.columns = ['base', 'reform']
merged.index = (merged.reset_index()).index
output_file('pr-948-fig-complexweight-F-MARS-ALL.html')
fig = mtr_plot(merged)
show(fig)
```

GoFroggyRun:
I'm very confused by the situation for the first script. The commands here 
```
df = get_mtr_data(calcx, calcy, MARS='ALL', income_measure='e00200', weight_by_income_measure=False)
df = get_mtr_data(calcx, calcy, MARS='ALL', income_measure='c00100', weight_by_income_measure=False)
```
seem to be duplicated. It will, however, generate different result if either line got removed. Have no idea what's going on. 

cc @MattHJensen

martinholmer:
@GoFroggyRun,  Let me be very clear about what I did.  This is the script I executed by entering `python pr-948-test.py` at the OS command prompt:
```
$ cat pr-948-test.py
from taxcalc import *
calcx = Calculator(policy=Policy(), records=Records())
calcy = Calculator(policy=Policy(), records=Records())
calcy.policy.implement_reform({2013: {'_SS_Earnings_c': [1e99]}})
df = get_mtr_data(calcx, calcy, income_measure='c00100')
fig = mtr_plot(df, plot_width=850, plot_height=500)
output_file('pr-948-fig.html')
show(fig)
```
Note that I have **never** called the `get_mtr_data` function twice in a row as you seem to incorrectly believe.

As I said before, here is the plot this script shows in my browser:

image

Why can't I produce what you say is the "correct" graph?

GoFroggyRun:
@martinholmer I didn't say that you called the `get_mtr_data` function twice in a row. 
I'm just posting what I have found, which is calling the `get_mtr_data` twice will yield the correct plot.

GoFroggyRun:
Ahh, looks like we forgot to call `calc_all`.
```
from taxcalc import *
calcx = Calculator(policy=Policy(), records=Records())
calcy = Calculator(policy=Policy(), records=Records())
calcy.policy.implement_reform({2013: {'_SS_Earnings_c': [1e99]}})
calcx.calc_all()
calcy.calc_all()
df = get_mtr_data(calcx, calcy, MARS='ALL', income_measure='c00100', weight_by_income_measure=False)
# print df
fig = mtr_plot(df, plot_width=850, plot_height=500)
output_file('pr-948-fig-complexweight-F-MARS-ALL.html')
show(fig)

```

feenberg:
On Fri, 21 Oct 2016, Sean.Wang wrote:

> 
> @martinholmer said:
>
>       No, this does not make any sense to me. All I know is that the
>       graph with the weight_by_income_measure=True does not make any
>       sense.
> 
> So the second plot doesn't any longer, after setting
> weight_by_income_measure=True, mean that
> 45% of (weighted) filing units would have higher taxes. It means that 45% of
> the aggregate income (sum of s006*e00200) is subject to higher taxes. This
> option is not about weighted vs unweighted, but rather about what to weight
> s006 by. This option allows us to investigate different activity
> (wage/agi/etc..) of interests, instead of filling units.

Another way to understand s006*e00200 is that it will give us the effect 
on the average dollar of wage income, rather than the average taxpayer. 
This is relevant if one wants to multiply the amount by an elasticity. If 
the average dollar of income is taxed at a high rate, that makes the 
elasticity important, even if the average taxpayer is taxed at a low rate. 
Consider the propensity of some to dismiss elasticities of taxable income 
with respect to income on the grouds that only high income taxpayers have 
elastic taxable income. That may be true, but it isn't an argument.

BTW, is there a link to the page that this is about?

dan

martinholmer:
@GoFroggyRun,  Thank your for your substantial contribution in pull request #948 that adds a marginal tax rate graphing capability.  I'm merging this pull request into the master branch.

PR #950, opened 2016-10-01
Improve pytest fixture and parametrization usage

zrisher:
I followed up on @talumbau's recommendation to use `@pytest.mark.parametrize` in the `TaxForm` tests. While reading about pytest, I discovered numerous areas in the current tests that could benefit from refactoring.

This PR simplifies the logic for all tests by increasing the use of pytest **fixtures**, **parametrization**, and **config files**. 

It does not change the test-specific logic, nor its coverage. 

It should make the tests more legible, though, and serve as a reference point for future contributors when creating pytest cases. I also noticed a small decrease in runtime of 2.8% on my machine over several runs, likely due to reading the 1991 puf.csv file fewer times.

**Notes:**
* This PR is broken into logically separate commits. Reading through them individually may be easier than  looking at all the changes as a whole.
* There are a lot of areas where I could have gone further with parametrization but it didn't seem like it would actually improve legibility, or I was unfamiliar enough with the code that it would take significant time.
* I attempted to maintain existing pylint compliance in the files I edited. However, some files were clearly written without pylint compliance in mind, so I didn't spend any time checking compliance there. Some files that were partially pylint compliant still had a few outstanding issues, and I didn't mess with those either.
* Parameterizing the child tax form class tests showed me some potential issues that I noted with `todo`s.
* I found pytest's loading and test creation logic fairly complex. There's a lot of features left we haven't used (e.g. hooks on the `metafunc` object) that might be helpful in the future. There's also a lot of peculiarities that make it difficult for pylint to understand the files (e.g. not using a package, fixture syntax). There's quite a few options for including helper functions (test classes, fixtures that return a function, supplying the fixture name as a parameter in the parameterization mark, etc), so if anyone would like to discuss that stuff for future tests, I'd be glad to help!

@martinholmer @talumbau @MattHJensen @GoFroggyRun @Amy-Xu


martinholmer:
@talumbau,  Thanks for your suggestion to @zrisher about using pytest parameterization to simplify the unit tests.  @zrisher,  Thanks for pursuing the suggestion and finding other ways to simplify the pytest code.

PR #950 is an excellent contribution to the Tax-Calculator project.  Not only is there less test code, but it is easier to read and executes noticeably faster (about a 13% reduction in unit-test runtime on my computer).

PR #968, opened 2016-10-06
Cgdiv exclusion

codykallen:
Exclude part of capital gains and dividend income from tax.


MattHJensen:
@codykallen, if you push more commits to this branch on your origin, they will automatically be incorporated into this PR. There is no need to close the PR and reopen unless you are making drastic changes and want to start over.

MattHJensen:
@codykallen, I wonder whether it would be better to apply the exclusion to all of the variables in one step before those variables get used in the form's logic (var1 = var1*exclusion), rather than slipping the exclusion into every instance of where the variable is used (var2 = var1*exclusion + var3).

codykallen:
That is easier, but it creates potential problems for calculating weighted average marginal tax rates on gains and dividends. If you want to do it that way, it would be better to create new local variables and use those in the form instead.
var1ex = var1 * (1 – cgdiv_ec)

martinholmer:
@codykallen,  Please revise your pull request #968 so that:
(a) the new policy parameter is named according to [these conventions](https://github.com/open-source-economics/Tax-Calculator/blob/master/docs/source/parameter_naming.rst), and
(b) the new policy parameter is placed in the `current_law_policy.json` file where it is first used in the `functions.py` code (see pull request #845 for details on the ordering of policy parameters).

The new parameter should be probably be named `CG_ec` and placed with the other `CG_*` parameters in the `current_law_policy.json` file.

@MattHJensen @feenberg @Amy-Xu @GoFroggyRun @andersonfrailey @zrisher

martinholmer:
@martinholmer said:
> The new parameter should be probably be named `CG_ec` and placed with the other `CG_*` parameters in the `current_law_policy.json` file.

Actually, this is incorrect now that I see this new parameter is a fraction rather than a dollar amount.  So, following the project's naming conventions, the new policy parameter should be named `CG_ec_rt`.

codykallen:
The interest exclusion is '_ALD_Interest_ec'. @martinholmer, are you sure that should be _ec_rt?

martinholmer:
@codykallen said:
> The interest exclusion is '_ALD_Interest_ec'. @martinholmer, are you sure that should be _ec_rt?

Good point.  My reading of the parameter naming conventions (which I did not write, but have found to be very useful) is that `_ALD_Interest_ec` should be `_ALD_Interest_ec_rt`.  I'll make that change in another pull request.

@MattHJensen @feenberg @Amy-Xu @GoFroggyRun @zrisher @andersonfrailey

martinholmer:
@codykallen,  I think your next step --- now that the new parameter name is placed correctly in the `current_law_policy.json` file --- is to use that new parameter name (`CG_ec_rt`) in the `functions.py` file.  All the checkin errors are caused by the fact that in the `functions.py` file you have `cgdiv_ec_rt` instead of `CG_ec_rt`.

Also, I don't understand why your pull request changes the `var_labels.txt` file.  You can return it to its original condition by entering the following at the command line in the tax-calculator directory:
`git checkout -- taxcalc/var_labels.txt`

martinholmer:
@codykallen,  This version of pull request #968 is much better in that it doesn't generated any unit-test errors.  But I have one question about how you've code this policy reform.  In the `functions.py` file's `GainsTax` function, you made this change:
```
 -        dwks33 = min(dwks9, e24518)
 +        dwks33 = min(dwks9, e24518) * (1 - CG_ec_rt)
```
Maybe I'm missing something, but this seems incorrect because the variable `dwks9` has already been multiplied by `(1 - CG_ec_rt)`.  Shouldn't this line be
```
 +        dwks33 = min(dwks9, e24518 * (1 - CG_ec_rt))
```
which is the way it was in an earlier version (c3c89d8) of your pull request?

Also, you still need to revert the `var_labels.txt` file back to its state on the master branch.

codykallen:
Martin, hold off on this for a bit. I've discovered a logic issue with my approach. The method in this pull request excludes half of capital gains and dividends from taxation, but it also excludes half of it when calculating dwks19, meaning that more total income is subject to regular tax rates instead of capital gains and dividend rates.

MattHJensen:
> but it also excludes half of it when calculating dwks19, meaning that more total income is subject to regular tax rates instead of capital gains and dividend rates.

@codykallen, I would have thought that that outcome is expected and *should* happen.

codykallen:
The idea of the exclusion is that only part of your gains and dividends are taxed. The logic I used results in a smaller value of gain/dividends but also reduces dwks19. Since dwks19 is passed to the Taxes function, the excluded investment income gets taxed at ordinary income rates, resulting in a tax increase on investment income.

PR #972, opened 2016-10-07
Investment exclusion

codykallen:
Excludes half of investment income. Note that this does not capture the complexity of Sch D.


martinholmer:
@codykallen,  You need to fix some minor formatting issues that generate the PEP8 errors.  You can see them by clicking on the Details and picking Python 2.7 and then scrolling all the way down to the bottom where the errors are identified by line/column number in the `functions.py` file.

Also,  I'm confused about the `_ALD_Investment_ec` name.  Is there a need for the ALD?  Why not just `_Investment_ec`?

martinholmer:
@codykallen,  Thanks for pull request #972.  I'm merging it into the master branch now.

PR #973, opened 2016-10-07
Add parameter to allow LTCG+QDIV income to be taxed as other income

martinholmer:
This pull request attempts to provide an easy-to-use policy parameter that causes long-term capital gains and qualified dividends to be taxed at the same rates as wage and salary income.  The changes in this pull request have no effect on current-law tax results, but these changes need review to determine if they do tax calculations correctly when the new `_CG_as_II` parameter is set to one.

Such a reform raises 2016 income tax revenue by almost $41.9 billion.  I have no idea if that is a plausible revenue increase when eliminating the special tax treatment of LTCG+QDIV
income.
```
tax-calculator$ cp ../tax-calculator-data/puf.csv .
tax-calculator$ python inctax.py puf.csv 2016 --weights
tax-calculator$ cat cgsame.json
{ "_CG_as_II": {"2016": [1] } }
tax-calculator$ python inctax.py puf.csv 2016 --weights --reform cgsame.json 
tax-calculator$ ll puf*
-rw-r--r--  1 mrh  staff  39385372 Oct  7 12:34 puf-16.out-inctax
-rw-r--r--  1 mrh  staff  39391761 Oct  7 12:37 puf-16.out-inctax-cgsame
-rw-r--r--@ 1 mrh  staff  52839666 Oct  7 12:32 puf.csv
tax-calculator$ awk '{rev+=$4*$29}END{print rev}' puf-16.out-inctax
9.27552e+11
tax-calculator$ awk '{rev+=$4*$29}END{print rev}' puf-16.out-inctax-cgsame 
9.69409e+11
```

@MattHJensen @feenberg @codykallen @Amy-Xu @GoFroggyRun @andersonfrailey

Issue #974, opened 2016-10-07
Activating the new _CG_nodiff parameter generates unanticipated results


MattHJensen:
These reforms generate different results. We should understand why and either adjust how _CG_as_II works or document the difference. 

p_yy = Policy()
reform1 = {2016: {
'_II_rt2': [0.1],
'_PT_rt2': [0.1],
'_II_rt4': [0.25],
'_PT_rt4': [0.25],
'_II_rt5': [0.25],
'_PT_rt5': [0.25],
'_II_rt6': [0.25],
'_PT_rt6': [0.25],
'_CG_as_II': [1],
}
}

$ -23.4 B

reform2 = {2016: {
'_II_rt2': [0.1],
'_PT_rt2': [0.1],
'_II_rt4': [0.25],
'_PT_rt4': [0.25],
'_II_rt5': [0.25],
'_PT_rt5': [0.25],
'_II_rt6': [0.25],
'_PT_rt6': [0.25],
'_CG_rt1': [0.1],
'_CG_rt2': [0.25],
'_CG_rt3': [0.396],
'_AMT_CG_rt1': [0.1],
'_AMT_CG_rt2': [0.25],
'_AMT_CG_rt3': [0.396]

}
}

$ -28.1 B 

When AMT is removed entirely from the baseline and reforms, the results are very close:

p_yy = Policy()
reform1 = {2016: {
'_II_rt2': [0.1],
'_PT_rt2': [0.1],
'_II_rt4': [0.25],
'_PT_rt4': [0.25],
'_II_rt5': [0.25],
'_PT_rt5': [0.25],
'_II_rt6': [0.25],
'_PT_rt6': [0.25],
'_CG_as_II': [1],
'_AMT_trt1':[0.0],
'_AMT_trt2':[0.0]
}
}

-67.4

reform2 = {2016: {
'_II_rt2': [0.1],
'_PT_rt2': [0.1],
'_II_rt4': [0.25],
'_PT_rt4': [0.25],
'_II_rt5': [0.25],
'_PT_rt5': [0.25],
'_II_rt6': [0.25],
'_PT_rt6': [0.25],
'_CG_rt1': [0.1],
'_CG_rt2': [0.25],
'_CG_rt3': [0.396],
'_AMT_trt1':[0.0],
'_AMT_trt2':[0.0]
}
}

-68.8 

cc @codykallen @martinholmer

martinholmer:
[...]

PR #975, opened 2016-10-08
Extend no-different tax treatment of LTCG+QDIV income to AMT

martinholmer:
In merged pull request #973, Tax-Calculator logic was extended to allow LTCG+QDIV income to be taxed using the same regular tax brackets and rates as ordinary income (like wages and salaries).  In this pull request, the effect of the new #973 policy parameter (the name of which has been changed from `CG_as_II` to `_CG_nodiff`) has been extended to AMT logic.  Under current law, LTCG+QDIV income is taxed with different brackets and rates in both the regular tax and AMT calculations, and therefore, the current-law-policy value of the `_CG_nodiff` is zero (that is, false).  Changing the value of `_CG_nodiff` to one (that is, true) will cause both regular income tax and AMT logic to include LTCG+QDIV income along with other income in AGI and tax that combined income using the ordinary tax brackets and rates for regular taxes and AMT.  Note that these changes do not affect the role of LTCG+QDIV income in the calculation of NIIT or EITC.

We have conducted a simple test to show that the logic in this pull request seems to work as designed. 

Consider two individuals both sixty years old with no dependents, with $500,000 income and with $200,000 of AMT-preference itemized deductions.  The first of the two has only qualified dividend income, while the second has only wage and salary income.  Using `simtax.py` input format, the two individuals are as follows:
```
$ cat cases.in
1 2015 0 1 0 6000 0 0 500000 0 0 0 0 0 0 200000 0 0 0 0 0 0
2 2015 0 1 0 6000 500000 0 0 0 0 0 0 0 0 200000 0 0 0 0 0 0
```
We specify a simple reform that eliminates the differential tax treatment of LTCG+QDIV income in a JSON reform file with the following contents:
```
$ cat same.json
{"_CG_nodiff":{"2015":[1]}}
```
Under current-law policy (which has a favorable differential tax treatment of LTCG+QDIV income), the individuals have the following tax liabilities:
```
$ python simtax.py cases.in 
$ awk '{print FILENAME,$1,$4,$26,$27,$28}' cases.in.out-simtax
cases.in.out-simtax 1.  85122.50 500000.00  1260.00 72462.50
cases.in.out-simtax 2. 136292.00 500000.00 51292.42 84999.58
```
The first individual pays much less income tax than the second even though they both have AGI equal to $500,000.

Next we simulate income tax liability for these two individuals under the `same.json` reform that eliminates the differential income tax treatment of LTCG+QDIV income.
```
$ python simtax.py cases.in --reform same.json
$ awk '{print FILENAME,$1,$4,$26,$27,$28}' cases.in.out-simtax-same
cases.in.out-simtax-same 1. 147692.00 0.00 500000.00 51292.42 84999.58
cases.in.out-simtax-same 2. 136292.00 0.00 500000.00 51292.42 84999.58
```
Notice that now the first individual pays more tax than before (as expected given this reform) and that the AMT tax liability (that sum of output variables 27 and 28) are exactly the same: $136,292.  But then the question is why does the first individual owe a total income tax (output variable 4) that is $11,400 more than the second individual.  The answer is that the first individual owes an NIIT (net investment income tax) liability of that amount: `11400 = 0.038 * (500000 - 200000)`.

### DIFF
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -132,13 +132,13 @@ class Calculator(object):
     def records(self):
         return self._records
 
-    def TaxInc_to_AMTI(self):
+    def TaxInc_to_AMT(self):
         TaxInc(self.policy, self.records)
         SchXYZTax(self.policy, self.records)
         GainsTax(self.policy, self.records)
         AGIsurtax(self.policy, self.records)
         NetInvIncTax(self.policy, self.records)
-        AMTInc(self.policy, self.records)
+        AMT(self.policy, self.records)
 
     def calc_one_year(self, zero_out_calc_vars=False):
         # calls all the functions except those in calc_all() function
@@ -160,14 +160,14 @@ class Calculator(object):
         item_no_limit = copy.deepcopy(self.records.c21060)
         self.records.c04470 = np.zeros(self.records.dim)
         self.records.c21060 = np.zeros(self.records.dim)
-        self.TaxInc_to_AMTI()
+        self.TaxInc_to_AMT()
         std_taxes = copy.deepcopy(self.records.c05800)
         # Set standard deduction to zero, calculate taxes w/o
         # standard deduction, and store AMT + Regular Tax
         self.records._standard = np.zeros(self.records.dim)
         self.records.c21060 = item_no_limit
         self.records.c04470 = item
-        self.TaxInc_to_AMTI()
+        self.TaxInc_to_AMT()
         item_taxes = copy.deepcopy(self.records.c05800)
         # Replace standard deduction with zero where the taxpayer
         # would be better off itemizing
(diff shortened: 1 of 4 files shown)


feenberg:
On Sat, 8 Oct 2016, Martin Holmer wrote:

> 
> In merged pull request #973, Tax-Calculator logic was extended to allow the regular tax treatment of
> LTCG+QDIV income to be exactly the same as ordinary income (like wages and salaries). In this pull

Wages and salaries have special treatment in numerous places in the tax 
law, especially EITC and the Medicare tax on earnings. I think interest 
income is the cleanest place to move capital gains if the special 
treatment of gains is to be eliminated. The only place interest income is 
"special" is in the limitation on non-mortgage interest deductions 
(Section 163d of the code) which I don't think we model.

dan

martinholmer:
Pull request #975 is being closed (without being merged into the master branch) because of all the recent changes to the `AMT` function logic.

martinholmer:
Pull request #975 was closed because no other code-revisions are needed to make the `_CG_nodiff` logic work as is shown in the discussion of issue #974.

PR #976, opened 2016-10-10
[WIP] Add function to calculate cost of Trump dependent care deduction

andersonfrailey:
This PR calculates the cost of the above-the-line deduction for child and elderly care [proposed by Trump:](https://www.donaldjtrump.com/policies/tax-plan/?/positions/tax-reform)

> Americans will be able to take an above-the-line deduction for children under age 13 that will be capped at state average for age of child, and for eldercare for a dependent. The exclusion will not be available to taxpayers with total income over $500,000 Married-Joint /$250,000 Single, and because of the cap on the size of the benefit, working and middle class families will see the largest percentage reduction in their taxable income.

It is dependent on variables that will be added to the forthcoming PUF file and I'll drop the WIP label when that new PUF is adopted. At this point I'm mostly looking for feedback on the logic and any other suggestions that could be offered while I work on getting the new PUF out.

I made a new function to calculate rather than adding it into the `Adj` function to keep things readable. I also put the average cost of childcare in the current law policy file, but if there is a better place for it I'm open to suggestions. 

The new PUF required for this function will be ready this week and will be the same as what is currently used just with the ages of the depends in the tax units included.

### DIFF
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -146,6 +146,7 @@ class Calculator(object):
             self.records.zero_out_changing_calculated_vars()
         # pdb.set_trace()
         EI_PayrollTax(self.policy, self.records)
+        DependentCare(self.policy, self.records)
         Adj(self.policy, self.records)
         CapGains(self.policy, self.records)
         SSBenefits(self.policy, self.records)
--- a/taxcalc/current_law_policy.json
+++ b/taxcalc/current_law_policy.json
@@ -145,6 +145,61 @@
         "value": [0.0]
     },
 
+        "_ALD_Dependents_HC":{
+        "long_name": "Deduction for childcare costs",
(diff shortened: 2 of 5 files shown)


MattHJensen:
@andersonfrailey, could you also include some documentation for how you calculate the weighted average childcare costs?

andersonfrailey:
@MattHJensen 
The weighted average of childcare costs was calculated using in formation from [this](https://usa.childcareaware.org/wp-content/uploads/2016/07/Full-Report_final.pdf) report by Child Care Aware of America.

The report includes average childcare costs by state in FCC homes and childcare centers as well as by age. I found the average cost of childcare for each age by state using the number of each type of facility (FCC homes and childcare centers) then the number of children in the state at each age level. Finally I found the national average weighted by the population of care aged children in each state.

Issue #977, opened 2016-10-11
GainsTax taxes all of taxable non ltcg or qdiv income at one rate


codykallen:
If the II_rt values do not equal the pt_rt values, the tax filers with positive pass-through income will have their taxable income taxed at two different rate schedules. If that filer also has long-term capital gains or qualified dividends, then their ltcg+qdiv income is subtracted from their taxable income, and is taxed at the CG_rt rate schedule. However, the GainsTax function passes their taxable income less ltcg and qdiv to the Taxes function using the II_rt and II_thd parameters, ignoring the separate taxation of pass-through income.

martinholmer:
@codykallen said:
> If the II_rt values do not equal the pt_rt values, the tax filers with positive pass-through income will have their taxable income taxed at two different rate schedules. If that filer also has long-term capital gains or qualified dividends, then their ltcg+qdiv income is subtracted from their taxable income, and is taxed at the CG_rt rate schedule. However, the GainsTax function passes their taxable income less ltcg and qdiv to the Taxes function using the II_rt and II_thd parameters, ignoring the separate taxation of pass-through income.

I don't understand what the point of your above statement is.  You are raising an issue.  Can you explain what your issue is?

codykallen:
Suppose we do the following reform:

reform1 = {
    2017: {
        '_II_rt7': [0.4],
        '_PT_rt7': [0.35]
    }
}

Suppose an individual has $300k taxable income, with $150k pass-through income. Suppose this filer has no deductions. The top tax rate on this individual is 35%. 

When calculating the MTR on long term gains, the calculator.mtr function adds 0.01 to p23250 and recalculates. Since this increases his taxable income by 0.01. The GainsTax function subtracts the 0.01 from his taxable income and passes it to the Taxes function, using only the II rates. Hence this person's top rate is now 40%, his full tax bill is several thousand dollars larger, and his MTR on ltcg is very large.

codykallen:
Correction: this only applies if the PT rates are greater than the II rates.

martinholmer:
@codykallen,  Thanks for a more complete bug report.  I'm starting to work on this issue by adding your test case to the suite of unit tests and then seeing if your proposed code change eliminates the new test failure and that all the old unit tests continue to pass.

martinholmer:
[...]

PR #979, opened 2016-10-11
Gains tax issue

codykallen:
This adds the pass-through logic used in SchXYZTax to the GainsTax function. This should correct Issue #977


martinholmer:
The bug-fix strategy in #979 has been incorporated in a somewhat different manner into #981.

PR #981, opened 2016-10-12
Add test that shows bug in GainsTax function and fix bug

martinholmer:
This pull request adds a test to illustrate a bug in the GainsTax function and then fixes that bug.  This bug was reported by @codykallen in issue #977.  The fix for this bug in this pull request is logically equivalent to the code revision suggested by @codykallen in pull request #979, but eliminates the duplication of code inherent in #979.  Thank you, @codykallen, for your helpful bug report in #977 and your correct diagnosis of the bug's cause in #979.

As an aside for other Tax-Calculator contributors, the bug report in #977 is an example to be emulated: it describes the problem and provides a simple test to demonstrate the problem.  #979 goes the extra mile by actually suggesting how to fix the bug reported in #977.

What I have done in this pull request is two things:
(1) created a new unit test based on the bug report in #977, and
(2) fixed the bug (that is, revised the code so that the new test does not fail).

@MattHJensen @feenberg @Amy-Xu @GoFroggyRun @andersonfrailey @zrisher

### DIFF
--- a/taxcalc/functions.py
+++ b/taxcalc/functions.py
@@ -440,25 +440,25 @@ def TaxInc(c00100, _standard, c21060, c21040, c04600, c04800):
     return c04800
 
 
-@iterate_jit(nopython=True)
-def SchXYZTax(c04800, MARS, e00900, e26270,
+@jit(nopython=True)
+def SchXYZamt(taxable_income, MARS, e00900, e26270,
               PT_rt1, PT_rt2, PT_rt3, PT_rt4, PT_rt5, PT_rt6, PT_rt7,
               PT_rt8, PT_brk1, PT_brk2, PT_brk3, PT_brk4, PT_brk5, PT_brk6,
               PT_brk7, II_rt1, II_rt2, II_rt3, II_rt4, II_rt5, II_rt6, II_rt7,
               II_rt8, II_brk1, II_brk2, II_brk3, II_brk4, II_brk5, II_brk6,
-              II_brk7, c05200):
+              II_brk7):
     """
-    SchXYZTax uses the tax rates in Schedule X, Y, or Z, to compute tax.
+    SchXYZamt uses the tax rates in Schedule X, Y, or Z, to compute tax amount.
     """
     # separate non-negative taxable income into two non-negative components,
     # doing this in a way so that the components add up to taxable income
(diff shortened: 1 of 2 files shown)


martinholmer:
Before the bug-fix code change in the `functions.py` file, the result of the new `test_Calculator_mtr_when_PT_rates_differ` test were as follows:
```
>       assert np.allclose(mtr1, mtr2, rtol=0.0, atol=1e-06)
E       assert <function allclose at 0x102f93410>(array([ 0.188]), array([ 0.438]), rtol=0.0, atol=1e-06)
E        +  where <function allclose at 0x102f93410> = np.allclose
```
So, before the bug fix, the marginal tax rate on LTCG (`p23250`) was different before and after the reform that made pass-through tax rates different from tax rates on non-pass-through income: instead of being the same the mtr on LTCG was 0.188 (the 15% CG rate plus the 3.8% NIIT rate) when II and PT tax rates were identical and 0.438 (the 40% II rate plus the 3.8% NIIT rate) after the reform that made all II rates equal to 40% and all PT rates equal to 30%.

After the bug fix, the mtr on LTCG is 0.188 both before and after the reform that makes II and PT rates different.

@MattHJensen @feenberg @Amy-Xu @GoFroggyRun @andersonfrailey @zrisher

MattHJensen:
@martinholmer, this looks good to me. @codykallen, thanks a lot for the bug report and suggested fix! 

@martinholmer, do you think it is ok to merge this now? I think it deserves a TC version bump and redeploy to TaxBrain asap given that I know there are currently users on TB who would be affected. 

cc @talumbau

martinholmer:
@MattHJensen said:
> @martinholmer, this [#981] looks good to me. @codykallen, thanks a lot for the bug report and suggested fix!
>
> @martinholmer, do you think it is ok to merge this now? I think it deserves a TC version bump and redeploy to TaxBrain asap given that I know there are currently users on TB who would be affected.

Yes, I'm comfortable with the changes in pull request #981, but was hoping to get a thumbs up from @codykallen before doing the merge.  But I haven't heard anything from his this morning.

martinholmer:
Merged pull request #981 resolves issue #977 and incorporates the bug-fix strategy (but not tactics) of pull request #979.

PR #987, opened 2016-10-14
Mtr e26270

codykallen:
Adds calculation of MTR on e26270 (partnership and S-corp income)

### DIFF
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -231,7 +231,8 @@ class Calculator(object):
                            'e01400', 'e01700',
                            'e02000', 'e02400',
                            'p22250', 'p23250',
-                           'e18500', 'e19200']
+                           'e18500', 'e19200',
+                           'e26270']
 
     def mtr(self, variable_str='e00200p',
             negative_finite_diff=False,
@@ -294,7 +295,8 @@ class Calculator(object):
         'p22250',  short-term capital gains;
         'p23250',  long-term capital gains;
         'e18500',  Schedule A real-estate-tax deduction;
-        'e19200',  Schedule A total-interest deduction.
+        'e19200',  Schedule A total-interest deduction;
+        'e26270',  Income from S-corps/partnerships (also in e02000).
         """
         # check validity of variable_str parameter
         if variable_str not in Calculator.MTR_VALID_VARIABLES:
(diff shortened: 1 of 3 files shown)


codykallen:
I tested this with a reform of pass-through rates:

reform_pt = { 2017: {
'_PT_rt5': [0.3],
'_PT_rt6': [0.3,
'_PT_rt7': [0.3]
}
}

I calculated average MTRs on e02000 and e26270 under the baseline and reform, weighting by the absolute value of the income type. 
Baseline MTR, e02000: 32.77%
Baseline MTR, e26270: 30.60%
Reform MTR, e02000: 29.08%
Reform MTR, e26270: 25.34%

taxcalc/calculate.py
```diff
@@ -322,6 +326,8 @@ def mtr(self, variable_str='e00200p',
             self.records.e00900 = seincome_var + finite_diff
         elif variable_str == 'e00650':
             self.records.e00600 = divincome_var + finite_diff
+        elif variable_str == 'e26270':
+            self.records.e02000 = schEincome_var + finite_diff
```
jdebacker: @codykallen Why add finite_diff to e02000 instead of e26270 here?  Seems like adding to e26270 is what would help this MTR enter `functions.SchXYZ()` in a way that accounts for potential differences between rates on ordinary income and pass-through income.

jdebacker:
@codykallen - thanks for opening this, it'll be helpful give the updated to pass-through income logic in taxcalc.

taxcalc/calculate.py
codykallen: finite_diff is already being added to e26270 in the line:
setattr(self.records, variable_str, variable + finite_diff)

However, e26270 is included in e02000, so you have to increase e02000 as well. If not, then you would be calculating the MTR from adding finite_diff to e26270 but substracting it from rental income.

jdebacker: Ah, I see. Thanks @codykallen, LGTM then.

martinholmer:
@codkallen,  Your pull request #987 looks fine except for one thing: you need to add a test to prevent a decline in code coverage.  Adding something like this at the end of the `test_Calculator_mtr` function (in the `taxcalc/tests/test_calculate.py` file) would be the minimal test:
```
    (_, _, mtr_combined) = calc.mtr(variable_str='e26270')
    assert type(mtr_combined) == np.ndarray
```

@MattHJensen @Amy-Xu @GoFroggyRun @andersonfrailey

martinholmer:
@codykallen,  Thanks for adding the test of the new mtr logic in pull request #987.

However, I don't understand why you have added the extra variables to the `Records.VALID_READ_VARS set`.  You're adding four variables: `'FDED', 'e11550', 'e11070', 'e11580'`.  Why?  None of these four are in the current `puf.csv` file.

martinholmer:
@codykallen,  Thanks so much for pull request #987, which has been merged into master branch.

Issue #37, opened 2014-09-30
C62100


iliakur:
This variable is in top 20 in terms of cumulative error. In the [Python code](https://github.com/OpenSourcePolicyCenter/Tax-Calculator/blob/master/translation.py#L1196) it is set with a long expression with a bunch of e-values. I looked at one taxpayer for whom **C62100** was off and found that all the e-values matched up (as a matter of fact, they were all zero), as were apparently the c-values (see table below). [Here](https://github.com/OpenSourcePolicyCenter/Tax-Calculator/blob/master/originaltaxcalc.sas#L583) is the relevant SAS code, for reference. Tagging @SameerSarkar and @feenberg.

| Variable Name | SAS value | Python Value |
| --- | --- | --- |
| _c62100_ | _4364782_ | _4352722_ |
| c60260 | 0 | 0 |
| c60000 | 421058 | 421058 |

PR #999, opened 2016-10-19
Dependent Care Deduction Function

andersonfrailey:
This PR adds a function to  calculate the cost of Trump's [proposed childcare deduction policy](https://www.donaldjtrump.com/policies/tax-plan/?/positions/tax-reform):

> Americans will be able to take an above-the-line deduction for children under age 13 that will be capped at state average for age of child, and for eldercare for a dependent. The exclusion will not be available to taxpayers with total income over $500,000 Married-Joint /$250,000 Single, and because of the cap on the size of the benefit, working and middle class families will see the largest percentage reduction in their taxable income...The eldercare exclusion would be capped at $5,000 per year. The cap would increase each year at the rate of inflation.

It uses variables that are not in our current PUF and I can send the PUF I used to anyone who would like to test this function. I'm not issuing a new PUF yet because I'm working on another one that will be created using new IRS target data and will remove variable `p87482` as requested by @martinholmer in [TaxData PR #37](https://github.com/open-source-economics/taxdata/pull/37) and want to cause the least amount of disruption. Adding the three additional variables to test this PR caused no change in any of the other outputs.

About the PR:

The proposal calls for an above the line deduction of the state average cost of childcare for up to four dependents and a $5,000 deduction for elderly dependents. Because our dataset does not contain information on the location of each tax unit, I found the national weighted average using [this](https://usa.childcareaware.org/wp-content/uploads/2016/07/Full-Report_final.pdf) report from Child Care Aware of America rather than using each state's average cost.

The PUF I have used to test it only had the ages of three dependents so I randomly assigned a fourth to some households. Looking through the 2014 CPS data I found that of all persons under 13, about 2.05% live in a household with three or more other dependents. During the DependentCare function, if a tax unit lists three children 13 or under, I generate a random float between 0.0 and 1.0 and if it is bellow 0.020545 I add a fourth dependent. I used a similar process to randomly assign elderly dependents.

Because of this random assignment, the projected decrease in tax liabilities is not always the exact same, but it is consistently in the $431-433 billion range from 2017-2026. This compares to an estimated $500 billion static and $461 billion dynamic decrease from the [Tax Foundation](http://taxfoundation.org/article/details-and-analysis-donald-trump-tax-reform-plan-september-2016) from 2016-2025, and a decrease of $131.5 billion from the [Tax Policy Center](http://www.taxpolicycenter.org/publications/analysis-donald-trumps-revised-tax-plan/full) from 2016-2026. The Tax Policy Center's analysis seems to included other provisions such as a large increase in the standard deduction while TC's does not, which likely explains at least part of the large difference.

To calculate these results I used this reform dictionary: 
```
careReform = {
    2017: {"_ALD_Dependents_HC": [0.0]}
}
```


martinholmer:
@andersonfrailey,  I have three sets of concerns about your pull request #999.

(1) Some of the new policy parameter names introduced by pull request #999 don't make sense to me.  The new `_ALD_Dependents_Child_ec` and `_ALD_Dependents_Elder_ec` parameters seem to be misnamed.  If you look at your proposed `DependentCare` function, each of these two parameters is a limit or ceiling, not an exclusion.  So, it would seem, if you were following the [Tax-Calculator coding style](https://github.com/open-source-economics/Tax-Calculator/blob/master/CODING.md), that they should end in `_c` (not end in `_ec`).  Am I confused about this reform proposal?  Or do you need to make changes in the parameter names?

(2) Your code is riddled with PEP8 errors.  You should be following the guidelines in the [Tax-Calculator coding style](https://github.com/open-source-economics/Tax-Calculator/blob/master/CODING.md) document, which calls for correcting these kinds of coding problems **before** submitting a pull request.

(3) I have serious reservations about your use of random numbers to impute a fourth child dependent and an elderly dependent.  This is a **data issue** (not a tax-calculating issue) and, therefore, needs to be handled in the preparation of a new `puf.csv` file in the taxdata repository.  This sort of data imputation has no place in the Tax-Calculator.  This is an important issue that needs to be fixed.  (And, as a side note, your use of random numbers is incorrect because you have not set the seed of the random number generator, which means you would get different tax results each time you ran Tax-Calculator.)

@MattHJensen @Amy-Xu @GoFroggyRun

feenberg:
On Wed, 19 Oct 2016, andersonfrailey wrote:

> 
> 
> The PUF I have used to test it only had the ages of three dependents so I randomly
> assigned a fourth to some households. Looking through the 2014 CPS data I found that
> of all persons under 13, about 2.05% live in a household with three or more other
> dependents. During the DependentCare function, if a tax unit lists three children 13
> or under, I generate a random float between 0.0 and 1.0 and if it is bellow 0.020545 I
> add a fourth dependent. I used a similar process to randomly assign elderly
> dependents.
>

1) We already have random variables in the calculator related to the dropq
    system. This does not cause random variation in output, because the
    seed for the random variable is chosen based on the full list of tax
    parameters. That way any change to the parameters changes the seed, but
    rerunning exactly the same parameters always generates exactly the same
    result. This is pretty important - it allows any run to be summarized
    by the parameters and reproduced exactly.

    I can add that one of the reasons given by Census for not pursuing
    dropq was the fear that random variation in output would embarrass the
    institution.

2) Is this going in to Taxbrain at some point? It seems like every day I
    see patches for campaign related tax reform proposals, but Taxbrain
    never changes. If they don't go in to Taxbrain, what is the point?

3) Can you say something about the source of the average child care
    expense by state?

dan feenberg

MattHJensen:
@feenberg asked:

> 2) Is this going in to Taxbrain at some point? It seems like every day I
    see patches for campaign related tax reform proposals, but Taxbrain
    never changes. If they don't go in to Taxbrain, what is the point?

Dan, we've been adding these features at the request of users of the Tax-Calculator package. They are also available on TaxBrain via the new file upload feature that allows the user to upload a json file of parameters. Here's a link to that discussion https://github.com/OpenSourcePolicyCenter/webapp-public/issues/312. This feature is live at www.ospc.org/taxbrain/file, but not well documented yet.

We haven't been adding new non-current-law features of the tax code to the standard TaxBrain interface because we don't currently have a good system for quickly adding new parameters to the user interface and organizing them once they are there. 

The file upload is a partial step towards [your suggestions](https://github.com/open-source-economics/Tax-Calculator/issues/429#issue-114166009) for user extensibility of TaxBrain. Although it doesn't allow the user to pass a custom reform snippet directly to the calculator via some TaxBrain input field, it does allow users to suggest a reform snippet to Tax-Calculator contributors, and then for  the capability to available very quickly for TaxBrain users without any modifications to the TaxBrain interface.

feenberg:
1) Requests for enhancements should go to the mailing list, so that
they are subject to comment.

2) Looking through the threads, I don't see any indication that the
JSON upload is a working feature - where would I click to get the file
upload box?

3) I don't see any list of parameter names. Shouldn't there be some
minimum of documentation enforced before patches are accepted?
Undocumented features are not valuable to other users. If anyone
distributes results based on an undocumented feature, how is that
different from a proprietary model?

dan

andersonfrailey:
@martinholmer 

> Am I confused about this reform proposal? Or do you need to make changes in the parameter names?

I wasn't entirely sure which way to go on naming the parameters. The amounts in `ALD_Dependents_Chid_ec` and `ALD_Dependents_Elder_ec` are the caps on the deduction, but from my reading of the policy proposal the Trump team is considering them exclusions. I'm happy to change it to ceiling if you believe that would be more appropriate.  

I have removed the random assignment from TaxCalc and added the necessary variables to the PUF. And thank you for pointing out the issue with setting the seed. I wasn't previously aware of the functionality of a seed.

andersonfrailey:
@feenberg 

> 3) Can you say something about the source of the average child care
    expense by state?

Child Care Aware is an organization that serves as a hub of information on child care for parents and providers. The report referenced gathered data from their state agencies and found the average cost of childcare in each state. I took those costs and weighted them by the size of the child population in each state to get a national average.

martinholmer:
@andersonfrailey said:
> I wasn't entirely sure which way to go on naming the parameters. The amounts in ALD_Dependents_Chid_ec and ALD_Dependents_Elder_ec are the caps on the deduction, but from my reading of the policy proposal the Trump team is considering them exclusions. I'm happy to change it to ceiling if you believe that would be more appropriate.

Yes, please change them because they are ceilings.  So, your need to do two things in pull request #999:

(1) Eliminate from the `current_law_policy.json` file the **duplication** of the new parameters you are adding.  There are two sets of new parameter currently.

(2) Change `_ALD_Dependents_Chid_ec` to `_ALD_Dependents_Chid_c` and
change `ALD_Dependents_Elder_ec` to `ALD_Dependents_Elder_c` and
replace in the description of these parameters the word "exclusion" with the word "ceiling".

martinholmer:
@andersonfrailey,  Your proposed code for the new `DependentCare` function is as follows:
```
def DependentCare(agedp1, agedp2, agedp3, agedp4, elderly_dependent, _earned,
                  MARS, ALD_Dependents_thd, ALD_Dependents_HC,
                  ALD_Dependents_Child_c, ALD_Dependents_Elder_c,
                  care_deduction):
    """
    Parameters
    ----------
    agedp1: Age of first dependent
    agedp2: Age of second dependent
    agedp3: Age of third dependent
    agedp4: Age of fourth dependent
    elderly_dependent: 1 if unit has an elderly dependent 0 otherwise
    _earned: Form 2441 earned income amount
    MARS: Marital Status
    ALD_Dependents_thd: Maximum income to qualify for deduction
    ALD_Dependents_HC: Deduction for dependent care haircut
    ALD_Dependents_Child_c: national weighted average cost of childcare
    ALD_Dependents_Elder_c: eldercare cost ceiling

    Returns
    -------
    care_deduction: Total above the line deductions for dependent care.
    """
    if _earned <= ALD_Dependents_thd[MARS - 1]:
        ages = np.array([agedp1, agedp2, agedp3, agedp4])
        # Count the number of qualifying dependents
        children = ((1 == ages) | (ages == 2)).sum()
        care_deduction = (((1 - ALD_Dependents_HC) * children *
                          ALD_Dependents_Child_c) +
                          ((1 - ALD_Dependents_HC) * elderly_dependent *
                          ALD_Dependents_Elder_c))
    else:
        care_deduction = 0.
    return care_deduction
```
Perhaps I'm confused, but there seems to be a documentation problem here.  It looks to me from your pending pull request in the taxdata repository that the four new `puf.csv` variables `agedp?` are **not** as described in the docstring.  They seem to be categorical variables with 1 meaning "under 5 years of age" and 2 meaning "from 5 to under 13 years of age", which is the way they are defined in the IRS-SOI 2009 PUF documentation on page 14 and that is the way your code handles these variables.  But the docstring describes them incorrectly.  You need to fix the description of these four variables in the `DependentCare` function docstring.

martinholmer:
@andersonfrailey,  Note also that the four new `agedp?` variables are defined incorrectly in the `taxcalc/var_labels.txt` file.  You need to fix these four definitions.

andersonfrailey:
@martinholmer 

> They seem to be categorical variables with 1 meaning "under 5 years of age" and 2 meaning "from 5 to under 13 years of age"

This is correct and I added the categories to the docstring for clarification. I have also referenced the 2009 PUF documentation in `var_labels.txt`. I referenced the documentation rather than listing out each category here to keep the file neat. The only agedp variable which this does not apply is agedp4 which was randomly assigned.

martinholmer:
@andersonfrailey,  Thanks for making these documentation changes to pull request #999; things look great now.  I guess the final step is to test the #999 code with the new version of the `puf.csv` file that includes the new variables, right?

andersonfrailey:
@martinholmer, I've been testing it with a `puf.csv` file that is the same as the one currently being used by everyone else just with these new variables and can send it to you if you would like to also test this code. This new file is not being widely distributed because we are making other updates and want to wait until these are finished to minimize any disruptions this would cause.

That being said, I found the reform specified by Donald Trump would decrease tax liabilities by $431 billion from 2017-2026. From 2016-2025 the Tax Foundation [found](http://taxfoundation.org/article/details-and-analysis-donald-trump-tax-reform-plan-september-2016) they would cost $500 billion and the Tax Policy Center [found](http://www.taxpolicycenter.org/publications/analysis-donald-trumps-revised-tax-plan/full) they would cost $131.5 billion from 2016-2016, though they appear to have also included his other reforms in their proposal.

martinholmer:
@andersonfrailey said:
> I've been testing it with a puf.csv file that is the same as the one currently being used by everyone else just with these new variables and can send it to you if you would like to also test this code. This new file is not being widely distributed because we are making other updates and want to wait until these are finished to minimize any disruptions this would cause.

Thanks for the offer, but I can wait for the wider distribution of the new `puf.csv` file.

@andersonfrailey also said:
> That being said, I found the reform specified by Donald Trump would decrease tax liabilities by $431 billion from 2017-2026. From 2016-2025 the Tax Foundation found they would cost $500 billion and the Tax Policy Center found they would cost $131.5 billion from 2016-2016, though they appear to have also included his other reforms in their proposal.

Sounds like you're in the ballpark.  Good work.

martinholmer:
@andersonfrailey,  After looking at your pending taxdata pull request logic and this #999 pull request, I have an efficiency concern.  Your are proposing to add five new variables to the `puf.csv` file when, as far as I can see, you need only two new variables to simulate DependentCare-like reforms.  There is even a comment in your proposed taxdata pull request that suggests you were once considering adding just two new variables: number of kids under 13 years of age and presence of an elderly dependent.  Here is the comment that leads me to this conjecture:
```
/* Add number of dependents under 13 from input file and parents */
CHILDREN = 0;
 IF( AGEDP1 = 1 ) OR ( AGEDP1 = 2) THEN CHILDREN = CHILDREN + 1;
 IF( AGEDP2 = 1 ) OR ( AGEDP2 = 2) THEN CHILDREN = CHILDREN + 1;
 IF( AGEDP3 = 1 ) OR ( AGEDP3 = 3) THEN CHILDREN = CHILDREN + 1;
```
There is a bug in the above code (the last line should have `OR (AGEDP3 = 2) THEN`, but you could correct that bug, give the new variable a more meaningful name (like `nu13`), impute the number of times `nu13 == 3` is incremented to have a value of four, and add just the `nu13` variable to the `puf.csv` file (instead of adding all four `agedp*` variables).

I think being careful about adding new `puf.csv` variables is important both for execution speed and for code clarity.  Let's face it, this kind of reform is not likely to be implemented anytime soon, so it pays to do the minimal amount of enhancement to estimate this reform accurately.

Have I missed some reason why we need all the `agedp*` variables?  If not, I think you should simplify along the lines I suggest above.

andersonfrailey:
@martinholmer when I was originally working on recreating our current `puf.csv` file with the additional variables I planned on adding a variable with the total number of dependents under 13 as you are suggesting. The problem I had was I couldn't get an exact recreation. My file always had 8 fewer records which slightly changed the other output from TaxCalc. @Amy-Xu then gave me a copy of `puf.csv` before all the unused variables were dropped, including the `agedp*` variables so I simply removed those three from the list of variables to drop and made the function work with that. 

I believe you're correct about the efficiency of this though and will create the an `nu13` variable before the next PUF distribution. I'll also update this PR accordingly.

martinholmer:
@andersonfrailey said:
> I believe you're correct about the efficiency of this though and will create the an nu13 variable before the next PUF distribution. I'll also update this PR accordingly.

That would be great.  Thank you.

andersonfrailey:
@martinholmer could you also give this PR final review and merge if you think its ready?

An update on the revenue estimates. After using the latest PUF which corrects the error with the number of houses with children under 13, this PR now shows a $440 billion loss in revenue.

martinholmer:
@andersonfrailey said:
> could you also give this PR final review and merge if you think its ready?

Please send me the pre-release `puf.csv` that you intend to distribute.  Let me check that out before I merge any new code.

martinholmer:
Merging Tax-Calculator pull request #999 into the master branch means that the newest (late Oct) version of the puf.csv must be used on the master branch.

@MattHJensen @feenberg @Amy-Xu @GoFroggyRun @andersonfrailey @codykallen @zrisher

PR #1022, opened 2016-11-01
Fix pytest deprc warning and add json r/w to utils

zrisher:
This small PR:

Changes the pytest section title in `setup.cfg` to fix this deprecation error: 
```
WC1 None [pytest] section in setup.cfg files is deprecated, use [tool:pytest] instead.
```

And adds to `taxcalc.utils`:
* a helper to read json from a file
* a helper to write json to a file
* a TAXCALC_PATH helper constant

All of which are used in an upcoming Luca work commit.

No changes to tax calculation logic.

Please let me know if any issues!

### DIFF
--- a/taxcalc/utils.py
+++ b/taxcalc/utils.py
@@ -13,7 +13,6 @@ import copy
 from collections import defaultdict, OrderedDict
 import json
 import six
-import os
 import numpy as np
 import pandas as pd
 try:
@@ -22,7 +21,6 @@ try:
 except ImportError:
     BOKEH_AVAILABLE = False
 
-TAXCALC_PATH = os.path.dirname(os.path.realpath(__file__))
 
 STATS_COLUMNS = ['_expanded_income', 'c00100', '_standard',
                  'c04470', 'c04600', 'c04800', 'c05200', 'c62100', 'c09600',
--- a/taxcalc/utils.py
+++ b/taxcalc/utils.py
@@ -11,7 +11,9 @@ Tax-Calculator utility functions.
 import math
 import copy
 from collections import defaultdict, OrderedDict
+import json
 import six
+import os
 import numpy as np
 import pandas as pd
 try:
@@ -20,6 +22,7 @@ try:
 except ImportError:
     BOKEH_AVAILABLE = False
 
+TAXCALC_PATH = os.path.dirname(os.path.realpath(__file__))
 
 STATS_COLUMNS = ['_expanded_income', 'c00100', '_standard',
                  'c04470', 'c04600', 'c04800', 'c05200', 'c62100', 'c09600',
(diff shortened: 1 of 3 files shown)


martinholmer:
@talumbau,  Does pull request #1022 make sense to you?

taxcalc/utils.py
```diff
@@ -19,6 +21,7 @@
 except ImportError:
     BOKEH_AVAILABLE = False
 
+TAXCALC_PATH = os.path.dirname(os.path.realpath(__file__))
```
talumbau: Just wondering about this variable here. What is the intended use case? This variable would not hold a useful value in the case of an installed `taxcalc` package.

martinholmer: I don't know, PR#1022 has been submitted by @zrisher.
Are you saying this line is extraneous?

martinholmer:
@zrisher,  Neither @talumbau nor myself can see the point of adding to `utils.py` this line:
```
TAXCALC_PATH = os.path.dirname(os.path.realpath(__file__))
```
If there is no reason, can this line be removed from your pull request #1022?

taxcalc/utils.py
zrisher: Hey @talumbau @martinholmer, sorry for the delay on this one. `TAXCALC_PATH` should always evaluate to the true (resolving symlinks) path to the taxcalc directory, in my case:

```
>>> taxcalc.utils.TAXCALC_PATH
'c:\\Sites\\ospc\\tax-calculator\\taxcalc'
```


There's quite a few places within the codebase where we use a direct child directory: 

```
CUR_PATH = os.path.abspath(os.path.dirname(__file__))
```
 in `validation/csv_in.py`, `tests/conftest.py`, 4 files in `taxbrain/`, and 2 files in `comparison/`.

But this exact string is only used twice within the taxcalc package:

* `CUR_PATH = os.path.abspath(os.path.dirname(__file__))` in `records.py`
* `path = os.path.join(os.path.abspath(os.path.dirname(__file__)), cls.DEFAULTS_FILENAME)` in `parameters.py`

My original idea was that we could define this in `taxcalc.utils`, only access the filesystem to resolve the directory once (slightly improving load times?), and improve code reuse by importing it and referencing it directly, e.g.

```
from taxcalc.utils import TAXCALC_PATH

...

path = os.path.join(TAXCALC_PATH, cls.DEFAULTS_FILENAME)
```

in `parameters.py` or 

```
TAXBRAIN_PATH = os.path.join(TAXCALC_PATH, 'taxbrain')
```

at the top of `taxbrain`, and then each of its 4 files would reference `TAXBRAIN_PATH`.

But I didn't realize that none of these sub directories are actually set up as sub-packages of `taxcalc` (using a `__init__.py` files). I'm also not sure if the file structure of these in relation to `taxcalc` changes when it's installed via a package manager, which is I believe what @talumbau was suggesting.

So there's my use case - if it's worth fixing, let me know @talumbau. Otherwise I will drop it from this PR.

MattHJensen:
@zrisher, I think this is good to go after dealing with https://github.com/open-source-economics/Tax-Calculator/pull/1022#issuecomment-257630107

zrisher:
@MattHJensen I replied to that comment above, just waiting for @talumbau's reply.

zrisher:
@talumbau I removed the `TAXCALC_PATH = ...` line, because while I think it has value, it'll be easier to demonstrate by simply coding up all its uses in a separate PR. This should now be good to merge.

Note the appveyor test is failing due to needing conda >= 4.2.12 per https://github.com/open-source-economics/Tax-Calculator/pull/1114

martinholmer:
Merged pull request #1022.  @zrisher,  thanks for this contribution.

PR #1027, opened 2016-11-02
Add _CG_dollar_ec and _CG_reinvest_ec policy parameters

martinholmer:
This pull request adds two new policy parameters that, together with the _CG_nodiff parameter, make it possible to simulate the capital gains and qualified dividends reform provision in a new tax reform proposal.  These code changes cause no changes in current-law tax results or in the reform comparison results.

Here is a description of this reform provision:
```
Tax all (not just long-term) capital gains and qualified dividends at ordinary income
tax rates, but provide a 50% exclusion for capital gains that are reinvested within 
the same tax year that the gains are realized as well as a small investor exclusion of
the first $10,000 in capital gains.
```
This description was clarified by these answers to my questions:
```
Question 1: Reinvested capital gains and qualified dividends that are invested 
should be treated equally. The 50% exclusion should apply to realized gains and
qualified dividends that are reinvested in the same calendar year in which they
are realized/received.

Question 2: The small investor exclusion should also apply to both capital gains
and qualified dividends.

Question 3: The small investor exclusion should be adjusted for inflation, so it 
should be a nominal $10,000 threshold in the first year of enactment.

Question 4: The $10k exclusion should apply first and the 50% exclusion for 
reinvested gains and qualified dividends should apply second.  If I have $10,500
in capital gains and dividends, and I reinvest at least $500, then I would have to
include $250 in my taxable income.

Yes, the exclusions will reduce AGI.
```
Please review these code changes to see if you think they capture the logic of this reform.

See below for some preliminary results on the 2018 income tax revenue effects of this reform provision considered in isolation.

```

$ python inctax.py puf.csv 2018 --blowup --weights
You loaded data for 2009.
Your data have been extrapolated to 2018.

$ awk '{r+=$4*$29}END{print r}' puf-18.out-inctax
1.564e+12

$ cat qdcg.json 
{
    // This reform taxes LTCG+QDIV no differently than other income and
    // provides a dollar and fractional exclusion of CG+QDIV income from AGI
    "_CG_nodiff": {"2018": [1]},
    "_CG_dollar_ec": {"2018": [10000]},
    "_CG_reinvest_ec": {"2018": [0.30]}
        // above 0.30 effective rate is derived as follows:
        // - assume statutory exclusion rate is 0.50
        // - assume 60% is reinvested within the year
        // - so 0.50 * 0.60 equals 0.30  
        // Vary the effective rate as desired as there is little or no
        // evidence about what fraction is reinvested within the year
        // of realization.
}

$ python inctax.py puf.csv 2018 --blowup --weights --reform qdcg.json
You loaded data for 2009.
Your data have been extrapolated to 2018.

$ awk '{r+=$4*$29}END{print r}' puf-18.out-inctax-qdcg 
1.17878e+12

Notice that this reform provision (along with the 60% reinvestment
assumption) causes an decrease in 2018 aggregate income tax revenue
of about $385.2 billion.

If we change the assumed reinvestment rate from 60% to 40% --- so the
effective exclusion rate is 0.20 (instead of 0.30) --- then we have this:

$ awk '{r+=$4*$29}END{print r}' puf-18.out-inctax-qdcg 
1.18577e+12

Aggregate income tax revenue is up by a little, but the decline in
2018 revenue is still $378.2 billion.

If the effective reinvestment exclusion rate is reduced to zero, then
we have these results:

$ awk '{r+=$4*$29}END{print r}' puf-18.out-inctax-qdcg 
1.20664e+12

Aggregate income tax revenue is up by a little more, but the decline
in 2018 revenue is still $357.4 billion.
```

### DIFF
--- a/taxcalc/current_law_policy.json
+++ b/taxcalc/current_law_policy.json
@@ -943,6 +943,32 @@
         "validations": {"min": 0, "max": 1}
     },
 
+    "_CG_dollar_ec": {
+        "long_name": "Dollar amount of all capital gains and qualified dividends that are excluded from AGI.",
+        "description": "Current-law value is zero; positive value used only if _CG_nodiff is non-zero.",
+        "irs_ref": "",
+        "start_year": 2013,
+        "col_var": "",
+        "row_var": "FLPDYR",
+        "row_label": ["2013"],
+        "cpi_inflated": true,
+        "col_label": "",
+        "value": [0.0]
+    },
+
+    "_CG_reinvest_ec": {
+        "long_name": "Fraction of all capital gains and qualified dividends in excess of _CG_dollar_ec that are excluded from AGI.",
+        "description": "Current-law value is zero; positive value used only if _CG_nodiff is non-zero.  If positive, set to statutory exclusion rate times the fraction of capital gains and qualified dividends in excess of _CG_dollar_ec that are assumed to be reinvested within the year.",
+        "irs_ref": "",
+        "start_year": 2013,
+        "col_var": "",
+        "row_var": "FLPDYR",
+        "row_label": ["2013"],
+        "cpi_inflated": false,
+        "col_label": "",
+        "value": [0.0]
+    },
+
     "_CG_rt1": {
         "long_name": "Long term capital gain and qualified dividends rate 1",
         "description": "The capital gain and dividends (stacked on top of regular income) that are below threshold 1 are taxed at this rate. ",
--- a/taxcalc/current_law_policy.json
+++ b/taxcalc/current_law_policy.json
@@ -943,7 +943,7 @@
         "validations": {"min": 0, "max": 1}
     },
 
-    "_CG_dollar_ec": {
+    "_CG_ec": {
         "long_name": "Dollar amount of all capital gains and qualified dividends that are excluded from AGI.",
         "description": "Current-law value is zero; positive value used only if _CG_nodiff is non-zero.",
         "irs_ref": "",
@@ -956,9 +956,9 @@
         "value": [0.0]
     },
 
-    "_CG_reinvest_ec": {
-        "long_name": "Fraction of all capital gains and qualified dividends in excess of _CG_dollar_ec that are excluded from AGI.",
-        "description": "Current-law value is zero; positive value used only if _CG_nodiff is non-zero.  If positive, set to statutory exclusion rate times the fraction of capital gains and qualified dividends in excess of _CG_dollar_ec that are assumed to be reinvested within the year.",
+    "_CG_reinvest_ec_rt": {
+        "long_name": "Fraction of all capital gains and qualified dividends in excess of _CG_ec that are excluded from AGI.",
+        "description": "Current-law value is zero; positive value used only if _CG_nodiff is non-zero.  If positive, set to statutory exclusion rate times the fraction of capital gains and qualified dividends in excess of _CG_ec that are assumed to be reinvested within the year.",
         "irs_ref": "",
         "start_year": 2013,
         "col_var": "",
@@ -1385,8 +1385,8 @@
     },
 
     "_NIIT_thd": {
-        "long_name": "Net Investment Income Tax income threshold (UIMC)",
-        "description": "If MAGI (modified AGI) is more than this threshold, tax unit is subject to the Net Investment Income Tax.",
+        "long_name": "Net Investment Income Tax income threshold",
+        "description": "If MAGI (modified AGI) is more than this threshold, filing unit is subject to the Net Investment Income Tax.",
         "irs_ref": "Form 8960, line 14, instructions. ",
         "notes": "",
         "start_year": 2013,
--- a/taxcalc/functions.py
+++ b/taxcalc/functions.py
@@ -173,7 +173,8 @@ def Adj(e03150, e03210, c03260,
 
 @iterate_jit(nopython=True)
 def CapGains(p23250, p22250, _sep, ALD_Investment_ec, ALD_StudentLoan_HC,
-             e00200, e00300, e00600, e00700, e00800,
+             e00200, e00300, e00600, e00650, e00700, e00800,
+             CG_nodiff, CG_dollar_ec, CG_reinvest_ec,
              e00900, e01100, e01200, e01400, e01700, e02000, e02100,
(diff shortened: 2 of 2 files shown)


codykallen:
Since this reduces AGI, it has interaction effects with other provisions that are phased out with income. This may be intentional, since they don't want to treat investment income that is immediately reinvested as ordinary income. However, if you want to avoid the interaction from phasing out other provisions, it may be useful to reduce taxable income by the amount to be excluded.

martinholmer:
@codykallen said:
> Since this reduces AGI, it has interaction effects with other provisions that are phased out with income. This may be intentional, since they don't want to treat investment income that is immediately reinvested as ordinary income. However, if you want to avoid the interaction from phasing out other provisions, it may be useful to reduce taxable income by the amount to be excluded.

The rest of the Lee reform will eliminate several of these interactions.  The staff was quite explicit about the CG+QDIV exclusions reducing AGI.  So, we'll stick with this formulation because it is what they asked for.  But I appreciate very much you quick review of this pull request.

MattHJensen:
@martinholmer, thanks for this one too. Besides comments above, everything looks good.

MattHJensen:
@martinholmer, The comments above were also pending. They should be visible now.

taxcalc/functions.py
```diff
@@ -184,11 +185,18 @@ def CapGains(p23250, p22250, _sep, ALD_Investment_ec, ALD_StudentLoan_HC,
     c23650 = p23250 + p22250
     # limitation on capital losses
     c01000 = max((-3000. / _sep), c23650)
-    # compute ymod* variables
+    # compute ymod1 variable that is included in AGI
     ymod1 = (e00200 + e00700 + e00800 + e00900 + e01400 + e01700 +
              (1 - ALD_Investment_ec) * (e00300 + e00600 +
                                         c01000 + e01100 + e01200) +
              e02000 + e02100 + e02300)
+    if CG_nodiff != 0.:
```
martinholmer: Yes.  Look at the sponsor's description of the reform in the first item in this discussion.

Both current law and this reform proposal give capital gains and qualified dividends preferential treatment, but they do that in different ways.  Current law taxes **long-term** capital gain and qualified dividends at lower rates than other income, while this proposal taxes them at the same rates as other income but gives allows an exclusion of **all** capital gains and qualified dividends.  Neither current law nor this reform proposal contemplates allowing exclusions and lower tax rates.

taxcalc/current_law_policy.json
```diff
@@ -943,6 +943,32 @@
...
+        "start_year": 2013,
+        "col_var": "",
+        "row_var": "FLPDYR",
+        "row_label": ["2013"],
+        "cpi_inflated": true,
+        "col_label": "",
+        "value": [0.0]
+    },
+
+    "_CG_reinvest_ec": {
```
martinholmer: Yes, this is a good suggestion.  This change has been made.

taxcalc/current_law_policy.json
```diff
@@ -943,6 +943,32 @@
         "validations": {"min": 0, "max": 1}
     },
 
+    "_CG_dollar_ec": {
```
MattHJensen: This all makes sense to me,  including separating the other parameter name fixes into a new PR. We will want to issue a release before that PR is merged as it will have breaking changes that require coordinating with tax-calculator users.  (cc @talumbau). 

@martinholmer, do you think our current use of "floor" is appropriate? It seems to be a special case of an exclusion where the excluded amount is keyed off of some special income type. 

(cc @Amy-Xu on this discussion as well).

taxcalc/functions.py
MattHJensen: Some users do want to have both preferential rates and an exclusion. For example, the current House Republican [Blueprint](https://abetterway.speaker.gov/_assets/pdf/ABetterWay-Tax-PolicyPaper.pdf) includes both preferential rates and a 50% exclusion. I believe @codykallen recently modeled this by just cutting the rates in half, but if we have a parameter called `qdcg_exclusion`, then some users might abuse it and try to implement it when CG_nodiff = 0. There is a warning in the notes about not doing that, but many users don't read the notes carefully and they won't receive any other warning that the reform they thought they were implementing wasn't implemented. 

If it adds significant complexity to relax the constraint that CG_nodiff!=0, then I understand taking the risk that some users make mistakes, but if it is simple to relax the constraint, then it could enhance usability.

martinholmer: @MattHJensen said:
> Some users do want to have both preferential rates and an exclusion. For example, the current House Republican Blueprint includes both preferential rates and a 50% exclusion. I believe @codykallen recently modeled this by just cutting the rates in half, but if we have a parameter called qdcg_exclusion, then some users might abuse it and try to implement it when CG_nodiff == 0. There is a warning in the notes about not doing that, but many users don't read the notes carefully and they won't receive any other warning that the reform they thought they were implementing wasn't implemented.
>
> If it adds significant complexity to relax the constraint that CG_nodiff!=0, then I understand taking the risk that some users make mistakes, but if it is simple to relax the constraint, then it is probably worthwhile for enhancing usability.

This pull request #1027 adds the `_CG_ec` parameter whose long name is:
```
Dollar amount of all capital gains and qualified dividends
that are excluded from AGI.
```
This makes sense when `_CG_nodiff` is not zero and all capital gains and qualified dividends are taxed at the same rates as other income.

But when `_CG_nodiff` is zero (as under current law), I don't know how to use the `_CG_ec` amount.  Don't we have to know what part of `_CG_ec` can be used to reduce taxable short-term capital gains as opposed to the part that can be used to reduce taxable long-term capital gains plus qualified dividends?  And how much of the part that reduces taxable LTCG+QDIV income can be used to reduce taxable special capital gains (like the gains on collectibles)?  Perhaps @codykallen has figured all this out, but there was nothing about this in the reform proposal that generated this pull request #1027.

MattHJensen: I understand the difficulties now. thanks @martinholmer. I say we merge this now and then think about relaxing the constraint later as necessary to other users.

martinholmer:
@MattHJensen said:
> I understand the difficulties now. thanks @martinholmer. I say we merge this now and then think about relaxing the constraint later as necessary to other users.

OK, I'll merge #1027 in the next few minutes.

PR #1036, opened 2016-11-06
Format-only changes that remove error-prone backslash continuations

martinholmer:
No change in logic or in results.


MattHJensen:
+1

PR #1029, opened 2016-11-03
Revise logic of the additional additional child tax credit

martinholmer:
This pull request adds one new policy parameter, renames others, and changes the logic of the new (formerly, "additional additional") child tax credit, which is not part of current-law policy.  These code changes do not cause any change in unit test results, in validation test results, or in reform comparison results.  These changes, however, will require the removal of the Lee-Rubio parameter from TaxBrain.

@MattHJensen @feenberg @talumbau @Amy-Xu @GoFroggyRun @andersonfrailey @codykallen

PR #1039, opened 2016-11-07
Implement Clinton CTC expansion

MattHJensen:
This allows for users to implement Clinton's CTC expansion in Tax-Calculator and resolves #1031. 

Here is how the Tax Policy Center [describes](http://www.taxpolicycenter.org/sites/default/files/alfresco/publication-pdfs/2000922-An-Updated-Analysis-of-Hillary-Clintons-Tax-Proposals.pdf) the Clinton proposal:

> NOTE: The Clinton campaign released a proposal for an expanded child tax credit (CTC) on
> October 11, 2016. Clinton proposes to increase the maximum credit to $2,000 and
> increase the phase-in rate from 15 percent to 45 percent for the refundable portion of the
> credit for families with children under age 5. For all families with eligible children, Clinton
> would eliminate the minimum earnings requirement (currently $3,000). 

Implementing all three reforms, TC generates a revenue cost of $15.6 B in 2020. TPC finds a revenue cost of $22 B in the same year, but this is stacked after several other reforms. 

# Test HC CTC Expansion Capabilities

### Import taxcalc package and other useful packages

```python
import sys
sys.path.append("../../")
from taxcalc import *
```

### Create Plan X and Plan Y  Policy objects containing current law policy and then implement reforms

```python
# The baseline includes AMT repeal. 
p_xx = Policy()

p_y1 = Policy()
reform1 = {2016: {
'_CTC_c_under5_bonus': [1000] }
}
p_y1.implement_reform(reform1)

p_y2 = Policy()
reform2 = {2016: {
'_CTC_c_under5_bonus': [1000],
'_ACTC_rt_bonus_under5family': [0.3],}
}
p_y2.implement_reform(reform2)

p_y3 = Policy()
reform3 = {2016: {
'_CTC_c_under5_bonus': [1000],
'_ACTC_rt_bonus_under5family': [0.3],
'_ACTC_Income_thd': [0]}
}
p_y3.implement_reform(reform3)
```

### Create calculator objects with default tax data and advance the calculator to 2020

```python
c_xx = Calculator(policy=p_xx, records=Records("../../puf.csv"))
c_xx.advance_to_year(2020)
c_y1 = Calculator(policy=p_y1, records=Records("../../puf.csv"))
c_y1.advance_to_year(2020)
c_y2 = Calculator(policy=p_y2, records=Records("../../puf.csv"))
c_y2.advance_to_year(2020)
c_y3 = Calculator(policy=p_y3, records=Records("../../puf.csv"))
[...]

PR #1028, opened 2016-11-02
Add _EITC_indiv policy parameter and associated logic

martinholmer:
This pull request adds the `_EITC_indiv` policy parameter that causes the current-law, filing-unit-based and number-of-kids-based EITC to be replaced by an individual-based EITC that does not vary by the number of kids.  These code changes cause no change in current-law tax results or in reform comparison results.

Here is a description of this reform provision:
```
Replace the Earned Income Tax Credit (EITC) with a credit specifically focused on 
workers (without an additional calculation for children in a household).
   This reform would create a subsidy for workers who earn less than $35,000
in wage income. The maximum subsidy would be $1000 and would be phased
in at a rate of 10% starting with the first dollar in wage income and hitting the
maximum subsidy at $10,000 in earned income.  Other tax provisions aside,
this feature would create a 10% negative income tax bracket up to the $10,000
earnings level.
   The subsidy would phase out at a 10% rate beginning at $25,000 with the 
credit fully phased out by the time a worker earns $35,000 in wages. 
   The subsidy would be calculated on a per-worker income basis rather than
on a household income basis like the current EITC. 
   Under the proposal, if a family has more than one wage earner, each would 
be eligible for the wage subsidy up to $35,000 in wages. This could result in
a married couple earning a total of $50,000 ($25,000 each) with $2,000 wage
subsidy.
```
Subsequent discussion confirmed that in the above description the phrase
"wage income" is the same concept as used in the current EITC: 
```
[We will not] be changing the definition of income for the redesigned EITC.
It should continue to include wage and salary income along with self-
employment income minus half the payroll tax [on self-employment income].
```

Please review these code changes to see if you think they capture the logic of this reform.

I'm assuming the property income eligibility rule under current law is not part of
this reform proposal because there is no way (with available information) to allocate
property income between the taxpayer and spouse.

See below for some preliminary results on the 2018 income tax revenue effects of this reform provision considered in isolation.

@MattHJensen @feenberg @Amy-Xu @GoFroggyRun @andersonfrailey @codykallen

```
$ cp ../tax-calculator-data/puf.csv .

$ python inctax.py puf.csv 2018 --blowup --weights
You loaded data for 2009.
Your data have been extrapolated to 2018.

$ awk '{r+=$4*$29}END{print r}' puf-18.out-inctax
1.564e+12

$ cat ind-eitc.json 
{
    // This reform replaces current filing-unit EITC with an individual EITC
[...]

PR #1040, opened 2016-11-07
[WAIT] Correct names of several policy parameters

martinholmer:
These changes bring the parameter names into closer alignment with the project's [parameter naming conventions](http://taxcalc.readthedocs.io/en/latest/parameter_naming.html).  The need for these changes arose in the discussion of pull request #1029.

@MattHJensen @talumbau @Amy-Xu @GoFroggyRun @andersonfrailey @codykallen @jdebacker


MattHJensen:
Thanks @martinholmer. Let's not merge this until after #1039, #1028, and #1027  are merged and a new release is issued as it will require changes to downstream users of Tax-calculator.

martinholmer:
@MattHJensen said:
> Thanks @martinholmer. Let's not merge this until after #1039, #1028, and #1027 are merged and a new release is issued as it will require changes to downstream users of Tax-calculator.

But if we don't include this in the new TaxBrain version, the reform file upload capability won't work, will it?  Why don't you get everybody on the same page now, so we can fix this?

MattHJensen:
> But if we don't include this in the new TaxBrain version, the reform file upload capability won't work, will it?
 
#1040 will only cause breaking changes for downstream users once it is included in a release. If we don't merge it now, then we can have release immediately w #1039, #1028, #1027, and the other changes since our last release. This first release won't break anything for anyone and can be deployed immediately. Then we can follow up with a second release that only includes #1040. The second release _will_ contain breaking changes for downstream users (renamed parameters), but we'll be able to get the fixes in place in a deliberate and careful manner without worrying about holding anything up.

MattHJensen:
+1

PR #1046, opened 2016-11-09
[WAIT] Rename _AMT_* params to conform with other parameter names

martinholmer:
This pull request brings the AMT tax bracket and rates parameter names into line with other parameter names.  There is no change in tax-calculating logic or in any tax results.


MattHJensen:
+1

PR #1047, opened 2016-11-09
[WAIT] Rename several CG_thd parameters as CG_brk

martinholmer:
This pull request change several parameter names so that they correspond to the naming of other parameters.  This move towards consistency does not change any tax logic or tax results.


MattHJensen:
+1

PR #1049, opened 2016-11-09
[WAIT] Rename _DCC_c parameter as _CDCC_c

martinholmer:
This pull request make the naming of the three Form 2441 policy parameters more consistent.
There is no change in tax-calculating logic, and thus, no changes in tax results.


MattHJensen:
+1

PR #1057, opened 2016-11-11
Fix calculation of expanded income

martinholmer:
This pull request adds several things into the `_expanded_income` variable that were previously left out of expanded income.  There is no change in tax-calculating logic, but this pull request will cause the value of `_expanded_income` to be higher for some filing units.  Obviously, these changes will affect the results of any post-simulation tabulation that uses the `_expanded_income` variable.

In the decision making about what new items to add into `_expanded_income`, the JCT principles discussed below in #1057 have been followed with respect to pension contributions and health insurance premiums.

@MattHJensen @feenberg @talumbau @Amy-Xu @GoFroggyRun @andersonfrailey @codykallen


feenberg:
On Fri, 11 Nov 2016, Martin Holmer wrote:

> 
> This pull request adds several things into the _expanded_income variable that were
> previously left out of expanded income. There is no change in tax-calculating logic,
> but this pull request will cause the value of _expanded_income to be higher for some
> filing units. Obviously, these changes will affect the results of any post-simulation
> tabulation that uses the _expanded_income variable.

1) What are our sources for "investment income not in AGI"?

2) Are we sure about the employer share of SETAX? If we use the c02900
    value, and a reform increases the tax rate, does that increase expanded
    income?

3) I am also concerned about the statement " Note: if behavioral responses
    to a policy reform are specified, then be sure this function is called
    after the behavioral responses are calculated." documenting the
    expanded income function. What is our use of expanded income? I thought
    it was the tab variable for tables, in which case it should use base
    law values.

4) I endorse adding CMBTP to the expanded income concept, but I wonder if
    JTC, TPC and CBO do the same.

dan

martinholmer:
Dan @feenberg asked the following questions about pull request #1057:
> 1) What are our sources for "investment income not in AGI"?

Zero under current-law policy; when positive it is computed this way:
```
 # compute ymod1 variable that is included in AGI
 invinc = e00300 + e00600 + c01000 + e01100 + e01200
 invinc_agi_ec = ALD_Investment_ec_rt * invinc
 ymod1 = (e00200 + e00700 + e00800 + e00900 + e01400 + e01700 +
          invinc - invinc_agi_ec +
          e02000 + e02100 + e02300)
```

2) Are we sure about the employer share of SETAX? If we use the c02900
    value, and a reform increases the tax rate, does that increase expanded
    income?

c02900, no matter what its value, is excluded from AGI, so adding it back in
makes sense to me.

3) I am also concerned about the statement " Note: if behavioral responses
    to a policy reform are specified, then be sure this function is called
    after the behavioral responses are calculated." documenting the
    expanded income function. What is our use of expanded income? I thought
    it was the tab variable for tables, in which case it should use base
    law values.

That's been there for many, many months.  I don't know much about the
behavioral response logic in Tax-Calculator, so somebody else will have to
answer your question.

4) I endorse adding CMBTP to the expanded income concept, but I wonder if
    JTC, TPC and CBO do the same.

Don't know about the others, but the logic of inclusion seems fairly strong.

@MattHJensen @Amy-Xu @GoFroggyRun @andersonfrailey @codykallen

MattHJensen:
@feenberg asked: 

> 3) I am also concerned about the statement " Note: if behavioral responses to a policy reform are specified, then be sure this function is called after the behavioral responses are calculated." documenting the expanded income function. What is our use of expanded income? I thought it was the tab variable for tables, in which case it should use base law values.

Yes, the purpose of expanded income is as the tab variable for tables, and we do use the base law values. Given that, I think the statement in question about behavioral responses should be removed.

MattHJensen:
@feenberg asked:

> 4) I endorse adding CMBTP to the expanded income concept, but I wonder if
JTC, TPC and CBO do the same.

@martinholmer replied:
> Don't know about the others, but the logic of inclusion seems fairly strong.

I think we should include them based on our reasoning alone. 

But for reference, JCT also includes them: see [page 7 of their report on income measures](https://www.jct.gov/publications.html?func=startdown&id=4408).

martinholmer:
@MattHJensen said about including `cmbtp` items in expanded_income:
> I think we should include them based on our reasoning alone.
>
> But for reference, JCT also includes them: see page 7 of their [report on income measures](https://www.jct.gov/publications.html?func=startdown&id=4408).

Thanks for this useful JCT reference.  It raises a couple of other issues:

(1) They don't add into expanded income deductible contributions to IRAs (or other non-employer-sponsored defined-contribution pensions?) as discussed on pages 9-10.  They do this because they don't have data on contribution to employer-sponsored pension plans.  They admit that if they did have data on employer-sponsored pension contributions that they would include all of these contributions in expanded income.  But because they don't have the data on contributions to employer-sponsored plans, they do not add into expanded income the known deductible contributions to IRAs or other non-employer-sponsored defined-contribution pensions (`e03150` + `e03300`).  Or at least this is the way it seems to me.  Do you agree that this is what JCT is doing?  Are they really not including in expanded income the `e03300` amount?  Do you want Tax-Calculator to do the same thing with respect to `e03150` and possibly `e03300`?

(2) On page 12, after discussing filing units with negative pass-through income (which is included in expanded income), JCT says: "For this reason, the Joint Committee staff eliminates observations with negative net expanded income from distributional analysis."  Does this make sense?  What, if anything, should we doing special with those filing units with negative expanded income?

feenberg:
On Fri, 11 Nov 2016, Martin Holmer wrote:

> 
> @MattHJensen said about including cmbtp items in expanded_income:
>
>       I think we should include them based on our reasoning alone.
>
>       But for reference, JCT also includes them: see page 7 of their report on income measures.
> 
> Thanks for this useful JCT reference. It raises a couple of other issues:
> 
> (1) They don't add into expanded income deductible contributions to IRAs (or other
> non-employer-sponsored defined-contribution pensions?) as discussed on pages 9-10. They do this
> because they don't have data on contribution to employer-sponsored pension plans. They admit that if
> they did have data on employer-sponsored pension contributions that they would include all of these
> contributions in expanded income. But because they don't have the data on contributions to
> employer-sponsored plans, they do not add into expanded income the known deductible contributions to
> IRAs or other non-employer-sponsored defined-contribution pensions (e03150 + e03300). Or at least this
> is the way it seems to me. Do you agree that this is what JCT is doing? Are they really not including
> in expanded income the e03300 amount? Do you want Tax-Calculator to do the same thing with respect to
> e03150 and possibly e03300?

I think the JCT procedure might make some sense if the two contribution 
streams had a high negative correlation. Then leaving both out makes the 
measurement error in income more random and less likely to bias 
conclusions. I think that since IRA eligibility is based on NOT having an 
employer provided pension, there may be a negative correlation.

> 
> (2) On page 12, after discussing filing units with negative pass-through income (which is included in
> expanded income), JCT says: "For this reason, the Joint Committee staff eliminates observations with
> negative net expanded income from distributional analysis." Does this make sense? What, if anything,
> should we doing special with those filing units with negative expanded income?
>

Back when there were tax shelters, losses were often artificial and it 
made sense to exclude them. I think currently they are excluded because 
they are thought to be temporary aberrations for otherwise wealthy people. 
But so are capital gains, and JCT doesn't exclude them. So I don't think 
this makes sense. However, we should probably follow JCT, perhaps adding a 
row for excluded records. I wonder if taxpayers with losses show less 
property tax than taxpayers with no losses and similar incomes?

martinholmer:
Dan @feenberg,  Thanks for your thoughtful expanded-income comments on pull request #1057.

taxcalc/functions.py
```diff
@@ -118,65 +118,72 @@ def Adj(e03150, e03210, c03260,
...
+        ALD_StudentLoan_hc : Student loan interest deduction haircut
+
+        ALD_SelfEmp_HealthIns_hc : Self-employed h.i. deduction haircut
 
-        ALD_KEOGH_SEP_hc :
-        Deduction for payment to either KEOGH or SEP plan haircut
+        ALD_KEOGH_SEP_hc : KEOGH/etc. plan contribution deduction haircut
 
-        ALD_EarlyWithdraw_hc : Deduction for forfeited interest penalty haricut
+        ALD_EarlyWithdraw_hc : Penalty on early withdrawl deduction haricut
```
MattHJensen: withdrawl - > withdrawal

taxcalc/functions.py
```diff
@@ -118,65 +118,72 @@ def Adj(e03150, e03210, c03260,
...
-              (1. - ALD_Alimony_hc) * e03500 +
-              e03220 + e03230 + e03240 + e03290 + care_deduction)
-    return c02900
+    c02900 : total Form 1040 adjustments, which are not included in AGI
+
+    c02900_less : total adjustments less those not included in expanded income
+    """
+    # Form 2555 foreign earned income deduction is assumed to be zero
+    # Form 1040 adjustments that are included in expanded income:
+    c02900_less = ((1. - ALD_StudentLoan_hc) * e03210 +
```
MattHJensen: What do you think about C02900_in_expanded_income? or C02900_EI?

MattHJensen:
@martinholmer, all of the substantive changes in this PR look good to me. I left a couple of small comments.

martinholmer:
@MattHJensen said:
> All of the substantive changes in this PR look good to me. I left a couple of small comments.

Thanks for the suggestions.  I've revised pull request #1057 along the lines you suggested.

PR #1055, opened 2016-11-11
add under5 bonus to ctc_new

MattHJensen:
fixes #1054 

Leaving WIP tag until I have a chance to generate and post reform results.

PR #1062, opened 2016-11-15
NIIT_PT_taxed -> boolean

MattHJensen:
In #1055, we started to use booleans instead of 1/0 indicators for parameter switches. I missed `NIIT_PT_taxed` before.


MattHJensen:
I removed the note "Current law has _NIIT_PT_taxed equal to zero" entirely since that is implied by the default value of the param and we don't include similar notes for other variables. 

cc @martinholmer

martinholmer:
@MattHJensen, #1062 looks good.  I never realized boolean policy parameters were OK.

MattHJensen:
Thanks @martinholmer for the speedy review!

Issue #1050, opened 2016-11-09
Proposal for proxy welfare measures


MattHJensen:
@gleiserson (Greg Leiserson) emailed me a detailed proposal for adding proxy welfare measures to Tax-Calculator and potentially TaxBrain output tables. 

[Proposal for proxy welfare measures.docx](https://github.com/open-source-economics/Tax-Calculator/files/581613/Proposal.for.proxy.welfare.measures.docx)

[Table Shells.xlsx](https://github.com/open-source-economics/Tax-Calculator/files/581616/Table.Shells.xlsx)

I will be studying and thinking about this over the next week or so, and I will leave comments here. I'm hoping that others will find this interesting and do the same. 

cc @martinholmer @feenberg @viard @andersonfrailey @Amy-Xu @GoFroggyRun @codykallen

MattHJensen:
**Proposal for proxy welfare measures**

This note outlines a methodology for computing proxy welfare measures
for use in assessing the normative implications of changes in tax
policy. It primarily relies on estimates of changes in tax liabilities
from Tax-Calculator. In addition, users may assume an ad hoc revenue
offset or use an estimate of the revenue offset from another economic
model to estimate a balanced-budget welfare impact.[^1] See Table
Shells.xlsx for sample output files.

User input: (i) welfare measures requested (step one only; steps one and
two; or steps one, two, and three), (ii) income classifiers and ranges,
(iii) choice of $\theta$ for use in estimating welfare impacts (default
is 0, 2, and 4), (iv) consumption level at the assumed upper bound for
marginal utility (default \$1,000), (v) assumed marginal utility of
income for tax units with zero or negative income in the baseline
(default zero), (vi) choice of revenue offset (default 15 percent), and
(vii) income elasticity (default 0.4).

Step 1: Static change in after-tax income without financing, welfare
impacts per assumed specification for utility of consumption

-   Using the incidence assumptions otherwise reflected in the model,
    compute after-tax income under the baseline and under the policy
    alternative for each tax unit.

-   For each income class compute the aggregate (for the class)
    percentage change in after-tax income.

-   For the specified set of values for $\theta,$ compute the change in
    welfare for each tax unit with positive after-tax income in the
    baseline as
`
[...]

PR #1066, opened 2016-11-15
Add lump-sum-tax policy parameters and associated logic

martinholmer:
This pull request adds the ability to specify a lump-sum tax as part of individual income tax liability.  Of course, there is no such tax under current-law policy, but such a tax can be useful in tax reform analysis because it provides a simple way to "finance" a revenue-reducing reform.  Adding this optional feature to Tax-Calculator will better enable us to evaluate the ideas being discussed in Tax-Calculator issue #1050.

Unless a non-zero value of the lump-sum tax is specified, this pull request does not change tax-calculating logic, and therefore, does not change any tax results.

@MattHJensen @feenberg @gleiserson @viard @codykallen @andersonfrailey @GoFroggyRun


feenberg:
On Tue, 15 Nov 2016, Martin Holmer wrote:

> 
> This pull request adds the ability to specify a lump-sum tax as part of
> individual income tax liability. Of course, there is no such tax under

Before or after credits, AMT?

Can the value be negative (for a negative income tax)?

Should we priviledge this particular financing method over others?

Do we anticipate users iterating to a revenue neutral level of tax? If we 
do, then should we consider automating the iteration, so that the user 
could, for instance enter a ? in the box and TB would determine the level 
of lump-sum tax corresponding to revenue neutrality and report it to the 
user?

dan

martinholmer:
@feenberg asked:
>> On Tue, 15 Nov 2016, Martin Holmer wrote:
>>  This pull request [#1066] adds the ability to specify a lump-sum tax as part of  individual income tax liability.  Of course, there is no such tax under current law ...
>
> Before or after credits, AMT?

Reading the proposed code change in #1066, we see this in the revised `IITAX` function:
```
    # compute lump-sum tax
     lumpsum_tax = LST
     if LST_per_adult and MARS == 2:
         lumpsum_tax += LST
     # compute final taxes
     _eitc = c59660
     _refund = _eitc + c11070 + c10960 + personal_credit + ctc_new
     _iitax = c09200 - _refund + lumpsum_tax
     _combined = _iitax + _payrolltax
     return (lumpsum_tax, ctc_new, _eitc, _refund, _iitax, _combined)
```
So, the lump-sum tax is added independent of everything else in the individual income tax calculation.

> Can the value be negative (for a negative income tax)?

Yes.

> Should we [prefer] this particular financing method over others?

No.  I see it as just one option.

> Do we anticipate users iterating to a revenue neutral level of tax? If we do, then should we consider automating the iteration, so that the user could, for instance enter a ? in the box and TB would determine the level of lump-sum tax corresponding to revenue neutrality and report it to the
user?

Yes and No.  It seems to me that Tax-Calculator/TaxBrain is fast enough that manual iteration is just fine.  I see automation of this process as a very low priority.

@MattHJensen @feenberg @gleiserson @viard @codykallen @andersonfrailey @GoFroggyRun

taxcalc/current_law_policy.json
```diff
@@ -2222,6 +2222,34 @@
...
+        "start_year": 2013,
+        "col_var": "",
+        "row_var": "",
+        "row_label": ["2013"],
+        "cpi_inflated": false,
+        "col_label": "",
+        "value": [0.0]
+    },
+
+    "_LST_per_adult": {
```
MattHJensen: Rather than adding `_LST_per_adult`, why not index `_LST` by MARS?

martinholmer:
@MattHJensen asked:
> Rather than adding [the] _LST_per_adult [parameter], why not index _LST by MARS?

Because we want the lump-sum tax to be non-distortionary.

The Wikipedia article on "Lump-sum tax" says, among other things, the following:
```
In economic theory, a head tax is a special type of lump-sum tax that is considered 
to be Pareto-efficient since it does not interfere with optimal market mechanisms.
A head tax will only reduce people's available income and therefore decrease their 
budget constraint, but leave the relative price of goods unchanged.  In basic 
microeconomics consumer theory, this will then lead to an income effect: 
consumers buy less in general (inwards shift of the budget line). There will be no
substitution effect.(Hindriks, Myles, 2006).

Similarly, a single tax on the unimproved value of land will also have the same
attributes since land itself exists in a fixed amount that can not be altered 
(although the improvements on it certain can vary, which is why a property 
tax does not have this attribute).

Therefore, this form of taxation may have the advantage of not contributing
to an excess burden of taxation, and loss in economic efficiency that results
from taxes reducing incentives for production or consumption.
```
It seem to me that indexing the lump-sum tax by MARS will create unwanted
incentives to change filing status if the lump-sum tax amounts are other than
the way we have them now in pull request #1066.

@MattHJensen @feenberg @gleiserson @viard @codykallen @andersonfrailey @GoFroggyRun

martinholmer:
Another non-theoretical reason why I think indexing the lump-sum tax amount by MARS is not a good idea, is that it will make the finding of `LST` values that "finances" the tax reform much more difficult.  This current specification involves changing a single-valued parameter and my experiece this afternoon with this single-valued `LST` parameter is that the iterations to find the `LST` value that finances any given tax reform go very quickly.  It would be much more cumbersome to be adjusting a vector of `LST` values.

@MattHJensen @feenberg @gleiserson @viard @codykallen @andersonfrailey @GoFroggyRun

martinholmer:
The more I think about pull request #1066, the more I think the option to assign `LST` to each filing unit is not a useful option.  The reason is that a lump-sum tax of that kind gives adults who are married filing jointly an incentive to break up.  Can anybody think of a situation in which the lump-sum tax should be levied on a filing unit (instead of on the taxpayer and spouse, if present, in a filing unit)?  If we can't think of a good reason to specify this kind of filing-unit-based lump-sum tax, then this pull request can be simplified by dropping the `LST_per_indiv` parameter (since it would always be true).  Thoughts?

@MattHJensen @feenberg @gleiserson @viard @codykallen @andersonfrailey @GoFroggyRun

gleiserson:
Agree. Per person and per adult (i.e. imposing some kind of age restriction) seem like the most useful approaches.

Also, with the caveat that I am unable to upload/download/read most files on github so this comment is based on the snippet in Martin’s earlier comment, there is potentially an issue related to dependent returns. Ideally, you would want the lump-sum tax to apply consistently to (for example) a 16 year-old who files and appears as a dependent and a 16 year-old who appears as a dependent but does not file. Since this is partially a question about what data is fed into the program, there isn’t necessarily a perfect solution. The cleanest might be to exclude dependent filers and impose it on dependents (possibly subject to an age restriction).

feenberg:
On Tue, 15 Nov 2016, Martin Holmer wrote:

> 
> @MattHJensen asked:
>
>       Rather than adding [the] _LST_per_adult [parameter], why not index _LST by MARS?
> 
> Because we want the lump-sum tax to be non-distortionary.
> 
> It seem to me that indexing the lump-sum tax by MARS will create unwanted
> incentives to change filing status if the lump-sum tax amounts are other than
> the way we have them now in pull request #1066.
>

If the LST is post-credits and post AMT, then it won't affect itemization 
status or AMT status, and there will be no incentive affects. If at some 
point in the future we model marital decisions we could revisit this, but 
I am not sure what is meant by "index _LST by MARS".

dan

martinholmer:
Thanks for all the thought-provoking comments on pull request #1066, which adds a lump-sum tax to Tax-Calculator.

One line of discussion was about how the lump-sum tax interacts with the income tax.  Of course, the intent is to have as little interaction as possible: ideally, no interaction.  The original version of #1066 had a poor placement of the lump-sum tax calculation in the sequence of Tax-Calculator logic.  This has been corrected in commit 937ef85, which does two things.  First, the calculation of the lump-sum tax has been put in its own function called `LumpSumTax` and that function is called after **all** income and payroll tax calculation are completed (including the BenefitLimitation, BenefitsSurtax, and FairShareTax calculations).  And second, the lumpsum_tax amount computed in the `LumpSumTax` function has been added to the `_combined` tax variable, but not included in either the `_iitax` or `_payrolltax` variables.

A second line of discussion concerns exactly who should have to pay the lump-sum tax.  There seems to be agreement that the lump-sum-tax eligibility rule should be as non-distortionary as possible and be practical given the data on filing units that are available to Tax-Calculator.  A major issue is how to treat older dependents especially if they file their own tax return.  The `DSI` variable in the IRS-SOI PUF has a value of one for filers who are claimed as a dependent on another tax return.  To help us reach some kind of resolution of this second issue, I've included below some tabulations of the current `puf.csv` file, which includes filers from the 2009 IRS-SOI PUF in addition to non-filer records constructed from CPS data.  I'll post another comment to #1066 that contains my own thoughts on how best to specify the lump-sum-tax eligibility rule, but I welcome suggestions from others based on the tabulations below.

@MattHJensen @feenberg @gleiserson @viard @codykallen @andersonfrailey @GoFroggyRun

```

$ ../csvvars puf.csv | grep MARS
8 MARS
$ ../csvvars puf.csv | grep DSI
3 DSI

$ awk -F, 'NR>1{n++}END{print n}' puf.csv
219814

$ awk -F, 'NR>1{n[$8]++}END{for(i in n)print i,n[i]}' puf.csv
1  85765
2 109265
3   4289
4  20495

$ awk -F, 'NR>1{n[$3]++}END{for(i in n)print i,n[i]}' puf.csv
0 211425
1   8389

$ awk -F, 'NR>1 && $3==1{n[$8]++}END{for(i in n)print i,n[i]}' puf.csv
1 8375
2    9
3    5
```
So, virtually all the `DSI==1` filers are single.
How old are these `DSI==1` filers?
```
$ awk -F, 'NR>1 && $3==1 && $8==1{n[$1]++}END{for(i in n)print i,n[i]}' puf.csv | sort
10 211
11 202
12 195
13 190
14 188
15 219
16 205
17 194
18 898
19 779
20 682
21 668
22 670
23 690
24 736
25 689
26 685
27 28
28 27
29 20
30 20
31 35
32 22
33 24
34 23
35 22
36 28
38 2
39 2
41 1
43 1
45 2
46 2
47 2
48 1
50 2
52 1
53 1
55 1
56 1
62 1
74 1
75 1
76 1
85 2

$ awk -F, 'NR>1 && $3==1 && $8==2{n[$1]+=1}END{for(i in n)print i,n[i]}' puf.csv | sort
23 1
26 3
27 1
28 1
30 1
35 1
36 1

$ awk -F, 'NR>1 && $3==1 && $8==3{n[$1]+=1}END{for(i in n)print i,n[i]}' puf.csv | sort
19 1
21 1
22 1
23 1
25 1
```

martinholmer:
@gleiserson said:
>> The more I think about pull request #1066, the more I think the option to assign LST to each filing unit is not a useful option.
>
> Agree. Per person and per adult (i.e. imposing some kind of age restriction) seem like the most useful approaches.

The problem with a lump-sum tax "per adult" is that it relies on dependent age data that is often censored in the IRS-SOI PUF, especially for the oldest dependents.  So, I'm reluctant to pursue that approach.

@gleiserson also said after identifying the dependent filer issue:
> The cleanest [approach] might be to exclude dependent filers [those with DSI=1] and impose it on dependents (possibly subject to an age restriction) [as well as the taxpayer and spouse, if present].

As mentioned above the age restriction part of "dependents (possibly subject to an age restriction)" does not seem feasible given the state of the IRS-SOI PUF data.

Given all this, I now think the best (but not perfect) approach is to levy the lump-sum tax on each individual no matter what their age.  In order to avoid double taxation, the lump-sum tax would not be levied on filing units who are dependents on another filer's return.  I'm going to revise pull request #1066 to implement this approach.

@MattHJensen @feenberg @gleiserson @viard @codykallen @andersonfrailey @GoFroggyRun

martinholmer:
After commit 97aac4d, the lump-sum tax is being computed as shown below.
(Note that the `_num` variable has a value of 2 when MARS==2 and a value of 1 otherwise; it is being used in case the Tax-Calculator input dataset has no XTOT variable.)
```
def LumpSumTax(DSI, _num, XTOT,
               LST,
               lumpsum_tax, _combined):
    """
    Compute lump-sum tax and add it to combined taxes.
    """
    if LST == 0.0 or DSI == 1:
        lumpsum_tax = 0.
    else:
        lumpsum_tax = LST * max(_num, XTOT)
    _combined += lumpsum_tax
    return (lumpsum_tax, _combined)
```
Comments?

@MattHJensen @feenberg @gleiserson @viard @codykallen @andersonfrailey @GoFroggyRun

martinholmer:
Using the proposed lump-sum-tax code in pull request #1066, here is an example of how a tax reform could be "financed" in two different ways.  The first way is to use lump-sum taxes to offset the revenue effect of the reform; the second way is to change tax rates to offset the revenue effects of the reform.  We pick a very simple reform in order to keep the example as simple as possible.  And we consider only two ways of financing the reform even though there are many other ways of doing this.

The example reform is to eliminate the OASDI maximum taxable earnings so that OASDI payroll taxes have the same tax base as Medicare (HI) payroll taxes.  This reform subjects all earnings to the 0.124 OASDI payroll tax rate.  I present the numerical results first and then at the bottom show the Python script used to produce the numerical results.
```

$ python reform-financing.py
Reform trt=[0.124]
Reform LST=[0.0]
Aggregate 2013 Tax Revenues ($billion)
                 base    reform      diff
payroll_tax     919.6    1050.7    131.06
iincome_tax    1192.6    1190.9     -1.63
combinedtax    2112.2    2241.6    129.43
```
The above results are static estimates of the reform's revenue impact without any financing.

The results below show how linear interpolation was used to quickly find the size of the lump-sum tax that "finances" this reform.
```
$ python reform-financing.py
Reform trt=[0.124]
Reform LST=[-900.0]
Aggregate 2013 Tax Revenues ($billion)
                 base    reform      diff
payroll_tax     919.6    1050.7    131.06
iincome_tax    1192.6    1190.9     -1.63
combinedtax    2112.2    1961.8   -150.38

$ python reform-financing.py
Reform trt=[0.124]
Reform LST=[-416.31]
Aggregate 2013 Tax Revenues ($billion)
                 base    reform      diff
payroll_tax     919.6    1050.7    131.06
iincome_tax    1192.6    1190.9     -1.63
combinedtax    2112.2    2112.2     -0.00
```
Notice that the -416.31 is computed using an HP-12C hand calculator by linear interpolation between the combined revenue estimates when LST is zero and -900.  This is quick and requires only two addition runs (one with LST at -900 and the other with LST at -416.31) to get the fully-financed results.  Notice that the aggregate lump-sum tax is equal to about $129.4 billion.

Now consider financing this reform by changing the OASDI payroll tax rate.  Several linear interpolations produce the following fully-financed results.
```
$ python reform-financing.py
Reform trt=[0.104886]
Reform LST=[0.0]
Aggregate 2013 Tax Revenues ($billion)
                 base    reform      diff
payroll_tax     919.6     920.7      1.04
iincome_tax    1192.6    1191.5     -1.03
combinedtax    2112.2    2112.2      0.00
```
So, to offset the reform's revenue effects the OASDI payroll tax rate would have to be lowered by roughly 1.91 percentage points, from 12.40 percent to about 10.49 percent.

And finally the `reform-financing.py` script:
```
$ cat reform-financing.py

import numpy as np
import pandas as pd
from taxcalc import Policy, Records, Calculator, weighted_sum

# specify baseline and reform Calculator objects and calculate results
calc1 = Calculator(policy=Policy(), records=Records(), verbose=False)
calc2 = Calculator(policy=Policy(), records=Records(), verbose=False)
reform = {
    2013: {"_SS_Earnings_c": [9e99],
           "_FICA_ss_trt": [0.124],
           "_LST": [0.0]}
}
# .......... trt: current-law 0.124 ; full-financing 0.104886
# .......... LST: current-law 0.0   ; full-financing -416.31
calc2.policy.implement_reform(reform)
calc1.calc_all()
calc2.calc_all()

# tabulate/print aggregate 2013 tax revenues for baseline and reform
record_columns = ['s006', '_payrolltax', '_iitax', '_combined']
out = [getattr(calc1.records, col) for col in record_columns]
df1 = pd.DataFrame(data=np.column_stack(out), columns=record_columns)
out = [getattr(calc2.records, col) for col in record_columns]
df2 = pd.DataFrame(data=np.column_stack(out), columns=record_columns)
print 'Reform trt={}'.format(reform[2013]['_FICA_ss_trt'])
print 'Reform LST={}'.format(reform[2013]['_LST'])
print 'Aggregate 2013 Tax Revenues ($billion)'
print '                 base    reform      diff'
resstr = '{}  {:8.1f}  {:8.1f}  {:8.2f}'
baseln = weighted_sum(df1, '_payrolltax') * 1.0e-9
reform = weighted_sum(df2, '_payrolltax') * 1.0e-9
print resstr.format('payroll_tax', baseln, reform, reform - baseln)
baseln = weighted_sum(df1, '_iitax') * 1.0e-9
reform = weighted_sum(df2, '_iitax') * 1.0e-9
print resstr.format('iincome_tax', baseln, reform, reform - baseln)
baseln = weighted_sum(df1, '_combined') * 1.0e-9
reform = weighted_sum(df2, '_combined') * 1.0e-9
print resstr.format('combinedtax', baseln, reform, reform - baseln)
```

@MattHJensen @feenberg @gleiserson @viard @codykallen @andersonfrailey @GoFroggyRun

martinholmer:
Merging pull request #1066 (which adds the calculation of a lump-sum tax) with the idea that, if the need arises, the lump-sum tax logic will be easy to revise.

PR #1060, opened 2016-11-14
Addition totals calculated for dropq calculation

talumbau:
- Instead of just a summary calculation of the delta between the policy
   and reform for total individual income tax, total payroll tax, and
   total combined tax for each budget year, we also compute these values
   for baseline and reform separately.

PR #1067, opened 2016-11-17
Avoid double-precision rounding error in dropq test

martinholmer:
This pull request makes no substantive changes in `test_dropq.py`, but does scale back expectations about the equality of two large floating-point numbers in a test that was added in pull request #1060.  That new test expected two numbers that are conceptually equal to be the same down to the third decimal place.  The new test passed on one local computer, but failed on another local computer (as described in the #1060 conversation).  Travis-CI does not run this test because the test uses the confidential `puf.csv` file.  But the two numbers being compared are large: about 1.081e11.  So the original test would fail if the two numbers were (for example):
```
108,100,000,000.001 and
108,100,000,000.002
```
The expectation was for 15 digits of accuracy, which is at the outer limits of double-precision accuracy.  This pull request simply scales back expectations for how close two large floating-point numbers must be to be considered equal.

@talumbau

### DIFF
--- a/taxcalc/tests/test_dropq.py
+++ b/taxcalc/tests/test_dropq.py
@@ -1,15 +1,14 @@
 import os
+import json
 import numpy as np
-import pandas as pd
-import pytest
 import numpy.testing as npt
+import pandas as pd
 from pandas import DataFrame, Series
-
+import pytest
 from taxcalc.dropq.dropq_utils import *
 from taxcalc.dropq import *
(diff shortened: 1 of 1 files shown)


martinholmer:
@talumbau,  Is commit 80de40e in pull request #1067 more like what you want?

talumbau:
Yes, I think that's better. The default relative tolerance in `assert_allclose` is 1e-7 which seems good for this use case.

talumbau:
+1

Issue #1068, opened 2016-11-17
Implementing Ryan-Brady plan


codykallen:
To implement the Ryan-Brady plan in the Tax-Calculator, we need to make two modifications: excluding part of only certain types of investment income (qdiv, ltcg and interest income), and a 500 credit for all dependents that functions with the child tax credit. 

The tax [blueprint](http://abetterway.speaker.gov/_assets/pdf/ABetterWay-Tax-PolicyPaper.pdf) calls for excluding half of investment income, and taxing the remainder as ordinary income, but the Blueprint is not clear on exactly what types of income this applies to. If we assume that this applies to all investment income, then the current setup of ALD_Investment_ec_rt is sufficient. Alex Brill decided to apply this only to ltcg, qdiv and interest, which requires a different structure of the CapGains function than what we currently have. We could implement this by modifying the application of ALD_Investment_ec_rt or by created a new exclusion parameter to apply only to ltcg, qdiv and interest. The latter approach should arguably include an assert statement to ensure that one does not apply ALD_Investment_ec_rt and the more limited exclusion in the same reform. 

The $500 credit for dependents is nonrefundable and is structured with the child tax credit. The plan technically calls for an increase in the CTC by $500 (the extra $500 is nonrefundable) and for a $500 nonrefundable credit for non-child dependents. We assumed that this should have the same phase-out structure as the CTC, but with the beginning of the phase-out set to begin when the phase-out of the CTC ends. I implemented it by hijacking the personal_credit, but a permanent solution should have its own credit (dependent_credit). We could implement it in two possible ways: implementing it as its own credit with the phase-out described above (I already have the code for it), or by adding it to the ChildTaxCredit function (prectc += dependent_credit) and then subtracting it from the refundable portion in the AdditionalCTC function (more complicated). 

Does anyone have any thoughts on the investment income exclusion portion? I already have the code to implement these changes, but I'd like feedback on the investment income exclusion portion. 

@martinholmer @MattHJensen @andersonfrailey @Amy-Xu @feenberg @GoFroggyRun

feenberg:
There is much "investment income" in partnerships and schedule c, but also 
much earned income. How do we distinguish those?

dan

codykallen:
Income from partnerships and Sch C gets taxed separately, with a maximum rate of 25%.

codykallen:
The $500 nonrefundable credit for dependents is implemented in PR #1069.

martinholmer:
[...]

PR #1069, opened 2016-11-17
Nonrefundable credit for dependents

codykallen:
Per the Ryan-Brady plan, this is structured to phase-out as though added to the Child Tax Credit. For example, for a head of household with one child, the $1000 CTC would phase out between $75k and $95k, and the additional $500 dependent credit would phase out at the same rate between $95k and $105k. For a head of household with two children, the $2000 CTC would phase out between $75k and $115k, and the $1000 of dependent credit would phase out between $115k and $135k. 

I used the existing parameters for the Child Tax Credit because that is the structure of this credit. Alternatively, we could add a phase-out rate and threshold specific to this credit, but implementation of the Ryan-Brady plan would require setting these parameters to the same as those for the CTC.

This pull request resolves the second part of issue #1068 

@MattHJensen @martinholmer @andersonfrailey @GoFroggyRun @feenberg @Amy-Xu

### DIFF
--- a/taxcalc/current_law_policy.json
+++ b/taxcalc/current_law_policy.json
@@ -670,7 +670,7 @@
         "validations": {"min": 0}
     },
 
-    "_DependentCredit_c":{
+    "_DependentCredit_c": {
 	"long_name": "Nonrefundable credit for dependents",
 	"description": "This nonrefundable credit is applied to dependents and phases out with the Child Tax Credit. The dependent credit's phaseout begins when the CTC's phaseout ends.",
         "irs_ref": "",
@@ -681,7 +681,7 @@
         "row_label": ["2013"],
         "cpi_inflated": false,
         "col_label": "",
-        "value": [0],
+        "value": [0]
     },
 
     "_II_rt1": {
--- a/taxcalc/current_law_policy.json
+++ b/taxcalc/current_law_policy.json
@@ -670,6 +670,20 @@
         "validations": {"min": 0}
     },
 
+    "_DependentCredit_c":{
+	"long_name": "Nonrefundable credit for dependents",
+	"description": "This nonrefundable credit is applied to dependents and phases out with the Child Tax Credit. The dependent credit's phaseout begins when the CTC's phaseout ends.",
+        "irs_ref": "",
+        "notes": "The Ryan-Brady tax plan sets this at $500",
+        "start_year": 2013,
+        "col_var": "MARS",
+        "row_var": "FLPDYR",
+        "row_label": ["2013"],
+        "cpi_inflated": false,
+        "col_label": "",
+        "value": [0.0],
+    },
+
     "_II_rt1": {
         "long_name": "Personal income tax rate 1",
         "description": "The lowest tax rate, applied to the portion of taxable income below tax bracket 1. ",
--- a/taxcalc/current_law_policy.json
+++ b/taxcalc/current_law_policy.json
@@ -676,12 +676,12 @@
         "irs_ref": "",
         "notes": "The Ryan-Brady tax plan sets this at $500",
         "start_year": 2013,
-        "col_var": "MARS",
+        "col_var": "",
         "row_var": "FLPDYR",
         "row_label": ["2013"],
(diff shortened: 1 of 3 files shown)


martinholmer:
@codykallen,  I don't understand why the new dependent credit is computed in the `AGI` function.  How is it related to the calculation of AGI?  If it is so closely related to the current-law CTC (with respect to its phase out), then why not move the code that calculates the new dependent credit to the `ChildTaxCredit` function?  That would seem to make more sense, but maybe I'm missing something.

@MattHJensen @feenberg

codykallen:
@martinholmer 
I put it in the AGI function because the personal credit is calculated in the AGI function (which I had previously used to implement the credit). If you prefer, I can easily switch the calculation to the ChildTaxCredit function.

martinholmer:
@codykallen said:
> I put it in the AGI function because the personal credit is calculated in the AGI function (which I had previously used to implement the credit). If you prefer, I can easily switch the calculation to the ChildTaxCredit function.

OK, I see your thought process.  But yes, I think it would be a big improvement to move the calculation of this new dependent credit to the `ChildTaxCredit` function.  Thanks for doing this.

codykallen:
@martinholmer 
I have switched the calculation to the ChildTaxCredit function.

martinholmer:
@codykallen,  Thanks.  Moving the calculation of the new dependent credit into the `ChildTaxCredit` function is a big improvement.  Now we can focus on the details of the calculation.  Here is what you're proposing:
```
    # calculate dependent credit
    if MARS == 2:
        dep_count = XTOT - 2
    else:
        dep_count = XTOT - 1
    dep_credit = DependentCredit_c * dep_count
    # phase-out dependent credit
    if CTC_prt > 0. and c00100 > CTC_ps[MARS - 1]:
        thresh = CTC_ps[MARS - 1] + n24 * CTC_c / CTC_prt
        excess = c00100 - thresh
        if _exact == 1:
            excess = 1000. * math.ceil(excess / 1000.)
        dep_phaseout = dep_phaseout = CTC_prt * (c00100 - excess)
        dep_credit = max(0., dep_credit - dep_phaseout)
```
All this looks good after a quick glance except for the first if-else statement.  There is one inelegance and one problem with this:
```
    if MARS == 2:
        dep_count = XTOT - 2
    else:
        dep_count = XTOT - 1
```
The inelegance is that we already have a calculated variable called `_num` that has a value of 2 for MARS==2 and a value of 1 for other MARS values.  It is defined in the `records.py` file as follows:
```
        self._num[:] = np.where(self.MARS == 2, 2, 1)
```

But more importantly, there is a substantive problem here.  This code will produce negative `dep_count` values for many filing units with DSI==1.  What does the Blueprint say about this reform provision?  Are filing units who are dependents on another return really going  to be able to get this new dependent tax credit?  That would seem pretty weird, but there are plenty of weird things in publicly-released tax reform proposals.  If DSI==1 makes you ineligible for this new dependent credit, then you need to build that logic into your proposed code.  If these dependent filers are eligible for the new dependent credit, then you need to prevent the value of `dep_count` from going negative.  Of the 219,814 records in `puf.csv`, 8,389 will have a negative value for `dep_count` in your proposed code.

codykallen:
@martinholmer 
Thanks for catching the DSI special case. I modified the code to:
dep_credit = DependentCredit_c * max(0, dep_count)

martinholmer:
@codykallen,  Thanks for that code correction, so that we don't calculate any negative credits.
Now your code is (in part) like this:
```
    # calculate dependent credit
    if MARS == 2:
        dep_count = XTOT - 2
    else:
        dep_count = XTOT - 1
    dep_credit = DependentCredit_c * max(0, dep_count)
    ...
```
Why not simplify to the following code?
```
    # calculate dependent credit
    dep_credit = DependentCredit_c * max(0, XTOT - _num)
    ...
```
Sometimes it really helps to use if-else statements in the function.py code, but I don't think is one of those cases.

martinholmer:
@codykallen,  Thanks for your contribution in #1069.  I'm merging it now.

PR #1084, opened 2016-12-02
add dollar limit on itemized deductions

codykallen:
This PR adds the ability to limit the amount of itemized deductions taken by a filer. The Trump plan sets this at $200,000 for married filing jointly and $100,000 for all others.

### DIFF
--- a/taxcalc/current_law_policy.json
+++ b/taxcalc/current_law_policy.json
@@ -356,7 +356,7 @@
         "value": [0.8]
     },
 
-    "_ID_cap": {
+    "_ID_c": {
         "long_name": "Itemized deduction cap",
         "description": "The amount of itemized deductions is capped at this dollar amount. ",
         "irs_ref": "",
--- a/taxcalc/current_law_policy.json
+++ b/taxcalc/current_law_policy.json
@@ -356,6 +356,26 @@
         "value": [0.8]
     },
 
+    "_ID_cap": {
+        "long_name": "Itemized deduction cap",
+        "description": "The amount of itemized deductions is capped at this dollar amount. ",
+        "irs_ref": "",
+        "notes": "",
+        "start_year": 2013,
+        "col_var": "MARS",
+        "row_var": "FLPDYR",
+        "row_label": ["2013",
+                      "2014",
+                      "2015",
+                      "2016"],
+        "cpi_inflated": true,
+        "col_label": ["single", "joint", "separate", "head of household", "widow", "separate"],
+        "value": [[1e99, 1e99, 1e99, 1e99, 1e99, 1e99],
+                  [1e99, 1e99, 1e99, 1e99, 1e99, 1e99],
+                  [1e99, 1e99, 1e99, 1e99, 1e99, 1e99],
+                  [1e99, 1e99, 1e99, 1e99, 1e99, 1e99]]
+    },
+
     "_ID_prt": {
         "long_name": "Itemized deduction phaseout rate (Pease)",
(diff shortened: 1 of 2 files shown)


codykallen:
@MattHJensen @martinholmer @andersonfrailey @Amy-Xu @feenberg

codykallen:
Never mind. It seems that this can be done using ID_BenefitSurtax. Closing.

MattHJensen:
> Never mind. It seems that this can be done using ID_BenefitSurtax. Closing.

@codykallen, I'm going through current_law_policy.json right now and realize that I'm a bit confused about your comment that you can limit the amount of itemized deductions to 100/200k using the ID_BenefitSurtax. 

My understanding of the logic in ID_BenefitSurtax is that you can limit the _benefit_ that accrues from itemized deductions to 100/200k using by setting the exemption at 100/200, the surtax rate at 1, and the credit rate at 0. I don't see how you can limit the amount of itemized deductions taken.

codykallen:
@MattHJensen, You are correct that it limits the benefit but not the actual amount of itemized deductions. From a scoring perspective, these are equivalent. However, this is not true if you intend to analyze the impact of capping itemized deductions on the actual deductions (beyond just scoring it). The implementation in this PR is more useful for the kind of analysis I was doing, but I closed the PR upon finding that the cap can be closely modeled by existing parameters.

MattHJensen:
> From a scoring perspective, these are equivalent. 

Could you explain that further? Here is how I've been thinking about this: 

Let's say my marginal rate is 30% and I have $100 of itemized deductions:

If we limit the "benefit" of itemized deductions to $50, then my tax liability is unchanged because the benefit of itemized deductions to me is only $30 to begin with. 

If we limit the "amount" of itemized deductions to $50, then my tax liability goes up by roughly $15 dollars because I lose $50 of itemized deductions at a 30% rate.

feenberg:
On Thu, 15 Dec 2016, Matt Jensen wrote:

>
>       From a scoring perspective, these are equivalent.
> 
> Could you explain that further? Here is how I've been thinking about this:
> 
> Let's say my marginal rate is 30% and I have $100 of itemized deductions:
> 
> If we limit the "benefit" of itemized deductions to $50, then my tax liability is
> unchanged because the benefit of itemized deductions to me is only $30 to begin with.
> 
> If we limit the "amount" of itemized deductions to $50, then my tax liability goes up
> by roughly $15 dollars because I lose $50 of itemized deductions at a 30% rate.

I agree, but it isn't more complicated then that - won't the benefit and 
cap parameters interact differently with the standard deduction, the 
AMT and the tax table? Also, if the benefit is post-credit then there will 
be a difference in credits.

dan

codykallen:
@MattHJensen, You're right, there is a difference. How did you model Trump's proposed deduction cap?

MattHJensen:
I haven't yet for publication. I tried to right when the proposal came out but misinterpreted the policy as limiting to taxpayers with income under 100/200. We can do that with the itemized deduction limitation parameters. But then jim tankersley (v. impressively) caught my error when I asked him to look through a reform file.

codykallen:
@MattHJensen, this PR should model it correctly. Do you want me to reopen it?

MattHJensen:
> @MattHJensen, this PR should model it correctly. Do you want me to reopen it?

Just did.

taxcalc/current_law_policy.json
```diff
@@ -356,6 +356,26 @@
         "value": [0.8]
     },
 
+    "_ID_c": {
+        "long_name": "Itemized deduction cap",
```
MattHJensen: I believe this ought to be "Itemized deduction ceiling".

MattHJensen:
This LGTM. Thanks @codykallen.

PR #1092, opened 2016-12-08
Fix #1091 by removing erroneous kwa delim_whitespace

### LINKED ISSUE #1091: Tests fail on pandas 0.19.1
`test_dropq.test_ascii_output_function()` calls `test_utils.ascii_output()`, which calls: 

```
.to_csv(ascii_filename, header=False, index=False, delim_whitespace=True, sep='\t')
```

on a pandas dataframe. This method doesn't support the `delim_whitespace` keyword arg, but normally doesn't raise an error here since it accepts extra keyword args. This changed in https://github.com/pandas-dev/pandas/pull/13804, so our code raises:

```
TypeError: to_csv() got an unexpected keyword argument 'delim_whitespace'
```

with the newest version of pandas. 

Fortunately, the fix is easy since this keyword argument is superfluous in this context. When reading CSV files, it provides faster parsing when the `sep` argument is ommitted, but it has no meaning when writing since the separator must be explicitly stated.

zrisher:
Fixes #1091 by removing erroneous keyword arg `delim_whitespace` in `pd.to_csv()` usage

### DIFF
--- a/taxcalc/utils.py
+++ b/taxcalc/utils.py
@@ -681,8 +681,7 @@ def ascii_output(csv_filename, ascii_filename):
     fstring = '{:' + str(col_size) + '}'
     out = out.applymap(fstring.format)
     # write ascii output to specified ascii_filename
-    out.to_csv(ascii_filename, header=False, index=False,
-               delim_whitespace=True, sep='\t')
+    out.to_csv(ascii_filename, header=False, index=False, sep='\t')
 
 
 def mtr_graph_data(calc1, calc2,


zrisher:
@martinholmer @talumbau

taxcalc/utils.py
```diff
@@ -681,8 +681,7 @@ def pdf_recid(recid):
     fstring = '{:' + str(col_size) + '}'
     out = out.applymap(fstring.format)
     # write ascii output to specified ascii_filename
-    out.to_csv(ascii_filename, header=False, index=False,
-               delim_whitespace=True, sep='\t')
+    out.to_csv(ascii_filename, header=False, index=False, sep='\t')
```
talumbau: yep. Bitten by the pandas philosophy "hmm, don't know that keyword arg, it's probably fine." :sad:

talumbau:
+1

MattHJensen:
Thanks @zrisher!

PR #1094, opened 2016-12-08
Fix #1093 by moving requirements into environment.yml

### LINKED ISSUE #1093: Conda dependency constraints overwritten by Pip requirements
In Conda's env config file `environment.yml` we specify the versions of `numpy` and `pandas` that Conda should install:

```
- numpy =1.10.4
- pandas =0.18.0
```

These versions were specified in #669 to fix issues encountered with older versions of those packages.

However, in Pip's env config file `requirements.txt` we say it may install the current default version of `pandas` and any `numpy` >= 1.7:

```
pandas
numpy >= 1.7
```

Pip determines what versions to fetch without knowing the versions specified in the Conda config.
When #669 was submitted, this had no effect because we were specifying the newest versions in `environment.yml`. However, there are now newer versions of each package, and these lines in `requirements.txt` render ineffective our controls in `environment.yml`. 

Some tests fail with the newest version of `pandas`, so this makes the initial environment unusable for a new contributor.

zrisher:
To fix #1093:
* I removed the contradicting `pandas` and `numpy` lines in `requirements.txt`. 
* I then found that `numba == 0.26.0` also implicity declares `numpy>=1.7`. Numba was also in the `environment.yml`, but without a version constraint, so I moved this there.
* `six` and `toolz` were declared in both places, so I removed them

That made `requirements.txt` empty, so I removed it. Happy to keep it in if preferred!

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,0 @@
-pandas
-numpy >= 1.7
-numba
-six
-toolz


talumbau:
Ah, I see you did remove the requirements.txt. Ok, very good. Let's try it this way and see what breaks. I think the only possible issue will be what happens when I try to upload a new release to PyPI (https://pypi.python.org/pypi/taxcalc). If this breaks, I will find a way to fix it that does not involve creating a requirements.txt, since that eventually leads to dependencies specified in two places, which is bad.

zrisher:
>I think the only possible issue will be what happens when I try to upload a new release to PyPI (https://pypi.python.org/pypi/taxcalc). If this breaks, I will find a way to fix it

I would really appreciate that. I have no idea how the environment is created when this is built as a package. I imagine it has something to do with this line in `setup.py`:

```
 'install_requires': ['numpy', 'pandas'],
```

The Travis environment doesn't use `environment.yml` either, but rather requirements entered directly via the command line:

```
 - conda create -n taxcalcdev python=$TRAVIS_PYTHON_VERSION pytest setuptools pandas=0.18 toolz six mock bokeh
```

It would be awesome if we could unify the environments for local users, package users, and testing at some point.

zrisher:
Just pushed a quick fix - the `numba == 0.26.0` requirement was in that file from my local windows dev branch and accidently slipped into this PR. I removed it, so actually no changes to the `environment.yml` now, just the removal of `requirements.txt`.

MattHJensen:
@talumbau, could you merge this if it looks good to you?

talumbau:
looks good. merging.

Issue #1095, opened 2016-12-08
Creating a new Conda env fails on Windows


zrisher:
If a Windows user builds a new Conda environment, they are presented a version of pandas greater than the specified working version, which makes tests fail. The reason for this is covered in #1093 and fixed by #1094.

However, when the specified versions of `pandas` and `numpy` are loaded via Conda on Windows, they no longer produce working versions. Both packages will fail to import:

```
(taxcalc-dev) c:\Sites\ospc\tax-calculator>conda list numpy
# packages in environment at C:\Dev\Miniconda3\envs\taxcalc-dev:
#
Using Anaconda Cloud api site https://api.anaconda.org
numpy                     1.10.4                   py35_2

(taxcalc-dev) c:\Sites\ospc\tax-calculator>conda list pandas
# packages in environment at C:\Dev\Miniconda3\envs\taxcalc-dev:
#
Using Anaconda Cloud api site https://api.anaconda.org
pandas 
                   0.18.0              np110py35_0
(taxcalc-dev) c:\Sites\ospc\tax-calculator>python
Python 3.5.2 |Continuum Analytics, Inc.| (default, Jul  5 2016, 11:41:13) [MSC v.1900 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Dev\Miniconda3\envs\taxcalc-dev\lib\site-packages\numpy\__init__.py", line 112, in <module>
    from ._globals import ModuleDeprecationWarning, VisibleDeprecationWarning
ImportError: No module named 'numpy._globals'
>>> import pandas
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Dev\Miniconda3\envs\taxcalc-dev\lib\site-packages\pandas\__init__.py", line 37, in <module>
    import pandas.core.config_init
  File "C:\Dev\Miniconda3\envs\taxcalc-dev\lib\site-packages\pandas\core\config_init.py", line 14, in <module>
    import pandas.core.config as cf
AttributeError: module 'pandas' has no attribute 'core'
```

This appears similar to the issues I experienced with conda builds of these packages in the past on Windows, i.e. #669. Previously, updating the included package version fixed the issue, and indeed that is true here: If I update pandas and numpy to their latest versions, I'm able to import both successfully.

[...]

Issue #1086, opened 2016-12-05
Introductory py.test failing


salimfurth:
The test of tax calculator, as suggested on the Contributor Guide, is giving 4 failures in the up-to-date tax calculator. Both @kdd0211 and I encountered this problem today, and we verified it on 3 separate systems. On all 3 systems, the tests give 4 failures.

Here's the output from @kdd0211's interface:

MINGW64 ~/Desktop/tax-calculator/taxcalc (master)
$ py.test -m "not requires_pufcsv"

============================= test session starts =============================
platform win32 -- Python 3.5.2, pytest-2.8.5, py-1.4.31, pluggy-0.3.1
rootdir: C:\Users\kdd0211\Desktop\tax-calculator, inifile: setup.cfg
collected 298 items

tests\test_behavior.py .......
tests\test_calculate.py ..................
tests\test_consumption.py .......
tests\test_decorators.py ...............F
tests\test_dropq.py ...................F.F..........
tests\test_functions.py ....
tests\test_growth.py ......
tests\test_incometaxio.py ..............
tests\test_macro_elasticity.py .
tests\test_parameters.py .
tests\test_policy.py ..........................
tests\test_records.py ..............
tests\test_reforms.py .
tests\test_simpletaxio.py ...................
tests\test_utils.py ...................................F......
tests\filings\forms\test_tax_form.py .............
tests\filings\forms\test_tax_forms.py ..........................................                                                                                                                                  ................................

================================== FAILURES ===================================
_____________________________ test_force_no_numba _____________________________

    def test_force_no_numba():
        """
        Force execution of code for non-existence of Numba
        """
        global Magic_calc6

        # Mock the numba module
>       from mock import Mock
E       ImportError: No module named 'mock'

tests\test_decorators.py:296: ImportError
___________________________ test_create_json_table ____________________________

    def test_create_json_table():
        df = DataFrame(data=[[1., 2, 3], [4, 5, 6], [7, 8, 9]],
                       columns=['a', 'b', 'c'])
>       ans = create_json_table(df)

tests\test_dropq.py:342:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
dropq\dropq_utils.py:67: in create_json_table
    row_out.append(format_print(df.loc[idx, col], _type, num_decimals))
[...]

PR #418, opened 2015-10-24
Fix Travis CI Python version for tests

talumbau:
- create the conda environment from TRAVIS_PYTHON_VERSION

PR #1100, opened 2016-12-10
Make Travis CI test env come from environment.yml

talumbau:


talumbau:
This seems like a trivial PR, but it is actually the proper solution for the less than ideal fix from #418. In #418, the .travis.yml was modified to no longer create the test environment from the environment.yml file because it was discovered that every test environment (Python 2.7, Python 3.4, and Python 3.5) were all actually running the tests with Python 3.5! This was happening silently, so the only way the problem was discovered was when a change was made that should have failed on Python 2.7 but Travis CI still passed. So, in #418 we started creating the test environment directly using the `TRAVIS_PYTHON_VERSION` and then explicitly specifying the required packages. This led to duplicate package specification in the environment.yml and the .travis.yml. A colleague recently had this same issue and told me the correct fix: remove `python` from the environment.yml only and create the environment from the `TRAVIS_PYTHON_VERSION`. So, now we test with exactly what is in the environment.yml. This is one step on the path towards a solution to the issues that @zrisher and @salimfurth have found (e.g. #1086, #1095). I think the next step will be to add AppVeyor.

talumbau:
If there are no additional comments, I'd like to merge in this change sometime tomorrow (Wednesday)

zrisher:
Looks great, big :+1:

zrisher:
@talumbau I think this is a very positive change, thanks for taking care of it.

PR #1104, opened 2016-12-13
EMTR on charitable contributions

codykallen:
This PR adds the ability to calculate the EMTR on charitable contributions. Since this is negative, this should be interpreted as the effective marginal subsidy on charitable contributions. 

Due to limited data on non-itemizers, this MTR should only be considered valid when called on a current law calculator or a reform that reduces or limits itemized deductions. It should also be considered valid only for small values of finite_diff (e.g. the default value of finite_diff of 0.01). 

This MTR is useful for estimating the impact of a reform on charitable giving. A reform that limits itemized deductions also reduces the marginal subsidy for charitable contributions, causing individuals to reduce charitable giving.

### DIFF
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -250,7 +250,7 @@ class Calculator(object):
                            'e02000', 'e02400',
                            'p22250', 'p23250',
                            'e18500', 'e19200',
-                           'e26270']
+                           'e26270', 'e19800']
 
     def mtr(self, variable_str='e00200p',
             negative_finite_diff=False,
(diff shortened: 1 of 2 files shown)


martinholmer:
@codykallen,  Thanks for your MTR contribution in pull request #1104.

As I remember from the last time we added a variable to the `MTR_VALID_VARIABLES` list, the addition of a variable automatically generates a new MTR histogram for that new variable in the `pufcsv_mtr_expect.txt` file.  Can you run the full set of unit tests (including those marked `requires_pufcsv`) to see if the new histogram for variable `e19800` looks the way you would expect?  If it does (and none of the other MTR results change), then add the new version of the `pufcsv_mtr_expect.txt` file to this pull request.

martinholmer:
@codykallen,  Thanks for the contribution.

PR #1111, opened 2016-12-20
Add Appveyor scripts

talumbau:


martinholmer:
@talumbau,  Are the Travis-CI processes and the AppVeyor process going to run simultaneously?  I certainly hope so because the check-in testing already takes too long.

talumbau:
yes they are different services provided by different companies. They are both free for open source projects.

martinholmer:
@talumbau said:
> yes they are different services provided by different companies. They are both free for open source projects.

I don't really care about the term of use.  My question is this: when I submit a pull request are the three Travis-CI processes and the AppVeyor process(es) going to start at the same time and run in parallel?

talumbau:
Github will notify both services at essentially the same time. Both services support the idea of "run these scripts on a virtual machine." Both services, therefore, need to be assigned a cloud instance for running the tests. Appveyor uses cloud machines provided by MS Azure. Travis CI uses cloud machines provided by AWS. If one service is experiencing a much heavier load than the other, they will not start at the same time, because one service will receive its virtual machine sooner than the other. There is no coordination between the services.

talumbau:
Resolves #1086 via https://github.com/open-source-economics/Tax-Calculator/pull/1111/commits/dced001fa1d882d38ebeed85c499d009ecadec6e

zrisher:
The fix for #1086 via dced001 looks great, 👍 .

The provision of automated testing for Windows would also be awesome. Would it be possible, though, to use the dependencies declared in `environment.yml` instead of declaring them within the build script itself, e.g. `continuous_integration/setup_conda_environment.cmd:L16-26`? Otherwise we're introducing for Windows the same problem you fixed for Linux in #1100.

talumbau:
> @talumbau The provision of automated testing for Windows would also be awesome. Would it be possible, though, to use the dependencies declared in environment.yml instead of declaring them within the build script itself, e.g. continuous_integration/setup_conda_environment.cmd:L16-26? Otherwise we're introducing for Windows the same problem you fixed for Linux in #1100.

Yes I agree. This was mostly a copy/paste job from an existing project. I think I can do that. If it is more difficult than it seems, I'll update here and see if doing that in a follow-up PR would be ok.

martinholmer:
@talumbau,  When do you expect to merge pull request #1111?

talumbau:
I will try to do as @zrisher suggests (create environment purely from the `environment.yml`) some time today. If that is successful, then hopefully this can get merged today. If it's nontrivial, other's can comment if they would like to wait for the proper fix, or would rather we merge this and work on a follow-up after that.

talumbau:
@zrisher have a look at https://github.com/open-source-economics/Tax-Calculator/pull/1111/commits/649d3957700595c7efe6ea6a02409281f0f373ad

I think that makes it so that on both AppVeyor and Travis CI, the config is done from environment.yml. If you are satisfied, I think we can merge.

continuous_integration/setup_conda_environment.cmd
```diff
@@ -15,17 +15,14 @@ conda remove --all -q -y -n %CONDA_ENV%
...
+conda create -n %CONDA_ENV% -q -y python=%PYTHON%
 
 call activate %CONDA_ENV%
 
-@rem Install optional dependencies for tests
-%CONDA_INSTALL% numpy=%NUMPY% pandas=%PANDAS% 
-%CONDA_INSTALL% numba bokeh=0.12.3 toolz six mock pep8 pylint
+%CONDA% env update -f environment.yml
 
 %PIP_INSTALL% pytest-pep8
```
zrisher: @talumbau This is handled by `environment.yml`:

```
- pip:
  - pytest-pep8
```

So we can remove it here, correct?

continuous_integration/setup_conda_environment.cmd
```diff
@@ -15,17 +15,14 @@ conda remove --all -q -y -n %CONDA_ENV%
...
 
-if %PYTHON% LSS 3.0 (%PIP_INSTALL% git+https://github.com/Blosc/castra)
 if %PYTHON% LSS 3.0 (%PIP_INSTALL% backports.lzma mock)
```
zrisher: @talumbau `mock` is installed via `environment.yml`. Do we need `backports.lzma` for Python < 3.0 on Windows? If so, should we include it in the environment file?

talumbau:
@zrisher I removed the additional external dependencies that you found, thanks! I think this is good to go.

zrisher:
Nice! :+1:

PR #1114, opened 2016-12-31
update numpy to 1.11.2 for windows compat

zrisher:
Solves #1095

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -25,7 +25,7 @@ init:
 install:
   # Update to a known good conda
   # (to workaround http://help.appveyor.com/discussions/problems/4910)
-  - conda install -q -y conda=4.2.9
+  - conda install -q -y conda=4.2.13
   - continuous_integration\\setup_conda_environment.cmd
 
 build_script:
--- a/environment.yml
+++ b/environment.yml
@@ -2,7 +2,7 @@ name: taxcalc-dev
 dependencies:
 - pytest
 - setuptools
-- numpy =1.10.4
+- numpy =1.11.2
 - pandas =0.18.0
 - numba
 - bokeh =0.12.3


zrisher:
It seems the appveyor tests failed because the new version of `numpy` triggered `conda-env` to rely on an unresolved dependency on `pyyaml`. Per https://github.com/ContinuumIO/anaconda-issues/issues/1205 we can fix this by upgrading to `conda=4.2.12`. It didn't occur on my machine because I already had `pyyaml` in my root environment.

Adding a new commit to upgrade the AppVeyor conda.

talumbau:
Very cool - the one line change in the environment.yml changed the numpy version in all of the Travis builds and on AppVeyor. D.R.Y. ftw! This looks good to me, and it now looks like we have moved to numpy 1.11. This is probably a necessary step for going to pandas 0.19, which I don't think is a rush for this project but can happen after we merge this. 
+1

talumbau:
One thing I just thought of: after we merge this in, most people doing development work won't know to update to numpy 1.11, so their dev environment will stay at numpy 1.10. We can likely maintain backwards compatibility with numpy 1.10 for a good while, but it might be overkill to keep testing with numpy 1.10 and numpy 1.11. I would prefer just to have the project on numpy 1.11 and move forward. This would be a good piece of info for a mailing list of some kind, just telling anyone who is currently developing for Tax-Calculator to `conda update numpy=1.11`. cc @MattHJensen

feenberg:
On Sat, 31 Dec 2016, T.J. Alumbaugh wrote:

> 
> One thing I just thought of: after we merge this in, most people doing development
> work won't know to update to numpy 1.11, so their dev environment will stay at numpy
> 1.10. We can likely maintain backwards compatibility with numpy 1.10 for a good while,
> but it might be overkill to keep testing with numpy 1.10 and numpy 1.11. I would
> prefer just to have the project on numpy 1.11 and move forward. This would be a good
> piece of info for a mailing list of some kind, just telling anyone who is currently
> developing for Tax-Calculator to conda update numpy=1.11. cc @MattHJensen

If we include code that requires 1.11 then government users who have 
persauded their IT group to install 1.10 will have to repeat that 
possibily difficult task. If they fail, they will continue to use the 
older version of the calculator past its prime. More likely, they will 
simply give up on using the project at all.

Does 1.11 bring something so valuable to the project that it is worth 
leaving out executive agency users? I say this because that will be the 
end result of issuing additional requirements with each release. So I 
would hope that "a good while" means more than "while we are paying 
attention". Presumably we can avoid uing new capabilities if we chose.

In the executive branch users are generally not allowed to install 
software on their machines. For good or bad, that is the rule. Once IT is 
persuaded to install software, updates are not automatic, but depend on 
will or whim of IT staff.

I know we have congressional users - I believe each office is independent 
and this is less of a problem.

Government users are especially desirable, as they are best able to lobby 
SOI for improvements to the data.

dan

zrisher:
@feenberg said:
> If we include code that requires 1.11 then government users who have
persauded their IT group to install 1.10 will have to repeat that
possibily difficult task. If they fail, they will continue to use the
older version of the calculator past its prime. More likely, they will
simply give up on using the project at all.

You could say the same thing about updating any package dependency. I don't know about your standard python project beyond TC work, but in Ruby and Node apps dependency versions change pretty frequently. Upstream code advances and users of the old versions often get left behind. If we can't do the same because we're worried about a large subset of users being unable to update their environments, that's quite a handicap.

Don't executive agency users already have to upgrade taxcalc itself when versions advance? Doesn't that step involve updating the environment?

feenberg:
On Sat, 31 Dec 2016, zrisher wrote:

> 
> @feenberg said:
>
>       If we include code that requires 1.11 then government users who have
>       persauded their IT group to install 1.10 will have to repeat that
>       possibily difficult task. If they fail, they will continue to use the
>       older version of the calculator past its prime. More likely, they will
>       simply give up on using the project at all.
> 
> You could say the same thing about updating any package dependency. I don't know about your standard python project beyond TC
> work, but in Ruby and Node apps dependency versions change pretty frequently. Upstream code advances and users of the old
> versions often get left behind. If we can't do the same because we're worried about a large subset of users being unable to
> update their environments, that's quite a handicap.


You haven't said what particular loss in capability will occur if 
features from 1.11 are used that are not in 1.10. If they are
significant, there is a serious decision to make. If they are
insignificant then why impose the restriction?
> 
> Don't executive agency users already have to upgrade taxcalc itself when versions advance? Doesn't that step involve updating the
> environment?
>
That is an interesting observation. taxsim.for and taxcalc.sas are 
imported as source code, and in most agencies I haVE dealt with that needs 
no approval for their Unix compatible environments, as far as can tell. 
BThe fortran compiler does require approval but it is usually already 
present. My code will compile in any fortran compiler, I don't specify a 
version. It would be a huge hurdle to get a different compiler at an 
agency that had one.

Note that at SOI fortran is forbidden, so I ran taxsim though f2c and gave 
them a C version of taxsim. At the BLS it is allowed by special permission 
(which must be renewed each year). Every year the BLS calls to ask me to 
do the state tax calculator in SAS (which is allowed) but I haven't done 
that because of the work involved.

A user downloading python source code for Anaconda, and compiling that 
locally, might be able to leverage one approved Python compiler into a 
working Anaconda system. But how many non-Anaconda systems will compile 
Anaconda? And it might be difficult in Windows, where some agencies will 
allow only whitelisted .exe files.

Again, the first question to answer is - what feature of 1.11 is so 
attractive that it is worth forcing the upgrade?

I can ask another question - do we have any executive agency users at all 
yet? Maybe my thoughts have no force because we don't have any affected 
users yet.

dan

MattHJensen:
@feenberg asked

> I can ask another question - do we have any executive agency users at all
yet? Maybe my thoughts have no force because we don't have any affected
users yet.

All executive branch users who I am aware of use Tax-Calculator via TaxBrain and therefore would not be affected when we update package dependencies. I suggest we keep moving forward with updates until it becomes a problem for someone.

As an aside, I don't expect us to have many agency users of the tax-calculator package itself until we support the IRS INSOLE full file (which we will do as soon as we have a publicly distributable variable list) _or_ until we support state taxes.

martinholmer:
Merged pull request #1114 so that Windows `py.test` works on GitHub when creating or updating a pull request.  @zrisher, thanks for this contribution.

@MattHJensen @feenberg @talumbau @zrisher

PR #1121, opened 2017-01-06
zero defaults for ALD_Dependent params

MattHJensen:
The Trump campaign proposal had an above the line deduction for dependents. Currently in Tax-Calculator, the current law values are set to the levels in the Trump proposal, and they are turned off with a haircut parameter. This means that a reform script won't show any changes to those parameters, just to the haircut. For the sake of clarity, this PR sets the values to zero so that the ALD_Dependent parameters can be set explicitly in a reform script.

### DIFF
--- a/taxcalc/current_law_policy.json
+++ b/taxcalc/current_law_policy.json
@@ -283,12 +283,12 @@
         "row_var": "",
         "row_label": ["2013"],
         "cpi_inflated": false,
-        "value": [1.0]
+        "value": [0.0]
     },
 
     "_ALD_Dependents_Child_c": {
         "long_name": "National average childcare costs. Ceiling for available childcare deduction.",
-        "description": "The weighted average of childcare costs in the US",
+        "description": "The weighted average of childcare costs in the US. 7165 is the weighted average from the 'Child Care in America: 2016 State Fact Sheets'",
         "section_1": "Above the line deductions",
         "section_2": "Child and elderly care",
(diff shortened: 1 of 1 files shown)


feenberg:
On Fri, 6 Jan 2017, Matt Jensen wrote:

> 
> The Trump campaign proposal had an above the line deduction for dependents.

What is the difference between that and a personal exemption?

dan

feenberg:
Silly me, the proposed exemption won't apply to the taxpayer and spouse.

dan

MattHJensen:
@feenberg said:

> the proposed exemption won't apply to the taxpayer and spouse.

That's correct. Here's what the [campaign document](https://www.donaldjtrump.com/policies/tax-plan) says.  

> Americans will be able to take an above-the-line deduction for children under age 13 that will be capped at state average for age of child, and for eldercare for a dependent. The exclusion will not be available to taxpayers with total income over $500,000 Married-Joint /$250,000 Single, and because of the cap on the size of the benefit, working and middle class families will see the largest percentage reduction in their taxable income.
> 
> The childcare exclusion would be provided to families who use stay-at-home parents or grandparents as well as those who use paid caregivers, and would be limited to 4 children per taxpayer. The eldercare exclusion would be capped at $5,000 per year. The cap would increase each year at the rate of inflation.

MattHJensen:
Does not affect results. Merging.

PR #1132, opened 2017-01-07
Fix elasticity of GDP sims from file reforms

talumbau:
- Take user specified elasticity of GDP and add to reform dictionary

 - Resolves https://github.com/OpenSourcePolicyCenter/webapp-public/issues/441


talumbau:
merging soon if no comments

PR #1109, opened 2016-12-16
Reorganize parameters -- will facilitate TaxBrain updates

MattHJensen:
In #1074, I described several disconnects between TaxBrain and current_law_policy.json:

> 
> 1. Sections and subsections
> - TB is organized by sections and subsections, (e.g Refundable Credits: EITC), whereas `current_law_policy.json` parameters are only informally grouped into sections by reference in the title name. 
> 
> 2. Order of the parameters. 
>   - TB is ordered roughly by where the params appear in the tax forms, while TC's `current_law_policy.json` is roughly ordered by where the parameters appear in `taxcalc/functions.py.`
> 
> 3. "Principle" parameters 
> - The TB input page does not include every parameter from `current_law_policy.json`, rather it only includes parameters of principle importance. 
> 
> 4. 1-2 relationships between TB fields and TC params
> - Two sets of fields on the TaxBrain input page have a 1-2 relationship with parameters in Tax-Calculator. 
>     - The personal income tax rate and brackets fields on TaxBrain govern both `II_rt`s and `II_brk`s as well as `PT_rt`s and `PT_brk`s
>     - the long term capital gains and qualified dividends fields on TaxBrain govern both `CG_rt`s and `CG_bk`s as well as `AMT_CG_rt`s and `AMT_CG_brk`s.
> 

This PR resolves those disconnects by 

1. Adding `section_1`, `section_2`, and in some cases `section_3` fields that specify the conceptual relationship among parameters and therefore how they ought to be grouped on TaxBrain. 

2. Sorting the parameters in current_law_policy.json to be easier to use.  (I have not written down/formalized this schema, but I hope it will be fairly easy to stick to by looking at the examples). 

3. Specifying which parameters *should not* be included in TaxBrain by leaving off the section fields. My intent was to include every parameter on TaxBrain, but some have not received any attention for a long time, and I figure they might need to be dusted off. 

4. Updating the names and descriptions of parameters such as II_rts and PT_rts to make their relationships more obvious and thereby allow us to do away with the 1-2 relationships between TaxBrain inputs and Tax-Calculator parameters.

[...]

PR #1134, opened 2017-01-09
cleanup _STD

MattHJensen:
Before #1109, the tax logic in `functions.py` for the maximum standard deduction for dependents was controlled by _STD[6]. 

In #1109, I added a new parameter to `current_law_params.json`, `_STD_Dep`, and refactored `functions.py` to use the new parameter. 

I forgot to remove _STD[6] from `current_law.params.json`. 

This PR fixes that.


MattHJensen:
No comments, so merging.

PR #1147, opened 2017-01-18
Read from JSON file using OrderedDict

talumbau:
- This guarantees that we preserve the order of key/value pairs as
   defined in the JSON files themselves.

### DIFF
--- a/taxcalc/parameters.py
+++ b/taxcalc/parameters.py
@@ -5,6 +5,7 @@ import os
 import json
 import numpy as np
 from abc import ABCMeta
+from collections import OrderedDict
 
 
 class ParametersBase(object):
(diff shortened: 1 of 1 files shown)


brittainhard:
@talumbau this is good. Not sure what the effect on the rest of the library will be, though.

MattHJensen:
+1 here.

MattHJensen:
@brittainhard @talumbau, I expect we'll be releasing a new tax-calculator today. Would you like this to be included?

talumbau:
Yep. Merging. Thanks for the heads up!

MattHJensen:
@talumbau, did you mean to merge?

talumbau:
yes thanks!

Issue #1142, opened 2017-01-13
Some Items in current_law_policy.json Lack Section Fields


brittainhard:
The fields without section names are:

- `_AMT_thd_MarriedS`
- `_ALD_Investment_ec_base_code_active`
- `_AMT_Child_em`
- `_LLC_Expense_c`
- `_ETC_pe_Married`
- `_AMT_em_pe`
- `_AMT_KT_c_Age`
- `_ETC_pe_Single`

@talumbau @MattHJensen I'm relying on thee section names to organize the UI, so this fields will have to be left out if they do not have section names.

brittainhard:
I've also noticed that some of the names for sections are not consistent, i.e. `Payroll Taxes` and `Payroll taxes`. I can work around this problem for now.

MattHJensen:
@brittainhard, leaving out params w o section labels is the right step for now. See #1141 for more info.

MattHJensen:
@brittainhard, I resolved the payroll taxes inconsistency w #1154. Since everything else is the same as #1141, I am closing this in favor of #1141.

PR #1154, opened 2017-01-24
Payroll Taxes -> Payroll taxes

MattHJensen:
fixes part of #1142. cc @brittainhard.


MattHJensen:
@brittainhard said: 

> I've also noticed that some of the names for sections are not consistent, i.e. Payroll Taxes and Payroll taxes. I can work around this problem for now.

This fixes that problem.

Merging.

PR #1167, opened 2017-01-31
Fix real GDP growth rates

martinholmer:
This pull request fixes the real GDP growth rates in the `growth.json` and `growth.py` files so that they  are consistent with the rates implied by the blowup factors in the `StageIfactors.csv` file.  The rates have changed slightly because of a correction to the test that compares these hard-coded rates to the implied rates.  That correction was introduced in pull request #1164.

These changes have no effect on any `requires_pufcsv` test results, validation test results, or reform comparison test results, because none of those tests use the Growth class to replace the default real GDP growth rates.

@MattHJensen @andersonfrailey @codykallen


codykallen:
The new REAL_GDP_GROWTH_RATES continue to differ substantially and consistently from the CBO projections (and the actual numbers), with all of the rates in REAL_GDP_GROWTH_RATES less than CBO projections except for 2013. 

If we look at the cumulative effect, the growth rates in REAL_GDP_GROWTH_RATES project that the economy in 2026 would be 24% larger than in 2013, whereas the CBO growth rates project cumulative growth by 2026 of 32% (August 2016 report) or 31% (January 2017 report).

I'm not sure why this PR reduces the values in REAL_GDP_GROWTH_RATES, when the current values are already generally lower than CBO rates.

martinholmer:
@codykallen said in PR#1167:
> The new REAL_GDP_GROWTH_RATES continue to differ substantially and consistently from the CBO projections (and the actual numbers), with all of the rates in REAL_GDP_GROWTH_RATES less than CBO projections except for 2013.
>
> If we look at the cumulative effect, the growth rates in REAL_GDP_GROWTH_RATES project that the economy in 2026 would be 24% larger than in 2013, whereas the CBO growth rates project cumulative growth by 2026 of 32% (August 2016 report) or 31% (January 2017 report).
>
> I'm not sure why this PR reduces the values in REAL_GDP_GROWTH_RATES, when the current values are already generally lower than CBO rates.

This PR "reduces the values in REAL_GDP_GROWTH_RATES" in order to be consistent with the blowup factors in the `StageIfactors.csv` file.  If the new `Growth.REAL_GDP_GROWTH_RATES` are different from CBO's then one of two things is true: either (1) I made a mistake in deriving the implied `REAL_GDP_GROWTH_RATES` from the `StageIfactors.csv` data, or (2) the `StageIfactors.csv` data are not consistent with CBO's projection.  Or perhaps there is another reason.  Can anybody suggest a third reason?

Can you see a problem in [how I derived the implied `Growth.REAL_GDP_GROWTH_RATES` here](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/tests/test_records.py#L173-L212)?

@MattHJensen @andersonfrailey

codykallen:
@martinholmer, I've finally found the cause of the discrepancy between the growth rates in REAL_GDP_GROWTH_RATES and the rates in CBO's economic projections. In StageIFactors.csv, AGDPN is the aggregate nominal GDP index. The growth rates in this index are nearly identical to those in CBO's economic projections. The difference is in the calculation of real GDP growth rates. CBO does this using inflation rates as measured by the GDP price deflator (which has a long-term rate of 2.0%), but we do it using the CPI (long-term growth rate of 2.4%). 

The use of the CPI to determine real growth rates instead of the GDP deflator is the proximate cause of the consistent difference between REAL_GDP_GROWTH_RATES and CBO's real GDP growth rate projections. I believe the best way to deal with this is to change the description of factor_target in growth.json to reflect that these are the real GDP growth rates implied from nominal rates using CPI-based inflation instead of chain-linked real growth rates.

martinholmer:
@codykallen said in #1167:
> I've finally found the cause of the discrepancy between the growth rates in `Growth.REAL_GDP_GROWTH_RATES` and the rates in CBO's economic projections. In the file `StageIFactors.csv`, AGDPN is the aggregate nominal GDP index. The growth rates in this index are nearly identical to those in CBO's economic projections. The difference is in the calculation of real GDP growth rates. CBO does this using inflation rates as measured by the GDP price deflator (which has a long-term rate of 2.0%), but we do it using the CPI (long-term growth rate of 2.4%).
>
> The use of the CPI to determine real growth rates instead of the GDP deflator is the proximate cause of the consistent difference between `REAL_GDP_GROWTH_RATES` and CBO's real GDP growth rate projections.  I believe the best way to deal with this is to change the description of factor_target in `growth.json` to reflect that these are the real GDP growth rates implied from nominal rates using CPI-based inflation instead of chain-linked real growth rates.

Thank you for solving this puzzle!  I agree that the solution is a documentation change in the `growth.json` file.  Does anybody disagree?  It seems to me that there is little or no use of this particular Growth class capability, so doing something more involved seems like overkill.

@MattHJensen @andersonfrailey @feenberg @Amy-Xu

MattHJensen:
> Thank you for solving this puzzle! 

Agreed!

> I agree that the solution is a documentation change in the growth.json file.

Agreed.

martinholmer:
Newest commit to pull request #1167 includes revised documentation in the `growth.json` file.
Any suggested revisions would be appreciated.

@codykallen @andersonfrailey @MattHJensen @Amy-Xu

codykallen:
@martinholmer, the revised documentation in the growth.json file looks good to me.

MattHJensen:
Back in https://github.com/open-source-economics/Tax-Calculator/pull/368#issuecomment-144751408, I said:

> one of the blowup factors we have is GDP growth. That GDP number was taken from a CBO document where we also took many other blowup factors (including, for example, wage growth), so, presumably, the ratio between wage growth and GDP growth is meaningful. Factor_target() takes advantage of this. We allow the user to say what they think the GDP growth rate is, and then adjust the GDP blowup factor to match that number and adjust all of the other blowup factors (including wage growth) to maintain the ratio between GDP growth and the other factor's growth.
> 
> So in summation, both ways of changing the baseline dataset provided by this PR will end up affecting wage growth. One method changes wage growth (and blowup factors) directly, the other method changes wage growth (and other blowup factors) to maintain a relationship between those factors and the GDP growth factor.
> 

As @martinholmer pointed out on the phone today, this logic isn't particularly economically literate. If growth is different than predicted, we shouldn't expect the other factors to move in lock step. In fact, we should expect some of them to move in contrary motion. While I wish this critique were constrained to factor target, it seems to also apply to factor adjustment as well. 

I wonder if we should instead just expose all of the factors to the user, and allow the user to adjust them at will. If the user wants to adjust them in lock step, that would be up to the user.

martinholmer:
@codykallen and @MattHJensen, Thanks for your comments.  Is seems as if revisiting the Growth class has raised several substantive issues.  I don't think we are going to resolve all these issues in this one pull request.  How about this as a course of action: we merge #1167 as it is now (with better documentation and a consistent test) and pursue the remaining substantive issues in other issues and pull requests.  Does this make sense?

@MattHJensen @codykallen @andersonfrailey @Amy-Xu

MattHJensen:
>  I don't think we are going to resolve all these issues in this one pull request. How about this as a course of action: we merge #1167 as it is now (with better documentation and a consistent test) and pursue the remaining substantive issues in other issues and pull requests. Does this make sense?

That makes sense to me.

martinholmer:
@MattHJensen said:
> That makes sense to me.

Merging pull request #1167.

PR #1192, opened 2017-02-13
hardcode 9e99 and 1e99 -> 9e99

### LINKED ISSUE #1191: Values in Current Law Policy
@MattHJensen @PeterDSteinberg 

I'm working on the problem that causes the output of the `_AGI_surtax_thd` paramater to look like this: 
screen shot 2017-02-13 at 11 39 00 am

The surtax parameter only has values for one year, so the [`expand_2D`](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/parameters.py#L406) function adds in the missing values up to 2017. The problem is that inflation rates get applied to the `9e99` values, and you end up with strange numbers in the UI. If you are using the app and set the year to 2013, then the default inputs look fine, but they get larger as you increase the year setting, which is why you get values like `9.71309658528e+99`.

Peter and I talked it out, and we both think that values ending in `e99` are arbitrary placeholders, so the main problem here is fixing how these values appear on the frontend. Is it the case that these are arbritrary placeholders?

If so, there are a couple of options for fixing this. First, we could prevent inflation rates from being applied to numbers like `1e99` or `9e99` by adding some logic in the function. Second, we could just update `current_law_policy.json` and give it values for the missing years for `_AGI_surtax_thd`.

Let me know what you think, or if you need more clarification.

MattHJensen:
Resolves #1191 

cc @PeterDSteinberg @britainhard


martinholmer:
Merging pull request #1192 which resolves issue #1191.

PR #1194, opened 2017-02-14
New grow logic -- Step 4 of 4

martinholmer:
This pull request is step 4 in a four-step redesign of the Growth class.

For **BACKGROUND** and **REDESIGN** see the pull request #1188.

**THIS PR**

This pull request adds a new Growdiff class that is derived from the ParametersBase class and uses two instances of that class (growdiff_baseline and growdiff_response) to specify changes in the default values in the `growfactors.csv` file.

While taxcalc logic is working fine after this redesign of the Growth class, the logic being used currently in `dropq.py` (which is not part of the taxcalc package) is not general enough to support two instances of the same class (or even the same parameter name in two different parameter classes).  @PeterDSteinberg, you can see my diagnosis of the problem by searching the `dropq.py` file for `PROBLEM`.  Maybe there is an easy fix to this problem, but it did not immediately come to my mind.  Also, note that the two extra uncovered statements in `dropq.py` are not possible to test because of the nature of the temporary work around this PROBLEM.

@MattHJensen @feenberg @Amy-Xu @andersonfrailey @zrisher


PeterDSteinberg:
@martinholmer Regarding the `PROBLEM` lines in `dropq.py`, one idea is:
* Make an extra keyword argument to `Growdiff` class [here](https://github.com/martinholmer/Tax-Calculator/blob/38f1ee97c8e2ee48070eaa5a4e709f9bf2eea1d5/taxcalc/growdiff.py#L47) such as `baseline_or_reform`.
* Move [this logic from `dropq.py`](https://github.com/martinholmer/Tax-Calculator/blob/38f1ee97c8e2ee48070eaa5a4e709f9bf2eea1d5/taxcalc/dropq/dropq.py#L453-L456) to inside the `Growdiff` class where it processes the `baseline_or_reform` argument.  
* Note [this line](https://github.com/martinholmer/Tax-Calculator/blob/38f1ee97c8e2ee48070eaa5a4e709f9bf2eea1d5/taxcalc/dropq/dropq.py#L455) is not needed - `.update` will never do anything as its argument here is an empty `dict`.

martinholmer:
@PeterDSteinberg said:
> Note this line [the last line in the following code fragment] is not needed - .update_growdiff will never do anything as its argument here is an empty dict.
```
# Specify growdiff_baseline and growdiff_response
    growdiff_baseline = Growdiff()
    growdiff_response = Growdiff()
    # PROBLEM: dropq makes no distinction between the two Growdiff instances
    # PROBLEM: code here is using growdiff_baseline assumptions
    # PROBLEM: code here is missing growdiff_response assumptions
    growdiff_base_assumps = only_growdiff_assumptions(user_mods, start_year)
    growdiff_resp_assumps = dict()  # PROBLEM: this is temporary "fix"
    growdiff_baseline.update_growdiff(growdiff_base_assumps)
    growdiff_response.update_growdiff(growdiff_resp_assumps)
```
Yes, you are correct that `growdiff_resp_assumps` is now an empty dictionary, but that is only because current dropq logic cannot support duplicate parameter names.  When that PROBLEM is fixed, the growdiff_resp_assumps dictionary will often be non-empty and it will need to be used to update the growdiff_response object.  That is why that statement in included in pull request #1194.

PeterDSteinberg:
@martinholmer Good point (`growth_resp_assumps` is not always going to be empty).  Still I think this logic should be handled within the `Growdiff` class as an keyword argument.

martinholmer:
@PeterDSteinberg suggested this solution to the dropq logic PROBLEM mentioned in PR#1194:
> Make an extra keyword argument to Growdiff class here such as baseline_or_reform.
>
> Move the following logic from `dropq.py` to inside the Growdiff class where it processes the baseline_or_reform argument.
```
    growdiff_resp_assumps = dict()  # PROBLEM: this is temporary "fix"
    growdiff_baseline.update_growdiff(growdiff_base_assumps)
    growdiff_response.update_growdiff(growdiff_resp_assumps)
```
This kind of Growdiff class change would leave it with a logical structure that varies substantially from the structure of all the other classes derived from the ParameterBase class. So, in your suggestion, where does the Growdiff constructor get growdiff_base_assumps and growdiff_resp_assumps?  Those are typically being read from a JSON-formatted assumption file and then being converted into a dictionary by other functions.  None of the other classes derived from the ParameterBase class are working the way you suggest.

The real problem is with dropq logic.  Specifically, the problem is that all the dropq functions accept an argument called user_mods, which throws away critical information based on the assumption that the thrown-away information can be reconstructed later because parameter names are unique.  Here is a typical use of a dropq function taken from the
`test_dropq.py` file:
```
def test_run_dropq_nth_year_from_file(is_strict, puf_1991_path,
                                      reform_file, assump_file):

    user_reform = Calculator.read_json_param_files(reform_file.name,
                                                   assump_file.name)
    user_mods = user_reform

    # Create a Public Use File object
    tax_data = pd.read_csv(puf_1991_path)
    first = 2016
    rjson = True

    (_, _, _, _, _, _, _, _,
     _, _, fiscal_tots) = dropq.run_models(tax_data,
                                           start_year=first,
                                           is_strict=is_strict,
                                           user_mods=user_mods,
                                           return_json=rjson,
                                           num_years=3)

    assert fiscal_tots is not None
```
The `read_json_param_files` function returns a tuple of five dictionaries, which are not kept separate when `dropq.run_models` is called.  The method of untangling this undifferentiated mass of data in dropq works as long as the individual parameters in each of those five dictionaries are unique.  That used to be the case by accident.  Now it is no longer true and so dropq does'nt work correctly.

Given that analysis, it seems as if the most direct fix is to change the read_json_param_files function return a dictionary of the five dictionaries and pass that dictionary of five dictionaries to the dropq functions.  Then those functions will be able to know easily which dictionary is which without resorting to the convoluted logic in the dropq.only_*_parameters functions.

I'll start implementing that approach in a "Part 4a of 4" pull request built on top of #1194.

@MattHJensen @andersonfrailey

PeterDSteinberg:
Hi @martinholmer - I see your points about complicating the `Growdiff` class.  In general I don't think it is a problem if the various Parameters classes have some specialized behavior, such as additional default data files, but am happy to defer if you want to place the logic elsewhere.  

Regarding your point:
> The read_json_param_files function returns a tuple of five dictionaries, which are not kept separate when dropq.run_models is called.

I agree that the passing of that tuple is cause for confusion.  It makes it less apparent to developers how to get started (e.g. which item is which parameter set?).  I found it confusing when I was modifying the `deploy` repo several weeks ago.  I think moving that information (the tuple of various parameter sets in a specific order) to a single data structure as you describe is a good idea.

martinholmer:
The increase in the number of uncovered statements from 35 to 37 in #1194 will be eliminated in the next pull request in this series: **Step 4a of 4**, which will fix the dropq logic problem in the way agreed to here by @martinholmer and @PeterDSteinberg.

PR #1199, opened 2017-02-16
New grow logic -- Step 4a of 4

martinholmer:
This pull request is step 4a in a four-step redesign of the Growth class.

For **BACKGROUND** and **REDESIGN** see the pull request #1188.

**THIS PR**

This pull request resolves the dropq logic problem discussed in #1194 using the approach agreed to by @PeterDSteinberg and @martinholmer in the #1194 discussion.  The key was to have the `Calculator.read_json_param_files` function return a single dictionary (instead of a tuple of five separate parameter dictionaries) and use that single dictionary in dropq functions.  Given this change, the `dropq.py` logic was able to be simplified: before these changes (in #1194) ` there were 836 lines in `dropq.py`; after the changes in this pull request there are 690 lines in the `dropq.py` file.

@PeterDSteinberg, please note that some of the dropq function names have been changed to make them be more clear about what the functions actually do.  I would imagine the TaxBrain code that calls dropq functions will need to be updated to reflect these new names.  Also, while the taxcalc unit tests for dropq all pass, they are not very strong tests, and so, you will need to test the new dropq functions in your TaxBrain test suite.

@MattHJensen @feenberg @Amy-Xu @andersonfrailey @zrisher


andersonfrailey:
@martinholmer, I've been looking over your code and the logic seems solid to me. I'd say it's ready to merge when you are.

taxcalc/growdiff.json
```diff
@@ -0,0 +1,205 @@
+{
+    "_ABOOK": {
+        "long_name": "ABOOK additive difference from default projection",
+        "description": "Default projection is in growfactors.csv file.  ABOOK extrapolates Records variables: e07300 and e07400.",
```
martinholmer: Yes, changes will be rare.  And when making any change the `growdiff.json` file will have to be edited, so the need to revise the "description" contents will be obvious.

@MattHJensen @andersonfrailey

taxcalc/growdiff.json
```diff
@@ -0,0 +1,205 @@
+{
+    "_ABOOK": {
+        "long_name": "ABOOK additive difference from default projection",
```
martinholmer: Yes, I think additive differences are the best way to express differences in grow factors.  First, this is the way most people talk about differences.  And second, any other way of expressing differences makes it very difficult (and therefore, potentially confusing) to apply **both** baseline and response differences to the grow factors.

@MattHJensen @andersonfrailey

MattHJensen:
Thanks for the explanations, @martinholmer. This series of PRs is a major improvement to the growth capabilities. +1

martinholmer:
@MattHJensen,  Thanks for the review of #1199 and its four predecessors back through #1188.
I'll wait until later today to merge these five PRs just in case there are any other questions or concerns.

Issue #1186, opened 2017-02-08
Why is there an empty taxcalc/conftest.py file?


martinholmer:
I understand the need for the `taxcalc/tests/conftest.py` file, but I don't understand the role of the `taxcalc/conftest.py` file.  Can somebody explain?

martinholmer:
@martinholmer asked:
> I understand the need for the taxcalc/tests/conftest.py file, but I don't understand the role of the `taxcalc/conftest.py` file. Can somebody explain?

@zrisher, Is `taxcalc/conftest.py` part of the useful test enhancements you made months ago?

zrisher:
@martinholmer Yep! Having the `conftest` file at `taxcalc` allows pytest to discover the `taxcalc` package no matter what directory the `pytest` command is run from. 

For example, if I run `pytest taxcalc/tests` from the root repo dir, or any dir above it, or a subdir of `taxcalc/tests`, without this file I'll get "No module named `taxcalc`". If I add the file back, the tests run successfully. That's why [the commit that added this file](https://github.com/open-source-economics/Tax-Calculator/commit/93f05fa470d5dc1a4c12342d09eb72c85cfbdec8) was also able to remove these lines from every test:

```
CUR_PATH = os.path.abspath(os.path.dirname(__file__))
sys.path.append(os.path.join(CUR_PATH, '..', '..', '..', '..'))
```

Perhaps we should add a comment to that effect within the file?

martinholmer:
@zrisher answered the question posed in issue #1186 and then said:
> Perhaps we should add a comment to that effect within the [`taxcalc/conftest.py`] file?

That would be great!  Can you do that when you have a moment?  You could just copy and paste your helpful explanation in #1186 into the file.

zrisher:
@martinholmer Yep, I will submit a PR for that by the weekend.

zrisher:
Implemented in #1201

martinholmer:
Issue #1186 has been resolved by pull request #1201.

PR #1201, opened 2017-02-20
comment taxcalc/conftest per #1186

zrisher:
Adds a comment to ` taxcalc/conftest.py` describing its purpose.


martinholmer:
@zrisher,  Thanks for pull request #1201.  It is a small addition, but is important step in moving toward the goal of helping people understand the role of each file in the Tax-Calculator repository.

PR #1204, opened 2017-02-21
Ryan brady2016

codykallen:
This PR adds a reform file for the Ryan-Brady Better Way tax blueprint. I included documentation on the scored reforms at the top, in the same manner as the Trump2016 reform JSON, and I included a list of unscored reforms at the end. 

@martinholmer, can you tell me if I am using the param_code statement correctly?

@MattHJensen

### DIFF
--- a/taxcalc/current_law_policy.json
+++ b/taxcalc/current_law_policy.json
@@ -273,7 +273,7 @@
         "value": [false]
     },
 
-    "_ALD_InvInc_ec_base_ryanbrady": {
+    "_ALD_InvInc_ec_base_RyanBrady": {
         "long_name": "Investment income base for Ryan-Brady tax blueprint",
         "description": "Exclusion is applied to long-term gains before loss limitation, taxable interest and qualified dividends.",
         "irs_ref": "",
@@ -284,7 +284,7 @@
         "row_label": ["2013"],
         "cpi_inflated": false,
         "col_label": "",
-        "value": [0.0]
+        "value": [false]
     },
 
     "_ALD_Dependents_hc": {
--- a/taxcalc/current_law_policy.json
+++ b/taxcalc/current_law_policy.json
@@ -273,6 +273,20 @@
         "value": [false]
     },
 
+    "_ALD_InvInc_ec_base_ryanbrady": {
+        "long_name": "Investment income base for Ryan-Brady tax blueprint",
+        "description": "Exclusion is applied to long-term gains before loss limitation, taxable interest and qualified dividends.",
+        "irs_ref": "",
+        "notes": "",
+        "start_year": 2013,
+        "col_var": "",
+        "row_var": "FLPDYR",
+        "row_label": ["2013"],
+        "cpi_inflated": false,
+        "col_label": "",
+        "value": [0.0]
+    },
+
     "_ALD_Dependents_hc": {
         "long_name": "Deduction for childcare costs",
         "description": "This decimal fraction, if greater than zero, reduces the portion of childcare costs that can be deducted from AGI",
--- a/taxcalc/functions.py
+++ b/taxcalc/functions.py
@@ -220,7 +220,7 @@ def ALD_InvInc_ec_base_code(calc):
 
 @iterate_jit(nopython=True)
 def CapGains(p23250, p22250, _sep, ALD_StudentLoan_hc,
-             ALD_InvInc_ec_rt, invinc_ec_base,
+             ALD_InvInc_ec_rt, invinc_ec_base, ALD_InvInc_ec_base_RyanBrady,
              e00200, e00300, e00600, e00650, e00700, e00800,
              CG_nodiff, CG_ec, CG_reinvest_ec_rt,
              e00900, e01100, e01200, e01400, e01700, e02000, e02100,
@@ -237,6 +237,11 @@ def CapGains(p23250, p22250, _sep, ALD_StudentLoan_hc,
     invinc = e00300 + e00600 + c01000 + e01100 + e01200
     # compute exclusion of investment income from AGI
     invinc_agi_ec = ALD_InvInc_ec_rt * max(0., invinc_ec_base)
+    # compute exclusion of investment income for Ryan-Brady plan
+    if ALD_InvInc_ec_base_RyanBrady:
+        CG_ec_RyanBrady = (c01000 - max(-3000. / _sep,
+                                        p22250 + ALD_InvInc_ec_rt * p23250))
+        invinc_agi_ec = ALD_InvInc_ec_rt * (e00300 + e00650) + CG_ec_RyanBrady
     # compute ymod1 variable that is included in AGI
     ymod1 = (e00200 + e00700 + e00800 + e00900 + e01400 + e01700 +
              invinc - invinc_agi_ec +
--- a/taxcalc/reforms/RyanBrady.json
+++ b/taxcalc/reforms/RyanBrady.json
@@ -4,7 +4,8 @@
 // Reform_Description:
 // -  New personal income tax schedule (regular/non-AMT/non-pass-through) (1) 
 // -  New pass-through income tax schedule (2) 
-// -  Tax long-term capital gains and qualified dividends as ordinary income, exclude half (3) 
+// -  Tax long-term capital gains and qualified dividends as ordinary income,
+//    - exclude half of interest, qualified dividends, and long-term gain (before loss limitation)(3) 
 // -  Repeal Alternative Minimum Tax (4) 
 // -  Repeal Net Investment Income Tax (5) 
 // -  Raise the Standard Deduction (6) 
@@ -57,15 +58,8 @@
             {"2017": [true]},
         "_ALD_InvInc_ec_rt":
             {"2017": [0.5]},
-        "_ALD_InvInc_ec_base_code_active":
+        "_ALD_InvInc_ec_base_RyanBrady":
             {"2017": [true]},
-        "param_code":
-            {"ALD_InvInc_ec_base_code":
-             ||returned_value = (e00300 + e00650 +
-                                 max(-3000. / _sep, p22250 + p23250) -
-                                 max(-3000. / _sep, p22250))
-             ||
-            },
         "_AMT_rt1":
             {"2017": [0]},
         "_AMT_rt2":
--- a/taxcalc/reforms/RyanBrady.json
+++ b/taxcalc/reforms/RyanBrady.json
@@ -4,7 +4,8 @@
 // Reform_Description:
 // -  New personal income tax schedule (regular/non-AMT/non-pass-through) (1) 
 // -  New pass-through income tax schedule (2) 
-// -  Tax long-term capital gains and qualified dividends as ordinary income, exclude half (3) 
+// -  Tax long-term capital gains and qualified dividends as ordinary income,
+//    - exclude half of interest, qualified dividends, and long-term gain (before loss limitation)(3) 
 // -  Repeal Alternative Minimum Tax (4) 
 // -  Repeal Net Investment Income Tax (5) 
 // -  Raise the Standard Deduction (6) 
@@ -57,10 +58,8 @@
             {"2017": [true]},
         "_ALD_InvInc_ec_rt":
             {"2017": [0.5]},
-        "_ALD_InvInc_ec_base_code_active":
+        "_ALD_InvInc_ec_base_RyanBrady":
             {"2017": [true]},
-        "param_code":
-            {"ALD_InvInc_ec_base_code": ||returned_value = e00300 + e00650 + max(-3000. / _sep, p22250 + p23250) - max(-3000. / _sep, p22250)||},
         "_AMT_rt1":
             {"2017": [0]},
         "_AMT_rt2":
--- a/taxcalc/reforms/RyanBrady.json
+++ b/taxcalc/reforms/RyanBrady.json
@@ -61,7 +61,6 @@
             {"2017": [true]},
         "param_code":
             {"ALD_InvInc_ec_base_code": ||returned_value = e00300 + e00650 + max(-3000. / _sep, p22250 + p23250) - max(-3000. / _sep, p22250)||},
-            //{"ALD_InvInc_ec_base_code": ||returned_value = e00300 + e00650 + p23250||},
         "_AMT_rt1":
             {"2017": [0]},
         "_AMT_rt2":
--- a/taxcalc/reforms/RyanBrady.json
+++ b/taxcalc/reforms/RyanBrady.json
@@ -1,0 +1,101 @@
+// Title: 2016 Better Way Tax Reform Blueprint (Ryan-Brady)
+// Reform_File_Author: Cody Kallen
+// Reform_Reference: http://abetterway.speaker.gov/_assets/pdf/ABetterWay-Tax-PolicyPaper.pdf  
+// Reform_Description:
+// -  New personal income tax schedule (regular/non-AMT/non-pass-through) (1) 
+// -  New pass-through income tax schedule (2) 
+// -  Tax long-term capital gains and qualified dividends as ordinary income, exclude half (3) 
+// -  Repeal Alternative Minimum Tax (4) 
+// -  Repeal Net Investment Income Tax (5) 
+// -  Raise the Standard Deduction (6) 
+// -  Repeal the Personal Exemption (7) 
+// -  New $500 nonrefundable credit, based on Child Tax Credit (8)
+// -  Eliminate marriage penalty in Child Tax Credit phase-out (9)
+// -  Repeal itemized deductions except mortgage interest and charitable contribution (10) 
+// Reform_Parameter_Map:
+// - 1: _II_rt*
+// - 2: _PT_rt*
+// - 3: _CG_nodiff, _ALD_InvInc_ec*
+// - 4: _AMT_*
+// - 5: _NIIT_rt
+// - 6: _STD
+// - 7: _II_em
+// - 8: _DependentCredit_c
+// - 9: _CTC_ps
+// - 10: _ID*
+{
+  "policy": {
+        "_II_rt1":
+            {"2017": [0.12]},
+        "_II_rt2":
+            {"2017": [0.12]},
+        "_II_rt3":
+            {"2017":  [0.25]},
+        "_II_rt4":
+            {"2017":  [0.25]},
+        "_II_rt5":
+            {"2017": [0.33]},
+        "_II_rt6":
+            {"2017": [0.33]},
+        "_II_rt7":
+            {"2017": [0.33]},
+        "_PT_rt1":
+            {"2017": [0.12]},
+        "_PT_rt2":
+            {"2017": [0.12]},
+        "_PT_rt3":
+            {"2017": [0.25]},
+        "_PT_rt4":
+            {"2017": [0.25]},
+        "_PT_rt5":
+            {"2017": [0.25]},
+        "_PT_rt6":
+            {"2017": [0.25]},
+        "_PT_rt7":
+            {"2017": [0.25]},
+        "_CG_nodiff":
+            {"2017": [true]},
+        "_ALD_InvInc_ec_rt":
+            {"2017": [0.5]},
+        "_ALD_InvInc_ec_base_code_active":
+            {"2017": [true]},
+        "param_code":
+            {"ALD_InvInc_ec_base_code": ||returned_value = e00300 + e00650 + max(-3000. / _sep, p22250 + p23250) - max(-3000. / _sep, p22250)||},
+            //{"ALD_InvInc_ec_base_code": ||returned_value = e00300 + e00650 + p23250||},
+        "_AMT_rt1":
+            {"2017": [0]},
+        "_AMT_rt2":
+            {"2017": [0]},
+        "_NIIT_rt":
+            {"2017": [0]},
+        "_STD":
+            {"2017": [[12000, 24000, 12000, 18000, 24000, 12000]]},
+        "_II_em":
+            {"2017": [0]},
+        "_DependentCredit_c":
+            {"2017": [500]},
+        "_CTC_ps":
+            {"2017": [[75000, 150000, 75000, 75000, 75000, 75000]]},
+        "_ID_StateLocalTax_hc":
+            {"2017": [1.0]},
+        "_ID_Medical_hc":
+            {"2017": [1.0]},
+        "_ID_Casualty_hc":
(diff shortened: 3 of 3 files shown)


martinholmer:
@codykallen asked:
> @martinholmer, can you tell me if I am using the param_code statement correctly?

First question to be answered is does the reform execute without error?
What happens when you execute the following command?
```
python inctax.py puf.csv 2018 --aging --reform taxcalc/reforms/RyanBrady.json
```

codykallen:
@martinholmer asked:

> First question to be answered is does the reform execute without error?

Yes, it executes without error.

martinholmer:
@codykallen said:
> Yes, it [that is, the `RyanBrady.json` reform file] executes without error.

Then all seems well, although the readability of the reform file could be
improved by writing the following entry
```
"param_code":
            {"ALD_InvInc_ec_base_code": ||returned_value = e00300 + e00650 + max(-3000. / _sep, p22250 + p23250) - max(-3000. / _sep, p22250)||},
```
with a format like this
```
"param_code":
            {"ALD_InvInc_ec_base_code":
||
returned_value = (e00300 + e00650 +
                  max(-3000. / _sep, p22250 + p23250) -
                  max(-3000. / _sep, p22250))
||
},
```
Then I would strongly recommend that you add a comment (using the `//` notation) to explain this reform provision.  It is not at all clear why the 3000 dollar loss limit is being applied twice.  Is that really what the RyanBrady reform calls for?

codykallen:
@martinholmer asked:

> It is not at all clear why the 3000 dollar loss limit is being applied twice. Is that really what the RyanBrady reform calls for?

The reform provision should exclude half of interest, qualified dividends, and long-term gains. Since it is just a blueprint, the reform does not clearly specify that. However, I believe that it is important to include the effect of the loss limitation, as the reform does not eliminate that. 

Ideally, I would prefer to apply the exclusion to p23250 _before_ the loss limitation, but this is as close as I can reasonably get to that application using ALD_InvInc_base_code. With the implementation in the JSON, the exclusion is applied to the difference between stcg+ltcg after the loss limitation and just stcg after loss limitation.

An alternative implementation, using
`returned_value = e00300 + e00650 + max(-3000. / _sep, p23250)`
would be ignore the interaction between the loss limitation, stcg and ltcg for those in a loss position for either stcg or ltcg.

MattHJensen:
It seems to me that the "param_code" functionality is very useful for allowing users to experiment with reform provisions. Once a prominent, published, reform includes a provision that requires "param_code", though, shouldn't we parameterize the provision so that our most novice tax-calculator and taxbrain users can both use it readily? 

That said, if @codykallen isn't interested in parameterizing this provision, I think we should merge this PR (once the loss limitation is sorted out), and I could follow up with a PR to parameterize the provision and update the json reform file. If he is interested in doing the paramterization, that's great.

martinholmer:
@MattHJensen said:
> It seems to me that the "param_code" functionality is very useful for allowing users to experiment with reform provisions.  Once a prominent, published, reform includes a provision that requires "param_code", though, shouldn't we parameterize the provision so that our most novice Tax-Calculator and TaxBrain users can both use it readily?

I don't understand the phrase "shouldn't we parameterize the provision so that our most novice Tax-Calculator and TaxBrain users can both use it readily?"  I thought the whole point of the pre-prepared JSON reform files was to make it easy to simulate these reforms either with Tax-Calculator on a local computer or with TaxBrain in the cloud.  All any user needs is the reform file; how can it get any easier?  So, unless I'm missing something here, I don't see how parameterizing would simplify things at all; it would just mean that the JSON reform file would need to be edited.

And there is a second important issue with respect to this particular reform: the proposal is not specific enough to parameterize.  This reform proposal, according to @codykallen, simply does not specify all the relevant details.  Doing the work to parameterize this reform provision could end up being a waste of time if we guess wrong on what the missing details are in this proposal.  If we want to parameterize, it seems as if the authors of the reform need to be asked about the missing details.

codykallen:
@MattHJensen said:

> Once a prominent, published, reform includes a provision that requires "param_code", though, shouldn't we parameterize the provision so that our most novice tax-calculator and taxbrain users can both use it readily?

I've just pushed a new commit that parameterizes this and modifies the JSON reform file to use this parameterization. 

@martinholmer noted:

> Doing the work to parameterize this reform provision could end up being a waste of time if we guess wrong on what the missing details are in this proposal. 

If they change the reform provision, then we would change the parameterization as well. 

If you would prefer to do that parameterization instead of here, I can just open a new PR for it.

martinholmer:
@codykallen,  Perhaps I've missed something but the only discussion of the investment income "deduction" in the Better Way document I can find is this sentence:
```
This Blueprint provides for reduced tax on investment income.
Families and individuals will be able to deduct 50 percent of
their net capital gains, dividends, and interest income, leading
to basic rates of 6 percent, 12.5 percent, and 16.5 percent on
such investment income depending on the individual’s tax bracket. 
```
If this is all there is, then you are correct in saying there is an ambiguity.
- Do they mean 50 percent of the sum of "net capital gains, dividends,
and interest income"?  
- Or do they mean the "deduction" is 50 percent
of net capital gains, 50 percent of [qualified?] dividends, plus 50
percent of interest income?

Which interpretation are you making in the new `CG_ec_RyanBrady`
calculation you have added?  Perhaps it is clear to everybody else,
but I need some help bridging the gap between what I read in the
reform proposal and the new code you've added to the CapGains
function.  Perhaps a simple example of a filing unit with both interest
and capital losses would help me understand your proposed code.

codykallen:
@martinholmer, The Blueprint discusses the current law structure:

> This rate structure applies to adjusted net capital gain and qualified dividends, with a top statutory tax
rate of 20 percent.

It then immediately discusses the new structure:
> Families and individuals will be able to deduct 50 percent of their net capital gains, dividends, and interest income, leading to basic rates of 6 percent, 12.5 percent, and 16.5 percent on such investment income depending on the individual’s tax bracket. 

It then proceeds to appFrom this context, we inferred that by "net capital gains" they are referring specifically to net long-term capital gain, which is p23250 in the PUF. Both Tax Foundation and TPC made this assumption as well.

In the implementation I used, I assumed that this deduction is 50 percent of net long-term capital gains, 50 percent of qualified dividends, and 50 percent of interest income. Since qualified dividends and interest cannot be negative, it is not difficult to apply the deduction to e00300 and e00650. Because of the possible losses on capital gains, and that the loss limitation applies to combined short- and long-term capital gains, the deduction is more complicated for long-term gains. 

@martinholmer requested:

> Perhaps a simple example of a filing unit with both interest and capital losses would help me understand your proposed code.

As I see it, this capital gain deduction can be implemented in 3 different ways:
`CG_ec_RyanBrady = ALD_InvInc_ec_rt * max(-3000 / _sep, p23250)`  (1)
`CG_ec_RyanBrady = ALD_InvInc_ec_rt * (c01000 - max(-3000 / _sep, p22250))`  (2)
`CG_ec_RyanBrady = c01000 - max(-3000 / _sep, p22250 + ALD_InvInc_ec_rt * p23250)`  (3)

Method 1 applies the deduction only to ltcg after loss limitation, Method 2 applies it to the difference between total capital gain after loss limitation and short-term capital gain after loss limitation, and Method 3 applies it to net long-term gains before applying the loss limitation. For convenience, let's assume that the filer is not separate, so `_sep = 1`. I will now walk through a few hypothetical combinations of ltcg (p23250) and stcg (p22250), and provide the amount of the capital gain deduction for each of the three methods, as well as the after-deduction taxable gain for each of the three methods and for current law.

Example 1: ltcg = 10000, stcg = 10000
Ec(1) = 5000
Ec(2) = 5000
Ec(3) = 5000
TG(1) = 15000
TG(2) = 15000
TG(3) = 15000

Example 2: ltcg = 10000, stcg = -10000
Ec(1) = 5000
Ec(2) = 1500
Ec(3) = 3000
TG(1) = -5000
TG(2) = -1500
TG(3) = -3000

Example 3: ltcg = -10000, stcg = 10000
Ec(1) = -1500
Ec(2) = -5000
Ec(3) = -5000
TG(1) = 1500
TG(2) = 5000
TG(3) = 5000

Example 4: ltcg = -10000, stcg = -10000
Ec(1) = -1500
Ec(2) = 0
Ec(3) = 0
TG(1) = -1500
TG(2) = -3000
TG(3) = -3000

Example 5: ltcg = 10000, stcg = -5000
Ec(1) = 5000
Ec(2) = 4000
Ec(3) = 5000
TG(1) = 0
TG(2) = 1000
TG(3) = 0

Method 1 results in an incorrect application of the loss limitation for those with net negative stcg (examples 2 and 4). Methods 2 and 3 are similar, but I prefer Method 3 because it is more in line with the idea of deducting half of long-term capital gains. The current version of this PR uses Method 3.

@martinholmer, does that answer your questions?

feenberg:
On Wed, 22 Feb 2017, Martin Holmer wrote:

> 
> @codykallen, Perhaps I've missed something but the only discussion of the
> investment income "deduction" in the Better Way document I can find is this
> sentence:
> 
> This Blueprint provides for reduced tax on investment income.
> Families and individuals will be able to deduct 50 percent of
> their net capital gains, dividends, and interest income, leading
> to basic rates of 6 percent, 12.5 percent, and 16.5 percent on
> such investment income depending on the individual’s tax bracket. 
> 
> If this is all there is, then you are correct in saying there is an
> ambiguity.
>
>  *  Do they mean 50 percent of the sum of "net capital gains, dividends,
>     and interest income"?
>  *  Or do they mean the "deduction" is 50 percent
>     of net capital gains, 50 percent of [qualified?] dividends, plus 50
>     percent of interest income?

The latter is the likely option. Consider the person with 1 million of 
capital loss, and no other income. There AGI would be (-3000+500000) under 
the first interpretation.

dan

martinholmer:
@feenberg said about my comment:
>> If this is all there is, then you [@codykallen] are correct in saying there is an ambiguity.
>> - Do they mean 50 percent of the sum of "net capital gains, dividends, and interest income"?
>> - Or do they mean the "deduction" is 50 percent of net capital gains, 50 percent of [qualified?] dividends, plus 50 percent of interest income?
>
> The latter is the likely option. Consider the person with 1 million of capital loss, and no other income. There AGI would be (-3000+500000) under the first interpretation.

Dan, Thanks for your observations on the Blueprint reform.  I'm still confused about the reform and about your observations.

First, about the reform itself.  On page 18 of the Blueprint document there is a *Simple, Fair "Postcard" Tax Filing* that could be used if the reform were to become law.  The first two lines on the postcard say this:
```
1  Wage and compensation income
2  Add 1/2 of investment income
```
This seems to me a pretty clear indication that the Blueprint is defining "investment income" as the sum of interest, dividends and capital gains, and that the exclusion is 50 percent of that total investment income.

Second, I don't understand your example.  A filing unit with a million dollar capital loss and no other income would have under the first interpretation an AGI of -3000.  The 50 percent of investment income is an exclusion and that exclusion has always been constrained in the source code to be non-negative.

Combining these two points leads me to interpret the Blueprint as follows:
```
 # compute exclusion of investment income for Ryan-Brady plan
 if ALD_InvInc_ec_base_RyanBrady:
     invinc_agi_ec = ALD_InvInc_ec_rt * max(0, e00300 + e00650 + p23250)
```

martinholmer:
@codykallen,  Thanks for all the examples.  But I'm still left wondering what you see in the Blueprint that leads you to believe the $3000 loss limit should be used to calculate the investment income exclusion.  What in the Blueprint leads you to think the code shouldn't be as follows?
```
 # compute exclusion of investment income for Ryan-Brady plan
 if ALD_InvInc_ec_base_RyanBrady:
     invinc_agi_ec = ALD_InvInc_ec_rt * max(0, e00300 + e00650 + p23250)
```

codykallen:
@martinholmer, The "Postcard" is a gross simplification, and should not be taken seriously. For example, the Blueprint calls for a top rate of 25% on pass-through income, but it has no place to enter pass-through income. This reform does not simplify nearly enough to fit on a postcard. 

@martinholmer, as you mentioned, this requires choosing between excluding 50 percent of the sum of net capital gains, dividends, and interest income, or excluding 50 percent of net capital gains, 50 percent of dividends, and 50 percent of interest income. As we've noted, the Blueprint is less than clear on this. However, I agree with @feenberg that it seems more plausible that the exclusion would apply to each of these categories separately (the latter option). @martinholmer suggests the former application:
`invinc_agi_ec = ALD_InvInc_ec_rt * max(0, e00300 + e00650 + p23250)`

I believe this is too simplistic. Moreover, I see no reason to think apply a strict minimum of zero to the exclusion. Suppose we use your suggested method, and apply this to some of the above examples, assuming that the individual has no interest or qualified dividends. 

Example 2: ltcg = 10000, stcg = -10000
Ec(4) = 5000
TG(4) = -5000

Example 3: ltcg = -10000, stcg = 10000
Ec(4) = 0
TG(4) = 0

The reason for including the interaction between stcg and ltcg is that the exclusion should apply in both directions. If you only have to apply half of ltcg toward taxable income when positive, you also only get to apply half of it when negative. 

We can undermine this with another example. Suppose an individual has no interest or qualified dividends, $10,000 in long-term capital gains and $50,000 in short-term capital losses. Under current law, this filer's taxable gain would be -$3000. From my post earlier, Methods 2 and 3 would also give this filer a -$3000 taxable gain. Under your formulation, this person would have a -$8000 taxable gain (from the -$3000 loss limitation on c01000 and the additional deduction of $5000). This definitely seems like an incorrect implementation to me.

feenberg:
On Wed, 22 Feb 2017, Martin Holmer wrote:

> 
> @feenberg said about my comment:
>
>             If this is all there is, then you [@codykallen] are
>             correct in saying there is an ambiguity.
>
>              o  Do they mean 50 percent of the sum of "net
>                 capital gains, dividends, and interest income"?
>              o  Or do they mean the "deduction" is 50 percent of
>                 net capital gains, 50 percent of [qualified?]
>                 dividends, plus 50 percent of interest income?
>
>       The latter is the likely option. Consider the person with 1
>       million of capital loss, and no other income. There AGI would be
>       (-3000+500000) under the first interpretation.
> 
> Dan, Thanks for your observations on the Blueprint reform. I'm still
> confused about the reform and about your observations.
> 
> First, about the reform itself. On page 18 of the Blueprint document there
> is a Simple, Fair "Postcard" Tax Filing that could be used if the reform
> were to become law. The first two lines on the postcard say this:
> 
> 1  Wage and compensation income
> 2  Add 1/2 of investment income
> 
> This seems to me a pretty clear indication that the Blueprint is defining
> "investment income" as the sum of interest, dividends and capital gains, and
> that the exclusion is 50 percent of that total investment income.

I haven't seen the blueprint. Can you provide a URL?

> 
> Second, I don't understand your example. A filing unit with a million dollar
> capital loss and no other income would have under the first interpretation
> an AGI of -3000. The 50 percent of investment income is an exclusion and
> that exclusion has always been constrained in the source code to be
> non-negative.

The previous email described it as a deduction. I agree as an exclusion my 
calculation would be wrong.

martinholmer:
@codykallen, Thanks for providing such an extensive explanation of how you interpret the Blueprint.

It seems as if about the only thing we agree on is that the Blueprint is vague about its proposed investment income exclusion.  You seem to think the Blueprint has no concept of "investment income", but rather has three different type of income that are each eligible for the 50 percent AGI exclusion.  While I read the Blueprint as saying that there is a concept of "investment income" (defined as the sum of taxable interest, qualified dividends and long-term capital gains) and an "investment income" AGI exclusion which is 50 percent of "investment income" subject to the standard IRS rule that AGI exclusions cannot be no less than zero.

All of the examples you've provided seem to want to treat short-term capital gains as having some kind of bearing on the calculation of the Blueprint's investment income AGI exclusion.  I think the Blueprint is very clear: short-term capital gains have no bearing on the investment income AGI exclusion.

I don't see why you have a difficult time believing that long-term capital losses cannot sensibly be used to offset positive interest and qualified dividend income when computing the proposed investment income AGI exclusion.  In your interpretation, there is no possibility of offset.  Consider somebody with only $1,000,000 in qualified dividends and $1,000,000 in long-term capital losses.  In my interpretation of the Blueprint reform, this person would have an investment income AGI exclusion of zero and an AGI of $997,000 (the dividends less the limited loss).  But the formula in this pull request would produce an investment income exclusion of $500,000 that reduces AGI to $497,000.  That's a big investment income exclusion to get when total investment income is zero.  Is this really what the Blueprint is proposing?

codykallen:
To provide some context, the relevant purpose here is to tax only half of investment income at ordinary rates, and not tax the other half. With interest and qualified dividends, subtracting half after adding all of it to AGI is equivalent to just adding half, but this is not true for capital gains.

@martinholmer, your interpretation in this example is still problematic and violates the intent of the Blueprint. 

> Consider somebody with only $1,000,000 in qualified dividends and $1,000,000 in long-term capital losses. In my interpretation of the Blueprint reform, this person would have an investment income AGI exclusion of zero and an AGI of $997,000 (the dividends less the limited loss). 

In your interpretation, this person has an exclusion of zero, so 99.7% of his qualified dividends are taxed at ordinary rates, effectively giving this filer an top rate of 32.9%, whereas that intended rate should be 16.5%, according to the Blueprint. This is definitely a violation of the intent of the reform, and $997,000 is a huge AGI (and the corresponding tax) to get when total investment income is zero. 

@martinholmer also said:
> All of the examples you've provided seem to want to treat short-term capital gains as having some kind of bearing on the calculation of the Blueprint's investment income AGI exclusion. I think the Blueprint is very clear: short-term capital gains have no bearing on the investment income AGI exclusion.

I have to disagree with you here. The Blueprint makes no mention of short-term capital gains. You interpret this as implying that they should not be included in the reform. However, there is an interaction between short-term and long-term capital gains under existing tax law.

The important piece here isn't the investment income exclusion; it's the investment income tax base. The interaction between ltcg and stcg is in the process of adding them to AGI. Therefore, it seems a bit absurd to retain the interaction when adding them to AGI but ignore it when subtracting off half of ltcg.

martinholmer:
@codykallen,  Thanks for your analysis of my simple example.
We are in agreement that the only guidance on the details of the investment income AGI exclusion provision in the Blueprint reform is this passage on page 18:
```
This Blueprint provides for reduced tax on investment income.
Families and individuals will be able to deduct 50 percent of
their net capital gains, dividends, and interest income, leading
to basic rates of 6 percent, 12.5 percent, and 16.5 percent on
such investment income depending on the individual’s tax bracket.
```
So, in your last set of comments you stress the importance of the final phrase in that passage:
```
leading to basic rates of 6 percent, 12.5 percent, and 16.5 percent on
such investment income depending on the individual’s tax bracket.
```
That phrase is the strongest evidence for your interpretation that the 50 percent exclusion is to be calculated separately for taxable interest, for qualified dividends and for long-term capital gains.
I had not grasped the importance of that phrase in your interpretation of the Blueprint.

Thanks for the helpful conversation on this matter.  I'm sorry if I was a little slow on the uptake.

I'll merge this pull request now.

codykallen:
@martinholmer, thanks for checking this. And you're not slow on the uptake; I've spent too many hours dealing with this annoyance (and Tax Foundation actually still has it wrong). And it's always good to check these.

feenberg:
On Wed, 22 Feb 2017, codykallen wrote:

> 
> To provide some context, the relevant purpose here is to tax only half of
> investment income at ordinary rates, and not tax the other half. With
> interest and qualified dividends, subtracting half after adding all of it to
> AGI is equivalent to just adding half, but this is not true for capital
> gains.
> 
> @martinholmer, your interpretation in this example is still problematic and
> violates the intent of the Blueprint.
>
>       Consider somebody with only $1,000,000 in qualified dividends
>       and $1,000,000 in long-term capital losses. In my interpretation
>       of the Blueprint reform, this person would have an investment
>       income AGI exclusion of zero and an AGI of $997,000 (the
>       dividends less the limited loss).
> 
> In your interpretation, this person has an exclusion of zero, so 99.7% of
> his qualified dividends are taxed at ordinary rates, effectively giving this
> filer an top rate of 32.9%, whereas that intended rate should be 16.5%,
> according to the Blueprint. This is definitely a violation of the intent of
> the reform, and $997,000 is a huge AGI (and the corresponding tax) to get
> when total investment income is zero.
> 
There is a long tradition in US tax law of not allowing a substantial 
benefit to long term losses. Consider the short term loss of revenue as 
everyone liquidated their short term losses as quickly as they could use 
the deduction. I believe there are a lot of outstanding losses. Is this 
something we could take up with the person(s) requesting this simulation?

dan

martinholmer:
@feenberg said:
> There is a long tradition in US tax law of not allowing a substantial benefit to long term losses. Consider the short term loss of revenue as everyone liquidated their short term losses as quickly as they could use the deduction.

Dan, Thanks for your comment, but I'm not sure what the implication of your observation is.  Are you saying that "long tradition in US tax law of not allowing a substantial benefit to long term losses" is based on a fear of people reducing taxes on other income by realizing long-term capital losses.  If that is a correct understanding of what you're saying, then it would seem the Ryan-Brady Blueprint reform (as currently implemented in Tax-Calculator) is at variance with that "long tradition".

Or maybe I've completely misunderstood the implications of your comment.  I'd be interested in more of your thoughts on this matter.

feenberg:
On Wed, 22 Feb 2017, Martin Holmer wrote:

> 
> @feenberg said:
>
>       There is a long tradition in US tax law of not allowing a
>       substantial benefit to long term losses. Consider the short term
>       loss of revenue as everyone liquidated their short term losses
>       as quickly as they could use the deduction.
> 
> Dan, Thanks for your comment, but I'm not sure what the implication of your
> observation is. Are you saying that "long tradition in US tax law of not
> allowing a substantial benefit to long term losses" is based on a fear of
> people reducing taxes on other income by realizing long-term capital losses.
> If that is a correct understanding of what you're saying, then it would seem
> the Ryan-Brady Blueprint reform (as currently implemented in Tax-Calculator)
> is at variance with that "long tradition".
> 
> Or maybe I've completely misunderstood the implications of your comment. I'd
> be interested in more of your thoughts on this matter.

I guess I don't know the answer, but estimating revenue when losses are 
deductible would be a research project. I would expect a huge revenue 
loss, short term and long term and it would surprise me if that were 
intended. You would have to ask the proposer.

I have seen economists complain about the loss limitation discouraging 
risk taking, and seen no sympathy displayed by any Washington insider.

dan

codykallen:
@feenberg asked:

> Is this something we could take up with the person(s) requesting this simulation?

If the Speaker's staff or the Ways and Means staff could clarify it, that would be great. Unfortunately, I don't have the clout to get those answers. If anyone can get a clarification, I would be happy to work with whatever the Hill staff says.

PR #1206, opened 2017-02-22
Select provisions from Clinton tax plan

codykallen:
This PR adds a reform file for select provisions of Clinton's tax plan. Unfortunately, most provisions of Clinton's tax proposals cannot be scored, for the following reasons:

- The provisions do not relate to the individual income or payroll tax systems.
- The provisions are too complicated to evaluate (e.g. cash accounting for small businesses).
- Scoring the provisions would require microdata that we do not have (e.g. taxing carried interest as ordinary income).
- The provisions would induce new behavior and filing activity that we cannot plausibly predict (e.g. a new credit for caregiver expenses).

For now, I have included at the bottom a list of provisions that we cannot score right now. I think we should merge this JSON in, and modify it if we add the capability to score these provisions.

@martinholmer @MattHJensen @andersonfrailey @feenberg @Amy-Xu

### DIFF
--- a/taxcalc/reforms/Clinton2016.json
+++ b/taxcalc/reforms/Clinton2016.json
@@ -2,7 +2,7 @@
 // Reform_File_Author: Cody Kallen
 // Reform Reference: https://www.hillaryclinton.com/briefing/factsheets/2016/01/12/investing-in-america-by-restoring-basic-fairness-to-our-tax-code/
 // Reform Description:
-// -  4% surcharge on AGI above $5 million (1)
+// -  4% surcharge on AGI above $5 million ($2.5 million for married filing separately) (1)
 // -  Establishment of the "Fair Share Tax" at a 30% rate (aka the Buffett rule) (2)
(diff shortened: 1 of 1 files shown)


MattHJensen:
Thanks @codykallen! @GoFroggyRun, I remember you modeling several of these over the winter -- could you review this one?

GoFroggyRun:
Sorry @codykallen @MattHJensen that I forgot to submit the review after commenting, so that you guys weren't able to see it.

codykallen:
@GoFroggyRun wrote:

> I can definitely understand why you would assign a threshold of 2.5e6 to separate filers, as otherwise high income joint filers will obviously get better-off by switching to separate filing status. But since Clinton's original plan wasn't detailed enough and (I assume) this number comes from our reasonable assumption, would it be helpful to provide a brief note regarding this?

Thanks for the suggestion.  I have modified the description of the provision to reflect that.

MattHJensen:
Thanks both. Merging.

PR #1107, opened 2016-12-15
Change param_code to have multiple-statement (rather than single-expression) format

martinholmer:
This is not so much a pull request as a continuation of the conversation that started with issue #429 and continued with merged pull request #1081 and WIP pull request #1103.  This extended conversation concerns the technical possibility (and the various pros and cons) of adding logic that handles policy parameters that are expressed as code rather than as numbers.

#1081 added the ability to handle very simple situations in which the code could be an single expression (that is, what is on the right-hand side of an assignment statement).  This ability was added using the Python `eval` function and by eliminating the availability of `__builtins__` and by rejecting code that contains high-security-risk strings (`lambda` and `__`, for example).

#1103 was more ambitious in that it tried to use the capabilities introduced in #1081 to compute with code a wide range of new refundable child tax credits.  This more complex problem was not able to be handled with the single expression capability introduced in #1081.

This WIP pull request #1107 extends #1103 so that a wide range of new refundable child tax credits can be computed with the new code parameter introduced by #1103.  The new refundable child tax credit example introduces several new issues: (a) the need to execute not just a single expression, but rather a series of statements, and (b) the need to handle policy parameters that are inflation indexed.  The revisions in this pull request solve both of these new issues.  The solution involves
using the Python `exec` function and continuing to eliminate the availability of `__builtins__` and to reject code that contains high-security-risk strings (`lambda` and `__`, for example).  The Policy class has been enhanced to provide information needed to index policy parameters.

This pull request is meant for discussion.  There are many questions that need to be discussed including the readability of the parameter code in the sample script below and the security risks involved in allowing multiple statements (rather than just a single expression) in the parameter code.

Below is a script that is similar to the notebook discussed in #1103.  It shows that the new parameter code feature produces exactly the same results as the (still present) numerical parameter characterization of two reforms that introduce new refundable child tax credits.

@MattHJensen @feenberg @talumbau @zrisher @Amy-Xu
@GoFroggyRun @andersonfrailey @codykallen 

Here is a command-line session that runs only under this pull request:
```
$ cp ../tax-calculator-data/puf.csv .

$ cat param-code-exec.py
from taxcalc import *

def print_results(desc, calc):
    record_columns = ['s006', '_combined']
    out = [getattr(calc.records, col) for col in record_columns]
    dfx = pd.DataFrame(data=np.column_stack(out), columns=record_columns)
    wsum = weighted_sum(dfx, '_combined') * 1.0e-9
[...]

PR #1221, opened 2017-03-07
Remove all logic and tests related to param_code capability

### LINKED ISSUE #429: User extensibility of Taxbrain

martinholmer:
The `param_code` capability, which was merged on January 13, 2017, into the master branch with pull request #1107, was an experimental attempt to resolve issue #429.  This experimental attempt to resolve #429 revealed that the approach adopted in #1107 had benefits, but its costs were larger than the benefits.  The #1107 approach raised security questions because of the use of the Python exec() function.  It was awkward to write multiple-line code in a JSON reform file.  And the #1107 approach required the use of the numpy `where` function, which was not as easy to read as the if-then-else statements used in the `functions.py` file.

Given this experience, my conclusion is that, if one wanted to eliminate the problems with the #1107 approach, it may require writing a simple parser that would translate more user-freindly parameter code into Python code.  But there may be different approaches that are effective at eliminating the problems with the #1107 approach.

In any case, this pull request removes the experimental `param_code` logic implemented in #1107.
And it also serves as an historical record of a failed experiment in resolving #429.

@PeterDSteinberg

### DIFF
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -143,10 +143,7 @@ class Calculator(object):
         EI_PayrollTax(self.policy, self.records)
         DependentCare(self.policy, self.records)
         Adj(self.policy, self.records)
-        if self.policy.ALD_InvInc_ec_base_code_active:
-            ALD_InvInc_ec_base_code(self)
-        else:
-            ALD_InvInc_ec_base_nocode(self.policy, self.records)
+        ALD_InvInc_ec_base(self.policy, self.records)
         CapGains(self.policy, self.records)
         SSBenefits(self.policy, self.records)
         AGI(self.policy, self.records)
@@ -188,10 +185,7 @@ class Calculator(object):
         NonrefundableCredits(self.policy, self.records)
         AdditionalCTC(self.policy, self.records)
         C1040(self.policy, self.records)
-        if self.policy.CTC_new_code_active:
-            CTC_new_code(self)
-        else:
-            CTC_new_nocode(self.policy, self.records)
+        CTC_new(self.policy, self.records)
         IITAX(self.policy, self.records)
 
     def calc_all(self, zero_out_calc_vars=False):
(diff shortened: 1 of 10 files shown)


MattHJensen:
This makes sense to me. +1

PR #1228, opened 2017-03-07
Move simtax code and test to taxcalc/validation/taxsim directory

martinholmer:
This pull request moves the `simtax.py`, `simpletaxio.py` and `test_simpletaxio.py` files to the taxcalc/validation/taxsim directory, which is the only place they are used (for Tax-Calculator validation against Internet TAXSIM).


martinholmer:
In addition to moving all simtax-related code and tests to the taxcalc/validation/taxsim directory, pull request #1228 switches from relative imports to absolute imports because the use of relative imports is strongly discouraged in the Python community.

Here is one example (a 2013 post on GitHub):
```
PEP 8 says: "Relative imports for intra-package imports are highly discouraged.
Always use the absolute package path for all imports. Even now that PEP 328
is fully implemented in Python 2.5, its style of explicit relative imports is 
actively discouraged; absolute imports are more portable and usually more readable."
```

Issue #83, opened 2014-12-17
Link output tables to calculator


MattHJensen:
@talumbau and @theofanislekkas, contingent on @feenberg's review, here are the first set of output tables we should provide. We will want to add several more options later and likely some charts, so we might want to have a page that links to each of these tables (or something similar) rather than putting them all on one page.

##### The first six/sixty-six output tables (variables listed below). 

1. Plan x vars, avg by AGI decile  
2. Plan x vars, avg by AGI group
3. Plan y vars, avg by AGI decile
4. Plan y vars, avg by AGI group
5. By-tax-record difference between plan y and plan x, avg by AGI decile
6. By-tax-record difference between plan y and plan x, avg by by AGI group

*The reason this could/should be sixty-six tables is that it'd be very nice to be able to see each of these tables by year, as well as see the ten-year average of the by-year tables.* 

*@feenberg, do you think the 10-year average should be the simple average of each cell across the 10 one-year tables, or should we do a weighted average (more important for AGI groups than deciles)?* 

##### Variables for first 6 tables:

`c00100` AGI
`c04100` Standard Deduction
`c04470` Itemized deductions
`c04800` taxable income
`c05200` Regular tax
`c09600` Alternative Minimum Tax
*[NEED TO CREATE VAR]      Non-Refundable Credits
`c09200` SOI Tax (Tax before refundable credits)
*[NEED TO CREATE VAR]      Refundable credits
*[NEED TO CREATE VAR] Tax after all credits

*Ignore for now; @feenberg, you should have an email from me on the 15th regarding these.*

##### AGI Groups
Below 10
10 -20
20-30
30-40
40-50
50-75
75-100
100-200
200 +
Total 

##### The seventh and eighth output table deal primarily with 'tax after all credits' 
7.  By AGI decile, 
    * percent of tax units in each decile with a tax cut 
    * percent of tax units in each decile with a tax increase
    * share of total tax change born by each decile (percent)
    * average tax change ($)
    * @feenberg, do you think we should include an average federal tax rate percentage point change and an average federal tax rate under the proposal as in the TPC example below? (average total-tax/average AGI?)

8.  Same by AGI group

*We won't be able to create these tables until we get the tax after all credits variable sorted out*

*Again, it would be nice to have these for each of the ten years and an average.* 

[...]

PR #1235, opened 2017-03-10
UBI Implementation

andersonfrailey:
Second time's the charm. This PR adds a feature to Tax-Calc that calculates the effects of a universal basic income. The UBI function relies on three new PUF variables (`nu18`, `n1821`, and `n21`) that hold the number of people in each age range in each tax unit. These have been discussed in TaxData PR [#83](https://github.com/open-source-economics/taxdata/pull/83).

Four new parameters are added into `current_law_policy.json`. `_UBI1-3` allow the user to specify different levels of UBI for each age group. `_UBI_hc` allows the user to specify what portion of the UBI is taxable (I'm not sure `hc` is the correct tag for this, but it seemed like the best option. If this needs to change let me know).

Using the new variables and parameters, the function calculates three values:
`ubi`: total income from the UBI received by the tax unit.
`taxable_ubi`: the portion of total UBI that is taxable. This is added to AGI.
`nontaxable_ubi`: the portion of total UBI that is not taxable. This is added to expanded income.

Adding only taxable UBI to AGI ensures that if the user specifies that UBI is not taxable, tax liability for the unit does not change. Nontaxable UBI, rather than total UBI is added to expanded income so that the taxable portion of UBI, which was already added to AGI, is not double counted. 

@MattHJensen @Amy-Xu @martinholmer

### DIFF
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -146,6 +146,7 @@ class Calculator(object):
         ALD_InvInc_ec_base(self.policy, self.records)
         CapGains(self.policy, self.records)
         SSBenefits(self.policy, self.records)
+        UBI(self.policy, self.records)
         AGI(self.policy, self.records)
         ItemDed(self.policy, self.records)
         AdditionalMedicareTax(self.policy, self.records)
--- a/taxcalc/current_law_policy.json
+++ b/taxcalc/current_law_policy.json
@@ -2753,5 +2753,69 @@
         "cpi_inflated": false,
         "col_label": "",
         "value": [0.0]
+    },
+
+    "_UBI1": {
+        "long_name": "UBI for those under 18",
+        "description": "Lump sum of money given to people under 18",
+        "section_1": "UBI",
+        "section_2": "Universal Basic Income",
+        "irs_ref": "",
+        "notes": "",
+        "start_year": 2013,
+        "col_var": "",
+        "row_var": "",
+        "row_label": ["2013"],
+        "cpi_inflated": true,
+        "col_label": "",
+        "value": [0.0]
+    },
+
+    "_UBI2": {
+        "long_name": "UBI for those between 18 and 21",
+        "description": "Lump sum of money given to people between 18 and 21",
+        "section_1": "UBI",
+        "section_2": "Universal Basic Income",
+        "irs_ref": "",
+        "notes": "",
+        "start_year": 2013,
+        "col_var": "",
+        "row_var": "",
+        "row_label": ["2013"],
+        "cpi_inflated": true,
+        "col_label": "",
+        "value": [0.0]
+    },
+
+    "_UBI3": {
+        "long_name": "UBI for those 21 and over",
+        "description": "Lump sum of money given to people 21 and over",
+        "section_1": "UBI",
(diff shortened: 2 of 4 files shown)


martinholmer:
@andersonfrailey said about the UBI benefits PR #1235:
>  `_UBI_hc` allows the user to specify what portion of the UBI is taxable (I'm not sure `_hc` is the correct tag for this, but it seemed like the best option.  If this needs to change let me know).

Well, I guess the proper name for this parameter all depends on what you view as the "normal" situation.  Searching through `functions.py`,  you can see that the other uses of `_hc` policy parameters are reducing some kind of tax benefit, which is why those parameters are always used like this `... * (1. - XXX_hc)`.  Notice that in the current version of #1235 you use the new parameter in the opposite way:
```
taxable_ubi = ubi * UBI_hc
```
This is not what `_hc` means in the rest of Tax-Calculator, so this is likely to be confusing to users.

If you look at the [Contributor Guide's parameter naming rules](http://taxcalc.readthedocs.io/en/latest/parameter_naming.html), it seems to me that what we are dealing with here is an AGI exclusion rate, the fraction of UBI benefits that are excluded from AGI.  Those rules suggest the new parameter might better be named `_UBI_ecrt`.  Using that name is, to me, clearer than "haircut" because I'm not sure what "hair" is being "cut".

Using the `_UBI_ecrt` parameter name, would require changes to the `functions.py` logic and the `current_law_policy.json` long_name and description.  The logic would be:
```
taxable_ubi = ubi * (1. - UBI_ecrt)
```
and the long_name and description would be:
```
 "long_name": "Fraction of UBI benefits excluded from AGI",
 "description": "One minus this fraction of UBI benefits is taxable and will be added to AGI",
```
Does anybody have a better idea?

@MattHJensen @feenberg @Amy-Xu @GoFroggyRun @codykallen @zrisher

andersonfrailey:
@martinholmer, I like your suggestion more than using `hc`. I'll apply and push the changes now.

martinholmer:
@andersonfrailey,  pull request #1235 looks very good now except for maybe one thing.  I'm not familiar with the UBI conversation, but wouldn't most people expect that UBI benefits would be included in AGI?  Right now, you have in `current_law_policy.json` that all of the UBI benefit would be *excluded* from AGI (that is, non-taxable).  Is that what most UBI advocates propose --- to give Bill Gates a tax-free UBI benefit?  If that is where the conversation stands, then setting the value of the `_UBI_ecrt` parameter to 1.0 (which is what you have in this pull request) makes sense.  If most would expect UBI benefits to be taxable, then you should probably change the value of the `_UBI_ecrt` parameter to 0.0.

Once we decide on the best default value, I'll be happy to merge @1235 into the master branch.
Thanks for this contribution to Tax-Calculator.

andersonfrailey:
@martinholmer, good catch. I haven't seen any true consensus among UBI proponents on whether or not it should be taxed, but I agree with your logic that most wouldn't be in favor of handing Bill Gates tax-free money. I'll make the change so it's ready to merge.

martinholmer:
@andersonfrailey, 
Thanks for the finishing touch on the UBI enhancement in pull request #1235.
Merging #1235 now.

PR #1247, opened 2017-03-15
Add additional haircuts for deductions and credits

andersonfrailey:
This PR adds additional haircut options for a number of credits and deductions and fixes a minor error in the UBI function's docstring. There was nothing in the naming convention regarding what to name variables related to credits in this situation so I went with `CR` to start the haircut variables applied to credits. I'm happy to change that if there is consensus on a better option.

The point is to allow a user to repeal credits and all above the line deductions. Here are the new haricuts and what they are applied to:

| Variable | Description                                 | Type      | functions                         | Hair Cut                   | 
|----------|---------------------------------------------|-----------|-----------------------------------|----------------------------| 
| e07240   | Retirement Saving Credit                    | Credit    | NonRefundable Credits             | _CR_RetirementSavings_hc   | 
| e07300   | Foreign Tax Credit                          | Credit    | NonRefundable Credits             | _CR_ForeignTax_hc          | 
| e07260   | Residential Energy Credit                   | Credit    | NonRefundable Credits             | _CR_ResidentialEnergy_hc   | 
| e07400   | General Business Credit                     | Credit    | NonRefundable Credits             | _CR_GeneralBusiness_hc     | 
| e07600   | Prior Year Minimum Tax Credit               | Credit    | NonRefundable Credits             | _CR_MinimumTax_hc          | 
| c10960   | American Opportunity Credit - Refundable    | Credit    | NonRefundable Credits, AmOppParts | _CR_AmOppRefundable_hc     | 
| c87668   | American Opportunity Credit - NonRefundable | Credit    | NonRefundable Credits, AmOppParts | _CR_AmOppNonRefundable_hc  | 
| c07200   | Schedule R                                  | Credit    | NonRefundable Credits, SchR       | _CR_SchR_hc                | 
| p08000   | Other Credits                               | Credit    | NonRefundable Credits             | _CR_OtherCredits_hc        | 
| e03220   | Educator Expenses Deduction                 | Deduction | Adj                               | _ALD_EducatorExpenses_hc   | 
| e03290   | HSA Deduction                               | Deduction | Adj                               | _ALD_HSADeduction_hc       | 
| e03150   | Total Deductable IRA plan contributions     | Deduction | Adj                               | _ALD_IRAContributions_hc   | 
| e03240   | Domestic Production Activity Deduction      | Deduction | Adj                               | _ALD_DomesticProduction_hc | 
| e03230   | Tuition and Fees                            | Deduction | Adj                               | _ALD_Tuition_hc            | 

[...]

PR #1252, opened 2017-03-20
Add Education Credit HC

andersonfrailey:
This PR is a continuation of PR #1247 and adds a haircut for education credits. The reason it was not included in #1247 is at the time I was under the impression that education credits could be zeroed out by simply setting `LLC_Expense` and `ETC_pe_Single/Married` to zero, but that is not the case. 

cc: @MattHJensen @martinholmer


martinholmer:
@andersonfrailey,  Thanks for adding an the additional haircut policy parameter in #1252.  Looks good, so I'm merging it now.

PR #1253, opened 2017-03-20
Make single CLI tool be an entry point to the taxcalc conda package

martinholmer:
This pull request changes the approach to developing a command-line interface (CLI) to Tax-Calculator to be in line with several suggestions @MattHJensen has offered over the past several weeks.  Thanks, Matt.

Now the CLI, which is called `tc`, is an integral part of the conda taxcalc package, which makes it easy to install after a user has install Anaconda on his/her computer:
```
conda install taxcalc
conda install bokeh         <---- needed as long as bokeh is not a dependency of taxcalc
```
After that installation, the command `conda list taxcalc` shows the Tax-Calculator version number and the command `tc --help` produces this:
```
usage: tc INPUT TAXYEAR [--reform REFORM] [--assump  ASSUMP]
                        [--exact] [--graph] [--ceeu] [--dump]

Writes to a file the federal income and payroll tax OUTPUT for each filing
unit specified in the INPUT file, with the OUTPUT computed from the INPUT for
the TAXYEAR using Tax-Calculator. The OUTPUT file is a CSV-formatted file that
contains tax information for each INPUT filing unit.

positional arguments:
  INPUT            INPUT is name of CSV-formatted file that contains for each
                   filing unit variables used to compute taxes for TAXYEAR.
  TAXYEAR          TAXYEAR is calendar year for which taxes are computed.

optional arguments:
  -h, --help       show this help message and exit
  --reform REFORM  REFORM is name of optional JSON reform file. No --reform
                   implies use of current-law policy.
  --assump ASSUMP  ASSUMP is name of optional JSON economic assumptions file.
                   No --assump implies use of no customized assumptions.
  --exact          optional flag that suppresses the smoothing of "stair-step"
                   provisions in the tax law that complicate marginal-tax-rate
                   calculations.
  --graph          optional flag that causes graphs to be written to HTML
                   files for viewing in browser.
  --ceeu           optional flag that causes normative welfare statistics,
                   including certainty-equivalent expected-utility (ceeu) of
                   after-tax income values for different constant-relative-
                   risk-aversion parameter values, to be written to screen.
  --dump           optional flag that causes OUTPUT to contain all INPUT
                   variables (possibly extrapolated to TAXYEAR) and all
                   calculated tax variables, where all the variables are named
                   using their internal Tax-Calculator names. No --dump option
                   implies OUTPUT contains minimal tax output.
```

This pull request also adds a `file-upload-tests/cli-results` bash script that uses the `tc` CLI to produce results that can be compared to results generated by the TaxBrain file-upload capability as explained in the `file-upload-tests/README.md` file in this pull request.


taxcalc/cli/tc.py
```diff
@@ -38,12 +44,12 @@ def main():
...
                               'No --reform implies use of current-law '
-                              'policy.'),
+                              'policy (clp).'),
                         default=None)
     parser.add_argument('--assump',
                         help=('ASSUMP is name of optional JSON economic '
-                              'assumption file.  No --assump implies use of '
-                              'static analysis assumptions.'),
+                              'assumptions file.  No --assump implies use of '
+                              'standard (std) analysis assumptions.'),
```
MattHJensen: I think "base" might be more appropriate here than "standard" as the word "standard" has normative implications.

martinholmer: Thanks for reviewing #1253.  What "normative implications" are you thinking of?

MattHJensen: To me, "standard" implies that that a usage is 'approved' usage and that other uses are deviant. 

For example, consider two Tax-Calculator users on CNN presenting their results. One person could say, "I got X results assuming Y". The second person could say, "Y is just your assumption; when I use the standard model, I got Z." 

In this case, Z would sound more reasonable than X. 

The same dialogue could take place if the name were "base" instead of "standard", but I think the second guy would have a weaker point. 

Something even more neutral than "base" would be preferable, but I can't think of it.

MattHJensen: "Static" solved this problem, but I can see how an analysis is still "static" when growdiff_baseline and growdiff_response have been applied.

taxcalc/taxcalcio.py
```diff
@@ -47,11 +47,11 @@ class TaxCalcIO(object):
         or string is name of optional ASSUMP file.
 
     growdiff_response: Growdiff object or None
-        growdiff_response Growdiff object used in dynamic analysis;
-        must be None when conducting static analysis.
+        growdiff_response Growdiff object used only in dynamic analysis;
+        must be None when not conducting dynamic analysis.
```
MattHJensen: Not sure whether it is worth splitting hairs here in the code, but growdiff_response must also be None when conducting a partial-equilibrium dynamic analysis.

taxcalc/cli/tc.py
martinholmer: @MattHJensen said:
> Something even more neutral than "base" would be preferable, but I can't think of it.

OK, I see your point.  Let me think about alternatives to "standard" and "base".

taxcalc/taxcalcio.py
martinholmer: @MattHJensen said:
> Not sure whether it is worth splitting hairs here in the code, but growdiff_response must also be None when conducting a partial-equilibrium dynamic analysis.

I thought your taxonomy was: static, partial-equilibrium(ie, behavior), and dynamic analysis, right?  I thought this was the implication of the various quotes you offered last week.  If I'm still confused, we need to talk on the phone.

Given that analysis taxonomy, non-dynamic analysis is static or partial-equilibrium, and the wording is sensible.

MattHJensen:
> I thought your taxonomy was: static, partial-equilibrium(ie, behavior), and dynamic analysis, right?

That's not right. My point was that partial-equilibrium analysis should be considered dynamic analysis since it changes GNP, and in the tax-scoring world anything that changes GNP is called dynamic. Will give you a call shortly.

taxcalc/taxcalcio.py
MattHJensen: > I thought your taxonomy was: static, partial-equilibrium(ie, behavior), and dynamic analysis, right?

That's not right. My point was that partial-equilibrium analysis should be considered dynamic analysis since it changes GNP, and in the tax-scoring world anything that changes GNP is called dynamic. Will give you a call shortly. 

Note that this is the taxonomy we use on TaxBrain since partial equilibrium analysis is available from the dynamic landing page.

MattHJensen: From the TaxBrain landing page:

> In the context of policy analysis, dynamic modeling incorporates behavior that affects the aggregate output of the economy. This broad definition includes nearly every type of economic analysis except for the approach most often used in the analysis of tax and spending policy proposals by agencies in the federal government.
> 
> There is no single "accepted" way to estimate the dynamic effects of policy reforms, so TaxBrain is designed to encourage users to try several approaches and to grapple seriously with the limitations and tradeoffs of each approach.

martinholmer:
@MattHJensen,  I'm hoping the two most recent commits respond effectively to your concerns about terminology in pull request #1253.

martinholmer:
Last call for review of pull request #1253.  If there are no concerns raised, I hope to merge #1253 at the end of the day on Thursday, March 23rd.

MattHJensen:
Very excited about the new CLI capabilities. :+1: :+1:

PR #1254, opened 2017-03-21
Correct Erroneous CTC_new_ps Field

brittainhard:
This fixes the issue in https://github.com/OpenSourcePolicyCenter/webapp-public/issues/507

The keys for this field incorrectly defined. `col_var` and `col_label` values had to be swapped in order to get the field to be processed correctly.

cc @MattHJensen @PeterDSteinberg @jdebacker


MattHJensen:
Thanks @brittainhard. Merging.

PR #1257, opened 2017-03-22
Add MTR with respect to spouse earnings

andersonfrailey:
This PR adds `e00200s` to the list of valid MTR variables in `calculate.py`. It allows a user to find the MTR for the secondary tax. It will be very similar to MTR found using `e00200p`, but they are slightly different, mostly due to payroll taxes.

@MattHJensen


andersonfrailey:
Forgot to make changes to the test suite to cover this code. I'll work on that and add a commit.

martinholmer:
@andersonfrailey said in conversation on PR #1257:
> This PR adds `e00200s` to the list of valid MTR variables in `calculate.py`. It allows a user to find the MTR for the secondary tax. It will be very similar to MTR found using `e00200p`, but they are slightly different, mostly due to payroll taxes.

I don't understand the logic of this pull request.

First, I don't know what you mean by "secondary tax".  The `e00200s` variable is the **spouse** wage and salary income.

Second, I don't see any logic in your pull request that would handle filing units with no spouse.  It seems to me that most people would expect a zero marginal tax rate on spouse earnings for each filing unit that contains no spouse.  What sense would a non-zero MTR make for those filing units?  I think most users would be very surprised and puzzled to see a positive MTR wrt `e00200s` for units with `MARS` not equal to 2.

Third, I don't understand the motivation for adding this logic.  What is the purpose of #1257?

andersonfrailey:
@martinholmer to address each of your questions:

1. This was poor wording on my part. It should say "secondary earner", so the spouse.

2. You're right and I didn't think to account for this. I will work on a way to limit the change to only those tax units with a spouse and update this PR.

3. This is for a project @MattHJensen and I are working on. We want to be able to calculate and plot MTR using both the primary earner or the spouse.

feenberg:
On Wed, 22 Mar 2017, Martin Holmer wrote:

> 
> @andersonfrailey said in conversation on PR #1257:
>
>       This PR adds e00200s to the list of valid MTR variables in calculate.py.
>       It allows a user to find the MTR for the secondary tax. It will be very
>       similar to MTR found using e00200p, but they are slightly different,
>       mostly due to payroll taxes.
> 
> I don't understand the logic of this pull request.
> 
> First, I don't know what you mean by "secondary tax". The e00200s variable is the
> spouse wage and salary income.

The marginal rate on the spouse is of interest as there are frequent 
suggestions to tax the spouse at a lower rate. See my working paper with 
Rosen or Feldstein, among many.

> 
> Second, I don't see any logic in your pull request that would handle filing units with
> no spouse. It seems to me that most people would expect a zero marginal tax rate on
> spouse earnings for each filing unit that contains no spouse. What sense would a
> non-zero MTR make for those filing units? I think most users would be very surprised
> and puzzled to see a positive MTR wrt e00200s for units with MARS not equal to 2.
>

0 is just as missleading as any other number. Is it possible to make it 
"missing"?

dan

martinholmer:
@feenberg said about setting MTR on spouse earnings to zero for filing units with no spouse:
> 0 is just as misleading as any other number. Is it possible to make it "missing"?

Because the intention is the graph spouse earnings MTR values, I don't think assigning a "missing" value is going to be a useful strategy.  The MTR graphing capability already has built into it the ability to graph MTR for a specified MARS.  So, I imagine that the users of the MTR on spouse earnings will have the sense to graph it only for MARS==2.  So the value of the MTR on spouse earnings for unit with MARS!=2 is sort of academic.  Setting it to zero makes the most sense to me because I can't imagine what a non-zero MTR on spouse earnings means when there is no spouse.

MattHJensen:
@martinholmer, @andersonfrailey, @feenberg,

Users frequently take averages of MTRs outside of plotting. If there were a way to use a missing value, that would be preferable to zeroes. As @martinholmer points out, the missing or zero shouldn't matter for plotting when the user specifies MARS==2, so we only need to be worried about other use cases, like averaging across the population.

andersonfrailey:
@martinholmer and @feenberg, thanks for the feedback. In my latest commit, I added a conditional for that will make the MTR 0 when using e00200s for all tax units where MARS != 2.

I also updated `pufcsv_mtr_actual.txt` to include MTR calculations for e00200s. 

Still working on getting the code coverage right.

andersonfrailey:
@MattHJensen, I just saw your comment. If the consensus is to use a missing value rather than a zero I'll make that change and offer a new commit.

MattHJensen:
> If the consensus is to use a missing value rather than a zero I'll make that change and offer a new commit.

If the plot works just fine w MARS==2 and missing values, then I feel more comfortable with that. I don't see any reason to use zeroes instead of missing values, though, so I'm afraid I'm missing something.

martinholmer:
@MattHJensen said:
> Users frequently take averages of MTRs outside of plotting.  If there were a way to use a missing value, that would be preferable to zeroes.  As @martinholmer points out, missing or zero shouldn't matter for plotting when the user specifies MARS==2, so we only need to worry about other use cases, like averaging across the population.

How does Python/Pandas treat missing values in situations where "users frequently take averages of MTRs outside of plotting"?   Are observations going to be dropped because MTR on spouse earnings is missing?  If so, this could lead to unexpected and confusing results?

andersonfrailey:
@MattHJensen @martinholmer,  plotting with missing variables works just fine because all of the records where MARS=2 will have a value. But if you want to perform an operation such as:
```
mtr = calc.mtr('e00200s')
np.mean(mtr)
```
The result will be `nan`.

However, if instead you use:
```
mtr = calc.mtr('e00200s')
np.nanmean(mtr)
```
All `nan` (missing) values will be ignored and you get the average of the non-missing values.

andersonfrailey:
@MattHJensen @martinholmer @feenberg, what are the final thoughts on the zero vs. missing number debate? I'm leaning toward using missing numbers and adding a note in the docstring informing users that they'll need to account for them when using `e00200s`.

MattHJensen:
> I'm leaning toward using missing numbers and adding a note in the docstring informing users that they'll need to account for them when using e00200s.

That makes sense to me.

andersonfrailey:
Ignore the last commit. I just noticed that the missing values cause one of the PUF tests to fail.

andersonfrailey:
Latest commit updates `tests_pufcsv.py` so that the MTR histogram bin counts for `e00200s` does not include any missing values. It also updates the `mtr_bin_counts` function so that a bin count error is not triggered when the histogram is created for `e00200s`.

martinholmer:
@andersonfrailey, @MattHJensen and @feenberg,
The choice between zero and NAN for MTR wrt e00200s for MARS!=2 is admittedly arbitrary.
But the NAN choice causes contorted code (see `test_pufcsv.py` changes that needed to be made in this pull request), which is a substantial negative.  Why not make the zero choice and eliminates all the extra code contortions?

martinholmer:
After more thought on the matter of nan vs zero for MTR wrt `e00200s` for MARS!=2, I can see an advantage to nan: it is more likely to alert unwary users to the fact that the only sensible values are for filing units with MARS==2.

PR #1270, opened 2017-03-30
Remove redundant _AMT_thd_MarriedS policy parameter

martinholmer:
This pull request replaces `_AMT_thd_MarriedS` for `MARS == 3` with `AMT_em[MARS - 1]`, which has the same value.  This means the following code:
```
if MARS == 3:
    amtsepadd = max(0.,
                    min(AMT_thd_MarriedS, 0.25 * (c62100 - AMT_em_pe)))
else:
    amtsepadd = 0.
```
 turns into the code below:
```
if MARS == 3:
    amtsepadd = max(0.,
                    min(AMT_em[MARS - 1], AMT_prt * (c62100 - AMT_em_pe)))
else:
    amtsepadd = 0.
```

@MattHJensen @feenberg @Amy-Xu @andersonfrailey @GoFroggyRun @codykallen


feenberg:
On Wed, 29 Mar 2017, Martin Holmer wrote:

> 
> 
> if MARS == 3:
>     amtsepadd = max(0.,
>                     min(AMT_em[MARS - 1], AMT_prt * (c62100 - AMT_em_pe)))

I am curious - why use "MARS - 1" instead of "2" in the subscript? The 
guard insures that MARS must equal "3" for the assignment to occur.

dan feenberg

martinholmer:
In principle, pull request #1270 should create no differences in tax results.  But in practice, it did create rounding-error changes in AMT Income for 2019 and 20121.  The reason for these minor differences is that #1270 replaces a parameter whose actual value was known only through 2016 with an equivalent variable whose actual value is known through 2017.  So, the changes in AMT Income are generated because now the variable indexing starts in 2018 while before it started in 2017.

@MattHJensen @feenberg @Amy-Xu @andersonfrailey @GoFroggyRun @codykallen

martinholmer:
@feenberg asked:
>> ```
>> if MARS == 3:
>>     amtsepadd = max(0.,
>>                     min(AMT_em[MARS - 1], AMT_prt * (c62100 - AMT_em_pe)))
>>                                ^^^^^^^^
>> ```
>
> I am curious - why use "MARS - 1" instead of "2" in the subscript?
> The guard insures that MARS must equal "3" for the assignment to occur.

I made the `MARS - 1` explicit for clarity: it looks exactly like the use of all the other MARS-indexed policy parameters in the `functions.py` file.

I thought for a moment that using `2` could produce a slight reduction in execution speed, but discounted that idea because all the `functions.py` code is run through a "compiler" that I image will do that simple optimization automatically.

@MattHJensen @feenberg @Amy-Xu @andersonfrailey @GoFroggyRun @codykallen

PR #1289, opened 2017-04-12
Fix Ryan-Brady json reform

codykallen:
I added the repeal of the additional standard deduction for the aged and blind (which the plan folds into its expansion of the standard deduction). I also removed some trailing whitespace.

@MattHJensen


martinholmer:
@codykallen, Thanks for the improvement in the RyanBrady.json reform file.

PR #1292, opened 2017-04-12
Fixed reference to non-existing variable.

brittainhard:


martinholmer:
@brittainhard,  Thanks for fixing the validation of the `_II_brk3` policy parameter in #1292.

PR #1298, opened 2017-04-15
Capitalize section titles and add test of section title consistency

martinholmer:
This pull request makes modifications in the capitalization of words in the section titles in the `current_law_policy.json` file and adds a test to ensure that section_1 and section_2 titles are consistent across all policy parameters in the same section and subsection.  Also, it adds section_1 and section_2 keys that have a value of "" (the empty string) to parameters without any section keys.  And for parameters that had only a section_1 key, it adds a section_2 key with a value equal to the parameter's capitalized long_name (which is the way things are displayed on TaxBrain).

These changes are being proposed to resolve three separate problems.

(1) The new test will check automatically for section name inconsistencies.  See pull requests #1154 and #1276 for examples of why adding this test makes sense.

(2) Capitalizing the words in the section_1 and section_2 values and adding missing section_2 values will make it easier for developers to specify section names in TaxBrain.  Although I have not looked at the TaxBrain source code, it seems likely that what is being done now is that the section values are being read from `current_law_policy.json` and every word in each section title is being capitalized.  This has the undesirable effect of converting every acronym (AMT, AGI and FICA, for example) in the headings into a capitalized "word" (Amt, Agi and Fica, for example).  This second set of changes to `current_law_policy.json` mean that TaxBrain developers can simply use the section values without making any changes, and therefore, the acronyms will remain all uppercase.

(3) Adding empty-string values for section_1 and section_2 for parameters that formerly had no section_1 key is required to support the development of the HTML user documentation.  See pull request #1297 for the (as yet incomplete) user documentation.  The user documentation needs to handle **all** the policy parameters that can be used by Tax-Calculator and the `tc` CLI and the TaxBrain file-upload capability, not just the parameters that appear in the TaxBrain GUI.  So, now the indicator of a parameter not appearing in TaxBrain is having a section_1 value of "" (the empty string).

There is one remaining inconsistency in the section values.  The four "Universal Basic Income" parameters have non-empty section values even though it seems they are not to appear in TaxBrain.   These UBI parameters were made to disappear in TaxBrain in [TaxBrain pull request 530](https://github.com/OpenSourcePolicyCenter/webapp-public/pull/530) by some means other than setting their section_1 value to the empty string (or formerly by removing the section_1 key).  This situation suggests that the way TaxBrain developers decide which policy parameters to show is not completely driven by the data in the `current_law_policy.json` file.  I assume that was the goal, so it would seem we are not quite there yet.

Note that this UBI and other related issues have been resolved in commit a05d520 as described in a subsequent comment in this conversation.

@MattHJensen @PeterDSteinberg @brittainhard


martinholmer:
Commit a05d520 changes the `current_law_policy.json` section_1 and section_2 values to empty strings for all parameters that do not appear in TaxBrain version 0.8.2.

MattHJensen:
> This situation suggests that the way TaxBrain developers decide which policy parameters to show is not completely driven by the data in the current_law_policy.json file. I assume that was the goal, so it would seem we are not quite there yet.

We have the capability on TaxBrain to hide parameters even if they have section labels in current_law_policy.json. We would use this capability if we find a problem with a parameter but don't want to release an entirely new version of taxcalc to deal with it  -- this recently happened with the UBI parameters when we realized that the current puf.csv doesn't have data to support them.  Afterwards, I should have immediately removed the corresponding section headers in a commit to tax-calculator, but I forgot to do so. In otherwords, thank you for a05d520! 

> Commit a05d520 changes the current_law_policy.json section_1 and section_2 values to empty strings for all parameters that do not appear in TaxBrain version 0.8.2.

PR #1299, opened 2017-04-18
Add content to Policy Parameters section of user documentation

martinholmer:
This pull request adds policy parameter documentation extracted from the `current_law_policy.json` file.  You can see the current state of the user documentation [here](http://open-source-economics.github.io/Tax-Calculator/).


martinholmer:
Now that pull request #1299 has been merged, the user documentation contains information on each policy parameter.  I'm hoping people can take a quick look to give reactions to the content of the paragraph for each policy parameter.  Most important question is this: are there parameter attributes in `current_law_policy.json`  that should be included in each parameter paragraph that are not now included?  Any other comments or suggestions are welcome.  The as-yet-incomplete user documentation is [here](http://open-source-economics.github.io/Tax-Calculator/).

MattHJensen:
@martinholmer, what do you think about using `tc name` and `TB name` or `TaxBrain name` as opposed to `CLI name` and `GUI name`? Seems like we can teach someone what `tc` and `TaxBrain` are without requiring that they really internalize that one is a CLI and the other is a GUI.

MattHJensen:
@martinholmer, I don't have a suggestion about this one, but "known value" seems to underplay our work somewhat. For example, the latest "known value" listed for the individual income rates is 2013, but in fact we check every year that they haven't changed and just choose not to list every years' value. So we "know" that the top rate in 2016 is 39.6, yet the docs indicate that we're not really sure past 2013.

martinholmer:
@MattHJensen said:
> What do you think about using `tc Name` and `TB name` or `TaxBrain name` as opposed to `CLI name` and `GUI name`? Seems like we can teach someone what tc and TaxBrain are without requiring that they really internalize that one is a CLI and the other is a GUI (two new acronyms they may not be familiar with).

Good point.  I'll change that in the next day or so.

martinholmer:
@MattHJensen said:
> I don't have a suggestion about this one, but the terminology of "Known Values" seems to underplay our work somewhat. For example, the latest "known value" listed for the individual income rates is 2013, but in fact we check every year that they haven't changed and just choose not to list every years' value. So we "know" that the top rate in 2016 is 39.6, yet the docs indicate that we're not really sure past 2013.

Excellent point.  I'll add all the years for the non-indexed parameters to show that they've been checked, and therefore, are "known".

Issue #1275, opened 2017-04-03
Make an event to test OG-USA + B-Tax on Tax-Calc master change


PeterDSteinberg:
* [ ] Run `get_micro_data.py` from OG-USA
 * [ ] Run B-Tax regression

Use Travis CI environment variables and webhook options to test these just on changes to master branch.

MattHJensen:
I think it would be preferable to run these from og-USA and b-tax respectively on new taxcalc releases.

In tax-calculator, it would be best to find the most parsimonious tests possible that would identify breaking changes for b-tax and og-USA so that we don't dramatically affect the time it takes to run the taxcalc test suite.

On Apr 3, 2017, at 6:01 PM, Peter Steinberg <notifications@github.com<mailto:notifications@github.com>> wrote:


  *    Run get_micro_data.py from OG-USA
  *    Run B-Tax regression

Use Travis CI environment variables and webhook options to test these just on changes to master branch.

jdebacker:
As an FYI, the breaking changes that I've noticed in the past have been changes to variable names, changes to the variables available in the PUF, and changes to the list of variables for which the mtr function is available.

PeterDSteinberg:
There have also been breaking changes in the deploy repo - the parts where the interface (`runner`) functions are called and are inconsistent with changes in the Tax-Calculator, B-Tax, and OG-USA repos.  I'll think further how we may better test that integration as well.

PeterDSteinberg:
Here's the related B-Tax regression testing PR: https://github.com/open-source-economics/B-Tax/pull/150

PeterDSteinberg:
Is there any interest, @jdebacker, @MattHJensen @martinholmer , in have private puf.csv.gz-related tests running in Travis?  There is a way we can encrypt things to allow an install of `taxpuf` on Travis CI tests safely.  If so, I can add that to the checklist at the top of this issue.

jdebacker:
@PeterDSteinberg Let me be sure I understand.

I thought we currently did Travis CI testing of OG-USA and B-Tax with a puf.csv.gz file.  Is the right?

I thought the current puf.csv.gz file was fake data for testing - and that it did not have to be private like the actually puf.csv file.  Do I have this wrong?

PeterDSteinberg:
[...]

PR #1303, opened 2017-04-20
Comply with Python convention on naming private variables

martinholmer:
There is a long-standing and widely-accepted convention in Python programming that class variables whose name contains a leading underscore character are considered private variables that should not be accessed from outside of their class.

Here is what the [official Python documentation](https://docs.python.org/3/tutorial/classes.html#private-variables) says about this:
```
“Private” instance variables that cannot be accessed except from inside an object
don’t exist in Python. However, there is a convention that is followed by most Python
code: a name prefixed with an underscore (e.g. _spam) should be treated as a non-public
part of the API (whether it is a function, a method or a data member). It should be
considered an implementation detail and subject to change without notice.
```

And the Google Python Style Guide echoes this convention in its [Naming section](https://google.github.io/styleguide/pyguide.html#Naming).

In Tax-Calculator there is adherence to this convention with respect to parameters in the Policy class.  But the Records class does not consistently follow this convention.  Most of the calculated variables in the Records class do not not start with an underscore character because they are often referenced outside of the Records class code.  However, there are sixteen calculated variables that begin with an underscore and they are referenced often outside of the Records class.

There is no apparent rhyme or reason for why some calculated variables start with an underscore but most don't.  This makes it unnecessarily difficult to remember which calculated variables start with an underscore and which don't.

This situation sends the wrong message to an experienced Python programmer who may be thinking of contributing to the project.  It also creates a need to litter the code with comments that disable the `protected-access` warnings generated by `pylint` which does follow the private-variable naming convention.

In order to eliminate these inconsistencies, this pull request simply drops the leading underscore from sixteen calculated variables in the Records class.  Here is a list of those sixteen variables:

1.  _combined
2.  _earned
3.  _earned_p
4.  _earned_s
5.  _eitc
6.  _exact
7.  _expanded_income
8.  _iitax
9.  _num
10.  _payrolltax
11.  _refund
12.  _sep
13.  _sey
14.  _standard
15.  _surtax
16.  _taxbc

This pull request does not change any tax logic or tax results.  All the tests --- including the notebook tests --- pass.  And the documentation has been updated to reflect these changes.


MattHJensen:
This will require changes to OG-USA and B-Tax. cc @jdebacker and @rickecon.

martinholmer:
@MattHJensen said:
> This [PR #1303] will require changes to OG-USA and B-Tax.    cc @jdebacker and @rickecon 

But that is not what their comments in issue #1275 say.  I'm confused.

But even if you are correct, the needed OG-USA and B-Tax changes are trivial.

MattHJensen:
> But that is not what their comments in issue #1275 say. I'm confused.

You're right, but I think @rickecon's comment about og-usa was incomplete. B-Tax may not need any changes. 

> But even if you are correct, the needed OG-USA and B-Tax changes are trivial.

Agreed.

feenberg:
On Thu, 20 Apr 2017, Martin Holmer wrote:

> 
> There is a long-standing and widely-accepted convention in Python
> programming that class variables whose name contains a leading underscore
> character are considered private variables that should not be accessed from
> outside of their class.
> 
...

>  1. _combined
>  2. _earned
>  3. _earned_p
>  4. _earned_s
>  5. _eitc
>  6. _exact
>  7. _expanded_income
>  8. _iitax
>  9. _num
> 10. _payrolltax
> 11. _refund
> 12. _sep
> 13. _sey
> 14. _standard
> 15. _surtax
> 16. _taxbc

Most of these variable names seem to be carried over from my SAS code, 
where the underscore was used to avoid inadvertent clashing with user 
defined variables, not so much because there is no way to define a local 
variable in a SAS macro but because we wanted to give users some access to 
the variable if they wanted it. Obviously in Python there are different 
considerations.

dan

martinholmer:
@feenberg said with respect the the calculated variable renaming in pull request #1303:
> Most of these [sixteen] variable names seem to be carried over from my SAS code, where the underscore was used to avoid inadvertent clashing with user defined variables, not so much because there is no way to define a local variable in a SAS macro but because we wanted to give users some access to the variable if they wanted it.  Obviously in Python there are different considerations.

Thanks for the historical background.  I'm pleased that you are comfortable with dropping the leading underscore from each of these sixteen variable names.

PR #1307, opened 2017-04-21
Add description for c21060, itemized deductions before phase-out

martinholmer:
Add another description to the `records_variables.json` file.


martinholmer:
How can it be that adding a few characters to a documentation string in the `records_variables.json` file causes a test failure under Python 3.4 (even though there are no test failures under Python 2.7 and Python 3.5)?

PR #1323, opened 2017-04-28
Trump2017

codykallen:
This PR adds a JSON file with the reform released by the Trump Administration on April 26, 2017. The details of this reform are based on a one-page handout from the press conference, as well as a separate document released shortly afterward from the Treasury Department. Due to lack of detail, several assumptions about specifics were made in order to score this plan. The following section has the details provided in the one-page handout, along with my italicized comments on the assumptions we will make to score these specific reforms. 
******************************
Reducing the 7 tax brackets to 3 tax brackets of 10%, 25%, and 35%
_Assume that this is implemented by merging the 10 and 15% brackets into one 10% bracket; merging the 25, 28 and 33% brackets into one 25% bracket; and merging the 35 and 39% brackets into one 35% bracket. Note that the proposal could potentially mean that the 33% bracket should be merged into the 35% bracket instead, but we will assume that the proposal does not raise tax rates on any bracket._

Doubling the standard deduction.
_Assume this does not apply to the additional standard deduction. Also, since the proposal does not eliminate the personal exemption (unlike the campaign proposal), I assume the personal exemption will remain._

Providing tax relief for families with child and dependent care expenses
_Cannot model this due to lack of detail._

Eliminate targeted tax breaks that mainly benefit the wealthiest taxpayers
_As Cohn mentioned in the press conference, this includes eliminating the state and local tax deduction. Since they don’t specify any provisions to eliminate beyond itemized deductions, we cannot model them._

Protect the home ownership and charitable gift tax deductions
_Based on the Treasury document titled “The 2017 Tax Reform Bill,” the proposal would eliminate all itemized deductions except for mortgage interest and charitable contributions. We will score this from a baseline of current law. Note that if the AHCA passes, the baseline would change due to the expansion of the medical expense deduction, and we would have to impute medical expenses for those not claiming the deduction in order to score it properly._ 

Repeal the Alternative Minimum Tax
_Can model this._

Repeal the death tax
_Cannot model this._

Repeal the 3.8% Obamacare tax that hits small businesses and investment income.
_Can model this easily. No assumptions needed._

15% business tax rate
_We only apply this to pass-through income (i.e. sole proprietorship,, partnership and S-corporation income). We model this under the assumption that the IRS somehow prevents reclassification of labor compensation as business income. Potential alternative assumptions would involve reclassifying all wage and salary income accruing to those who also have pass-through income as pass-through income, or reclassifying all wage and salary income as pass-through income if firms and workers restructure their employment relationships to take advantage of the preferential tax rate. Note that the current approach can be modeled as is, but the income reclassification scenarios would require adding new capacity to Tax-Calculator to model._


codykallen:
@MattHJensen, could you review?

@martinholmer @andersonfrailey

taxcalc/reforms/Trump2017.json
```diff
@@ -0,0 +1,75 @@
...
+        "_II_rt4":
+            {"2017":  [0.25]},
+        "_II_rt5":
+            {"2017": [0.25]},
+        "_II_rt6":
+            {"2017": [0.35]},
+        "_II_rt7":
+            {"2017": [0.35]},
+        "_PT_rt1":
+            {"2017": [0.15]},
```
MattHJensen: I think _PT_rt1 and _PT_rt2 should probably be 0.1. I doubt they want to tax pass through income at a higher rate than wage and salary income

codykallen:
@MattHJensen commented

> I think _PT_rt1 and _PT_rt2 should probably be 0.1. I doubt they want to tax pass through income at a higher rate than wage and salary income

I agree that they likely wouldn't want to tax pass-through income at a higher rate. But the plan only states

> 15% business tax rate

If we apply your approach, do we apply a 10% rate only to the first bracket (10% under current law), or do we also apply it to the second bracket (15% under current law but 10% ordinary tax rate under the Trump proposal). 

@MattHJensen also said

> Although I still think it could make more sense to keep the Trump 2016 child care ald until it gets swapped out with something new.
> 

I disagree on this. Based on media reports and private discussions, it seems highly unlikely that the administration will proceed with the child care ALD. They are exploring a CDCC expansion instead.

codykallen:
@MattHJensen said 

> I think _PT_rt1 and _PT_rt2 should probably be 0.1. I doubt they want to tax pass through income at a higher rate than wage and salary income

On further consideration, I believe you are correct. Although the handout only states "15% business tax rate," the Treasury document states

> Small Business Owner/Operators Eligible for Business Rate

That seems to imply that they could opt for the lower wage-income rate if they choose. I will modify the json to make the 15% rate a top rate instead of a flat rate.

MattHJensen:
This looks great. Thanks for the contribution @codykallen.

MattHJensen:
Here is the TaxBrain version. @codykallen, look right? http://www.ospc.org/taxbrain/11522/

codykallen:
@MattHJensen, that looks right to me. That's what I got when I ran it.

MattHJensen:
Thanks @codykallen

Issue #1311, opened 2017-04-22
Questions about Itemized Deduction amounts


martinholmer:
Among the calculated output variables, we have the following deduction-related variables:
```
standard : standard deduction amount (zero for itemizers)
c04470   : itemized deduction amount after phase-out (zero for non-itemizers)
c21060   : itemized deduction amount before phase-out
c21040   : itemized deduction amount phased out
```
By definition, the following equation should hold for each filing unit's itemized deductions:
```
(1)  c21040 = c21060 - c04470
```
And also, it would seem (without thinking about it too much) that these relationships should hold:
```
(2)  standard > 0  <==>  c04470 = 0
(3)  standard == 0  <==> c04470 > 0
```
I know we have been around on this matter before, so I'm hoping somebody with a better memory (or ability to search GitHub for long past issues and pull requests) can help me out with the following two questions: (A) and (B).

These two questions are based on tabulations of the `--dump` output from this CLI run:
```
tc puf.csv 2017 --dump
```
So, the dump output is for 2017 current-law policy.  The dumped CSV file was imported into the dump table in an SQLite database (which is part of the Anaconda distribution we are using for Tax-Calculator) where it was tabulated using structured query language (SQL).  Here are some basic tabulation results:
```
sqlite> select count(*) from dump;
219814

sqlite> select count(*) from dump where standard==0.0 and c04470>0.0;
89522

sqlite> select count(*) from dump where standard>0.0 and c04470==0.0;
130054

sqlite> select count(*) from dump where standard==0.0 and c04470==0.0;
238

sqlite> select count(*) from dump where standard==0.0 and c04470==0.0
                                        and c21060==0.0 and c21040==0.0;
238

sqlite> select count(*) from dump
        where abs(c21040 - (c21060 - c04470))>0.011;
2019
sqlite> select RECID,standard,c21060,c21040,c04470 from dump
        where abs(c21040 - (c21060 - c04470))>0.011 limit 9;
144110,12700.0,0.0,13415.04,0.0
3242611,15200.0,0.0,9135.92,0.0
3569860,12700.0,0.0,4836.43,0.0
1349180,15200.0,0.0,2285.83,0.0
1349181,15200.0,0.0,2285.83,0.0
3418830,12700.0,0.0,35737.24,0.0
833050,12700.0,0.0,12079.3,0.0
520180,12700.0,0.0,11781.05,0.0
1498650,13950.0,0.0,66400.81,0.0

[...]

PR #1327, opened 2017-04-30
Fix c21040 logic and add several output consistency tests

martinholmer:
This pull request resolves issue #1311.  Now `c04470 == c21060 - c21040` is true for all records.
This change does not have any visible effect on the unit test results because the fix affects so few people (`c21040` was corrected for over 2000 units but they constitute less than 0.1 percent of the `puf.csv` sample weight) and not all of them have a positive AMT liability.


MattHJensen:
@martinholmer, thanks for this bug fix.

Issue #1325, opened 2017-04-28
Questions about the summary tax tables


martinholmer:
As part of developing the user documentation, I've been adding descriptions of several calculated output variables.  In the course of doing this, I've noticed some inconsistencies in the aggregate tables prepared by the Tax-Calculator (see, for example, the `create_distribution_table` utility function) and in the tables shown on the TaxBrain static results page.

Here is a TaxBrain results table for [run 11490](http://www.ospc.org/taxbrain/11490/):

![image](https://cloud.githubusercontent.com/assets/12170745/25550155/d8f9bec4-2c44-11e7-998b-5b5c9f8f6275.png)

The problem is that the numbers seem inconsistent as you move across the columns.  Focusing on the `All` row, we see two unexpected results.

First, the `Regular Tax` and the `AMT` amounts don't add up to the `Tax before Credits` amount --- most people would expect that sum to be equal to the `Tax before Credits` amount.

Second, when the `Non-refundable Credits` amount is subtracted from the `Tax before Credits` amount, the difference is not equal to the `Tax before Refundable Credits` amount --- most people would expect that difference to be equal to the `Tax before Refundable Credits` amount.

These inconsistencies are caused by using `c05200` rather than `taxbc` for the `Regular Tax` amount and by using `c07100` rather than `(c05800 - c09200)` for the `Non-refundable Credits` amount.  The latest version of the [user documentation](http://open-source-economics.github.io/Tax-Calculator/#output) clarifies the difference between the variables used now and the variables that seem more sensible.

I'd be happy to prepare a pull request that makes these changes, but I'm hoping for some kind of discussion here before I start doing that.

@MattHJensen @feenberg @Amy-Xu @andersonfrailey @GoFroggyRun @codykallen @zrisher 

P.S.  When @GoFroggyRun wrote the [`ODD_RECORDS.md`](https://github.com/open-source-economics/Tax-Calculator/blob/master/ODD_RECORDS.md) documentation many months ago, his discussion made it clear that `taxbc` (not `c05200`) represents `Regular Tax`.

MattHJensen:
@martinholmer, your assessment of the problem and solution makes sense to me. Very happy you found this.

martinholmer:
After further investigation, my analysis of the situation regarding non-refundable tax credits was wrong in issue #1325.  The `c07100` variable has always been total non-refundable credits used to reduce positive tax liability (which might be less than the total potential non-refundable credits).  

[...]

PR #1328, opened 2017-04-30
Fix calculated-variable documentation and fix diagnostic table

martinholmer:
This pull request partially resolves issue #1325.  Now the Tax-Calculator diagnostic table and TaxBrain distribution tables use `taxbc` (instead of `c05200`) for Regular Tax.  Also, the Tax-Calculator diagnostic table adds an Other Taxes row so that now the table variables "add up" in the way most users would expect.

What this pull request does not do is add an Other Taxes component to TaxBrain distribution tables, which means that the table columns will not "add up" the way most users would expect.  As mentioned in issue #1325, the pros and cons of doing this should be discussed in #1325 or in a separate issue.


martinholmer:
@MattHJensen, commit 67c4b87 in pull request #1328 adds a fix for the dropq, and hence TaxBrain, distribution tables.  Let me know if you think #1328 needs more work or if it can be merged as is.

MattHJensen:
This looks great. +1

PR #1332, opened 2017-05-02
Add template for economic responses

codykallen:
Since a previous PR separated the economic assumptions from the policy assumptions in the JSON reform files, this PR adds a new template for implementing specific economic responses for consumption, behavior and growdiff. This will hopefully make it easier for users to identify what economic assumptions they would like to make or experiment with to analyze specific reforms. 

As a follow-up to this, based on a conversation with @MattHJensen, we could solicit specific sets of economic parameters, and store these along with the template in a folder taxcalc/reforms/econ_assumptions. 

@martinholmer, could you verify that the descriptions for the growdiff parameters are correct?

@andersonfrailey @Amy-Xu @GoFroggyRun @feenberg


martinholmer:
@codykallen,  It's great that you've prepared an example of an assumption file in pull request #1332.
This task was coming up to the top of my to-do list, but you beat me to it.

Here are a few suggestions:

1.  Move the new file into a new directory called `taxcalc/responses` (on the same level as the `taxcalc/reforms` directory).  This move should eliminate the pytest errors you are encountering.  (As you seem understand, the errors are caused by the fact that `taxcalc/tests/test_reforms.py` assumes all the `*.json` files in the `taxcalc/reforms` directory are **reform** files.  You tried to move the new file but it didn't work, probably because you did `mv ... ...` at the command prompt instead of `git mv ... ...`.  Make sure your PR has only one copy of this new file.  If that doesn't work, send me an email and will figure out the problem.)

2. In the new JSON assumption response file, I strongly recommend that you remove almost all the comments (some of which are wrong) and point the reader to the Response Parameters section of the user documentation, which is [here](http://open-source-economics.github.io/Tax-Calculator/#params).  This approach keeps all the documentation in one place.  That means you (as the original author) will not have to constantly monitor things to make sure the comments in the new file are in sync with the evolution of Tax-Calculator.   If anybody wants to suggest improvements in the user document, they can raise an issue or prepare a pull request.  The documentation is prepared automatically by the `taxcalc/docs/make.py` script.

3. Add a new test to the `taxcalc/tests/test_reforms.py` file that does for all the JSON files in the `taxcalc/responses` directory what the failing test does for all the JSON files in the `taxcalc/reforms` directory.  Or, if you want to be totally systematic about adding the new test, you could put the new test in a new file called `taxcalc/tests/test_responses.py`.

@MattHJensen @feenberg @Amy-Xu @andersonfrailey @GoFroggyRun

codykallen:
@martinholmer, thanks for catching the moving error. I wasn't sure what was going wrong, but I've cleared that up now. 

@martinholmer said:

> I strongly recommend that you remove almost all the comments (some of which are wrong) and point the reader to the Response Parameters section of the user documentation

I expect that my implementation is incorrect for the growdiff_baseline dictionary. As for the descriptions of the specific parameters, this is intended to provide a simplified description to allow a user to identify the general effects of each parameter. That is why I  direct them to the source file for more detail. 

However, I believe it is important to include the comments in the template file. Including these details in the template makes it much easier to choose economic responses than forcing users to search through our documentation. I'm open to reorganizing the file for clarity, but it's important to have the parameter descriptions with the template. 

@martinholmer also suggested:

> Add a new test to the taxcalc/tests/test_reforms.py file that does for all the JSON files in the taxcalc/responses directory what the failing test does for all the JSON files in the taxcalc/reforms directory. 

I'll add a test for this shortly.

martinholmer:
@codykallen,  I don't see how the first line in the top comment section can be correct.
```
//Parameters for Partial-equilibrium dynamic analysis
```
What TaxBrain calls partial-equilibrium analysis is just the behavior parameters.  All the rest of the assumption parameters describe other kinds of responses.  For example, the growdiff_response parameters can be used to feed-back the macroeconomic general-equilibrium effects of a reform to the incomes of each tax filing unit.  And the growdiff_baseline parameters specify alternative baseline economic projections.  Neither of those sets of response parameters have anything to do with "partial-equilibrium analysis".

If you want to have duplicate documentation in your example response assumption file (and I can see your reasoning), then are you going to commit to keeping it up-to-date in the future?  Sometimes there is a trade-off between making the model easy for people to understand (because they don't have to point their web browser to a URL) and the staff effort required to keep the duplicate documentation correct as the model evolves.  It seems to me that users who get this far into using the model are not going to be put off by using a browser to read detailed documentation.  But if you want to pay the price of keeping this duplicate documentation up-to-date, then go ahead and include it.

martinholmer:
@codykallen,  Thanks for the template assumption file in pull request #1332.
I'm going to merge it now and then I'll update the `ASSUMPTIONS.md` file to point to it.

Issue #1350, opened 2017-05-05
Overly restrictive and inconsistent "pinned" package versions


martinholmer:
This afternoon I tried to follow the taxcalc package installation instructions in the [user documentation](http://open-source-economics.github.io/Tax-Calculator/#cli). I did this by creating a new user account on my Mac (which had the Apple version of Python and nothing from Continuum Analytics) and following the installation instructions.

I had no problem installing the offered Anaconda Python 2.7 distribution, which is the 4.3.1 version.
That version of Anaconda came with all the packages used by Tax-Calculator except for the `mock` package, which is not used in Tax-Calculator calculations, but is used in one of the thee-hundred some unit tests.  Among those packages included in Anaconda2 version 4.3.1 are the three packages that have "pinned" versions in the `environment.yml` file (which sets the development environment).  But those three packages have newer versions in 4.3.1 than in the `environment.yml` file.  Also, the `conda.recipie/meta.yaml` (which specifies package versions for the taxcalc package) has "pinned" versions that differ from those in the `environment.yml` file.  Here is a summary of the the three different sets of "pinned" package versions.
```
------------------------------------------------------------
package    environment.yml   meta.yaml   Mac Anaconda2 4.3.1
------------------------------------------------------------
numpy          =1.11.2        =1.11.2         1.11.3
pandas         =1.18.0       <=0.18.0         1.19.2
bokeh          =0.12.3        MISSING         0.12.4
------------------------------------------------------------
```
But, at least on a Mac, the Tax-Calculator works just fine with the more recent versions of these three packages.  After downloading a zip file containing the Tax-Calculator code at the tip of the master branch, bringing to the new tcuser account a copy of the `puf.csv` file, and conda installing the `mock` package, I passed all 305 unit tests.

So, this experiment seems to prove (at least on a Mac) that the Tax-Calculator will work just fine when using the newest versions of `numpy` and `pandas` and `bokeh`.  Given this, why shouldn't all the `=` and `<=` symbols in the `environment.yml` and `conda.recipie/meta.yaml` files be converted to `>=` symbols?

Unless I get a cogent reason for not making these changes soon, I'm going to prepare a pull request that makes those symbol changes.  These changes are essential otherwise the users who download the taxcalc package after installing Anaconda are going to be subjected to a whole set of DOWNGRADE messages when doing the conda install of taxcalc.  Here is what I got before I installed the `mock` package:
```
[...]

Issue #1339, opened 2017-05-03
No taxcalc package for Python 3.6


martinholmer:
For some time now the taxcalc package has been made for Python 2.7, Python 3.4, and Python 3.5.
You can see that from the following command:
```
$ conda search -c ospc taxcalc* | grep 0.8
                             0.8.0               np111py35_0  ospc            
                             0.8.0               np111py34_0  ospc            
                             0.8.0               np111py27_0  ospc            
                             0.8.1               np111py27_0  ospc            
                             0.8.1               np111py35_0  ospc            
                             0.8.1               np111py34_0  ospc            
                             0.8.2               np111py34_0  ospc            
                             0.8.2               np111py27_0  ospc            
                             0.8.2               np111py35_0  ospc            
```
However, Anaconda distributions of Python 3.4 and Python 3.5 are not available on the Continuum Analytics [download page](https://www.continuum.io/downloads); only Python 3.6 and Python 2.7 are available there.

So it seems as if we can recommend to CLI users only Python 2.7 because we have no conda taxcalc package for Python 3.6.  Right?  If not correct, what am I missing?

Why are we still making the out-of-date Python 3 packages?  Shouldn't we be making taxcalc packages only for the versions of Anaconda Python available from Continuum Analytics?

Answering those questions by saying we need to serve people using older versions of Python 3 is not convincing because if that were really the goal we would have been making packages for 3.0, 3.1, 3.2 and 3.3 as well.

@MattHJensen @feenberg @Amy-Xu @andersonfrailey @GoFroggyRun @codykallen 
@PeterDSteinberg @brittainhard

PeterDSteinberg:
The build is currently dying on Python 3.6 and has never succeed on Python 3.6.  There is a dependency issue that prevents successful conda build.  This issue can be a reminder to check into the package specs for taxcalc as they related to Python 3.6 and see what is the conflict.

martinholmer:
@PeterDSteinberg said:
> The build is currently dying on Python 3.6 and has never succeed on Python 3.6. There is a dependency issue that prevents successful conda build.

Can you tell us more about the "dependency issue"?

@MattHJensen @feenberg @Amy-Xu @andersonfrailey @GoFroggyRun @codykallen @brittainhard

PeterDSteinberg:
@martinholmer This is what I just got in Jenkins:
```
The following specifications were found to be in conflict:
[...]

PR #1300, opened 2017-04-18
Add content to Input Variables and Output Variables sections

martinholmer:
Derived content from the `records_variables.json` file.

Issue #381, opened 2015-10-10
taxcalc incompatible with numpy=1.10.0-py27_0


MattHJensen:
I tried updating numpy and had to reverse. When I upgrade to 1.10.0 I get these errors;
    
    def _blowup(self, year):
        >       self.e00200 *= self.BF.AWAGE[year]
        E       TypeError: Cannot cast ufunc multiply output from dtype('float64') to dtype('int64') with casting  rule 'same_kind'

talumbau:
Digging in to this. We currently rely on this behavior in numpy, valid up to v1.9.x:

```
>>> import numpy as np
>>> x = np.array([1000], dtype='int64')
>>> x *= 1.02
>>> x
array([1020])
```
This is now gone in v1.10, see the release notes on 1.9 for Compatibility Notes:

http://docs.scipy.org/doc/numpy/release.html

In our case, we are fine with this "unsafe" behavior (at least, I'm assuming we are). A couple solutions occur to me:

- change the dtype for all `e`-codes to `float64' instead of its current `int64`.

- modify `_blowup` to use the "unsafe" casting behavior we have been using so that we preserve the arithmetic here ( int type * floating point type -> floating point result -> stored into int type)

@MattHJensen, do you have a preferred solution?

MattHJensen:
My preferred solution would be to change the dtype for all e codes to float64 instead of int64

cc @martinholmer

talumbau:
Just got this idea from reading the numpy docs. For all operations where we currently do:
```
 a *= b
```
we could do

```
np.multiply(a, b, out=a, casting='unsafe')
```

this would preserve the aliasing (which is the reason why we do the in-place `*=` operations) and it would allow us to keep the dtypes the same. This is essentially what we were doing before (just with different syntax). One way to think about it is that in numpy v1.10, the argument `casting` was changed to `same_kind` instead of `unsafe`.

martinholmer:
@MattHJensen and T.J. (@talumbau),
Where do we stand on Tax-Calculator issue #381 and the use of numpy ```casting='unsafe'```?

I notice that as of 14-Feb-2016 the Records class does use ```casting='unsafe'```.  Is that wise?  The numpy developers presumably selected the adjective ```unsafe``` to send a not too subtle message.

[...]

PR #1351, opened 2017-05-06
New package version requirements

martinholmer:
This pull request started out as an experiment to see how the tests run on GitHub when using less restrictive package versions in the `environment.yml` and `conda.recipe/meta.yaml` files.

See subsequent comments in the #1351 conversion for details on how this pull request has evolved.

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -11,7 +11,6 @@ requirements:
     - pytest
     - setuptools
     - python
-    - mock
     - numpy >=1.11.2
     - pandas >=0.18.0
     - numba
@@ -21,7 +20,6 @@ requirements:
     - pytest
     - setuptools
     - python
-    - mock
     - numpy >=1.11.2
     - pandas >=0.18.0
     - numba
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -12,8 +12,8 @@ requirements:
     - setuptools
     - python
     - mock
-    - numpy ==1.11.2
-    - pandas <=0.18.0
+    - numpy >=1.11.2
+    - pandas >=0.18.0
     - numba
     - toolz
     - six
@@ -22,8 +22,8 @@ requirements:
     - setuptools
     - python
     - mock
-    - numpy ==1.11.2
-    - pandas <=0.18.0
+    - numpy >=1.11.2
+    - pandas >=0.18.0
     - numba
     - toolz
     - six
--- a/environment.yml
+++ b/environment.yml
@@ -2,16 +2,14 @@ name: taxcalc-dev
 dependencies:
 - pytest
 - setuptools
-- numpy =1.11.2
-- pandas =0.18.0
+- mock
+- numpy >=1.11.2
+- pandas >=0.18.0
 - numba
-- bokeh =0.12.3
 - toolz
 - six
-- ipython
-- ipython-notebook
+- bokeh >=0.12.3
 - pep8
 - pylint
-- mock
 - pip:
   - pytest-pep8
--- a/environment.yml
+++ b/environment.yml
@@ -2,13 +2,13 @@ name: taxcalc-dev
 dependencies:
 - pytest
 - setuptools
-- mock
 - numpy >=1.11.2
 - pandas >=0.18.0
 - numba
 - toolz
 - six
 - bokeh >=0.12.3
+- mock
 - pep8
 - pylint
 - coverage
--- a/environment.yml
+++ b/environment.yml
@@ -11,5 +11,6 @@ dependencies:
 - bokeh >=0.12.3
 - pep8
 - pylint
+- coverage
 - pip:
   - pytest-pep8
--- a/taxcalc/behavior.py
+++ b/taxcalc/behavior.py
@@ -183,9 +183,8 @@ class Behavior(ParametersBase):
             new_ltcg = calc_x.records.p23250 * exp_term
             ltcg_chg = new_ltcg - calc_x.records.p23250
         # calculate charitable giving effect
-        no_charity_response = \
-            calc_y.behavior.BE_charity.tolist() == \
-            [0.0, 0.0, 0.0]
+        no_charity_response = (calc_y.behavior.BE_charity.tolist() ==
+                               [0.0, 0.0, 0.0])
         if no_charity_response:
             c_charity_chg = np.zeros(calc_x.records.dim)
             nc_charity_chg = np.zeros(calc_x.records.dim)
@@ -301,6 +300,7 @@ class Behavior(ParametersBase):
         # assume behv response only for filing units with positive agi_m_ided
         delta_income = np.where(agi_m_ided > 0., taxinc_change, 0.)
         # allocate delta_income into three parts
+        old_settings = np.seterr(invalid='ignore')  # ignore runtime warnings
         delta_wage = np.where(agi_m_ided > 0.,
                               delta_income * calc.records.e00200 / agi_m_ided,
                               0.)
@@ -311,6 +311,7 @@ class Behavior(ParametersBase):
         delta_ided = np.where(agi_m_ided > 0.,
                               delta_income * ided / agi_m_ided,
                               0.)
+        np.seterr(**old_settings)  # reset warnings to old_settings
         # confirm that the three parts are consistent with delta_income
         assert np.allclose(delta_income, delta_wage + delta_oinc - delta_ided)
         # add the three parts to different calc.records variables
--- a/taxcalc/behavior.py
+++ b/taxcalc/behavior.py
@@ -298,7 +298,7 @@ class Behavior(ParametersBase):
                         calc.records.c04470)
         agi_m_ided = agi - ided
         # assume behv response only for filing units with positive agi_m_ided
-        pos = agi_m_ided > 0.
+        pos = np.array(agi_m_ided > 0., dtype=bool)
         delta_income = np.where(pos, taxinc_change, 0.)
         # allocate delta_income into three parts
         winc = calc.records.e00200
--- a/taxcalc/behavior.py
+++ b/taxcalc/behavior.py
@@ -298,25 +298,22 @@ class Behavior(ParametersBase):
                         calc.records.c04470)
         agi_m_ided = agi - ided
         # assume behv response only for filing units with positive agi_m_ided
-        delta_income = np.where(agi_m_ided > 0., taxinc_change, 0.)
+        pos = agi_m_ided > 0.
+        delta_income = np.where(pos, taxinc_change, 0.)
         # allocate delta_income into three parts
-        old_settings = np.seterr(invalid='ignore')  # ignore runtime warnings
-        delta_wage = np.where(agi_m_ided > 0.,
-                              delta_income * calc.records.e00200 / agi_m_ided,
-                              0.)
-        other_income = agi - calc.records.e00200
-        delta_oinc = np.where(agi_m_ided > 0.,
-                              delta_income * other_income / agi_m_ided,
-                              0.)
-        delta_ided = np.where(agi_m_ided > 0.,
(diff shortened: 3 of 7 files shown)


talumbau:
I see this PR as less of an "experiment" and more of a philosophical decision on how to move forward with the requirements for Tax-Calculator. This PR positions Tax-Calculator to be "forward compatible" with numpy and pandas. That is, by specifying a version that is greater than or equal to some static version, we can take advantage of new versions of these packages, and only insist that the environment have versions that are no older than some set version. This is a very reasonable path, since we can take advantage of bug fixes and performance improvements in those packages.

However, it is not without danger. The danger is that a new version of numpy or pandas will exhibit behavior that breaks the current Tax-Calculator. We have already seen this situation in the past. In issue #381, we had a situation where Tax-Calculator code worked with numpy 1.9, but not numpy 1.10, because there was a behavior change in numpy.

I don't think there is one "right" answer here. If we make the conda recipe such that Tax-Calculator packages are "forward compatible" with numpy and pandas, a user's environment will likely be modified less than if we insist on some set version. However, we increase the risk that, over time, the Tax-Calculator code will actually be incompatible with some future version of numpy or pandas (or some other package) and the package will still successfully install and run. This can lead to issues like #381 or perhaps more nefarious situations where the code does not give an error but executes incorrectly (an admittedly unlikely scenario).

martinholmer:
Pull request #1351 has grown out of an assignment given to me several months ago by @MattHJensen.  Matt asked me to develop user (not developer) documentation and a way to use Tax-Calculator that requires no Python programming knowledge.   I told him I could write the HTML documentation, but that others with more web programming experience would have provide a CSS if he wanted the user documentation to look fancy.

When giving me that assignment, Matt recommended a book to guide my effort: Karl Fogel, [Producing Open Source Software: How to Run a Successful Free Software Project, Second Edition](http://producingoss.com/). 

Even though I have read only through page 36 of Fogel's  book, it is clearly an excellent document with many useful observations.  Before explaining the thinking behind pull request #1351, I want to quote extensively from page 11 of Fogel's book:
> Chapter 2. Getting Started
>
> Starting a free software project is a twofold task. The software needs to acquire users, and to acquire developers.  These two needs are not necessarily in conflict, but the interaction between them adds some complexity to a project's initial presentation. Some information is useful for both audiences, some is useful only for one or the other. Both kinds of information should subscribe to the principle of scaled presentation: the degree of detail presented at each stage should correspond to the amount of time and effort put in by the reader at that stage. More effort should always result in more reward. When effort and reward do not correlate reliably, most people will lose faith and stop investing effort.
>
> The corollary to this is that appearances matter. Programmers, in particular, often don't like to believe this. Their love of substance over form is almost a point of professional pride. It's no accident that so many programmers exhibit an antipathy for marketing and public relations work, nor that professional graphic designers are often horrified at the designs programmers come up with on their own.
>
> This is a pity, because there are situations where form is substance, and project presentation is one of them. For example, the very first thing a visitor learns about a project is what its home page looks like.  This information is absorbed before any of the actual content on the site is comprehended — before any of the text has been read or links clicked on. However unjust it may be, people cannot stop themselves from forming an immediate first impression. The site's appearance signals whether care was taken in organizing the project's presentation. Humans have extremely sensitive antennae for detecting the investment of care. Most of us can tell in one glance whether a home page was thrown together quickly or was given serious thought. This is the first piece of information your project puts out, and the impression it creates will carry over to the rest of the project by association.
>
> Thus, while much of this chapter talks about the content your project should start out with, remember that its look and feel matter too. Because the project web site has to work for two different types of visitors — users and developers — special attention must be paid to clarity and directedness. Although this is not the place for a general treatise on web design, one principle is important enough to deserve mention, particularly when the site serves multiple (if overlapping) audiences: people should have a rough idea where a link goes before clicking on it. For example, it should be obvious from looking at the links to user documentation that they lead to user documentation, and not to, say, developer documentation.  Running a project is partly about supplying information, but it's also about supplying comfort. The mere presence of certain standard offerings, in expected places, reassures users and developers who are deciding whether they want to get involved. It says that this project has its act together, has anticipated the questions people will ask, and has made an effort to answer them in a way that requires minimal exertion on the part of the asker. By giving off this aura of preparedness, the project sends out a message: "Your time will not be wasted if you get involved," which is exactly what people need to hear.

One of my take-aways from this passage is that the getting-started experience for new users should be as simple as possible and, as Fogel says, supply comfort and reassure users.

So, in order to get an idea about what the new-user experience would be like when following the [installation directions](http://open-source-economics.github.io/Tax-Calculator/#cli-install-test), I created a new account on my Mac that had no Anaconda or Tax-Calculator and followed the installation directions.  

One thing I found is that Continuum Analytics offers Anaconda distributions for Python 2.7 and for Python 3.6, but there is no taxcalc package for Python 3.6 (see issue #1339 for more details).

So I followed the instruction at the Continuum Analytics download page and I had the Anaconda2 (version 4.3.1) Python 2.7 distribution installed in no time as all.  Everything worked smoothly.

Then I went to the next step and ran the `conda install -c ospc taxcalc` command.  Here is what I got (which, in my opinion, does not supply comfort and reassure new users):
```
iMac2:~ tcuser$ conda install -c ospc taxcalc
Fetching package metadata ...........
Solving package specifications: .

Package plan for installation in environment /Users/tcuser/anaconda:

The following NEW packages will be INSTALLED:
  Do
    mock:         2.0.0-py27_0           
    pbr:          1.10.0-py27_0          
    taxcalc:      0.8.2-np111py27_0  ospc

The following packages will be UPDATED:

    anaconda:     4.3.1-np111py27_0       --> custom-py27_0     
    conda:        4.3.14-py27_0           --> 4.3.17-py27_0     

The following packages will be DOWNGRADED due to dependency conflicts:

    mkl:          2017.0.1-0              --> 11.3.3-0          
    numexpr:      2.6.1-np111py27_2       --> 2.6.1-np111py27_1 
    numpy:        1.11.3-py27_0           --> 1.11.2-py27_0     
    pandas:       0.19.2-np111py27_1      --> 0.18.0-np111py27_0
    scikit-learn: 0.18.1-np111py27_1      --> 0.18.1-np111py27_0
    scipy:        0.18.1-np111py27_1      --> 0.18.1-np111py27_0
```
The first thought new users will probably have is "Why doesn't Tax-Calculator work with the version of Anaconda being offered by Continuum Analytics?"  or "Why do I have to downgrade my Anaconda distribution in order to use Tax-Calculator?"  This is not a good initial impression to leave with potential new users.

The goal of this pull request is to eliminate this downgrading experience.

@feenberg @PeterDSteinberg @brittainhard @talumbau
@Amy-Xu @andersonfrailey @GoFroggyRun @codykallen @zrisher

martinholmer:
#### Details of what was done in the development of pull request #1351

I followed Continuum Analytics instructions about how to uninstall my old Anaconda2 (version 2.3.0) and then installed Anaconda2 (version 4.3.1).  I did this because I could not figure out how to use the `conda update anaconda` command (outside of any conda environment) to get from version 2.3.0 to the newest version 4.3.1.

Then I edited the `environment.yml` and `conda.recipe/meta.yaml` files as shown in this pull request #1351.  The unit tests all passed on my computer after doing `conda install mock` and they all passed when I first submitted #1351 to GitHub (which showed that all the unit tests passed on Linux for Python 2.7, 3.4 and 3.5 and on Windows for Python 3.5).

However, I noticed two sets of warnings that were printed to the screen that I had never seen before: one set of warnings occurred when using the `--graphs` option of tc, the command-line interface to Tax-Calculator, and the other set of warnings occurred when executing the one reform in the `python reform_results.py` test that involves Behavior.  These are just warning messages about floating-point division and Tax-Calculator results are unchanged from before the upgrade from Anaconda2 2.3.0 to Anaconda2 4.3.1.

It turns out that both sets of warnings are caused by improvements in numpy warning logic that are in numpy 1.11.3, but not in numpy 1.11.2.  This pull request fixes the `utils.py` and `behavior.py` code so that there warnings are not generated.  Doing this forced me to learn more about the operation of the numpy.where method and how to avoid some of its shortcomings.  The changes I've made leave the code more streamlined and (probably) more efficient and I've learned more about how to use the numpy package efficiently.  I consider this to have been a positive experience.

I also learned the the conda-build package is not part of the core Anaconda distribution, so I added code to the `taxcalc/cli/install_local_taxcalc_package` bash script to install that package if it has not already been installed.  This is another small improvement from going through the Anaconda upgrade process.

And finally, my new appreciation for the numpy.seterr method has led me to strengthen the unit tests by converting all numpy warnings into errors so they can be detected in the tests.  This is done in `taxcalc/tests/config.py` so this conversation happens only for the pytest unit tests.

I have experienced just one flaw with Anaconda2 version 4.3.1: the pylint utility does not work correctly without an additional `conda install` because the conda recipe for pylint is incorrect as described [here](https://github.com/ContinuumIO/anaconda-recipes/issues/79).  As mentioned in that Continuum issue, the work-around is as follows:
```
conda install --channel conda-forge backports.functools_lru_cache
```

@MattHJensen @feenberg @PeterDSteinberg @brittainhard @talumbau
@Amy-Xu @andersonfrailey @GoFroggyRun @codykallen @zrisher

MattHJensen:
I see the advantage of allowing most users to work with taxcalc without needing to create a conda environment and without downgrading their other packages, but I'm still worried about the question I asked over at #1350:

> If we move to `>=` in `environment.yml`, will we _ever_ be able to guarantee that the taxcalc works, given that the packages on which we depend can push updates anytime w/o triggering our tests?

I think my problem is ignorance, because other projects also specify dependencies this way. For example, a [dependency for pandas](http://pandas.pydata.org/pandas-docs/stable/install.html#dependencies) is "NumPy: 1.7.1 or higher"

martinholmer:
@MattHJensen said about pull request #1351:
> I see the advantage of allowing most users to work with taxcalc without needing to create a conda environment and without downgrading their other packages, but I'm still worried about the question I asked over at #1350:
>
>> If we move to >= in `environment.yml`, will we ever be able to guarantee that the taxcalc works, given that the packages on which we depend can push updates anytime w/o triggering our tests?

That's a good question to pose.  Here is what I think the answer is.

There are two ways to guard against the problem Matt identified above.

1.  **Have all (most) developers of Tax-Calculator use the newest version of Anaconda.**  I've upgraded to Anaconda2 version 4.3.1 and will be happy to check each month to see when a newer version is available.  When that happens I can raise an issue so that everybody knows about the newer Anaconda version.

2. **Have the GitHub check-in tests always use the newest version of Anaconda.**  This seems to be happening already.  When I look at the tests run on pull request #1300 (three weeks ago), I see that on Linux python 2.7.13 was used to run the tests even though at the time I had python 2.7.12 installed on my computer.  I'm not one hundred percent sure about this because I can't find the code that sets up the GitHub tests anywhere in the Tax-Calculator.

If we do these two things (especially the second thing), any new version of numpy, pandas, bokeh, or any other package, that breaks Tax-Calculator will immediately be exposed.

This is the way I understand the situation.  Have I missed something?

@feenberg @Amy-Xu @andersonfrailey @GoFroggyRun @codykallen @zrisher
@PeterDSteinberg @brittainhard @talumbau

MattHJensen:
> If we do these two things (especially the second thing), any new version of numpy, pandas, bokeh, or any other package, that breaks Tax-Calculator will immediately be exposed.

Thanks for the explanation. My takeaway is that a dependency's update might break taxcalc, but we'll catch it very quickly if it does. That seems like an ok state of things for now, and we can always reconsider if it causes too many problems.

martinholmer:
@MattHJensen said about pull request #1351:
> My takeaway is that a dependency's update might break taxcalc, but we'll catch it very quickly if it does. That seems like an ok state of things for now, and we can always reconsider if it causes too many problems.

Yes, I think that is a reasonable way to think about it.

Shall we proceed and merge pull request #1351?

MattHJensen:
> Shall we proceed and merge pull request #1351?

+1

PR #1356, opened 2017-05-09
Update pylint-disable comments after upgrading pylint from 1.5.4 to 1.6.4

martinholmer:
This pull request changes (mostly deletes) several pylint-disable comments to reflect the new pylint version 1.6.4 that is part of the Anaconda 4.3.1 distribution.  One example of the changes is that the default maximum number of branches in a method/function allowed before a pylint warning is raised has been increased (so some disable=too-many-branches comments have been removed).

This pull request causes no change in tax logic or results.

### DIFF
--- a/taxcalc/behavior.py
+++ b/taxcalc/behavior.py
@@ -263,7 +263,6 @@ class Behavior(ParametersBase):
         """
         Check that behavioral-response elasticities have valid values.
         """
-        # pylint: disable=too-many-branches
         msg = '{} elasticity cannot be {} in year {}; value is {}'
         pos = 'positive'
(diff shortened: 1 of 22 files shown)


martinholmer:
Commit de096a8 fixes three utility tests that had delved deep into the internals of the pandas GroupBy object.  These three tests failed because of the pandas upgrade to 0.20.1, which just became available and which changed some of the internals of the GroupBy object.   The tests have been rewritten to use the new internals, which are described [in this pandas documentation](https://github.com/pandas-dev/pandas/blob/v0.20.1/pandas/core/indexes/interval.py#L95-L1014).

Issue #403, opened 2015-10-17
report deadweight loss


MattHJensen:
It would be nice to report estimates of deadweight loss given 
1. user-specified assumptions for the elasticity of taxable income (following [Feldstein 1999](http://www.jstor.org/stable/2646716?seq=1#page_scan_tab_contents)) or
2.  user-specified assumptions for taxable and total income elasticities and the marginal resource cost of sheltering (following [Chetty 2009](https://www.aeaweb.org/articles.php?doi=10.1257/pol.1.2.31)). 

cc @feenberg, @martinholmer, @viard.

MattHJensen:
Would this also give us the implied cost of an additional dollar of government spending, assuming  revenue is raised by the income+payroll taxes on the margin?

feenberg:
I know that Marty has asked me for DWL numbers in the past, but I don't 
think they are legitimate in this context.

The incrmental value of deadweight loss from a change in the income tax 
depends on all the taxes and increasing returns everywhere in the economy, 
not just on the ones in the tax model. The change in DWL from the sales 
tax, when the income tax is changed can easily outstrip the loss we would 
calculate in the model. Add in the changes in all the other taxes and 
increasing returns industries, and the delta DWL we would calculate is not 
dominant at all.

This is something that Kevin and Alan should have a role in deciding.

dan

Viard:
That’s an excellent point.

All taxes need to be included to compute deadweight loss of any tax.

MattHJensen:
@feenberg said:

>The incremental value of deadweight loss from a change in the income tax
depends on all the taxes and increasing returns everywhere in the economy,
not just on the ones in the tax model. The change in DWL from the sales
tax, when the income tax is changed can easily outstrip the loss we would
calculate in the model. Add in the changes in all the other taxes and
increasing returns industries, and the delta DWL we would calculate is not
dominant at all.

@feenberg, my understanding is that these factors mean we can only calculate a lower bound for the delta DWL. Is that right? 

cc @khecon.

Viard:
Of course, you actually need to know all of the other taxes in order to do a correct dynamic revenue estimate – the problem is not unique to deadweight loss.

feenberg:
On Sat, 17 Oct 2015, Viard wrote:

> Of course, you actually need to know all of the other taxes in order to do a correct
> dynamic revenue estimate – the problem is not unique to deadweight loss.
>

Well, I would characterize "other taxes" as second order for revenue 
[...]

PR #1365, opened 2017-05-12
Weighted average MTR function

codykallen:
This PR adds a function to calculate the weighted average effective marginal tax rate for all filers.

Based on a previous discussion with @jdebacker, there are complications to calculating a weighted average if the income type used for the weighting can be negative (capital gains, Sch. C, etc.). I give three methods to deal with this. 
- Use the absolute value of the income type for the weight.
- Restrict the average to only filers with positive income of that type.
- Restrict the average to only filers with positive taxable income.

I also added a test for the wmtr function.

### DIFF
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -441,22 +441,11 @@ class Calculator(object):
         Returns
             wmtr: overall weighted average marginal tax rate
         """
-        # check validity of income_type parameter
-        if income_type not in Calculator.MTR_VALID_VARIABLES:
-            msg = 'mtr income_type="{}" is not valid'
-            raise ValueError(msg.format(income_type))
-        # check validity of mtr_measure
-        if mtr_measure not in ['itax', 'ptax', 'combined']:
-            msg = 'mtr_measure is not "itax", "ptax" or "combined"'
-            raise ValueError(msg)
-        # check validity of weight_type
-        if weight_type not in ['none', 'income', 'abs']:
-            msg = 'weight_type is not "none", "income", or "abs"'
-            raise ValueError(msg)
-        # check validity of pos_control
-        if pos_control not in ['none', 'posti', 'posinc']:
-            msg = 'pos_control is not "none", "posti", or "posinc"'
-            raise ValueError(msg)
+        # check validity of parameter inputs
+        assert income_type in Calculator.MTR_VALID_VARIABLES
+        assert mtr_meaure in ['itax', 'ptax', 'combined']
+        assert weight_type in ['none', 'income', 'abs']
+        assert pos_control in ['none', 'posti', 'posinc']
         # collect mtr array
         (mtr_ptax, mtr_itax,
          mtr_combined) = self.mtr(variable_str=income_type)
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -443,7 +443,7 @@ class Calculator(object):
         """
         # check validity of parameter inputs
         assert income_type in Calculator.MTR_VALID_VARIABLES
-        assert mtr_meaure in ['itax', 'ptax', 'combined']
+        assert mtr_measure in ['itax', 'ptax', 'combined']
         assert weight_type in ['none', 'income', 'abs']
         assert pos_control in ['none', 'posti', 'posinc']
         # collect mtr array
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -406,6 +406,84 @@ class Calculator(object):
         # return the three marginal tax rate arrays
         return (mtr_payrolltax, mtr_incometax, mtr_combined)
 
+    def wmtr(self, income_type='e00200p', mtr_measure='combined',
+             weight_type='income', pos_control='none'):
+    	"""
+        Calculates the weighted average marginal tax rate on an income type.
+
+        Parameters
+        ----------
+        income_type: string
+            specifies type of income or expense that is increased to compute
+            the marginal tax rates. Valid values are the same as for
+            variable_str in the Calculator.mtr function.
+
+        mtr_measure: string
+            options:
+                'itax': marginal individual income tax rate;
+                'ptax': marginal payroll tax rate; and
+                'combined': sum of marginal income and payroll tax rates.
+            specifies which type of marginal tax rate to use
+
+        weight_type: string
+            options:
+                'none': simple average of MTRs
+                'income': average of MTRs, weighted by income_type
+                'abs': average of MTRs, weighted by abs(income_type)
+            specifies the weighting formula to use
+        pos_control: string
+            options:
+                'none': include all filing units
+                'posti': exclude filing units with negative taxable income
+                'posinc': exclude filing units with negative of income type
+            excludes filing units with negative income
+
+        Returns
+            wmtr: overall weighted average marginal tax rate
+        """
+        # check validity of income_type parameter
+        if income_type not in Calculator.MTR_VALID_VARIABLES:
+            msg = 'mtr income_type="{}" is not valid'
+            raise ValueError(msg.format(income_type))
+        # check validity of mtr_measure
+        if mtr_measure not in ['itax', 'ptax', 'combined']:
+            msg = 'mtr_measure is not "itax", "ptax" or "combined"'
+            raise ValueError(msg)
+        # check validity of weight_type
+        if weight_type not in ['none', 'income', 'abs']:
+            msg = 'weight_type is not "none", "income", or "abs"'
+            raise ValueError(msg)
+        # check validity of pos_control
+        if pos_control not in ['none', 'posti', 'posinc']:
+            msg = 'pos_control is not "none", "posti", or "abs"'
+            raise ValueError(msg)
+        # collect mtr array
+        (mtr_ptax, mtr_itax,
+         mtr_combined) = self.mtr(variable_str=income_type)
+        income = getattr(self.records, income_type)
+        if mtr_measure == 'itax':
+            mtr1 = mtr_itax
+        elif mtr_measure == 'ptax':
+            mtr1 = mtr_ptax
+        elif mtr_measure == 'combined':
+            mtr1 = mtr_combined
+        # select weights
+        if weight_type == 'none':
+            wgt = self.records.s006
+        elif weight_type == 'income':
+            wgt = self.records.s006 * income
+        elif weight_type == 'abs':
+            wgt = (self.records.s006 *
+                   np.abs(income))
+        # select units to exclude (or not)
+        if pos_control == 'posti':
+            wgt = np.where(self.records.c04800 > 0, wgt, 0.)
+        elif pos_control == 'posinc':
+            wgt = np.where(income > 0, wgt, 0)
+        # calculate weighted average marginal tax rate
+        wmtr = sum(mtr1 * wgt) / sum(wgt)
+        return wmtr
+
     def current_law_version(self):
         """
         Return Calculator object same as self except with current-law policy.
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -455,7 +455,7 @@ class Calculator(object):
             raise ValueError(msg)
         # check validity of pos_control
         if pos_control not in ['none', 'posti', 'posinc']:
-            msg = 'pos_control is not "none", "posti", or "abs"'
+            msg = 'pos_control is not "none", "posti", or "posinc"'
             raise ValueError(msg)
         # collect mtr array
         (mtr_ptax, mtr_itax,
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -408,7 +408,7 @@ class Calculator(object):
 
     def wmtr(self, income_type='e00200p', mtr_measure='combined',
              weight_type='income', pos_control='none'):
-    	"""
+        """
         Calculates the weighted average marginal tax rate on an income type.
 
         Parameters
@@ -473,8 +473,7 @@ class Calculator(object):
         elif weight_type == 'income':
             wgt = self.records.s006 * income
         elif weight_type == 'abs':
-            wgt = (self.records.s006 *
-                   np.abs(income))
+            wgt = (self.records.s006 * np.abs(income))
         # select units to exclude (or not)
         if pos_control == 'posti':
             wgt = np.where(self.records.c04800 > 0, wgt, 0.)
--- a/taxcalc/tests/test_calculate.py
+++ b/taxcalc/tests/test_calculate.py
@@ -214,9 +214,9 @@ def test_Calculator_wmtr(records_2009):
     calc = Calculator(policy=Policy(), records=records_2009)
     recs_pre_e00200p = copy.deepcopy(calc.records.e00200p)
     # test calling Calculator.wmtr on several income types
-    assert type(calc.wmtr(income_type='e00200p')) == numpy.float64
-    assert type(calc.wmtr(income_type='p23250')) == numpy.float64
-    assert type(calc.wmtr(income_type='e00650')) == numpy.float64
+    assert type(calc.wmtr(income_type='e00200p')) == np.float64
+    assert type(calc.wmtr(income_type='p23250')) == np.float64
+    assert type(calc.wmtr(income_type='e00650')) == np.float64
     # Since e00200p is non-negative, test that different methods
     #    to control for that produce the same result
     wmtr1 = calc.wmtr(income_type='e00200p', weight_type='income')
--- a/taxcalc/tests/test_calculate.py
+++ b/taxcalc/tests/test_calculate.py
@@ -214,14 +214,14 @@ def test_Calculator_wmtr(records_2009):
     calc = Calculator(policy=Policy(), records=records_2009)
     recs_pre_e00200p = copy.deepcopy(calc.records.e00200p)
     # test calling Calculator.wmtr on several income types
-    assert type(wmtr(income_type='e00200p')) == numpy.float64
-    assert type(wmtr(income_type='p23250')) == numpy.float64
-    assert type(wmtr(income_type='e00650')) == numpy.float64
+    assert type(calc.wmtr(income_type='e00200p')) == numpy.float64
+    assert type(calc.wmtr(income_type='p23250')) == numpy.float64
+    assert type(calc.wmtr(income_type='e00650')) == numpy.float64
     # Since e00200p is non-negative, test that different methods
     #    to control for that produce the same result
-    wmtr1 = wmtr(income_type='e00200p', weight_type='income')
-    wmtr2 = wmtr(income_type='e00200p', weight_type='abs')
-    wmtr3 = wmtr(income_type='e00200p', pos_control='posinc')
+    wmtr1 = calc.wmtr(income_type='e00200p', weight_type='income')
+    wmtr2 = calc.wmtr(income_type='e00200p', weight_type='abs')
+    wmtr3 = calc.wmtr(income_type='e00200p', pos_control='posinc')
     assert np.allclose(wmtr1, wmtr2, wmtr3)
 
 
--- a/taxcalc/tests/test_calculate.py
+++ b/taxcalc/tests/test_calculate.py
@@ -216,7 +216,7 @@ def test_Calculator_wmtr(records_2009):
     # test calling Calculator.wmtr on several income types
     assert type(calc.wmtr(income_type='e00200p')) == np.float64
     assert type(calc.wmtr(income_type='p23250')) == np.float64
-    assert type(calc.wmtr(income_type='e00650')) == np.float64
+    assert type(calc.wmtr(income_type='e00900p')) == np.float64
     # Since e00200p is non-negative, test that different methods
     #    to control for that produce the same result
     wmtr1 = calc.wmtr(income_type='e00200p', weight_type='income')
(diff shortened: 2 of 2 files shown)


codykallen:
@MattHJensen @martinholmer @feenberg @andersonfrailey @Amy-Xu @GoFroggyRun

feenberg:
On Fri, 12 May 2017, codykallen wrote:

> 
> This PR adds a function to calculate the weighted average effective marginal
> tax rate for all filers.
> 
> Based on a previous discussion with @jdebacker, there are complications to
> calculating a weighted average if the income type used for the weighting can
> be negative (capital gains, Sch. C, etc.). I give three methods to deal with
> this.
>
>  *  Use the absolute value of the income type for the weight.
>  *  Restrict the average to only filers with positive income of that type.
>  *  Restrict the average to only filers with positive taxable income.
>

I usually increase all income with a factor of 1.01 and then divide the 
weighted aggregate tax by the weighted aggregate change in income. I think 
this is the same as your 2nd method. It has the advantage that wild 
marginal tax rates are unlikely to skew the result but notches still 
contribute an appropriate amount to the average.

Your third method (using only non-zero incomes) is probably equivalent to 
my method, because 1.01*zero is still zero.

The advantage of these marginal tax rates is that such a rate will give a 
good prediction of the revenue effect of income changes that are 
proportional to existing income. In long run growth both income and losses 
will grow, so there is a good argument for using the signed value of each 
income time. I don't see a value to using the absolute value. It might be 
argued that in the short run an improvement in economic conditions will 
increase income and reduce losses, which might be the justification for 
using absolute values or just neglecting negative values, but I have never 
actually seen that done and would not favor it.

dan

martinholmer:
@feenberg said about pull request #1365:
> I usually increase all income with a factor of 1.01 and then divide the weighted aggregate tax by the weighted aggregate change in income.  I think this is the same as your 2nd method [Restrict the average to only filers with positive income of that type].  It has the advantage that wild marginal tax rates are unlikely to skew the result but notches still contribute an appropriate amount to the average.

It seems to me that Dan's method is more straightforward to explain and easier to program.

We just work with aggregates so that the weighted MTR is just the sum of extra tax revenue generated by a marginal increase in income divided by the sum of the marginal increase in income across all the filing units.  Isn't it as simple as that?

However, I don't think Dan's view that his simple method is equivalent to the proposed second method is true.  Dan never drops any filing units from his calculation while the second method does drop filing units with zero and with negative taxable incomes, so it seems as if the answers cannot be the same.

@codykallen, will Dan's simple method meet your tax analysis needs?  Speaking of analysis needs, can you give us an example of how the weighted marginal tax rate (for the whole population) would come into play in your tax analysis?

@MattHJensen @Amy-Xu @andersonfrailey @GoFroggyRun

codykallen:
@feenberg said

> In long run growth both income and losses will grow, so there is a good argument for using the signed value of each income time. I don't see a value to using the absolute value. 

Although in the long run income and losses will grow, there remains an important consideration in dealing with negative incomes, largely due to the loss limitations in the tax system. Consider a filer with negative long-run capital gains. Suppose that this loss exceeds the loss limitation of $3000. In this case, a marginal change in this person's income would have no effect on his taxes, so his MTR on ltcg is zero. However, when computing the weighted average, this person contributes zero to the numerator (since his MTR is zero), but his negative income decreases the denominator, so the overall weighted average actually increases even though this unit's MTR is zero. This problem actually results in substantial differences by weighting method, and it is why we introduced the absolute value weighting. 

@feenberg suggested

> I usually increase all income with a factor of 1.01 and then divide the weighted aggregate tax by the weighted aggregate change in income. 

This is not unreasonable, but it is a proportional approach, rather than a marginal approach. I believe a more common standard is to increase by the same dollar amount, not by the same proportion. For example, TPC uses a marginal increase of $1000, and CBO uses a marginal increase of $100 or $1000, depending on the publication. 

@feenberg said

> In long run growth both income and losses will grow, so there is a good argument for using the signed value of each income time. 

Using a proportional increase also makes losses more severe, but the intent of the MTR calculation is to identify the tax disincentive to earning additional income of the given type, whereas the proportional approach actually measures the marginal tax change to income growth. 

@martinholmer asked

> Can you give us an example of how the weighted marginal tax rate (for the whole population) would come into play in your tax analysis?

These aggregates are used in B-Tax for calculating the METR on new corporate and non-corporate investment. It's also common to produce weighted average MTRs when analyzing tax reform proposals. These are useful for understanding the aggregate labor supply incentives and to save/invest. As examples, see Table 7 in TPC's analyses of the [House GOP tax plan](http://www.taxpolicycenter.org/publications/analysis-house-gop-tax-plan/full) and [Trump's tax plan](http://www.taxpolicycenter.org/publications/analysis-donald-trumps-revised-tax-plan/full).

martinholmer:
@martinholmer said:
> Why don't we just work with aggregates so that the weighted MTR is just the sum of extra tax revenue generated by a marginal increase in income divided by the sum of the marginal increase in income across all the filing units. Isn't it as simple as that?

I was thinking that the "marginal increase in income" would be $0.01 (the same as used by default in the `Calculator.mtr` method) or you could make it any other fixed amount (like the 100.00 used by CBO).  I agree that using a proportional increase in income is not the best idea.

@codykallen, why not just calculate weighted MTR this way and then you don't have to mess around with many of the problems you're dealing with in the current pull request?  This approach would seem to require only three parameters:  `income_type` and `mtr_measure` plus `marginal_amount` which would default to 0.01 (or some other value of your choice) to give you the flexibility of specifying a large "marginal increase in income" (like CBO and TPC do).  Wouldn't a statistic computed in this way, meet you tax analysis needs?

Actually, if you are going to use a large "marginal increase in income" (like CBO's 100 and TPC's 1000), then building up the weighted MTR from the filing unit MTR values (as in the current version of your pull request) is not absolutely correct because those MTR value are based on a small (one cent) "marginal increase in income".

@MattHJensen @feenberg

feenberg:
On Sat, 13 May 2017, Martin Holmer wrote:

> 
> @feenberg said about pull request #1365:
>
>       I usually increase all income with a factor of 1.01 and then
>       divide the weighted aggregate tax by the weighted aggregate
>       change in income. I think this is the same as your 2nd method
>       [Restrict the average to only filers with positive income of
>       that type]. It has the advantage that wild marginal tax rates
>       are unlikely to skew the result but notches still contribute an
>       appropriate amount to the average.
> 
> It seems to me that Dan's method is more straightforward to explain and
> easier to program.
> 
> We just work with aggregates so that the weighted MTR is just the sum of
> extra tax revenue generated by a marginal increase in income divided by the
> sum of the marginal increase in income across all the filing units. Isn't it
> as simple as that?
> 
> However, I don't think Dan's view that his simple method is equivalent to
> the proposed second method is true. Dan never drops any filing units from
> his calculation while the second method does drop filing units with zero and
> with negative taxable incomes, so it seems as if the answers cannot be the
> same.

I wouldn't think it desirable to drop tax units with no taxable income, as 
many will have AMT income and owe AMT. We don't want to ignore that tax.

dan feenberg

feenberg:
On Sat, 13 May 2017, codykallen wrote:

> 
> @feenberg said
>
>       In long run growth both income and losses will grow, so there is
>       a good argument for using the signed value of each income time.
>       I don't see a value to using the absolute value.
> 
> Although in the long run income and losses will grow, there remains an
> important consideration in dealing with negative incomes, largely due to the
> loss limitations in the tax system. Consider a filer with negative long-run
> capital gains. Suppose that this loss exceeds the loss limitation of $3000.
> In this case, a marginal change in this person's income would have no effect
> on his taxes, so his MTR on ltcg is zero. However, when computing the
> weighted average, this person contributes zero to the numerator (since his
> MTR is zero), but his negative income decreases the denominator, so the
> overall weighted average actually increases even though this unit's MTR is
> zero. This problem actually results in substantial differences by weighting
> method, and it is why we introduced the absolute value weighting.

I limit capital losses to $3,000 (or whatever the statutory limit is for 
the year in question) when calculating income. Then a change in the loss 
doesn't affect income unless it also affects taxable income.

dan

feenberg:
On Sat, 13 May 2017, Martin Holmer wrote:

> 
> @martinholmer said:
>
>       Why don't we just work with aggregates so that the weighted MTR
>       is just the sum of extra tax revenue generated by a marginal
>       increase in income divided by the sum of the marginal increase
>       in income across all the filing units. Isn't it as simple as
>       that?
> 
> I was thinking that the "marginal increase in income" would be $0.01 (the
> same as used by default in the Calculator.mtr method) or you could make it
> any other fixed amount (like the 100.00 used by CBO). I agree that using a
> proportional increase in income is not the best idea.

Do we need an MTR for each taxpayer, or only for aggregates? If the 
latter, my method is best and easiest. If the former, you need to decide 
what to do about notches in the law. If you add a penny and tax changes by 
$10, that is a 100,000% tax rate. A few such will come up. You can 
top-code the MTR, or use a larger finite difference or try subtracting a 
penny for those records - I have used all of those methods at times.

Another technique I have used to make sure all AGI amounts are not an even 
dollar amount by starting each taxpayer with .001 in other income. That 
can reduce the number of notches encountered without introducing much 
bias.

dan

codykallen:
@feenberg said

> I wouldn't think it desirable to drop tax units with no taxable income, as many will have AMT income and owe AMT. We don't want to ignore that tax.

I agree. I only included it here because that option is included in the mtr calculation in B-Tax. 

@feenberg suggested

> I limit capital losses to $3,000 (or whatever the statutory limit is for the year in question) when calculating income. Then a change in the loss doesn't affect income unless it also affects taxable income.

That seems like it could introduce some problems with the interaction between long term losses, short term gains and the loss limitation. If you're doing aggregates, there seems no reason to limit losses.

codykallen:
@feenberg asked

> Do we need an MTR for each taxpayer, or only for aggregates? If the latter, my method is best and easiest. If the former, you need to decide what to do about notches in the law. If you add a penny and tax changes by $10, that is a 100,000% tax rate. 

We already have a function that calculates the MTR for each taxpayer. I'm fine with switching this to calculate using aggregates (and that would largely eliminate those outliers from notches), but the "marginal" income change should be from adding (or subtracting) a fixed amount to each income type for each person. 

The proportional approach for the aggregate measure is not actually a marginal tax rate measure, but rather a measure of marginal tax revenue from additional economic activity (since it increases positive incomes but also decreases negative incomes). The weighted average marginal tax rate is intended as a measure of the aggregate disincentive to earning additional income.

martinholmer:
@codykallen said about calculating weighted marginal tax rates in #1365:
> I'm fine with switching [#1365] to calculate using aggregates (and that would largely eliminate those outliers from notches), but the "marginal" income change should be from adding (or subtracting) a fixed amount to each income type for each person.

I agree with you completely, @codykallen.  I think you'll find an aggregate approach much easier to implement and to explain.

@MattHJensen @feenberg

feenberg:
On Sat, 13 May 2017, codykallen wrote:

> 
> 
> @feenberg suggested
>
>       I limit capital losses to $3,000 (or whatever the statutory limit is for the year in
>       question) when calculating income. Then a change in the loss doesn't affect income unless
>       it also affects taxable income.
> 
> That seems like it could introduce some problems with the interaction between long term losses, short
> term gains and the loss limitation. If you're doing aggregates, there seems no reason to limit losses.

Please use AGI plus and minus whatever adjustments seem useful. That way 
capital gains enter income only to the extent that they enter the 
calculation of tax.

dan feenberg

codykallen:
@feenberg said

> Please use AGI plus and minus whatever adjustments seem useful. That way capital gains enter income only to the extent that they enter the calculation of tax.

That seems incorrect as well. If we are using aggregates, then we should divide the change in taxes paid by the change in income. But AGI is not the right measure for full income, but rather an income measure for tax purposes. For example, suppose that we implemented a reform to make charitable contributions an above the line deduction (ALD) instead of an itemized deduction, so that anyone, including non-itemizers, could use it. If we then attempted to calculate the weighted average marginal tax rate on charitable giving (which should be negative), we would decrease the total amount of taxes paid but would also decrease total AGI, resulting in a positive weighted MTR calculation, which is clearly incorrect. 

The change in AGI is the wrong measure to use. As I said, if we are using aggregates, then we should divide the change in taxes paid by the change in income. It would seem to me that the appropriate calculation would use the aggregate income of the type for which we are trying to calculate the weighted average MTR. 

For example, suppose we wanted to calculate the weighted average MTR on long term gains (p23250). Let's suppose calc1 is the base and calc2 is the version with some incremental income change applied to p23250. 

`Change in total tax = ((calc2.records.combined - calc1.records.combined) * calc1.records.s006).sum()`
`Change in total income = ((calc2.records.p23250 - calc1.records.p23250) * calc1.records.s006).sum()`
`WMTR = (Change in total tax) / (Change in total income)`

It seems to me that for the denominator we should use the entire change in the income base, not just the change in taxable base.

martinholmer:
@codykallen said:
> As I said, if we are using aggregates, then we should divide the change in taxes paid by the change in income.  It would seem to me that the appropriate calculation would use the aggregate income of the type for which we are trying to calculate the weighted average MTR.

I agree.

> For example, suppose we wanted to calculate the weighted average MTR on long term gains (p23250). Let's suppose calc1 is the base and calc2 is the version with some incremental income change applied to p23250.
>
>```
>Change in total tax = ((calc2.records.combined - calc1.records.combined) * calc1.records.s006).sum()
>Change in total income = ((calc2.records.p23250 - calc1.records.p23250) * calc1.records.s006).sum()
>WMTR = (Change in total tax) / (Change in total income)
>```
>
> It seems to me that for the denominator we should use the entire change in the income base, not just the change in taxable base.

What @codykallen is saying here makes perfect sense to me.

@MattHJensen @feenberg

codykallen:
@martinholmer, @feenberg, I've discovered a new problem with using aggregates. As it turns out, applying the aggregate approach with a uniform increment (e.g. $100) is equivalent to taking a simple average of all MTRs. However, applying the aggregate approach with a proportional increment (e.g. 1%) is equivalent to taking a weighted average of all MTRs if they were calculated with a proportional approach. See the attached file for the math behind it. Coding in these different approaches confirms this. 

[WMTR math.docx](https://github.com/open-source-economics/Tax-Calculator/files/999622/WMTR.math.docx)

Unfortunately, this implies that the aggregate proportional approach is subject to the same problems that occur when taking a weighted average of MTRs that includes filing units with negative income. As I mentioned earlier in this thread,

> Although in the long run income and losses will grow, there remains an important consideration in dealing with negative incomes, largely due to the loss limitations in the tax system. Consider a filer with negative long-run capital gains. Suppose that this loss exceeds the loss limitation of $3000. In this case, a marginal change in this person's income would have no effect on his taxes, so his MTR on ltcg is zero. However, when computing the weighted average, this person contributes zero to the numerator (since his MTR is zero), but his negative income decreases the denominator, so the overall weighted average actually increases even though this unit's MTR is zero. This problem actually results in substantial differences by weighting method. 

For example, if you use aggregate proportional approach for p23250, using the equation from my previous comment:

> WMTR = (Change in total tax) / (Change in total long-term gains)

This results in a wMTR of 8.15, which is absurd. Note that taking a weighted average of MTRs calculated using a proportional increase produces the same wMTR, and taking a weighted average of MTRs calculated using a uniform increment (increase of $0.01) gives a wMTR of 8.14.

This brings us back to the original screening methods I included. For example, although there isn't a good economic reason for using an absolute value weighting, it allows us to include the zero MTRs for those with losses that exceed the loss limitation.

martinholmer:
@codykallen said:
> Unfortunately, this implies that the aggregate proportional approach is subject to the same problems that occur when taking a weighted average of MTRs that includes filing units with negative income. 

OK.  But you already said you didn't think an "aggregate proportional approach" made sense and I agreed.  Your finding just seems to strengthen the case against the "aggregate proportional approach".

> As it turns out, applying the aggregate approach with a uniform increment (e.g. $100) is equivalent to taking a simple average of all MTRs.

OK.  That's good to know.  But this is the approach that CBO and TPC use, right?
Why not proceed with this uniform-increment aggregate approach?

@MattHJensen @feenberg

codykallen:
@martinholmer said

> But this is the approach [aggregate approach with a uniform increment] that CBO and TPC use, right? Why not proceed with this uniform-increment aggregate approach?

No. The approach used by CBO and TPC is to calculate individual MTRs using a uniform increment approach, and then calculate a weighted average of those MTRs. 

I found some CBO [estimates](https://www.cbo.gov/sites/default/files/113th-congress-2013-2014/reports/49817-taxingcapitalincome0.pdf) (Table A-4), and I tested the wmtr function with different screening/weighting measures, the aggregate proportional approach using different income base measures (the income type, AGI, and taxable gain) in the denominator, and a separate weighted average using taxable gain. 

Using the aggregate proportional approach, the income type base worked well for qualified dividends but very badly for long-term gains and short-term gains. Using AGI in the base worked fairly well for qualified dividends and long-term gains but is 8 percentage points off for short-term gains. 

Using the weighted average approach with a uniform incremental change, is very accurate for qualified dividends (as is the proportional aggregate using the income type as the base). Weighting by taxable gain was generally the most accurate method for long-term and short-term gains (not applicable to dividends).

feenberg:
On Sun, 14 May 2017, Martin Holmer wrote:

> 
> @codykallen said:
>
>       Unfortunately, this implies that the aggregate proportional
>       approach is subject to the same problems that occur when taking
>       a weighted average of MTRs that includes filing units with
>       negative income.

The purpose of using averge MTR without weighting by income is to minimize 
the importance of behavioral effects. The purpose of weighting by income 
is to either highlight behavioral effects for to get a more reasonable 
estimate of the revenue consequence of economic growth. Putting large 
capital losses in the denominator does not serve either of these 
functions, especially since such large losses can never be recouped at 
$3,000/year.

dan

martinholmer:
@codykallen said:
> No. The approach used by CBO and TPC is to calculate individual MTRs using a uniform increment approach, and then calculate a weighted average of those MTRs.
>
> I found some CBO estimates (Table A-4), and I tested the wmtr function with different screening/weighting measures ...

Thanks for the link to the Dec-2014 CBO research article and in particular to Table A-4 on page 34.  When I look at Table A-4 and the text that points to it, I don't see anywhere where it says the rates in Table A-4 are calculated as "a [dollar] weighted average of [individual] MTRs" as you assert.  Did I miss something in the CBO report?  CBO does say on page 34 that the rates in Table A-4 are "the average of the marginal tax rates faced by taxpayers with positive amounts of that particular type of capital income and positive taxable income overall (see Table A-4)."

The goal of the CBO paper (page 2) is this:
```
For this report, the Congressional Budget Office (CBO) estimated
[effective marginal tax rates] on income from marginal investments
(those expected to earn just enough, after taxes, to attract investors)
in such tangible capital assets as equipment, structures, land, and 
inventories (assets held for resale).
```

My major conclusion from glancing over this CBO report is that this is advanced tax research, in which experienced scholars can differ on the best approach.  The CBO paper is far more complex than simply computing weighted marginal tax rates on dividends and capital gains.

This means to me that these calculations do not belong in Tax-Calculator, but would be useful in a standalone research project that uses Tax-Calculator.  You haven't told us much about the B-Tax development plans, but if that model needs these sort of weighted-MTR calculations, then those calculations should be done in the B-Tax model using the general capabilities of Tax-Calculator.

Back in the fall of 2015, there was an extended discussion of whether or not to add deadweight loss calculations to Tax-Calculator (see issue #403).  In the end, @feenberg said this:
> I think this [providing deadweight loss calculation in Tax-Calculator] is getting away from "providing a tool" and moving towards "doing a research project".

I have the same view as Dan did then about weighted marginal tax rates in pull request #1365.  If they are important for B-Tax then, by all means, compute them there.  Or somebody could use the current version of Tax-Calculator to do an interesting standalone research paper on this topic. Or both.

@MattHJensen @feenberg @codykallen

MattHJensen:
@martinholmer said:

> My major conclusion from glancing over this CBO report is that this is advanced tax research, in which experienced scholars can differ on the best approach. The CBO paper is far more complex than simply computing weighted marginal tax rates on dividends and capital gains.

> This means to me that these calculations do not belong in Tax-Calculator, but would be useful in a standalone research project that uses Tax-Calculator. You haven't told us much about the B-Tax development plans, but if that model needs these sort of weighted-MTR calculations, then those calculations should be done in the B-Tax model using the general capabilities of Tax-Calculator.

I'm halfway to where Martin is but not entirely there. 

@codykallen, the point of this PR is to be able to report MTRs on income items from tax-calculator for the purpose of publication, right? Not as inputs to b-tax? 

If so, I think it could be valuable to support that effort in Tax-Calculator, but an essential question is which direct-from Tax-Calculator MTRs are actually meaningful for publication. 

Given that the B-Tax project provides EMTRs on pass through and corporate investment, I don't think it adds much to calculate average MTRs on business income, capital gains, or dividends in Tax-Calculator -- it will always be more meaningful to report the B-Tax results than the Tax-Calculator results. 

On the other hand, there are many income and deductible expenses for which it would be valuable to publish average marginal tax rates directly from Tax-Calculator, like wages and salary, charitable giving, mortgage interest, state and local tax, etc. 

Based on a simple mental checklist, it doesn't seem like any of the income or deductible expenses for which it would be valuable to publish average marginal tax rates directly from tax-calculator every go negative. 

So would it make sense to focus this PR on those income items and lose all of the complication of dealing w stuff that goes negative?

codykallen:
Since there isn't a consensus on a best approach to this, I am closing this PR.

MattHJensen:
@codykallen, if you send me an email describing the issue I am happy to forward to friends at JCT/CBO.

Issue #1373, opened 2017-05-19
Congressional Staffer Requested Preset 


ecbrown5:
A congressional staffer requested a preset developed for the reform at the following link: 

http://renacci.house.gov/_cache/files/7eeb81fe-f245-42c0-8de4-055ebf4c12e9/sats-white-paper.pdf

andersonfrailey:
@MattHJensen putting together a JSON file for this reform could be another good first project for Hank.

As a side note, as we get more requests for presets and create the files, adding them to the list of presets at the top of TaxBrain could cause the page to get cluttered. What are people's thoughts about creating a new section of the website that features the tax plans of Congresspeople along with notes and links to documents such as the one above? Staffers/Congresspeople could submit their proposals and associated literature and have it all in one spot for people to look at.

@martinholmer @codykallen @ecbrown5

codykallen:
@andersonfrailey suggested

> As a side note, as we get more requests for presets and create the files, adding them to the list of presets at the top of TaxBrain could cause the page to get cluttered. What are people's thoughts about creating a new section of the website that features the tax plans of Congresspeople along with notes and links to documents such as the one above? 

If you go to http://www.ospc.org/taxbrain/file/ you will see a separate link to the reform examples. Is this the type of approach you're thinking of?

feenberg:
On Fri, 19 May 2017, andersonfrailey wrote:

> 
> @MattHJensen putting together a JSON file for this reform could be another good first project for Hank.
> 
> As a side note, as we get more requests for presets and create the files, adding them to the list of presets at
> the top of TaxBrain could cause the page to get cluttered. What are people's thoughts about creating a new
> section of the website that features the tax plans of Congresspeople along with notes and links to documents
> such as the one above? Staffers/Congresspeople could submit their proposals and associated literature and have
> it all in one spot for people to look at.

Yes a new page for presets would be fine.

I notice that the presets are the same as any tax plan an ordinary user 
could prepare, except for the ability to link to notes. Perhaps we could 
allow a text box into which users could put comments that would travel 
with the plan. Weren't we going to give people a box to allow them to name 
plans?

The collapsable menus are great - I don't recall any discusion of that 
though.

dan

andersonfrailey:
@codykallen 
> If you go to http://www.ospc.org/taxbrain/file/ you will see a separate link to the reform examples. Is this the type of approach you're thinking of?

[...]

PR #1376, opened 2017-05-23
Add Renacci reform

hdoupe:
Close #1376

### DIFF
--- a/taxcalc/reforms/Renacci.json
+++ b/taxcalc/reforms/Renacci.json
@@ -7,6 +7,7 @@
 //	- Personal exemption increased (3)
 //	- Eliminate all deductions except charitable and mortgage interest (4)
 //		- Cap mortgage interest deduction (4)
+//			- This limit was calculated by multiplying $500,000 by the current APR (%3.946 as of 5/24/2017)
 //	- Repeal Alternative Minimum Tax (5)
 //	- Capital gains and dividends are treated the same as ordinary income (6)
 //	- Increase in EITC (7)
--- a/taxcalc/reforms/Renacci.json
+++ b/taxcalc/reforms/Renacci.json
@@ -1,0 +1,128 @@
+// Title: Simplifying America's Tax System (Renacci)
+// Reform_File_Author: Hank Doupe
+// Reform_Reference: http://renacci.house.gov/_cache/files/7eeb81fe-f245-42c0-8de4-055ebf4c12e9/sats-white-paper.pdf
+// Reform_Description:
+//	- Number of brackets reduced to 3 (1)
+// 	- Standard deduction increased
+//		- Additional standard deduction for blind and aged is reduced to zero (2)
+//	- Personal exemption increased (3)
+//	- Eliminate all deductions except charitable and mortgage interest (4)
+//	- Repeal Alternative Minimum Tax (5)
+//	- Capital gains and dividends are treated the same as ordinary income (6)
+//	- Increase in EITC (7)
+// Reform_Parameter_Map:
+//	-1: _II_rt*,_II_brk*
+//	-2: _STD*
+//	-3: _II_em
+//	-4: _ID*
+//	-5: _AMT_rt*
+//	-6: _CG_rt*,_CG_brk*
+//	-7: _EITC_c
+{
+	"policy":{
+		"_II_rt1" : {
+			"2018" : [0.1]
+		},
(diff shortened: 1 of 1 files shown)


andersonfrailey:
Thanks for the contribution @hdoupe, taking a second look at the reforms and the proposal, there are three things that need fixing.

First, can you change the year of these reforms to 2017.

Second, in your file, you eliminate the additional standard deduction for aged and blind. Was this mentioned in the proposal? I only remember seeing an increase in the base standard deduction.

Finally, there is one more provision you can account for. You can use the `_ID_BenefitCap_Switch` and `_ID_c` provisions to implement the $500K limit on mortgage interest deductions.

From page 6:
>SATS eliminates all itemized deductions except two, charitable contributions and mortgage interest up to $500,000 of debt

After you've applied the haircut to each of the eliminated itemized deductions, you can set a cap on itemized deductions of $500k and exempt charitable deductions from that cap.

Once you've made those changes this should be ready to merge.

MattHJensen:
@hdoupe, when you open a PR, in the top comment, include "closes #1373" so that everyone can see which issue you are addressing. GH will also automatically close the referenced issue when the PR is merged.

MattHJensen:
@andersonfrailey, the cap you are suggesting would limit the amount of mortgage interest deducted every year to $500,000, whereas the proposal calls for "limiting mortgage interest up to $500,00 of debt."

What do you think about something like setting the cap as the prevailing APR on mortgage interest debt multiplied by $500,000?

Interested to also hear what @codykallen and @feenberg think about this.

hdoupe:
@andersonfrailey, I made the first two changes that you suggested.  I'll follow this feed to see what the final word is on the itemized deductions.

@MattHJensen, thanks for the heads up on closing the issue.

MattHJensen:
@hdoupe, you can also do some digging to see if you can find any other organization who has modeled a similar cap on mortgage interest deduction. Check TPC, Tax Foundation, CBO Budget Options

codykallen:
@hdoupe, there is an easier way to tax capital gains and dividends as ordinary income. You can simply set
`"_CG_nodiff": {"2018": [true]}`

@MattHJensen said

> The proposal calls for "limiting mortgage interest up to $500,00 of debt." What do you think about something like setting the cap as the prevailing APR on mortgage interest debt multiplied by $500,000?

That is not an unreasonable approach, but it raises an interesting issue over the correlation between income and borrowing costs. A high-income individual would have a lower mortgage interest rate than a low-income or middle-income individual. Thus a cap on the debt usable for the interest deduction would  allow low and middle-income individuals to deduct more than high-income individuals, although deductions are relatively more valuable for high-income individuals. So this reform would actually make the mortgage interest deduction more progressive than under current law. 

That being said, I think @MattHJensen's approach is reasonable. Does anyone have a good idea of what interest rate should be used? Especially if rates are set to rise over the 3 years or more.

hdoupe:
I think I understand.  Renacci wants to limit the deduction for interest on a mortgage to $500,000 of debt.  So the actual limit is the APR*500000.  The current APR is 3.946% ([https://www.usbank.com/home-loans/mortgage/mortgage-rates.aspx](url)).  

Then, as @andersonfrailey suggested, the reform would be

`"_ID_BenefitCap_Switch": {
	"2017": [[false, false, true, false, false, false, false]]
},
"_ID_c": {
	"2017": [[19730.0, 19730.0, 19730.0, 19730.0, 19730.0]]
}`

Is this what you all are thinking?
@MattHJensen @codykallen 

I tried to find a similar example, but I didn't have much luck.  The Tax Foundation analyzed this plan.  However, it's not very transparent. ([https://taxfoundation.org/details-and-analysis-rep-jim-renacci-s-tax-reform-proposal/](url))

codykallen:
@hdoupe, I think that is a good approach. 

Also, check your indentation in the JSON. It looks off for the _ID_* variables. Also, as I mentioned earlier, you can replace the 7 _CG_* variables in your JSON with just _CG_nodiff.

andersonfrailey:
@hdoupe, can you add a note in the file about your approach for setting the interest deduction? Other than that, if there's no other objection, this looks ready to merge.

@MattHJensen

MattHJensen:
Thanks for the contribution @hdoupe. Merging. 

Please update https://github.com/open-source-economics/Tax-Calculator/issues/1364 with your contribution.

PR #1380, opened 2017-05-24
specify that CG elasticity is a semi-elasticity

MattHJensen:
@salimfurth asked: Can we change the [TaxBrain] interface to say semi-elasticity of long-term capital gains as the parameter label?

The TaxBrain interface pulls from the _BE_cg long_name. 

@codykallen, I'll merge w your +1.


codykallen:
Good change. +1

PR #1382, opened 2017-05-26
Personal exemption payroll tax

### LINKED ISSUE #1369: request for personal exemption to payroll tax
A congressional staff member made this request after receiving an email about TaxBrain:

> I love this!  I had one random idea for an addition:  Tim Carney proposed concept of a personal exemption to payroll tax.  I think that would be an interesting variable to add to this.  http://www.washingtonexaminer.com/the-case-for-a-payroll-tax-cut/article/2560211

econ02:
closes #1369  

This allows implementation of a personal exemption to the payroll tax through as outlined in [Timothy P. Carney, "The case for a payroll tax cut"](http://www.washingtonexaminer.com/the-case-for-a-payroll-tax-cut/article/2560211). The exemption amount is set by both filing status and number of kids for both Social Security and Medicaid payroll taxes. Income below the sum of the filing status and number of kids exemptions is not subject to the payroll tax rate.

The four policy parameters created are: 

1. _FICA_ss_emk : Social Security Payroll Tax Exemption by Number of Kids
2. _FICA_ss_emf : Social Security Payroll Tax Exemption by Filing Status
3. _FICA_mc_emk : Medicare Payroll Tax Exemption by Number of Kids
4. _FICA_mc_emf : Medicare Payroll Tax Exemption by Filing Status

Three records variables are also created. These are:
1. 	ss_em : The total Social Security exemption (sum of filing status exemption and per kid exemption). 
2. 	mc_em: Same as above, for Medicaid
3. 	pte: Summed differences between the income and non-exempted income. This is the total income not subject to the payroll taxes. 

Notes:
* Applies to wage income and self-employment income
* Social Security and Medicaid taxes can be treated with different exemption amounts. 
* The filing status and kid exemption policies are separate to accomplish the goals outlined in the post: "A single earner pays no payroll tax on his first $x. A couple gets $2x of income before they start paying payroll taxes. If you've got a husband and three kids, your first $5x of income is payroll-tax-free."

@andersonfrailey


econ02:
Currently the added policy paremters (_FICA_ss_emk, _FICA_mc_emk, _FICA_ss_emf, and _FICA_mc_emf) are classified under "section_1": "Personal Exemptions" and "section_2": "Personal And Dependent Exemption Amount". 

Would it be more appropriate to classify them under the "Payroll Taxes" category?

econ02:
This method calculates a total payroll tax exemption from Social Security and Medicare by summing a filer-status exemption and a number-of-kids exemption. These exemptions are then subtracted from an individual's earnings which are subject to FICA taxes. The function has a lower limit of zero so an individual can at most be exempt from all FICA taxes, but not receive a credit/subsidy. This reform has no effect on the corporate half of the payroll tax. The "pte" variable is added to calculate the total dollars not subject to payroll taxes, given the exemption settings.

MattHJensen:
> Would it be more appropriate to classify them under the "Payroll Taxes" category?

Yes, that makes more sense to me.

martinholmer:
I have several questions about this PR.

econ02:
@MattHJensen I made changes to the section one classification. n

@martinholmer Please let me know, and I will try my best to answer them.

martinholmer:
I'll be back in the office on June 1st. 
Meanwhile, you could respond to Dan's comments on #1382, which seem to be misplaced in #1383.

MattHJensen:
(The conversation for #1382 was misplaced in #1383. This comment reproduces the lost conversation)

@willensor asked:

> > OK, the exemption is applied to the sum of the husband's and wife's labor
> > income. Note that the FICA tax is not part of the income tax, so a
> > taxpayer might have a lot of FICA and no income tax, so unless the
> > exemption is refundable it is not very valuable/expensive.
> > 
> > This isn't really an exemption on the FICA, it is a credit on the income
> > tax for FICA tax already paid.
> > 
> > dan
> 
> @feenberg is there any way you can clarify your comment on #1381 ?  

@feenberg replied: 

> 
> I have to say that the EITC was initially characterized as a way to return
> FICA to low income earners, and was initially scaled to do about that.
> This proposal is a lot like an increase in the EITC if it is refundable.
> 
> Is it one of the credits that is refundable on the income tax? If it is,
> then it is very much like the EITC. If it is not refundable then it is
> much less expensive. Recall that there are separate child credits
> (refundable, a Democratic initiative, and non-refundable, a Republican
> initiative) depending on refundability status.
> 
> Should it be called an exemption at all? It is really a credit on the
> income tax of an amount calculated to have some relation to the FICA tax
> and some demographics. Shouldn't it be called a credit?
> 
> Note that for employees FICA is not collected on the 1040 but is
> calculated and remitted entirely by the employer. We can do the
> calculation for both, but they are separate things.
> 
> dan

econ02:
@feenberg I am responding to your comment in #1383 which states

> I have to say that the EITC was initially characterized as a way to return
> FICA to low income earners, and was initially scaled to do about that.
> This proposal is a lot like an increase in the EITC if it is refundable.
> 
> Is it one of the credits that is refundable on the income tax? If it is,
> then it is very much like the EITC. If it is not refundable then it is
> much less expensive. Recall that there are separate child credits
> (refundable, a Democratic initiative, and non-refundable, a Republican
> initiative) depending on refundability status.

This is not refundable. According to the proposal, it simply exempts $X first dollars of income from being subject to Medicare and Social Security FICA taxes. After $X, wages do face Medicare and Social Security FICA taxes. 

> Should it be called an exemption at all? It is really a credit on the
> income tax of an amount calculated to have some relation to the FICA tax
> and some demographics. Shouldn't it be called a credit?

I do not believe so. It is reducing one's liability to Medicare and Social Security payroll taxes by a given amount. Now, that exemption amount is arrived at (for both Medicare and Security) given one's filing status and number of children. I still don't believe that makes it a credit; it just determines the exemption level. 

> Note that for employees FICA is not collected on the 1040 but is
> calculated and remitted entirely by the employer. We can do the
> calculation for both, but they are separate things.

Certainly. The proposal only addressed exempting individual Medicare and Social Security FICA taxes, not the employer half. That can certainly be added in. 

The main thrust of the proposal aimed to allow individuals based on their filing status and number of kids to receive their first $X without paying payroll taxes. As for the policies effectiveness, I am unsure.

andersonfrailey:
From my understanding of the policy proposal, the goal is to exempt the first $X in income from the payroll tax, and it doesn't discriminate between payroll taxes paid on wages and those on self-employment income. In this PR the full exemption is applied to both wage and self-employment income. 

Won't that overstate the benefit of this policy to those with both wage and self-employment income?

Let's say an individual's exemption exceeds their wage income, shouldn't only that excess be applied to their self-employment income subject to the payroll tax rather than the entire exemption again?

@econ02

econ02:
@andersonfrailey Yes, I agree, that makes sense.

econ02:
@andersonfrailey 

Commit [8040f6dd32615923d3c6fad42a9819aee49c6a58  ](https://github.com/open-source-economics/Tax-Calculator/pull/1382/commits/8040f6dd32615923d3c6fad42a9819aee49c6a58) solves the issue of double-exemption. The exemption is first applied to wage and salary income. The residual exemption is then applied to self-employment income.

feenberg:
FICA is collected from each paycheck and there is no "lookback" based on 
annual income. I thought this exemption was intended to apply only to 
low-income earners. If that is the case, how is the exemption applied when 
annual income is only known at the end of the year? I assumed that it 
would be applied as a credit on the income tax. But FICA could be modified 
to be withholding have a lookback for the final calculation of liability. 
Or it could be applied without any means test.

I think we need to ask the congressperson if this is intended to be an 
exemption or a credit. Barring that, Matt should make a decision.

dan

feenberg:
On Tue, 30 May 2017, andersonfrailey wrote:

> 
> From my understanding of the policy proposal, the goal is to exempt the
> first $X in income from the payroll tax, and it doesn't discriminate between
> payroll taxes paid on wages and those on self-employment income. In this PR
> the full exemption is applied to both wage and self-employment income.
> 
> Won't that overstate the benefit of this policy to those with both wage and
> self-employment income?
> 
> Let's say an individual's exemption exceeds their wage income, shouldn't
> only that excess be applied to their self-employment income subject to the
> payroll tax rather than the entire exemption again?

Surely the exemption amount must apply to total earnings, not each 
component separately.

What is the decision about the situation where the exemption amount times 
the payroll tax rate exceeds the income tax alone? Does that result in 
negative income tax? That is a really important decsion and I don't think 
it is obvious which way the decsion goes. There are lots of non-refundable 
credits, and some that are refundable.

dan

econ02:
@feenberg The [proposal](http://www.washingtonexaminer.com/the-case-for-a-payroll-tax-cut/article/2560211) makes no mention of a negative income tax, rebate, or making it refundable so I don't think that is an issue. 

It seems we are getting caught on the implementation issue. I see two remedies: (1) payroll tax could be paid in April like any other tax (instead of withheld, similar to how the self-employed pay) or (2) the payroll tax exemption could be administered by employers (who would still be paying their half of the tax, and could track when the exemption threshold is met). I am not sure if either of those solutions are satisfying.

MattHJensen:
@feenberg and @econ02, 

Based on my reading of the [article cited by the staffer](http://www.washingtonexaminer.com/the-case-for-a-payroll-tax-cut/article/2560211), I agree with @econ02's characterization that this is truly a personal exemption to FICA itself and not an income tax credit against FICA paid. 

Here is a telling quotation from the article that confirms that this would reduce FICA revenue rather than income tax revenue:

> The primary objection to this or any payroll tax cut is that it pillages the Social Security trust fund. But the trust fund is already an accounting fiction — as of now the general fund is covering the marginal dollar in entitlement benefits.

There is also no discussion of an income phaseout in the article, and so I think the exemption should apply to everyone for now. If a future user requests an income phaseout, then we could add one then (and leave implementation details to the IRS.)

MattHJensen:
Having two separate parameters, one by the number of kids and the other by filing status introduces a couple of complexities: 

1. We don't have a variable naming scheme to handle it. 
2. Neither parameter matches the income tax personal exemption and will confuse users. 

I'd recommend mirroring the income tax personal exemption instead, and having one personal exemption parameter based on the combined number of taxpayers and dependents.

MattHJensen:
I also don't see any indication in the cited article for applying the exemption separately to social security and medicare. I'd recommend using a single exemption parameter for both taxes.

feenberg:
OK then, that settles it. It doesn't affect the income tax at all.

dan

econ02:
@MattHJensen @feenberg 

The most recent commits change the code to one exemption policy parameter (FICA_em) implemented on both Social Security and Medicare payroll taxes.

andersonfrailey:
The latest commit doesn't quite mirror the income tax personal exemption as @MattHJensen suggested. It still bases the total exemption on filing status rather than the total number of taxpayers and dependents in the tax-unit.

`_FICS_em` can be one number, rather than five, and multiplied by `XTOT` to get the total exemption for that unit.

For the sake of readability/simplicity, I would recommend creating one variable equal to the total exemption for each unit (`XTOT * FICA_em`), subtract that from payroll taxes on WAS, find the residual, then subtract that from payroll taxes on self-employment income.

@econ02 @feenberg

econ02:
When calculating the payroll taxes, `txearn_was_p` is used for the primary earner's social security taxes while `e00200p` is used for the secondary earner's medicare taxes. It is possible that the exemption amount is larger than one of these measures, but less than another. This creates a problem when calculating the remaining exemption to apply to self-employment income (currently the exemption amount net of the exemption used toward wage and salary income).

Suppose `txearn_was_p = 100`,  `e00200p = 70`, and the exemption `tot_em = 80`. The residual exemption amount would therefore be zero for `txearn_was_p` and ten for `e00200p`. The question is whether to carry on the residual ten apply it to self-employment income, or carry on the zero.

econ02:
@andersonfrailey @MattHJensen @feenberg 

The code has been changed to take a single number from parameter `FICA_em`.

When the residual exemption amounts differ between Social Security and Medicare calculations, the minimum value is applied to self-employment income. 

**Note:** Changing the order in which the exemption is applied to avoid facing different residual amounts requires a re-calculation of `txearn_sey_p` and `txearn_sey_s` without wage and salary income parameters. The current formulas are as follows:

```
txearn_sey_p = min(max(0., sey_p * sey_frac), SS_Earnings_c - txearn_was_p)
txearn_sey_s = min(max(0., sey_s * sey_frac), SS_Earnings_c - txearn_was_s)
```

andersonfrailey:
@econ02 I believe this is almost ready. There are two block of code I'm not sure about:
```python
txearn_was_p = max(0., min(SS_Earnings_c, e00200p) - was_em)
txearn_was_s = max(0., min(SS_Earnings_c, e00200s) - was_em)
e00200p_was_p = max(0., e00200p - was_em)
e00200s_was_s = max(0., e00200s - was_em)
```

and 

```python
txearn_sey_p = max(0., min(max(0., sey_p * sey_frac),
                               SS_Earnings_c - txearn_was_p) - resid_em_p)
txearn_sey_s = max(0., min(max(0., sey_s * sey_frac),
                               SS_Earnings_c - txearn_was_s) - resid_em_s)
```


`FICA_em * XTOT` gives you the exemption for the entire tax unit, so by applying the exemption to primary and secondary earners separately without first finding the residuals in between aren't you effectively doubling the exemption for units with both?

econ02:
@andersonfrailey So we need the residual after exempting the primary earner, and then apply to the secondary earner? I can make that change. 

I had operated under the assumption both household earners would receive the same exemption. If we do it as you suggest, wouldn't that have some effect on the extensive margin for secondary earners? Any thoughts on this would be appreciated.

martinholmer:
@andersonfrailey said:
> `was_em = FICA_em * XTOT` gives you the exemption for the entire tax unit, so by applying the exemption to primary and secondary earners separately without first finding the residuals in between aren't you effectively doubling the exemption for units with both?

This is just one of several examples that show how poorly thought out is the "proposal" in [Timothy P. Carney, "The case for a payroll tax cut"](http://www.washingtonexaminer.com/the-case-for-a-payroll-tax-cut/article/2560211).  I would argue that the Carney "proposal" is so logically inconsistent that it makes it impossible to simulate.  I think pull request #1382 should be closed (without merging) and our efforts focused on asking Carney (or the Congressional staff person who suggested we add the Carney "proposal" to Tax-Calculator) to make the nature of this reform more clear.

The fundamental problem that @andersonfrailey is pointing out here is that the total number of dependents in the filing unit (`XTOT`) is a concept from the **income tax**.  There is no concept in FICA of a filing unit because FICA is a tax on individuals, not filing units.  And the definition of who is an income tax filing unit dependent is part of the income tax law, not FICA.

Apparently, these distinctions escaped Carney's notice.  But they have not escaped the notice of others working on tax reforms.  That is why, in Carney's own words, `Many conservatives, led by Sen. Mike Lee, R-Utah, and many of my colleagues at the American Enterprise Institute, are looking for ways to reduce taxes on middle-class families.  Lee, along with Florida Sen. Marco Rubio, are pushing a plan to expand the child tax credit, and to allow families to use the money to offset the payroll taxes they paid.  A simpler approach, in my view, is simply to cut people's payroll taxes.`  I guess Carney never thought much about why all these people do not propose the "simpler approach".  As @feenberg explained, the EITC was originally designed to refund a portion of FICA taxes in a way that targets the refund on those with low earnings.  My suspicion is that Carney wants to refund FICA taxes to **all** earners (even those earning more than a million dollars), but that can be done in the current version of Tax-Calculator by eliminating the EITC phaseout.

@MattHJensen @feenberg @Amy-Xu @andersonfrailey @econ02 @GoFroggyRun @codykallen

MattHJensen:
@martinholmer, what questions would you propose asking Carney or the congressional staff member? 

My first reaction to @andersonfrailey's suggestion (to apply the exemption to primary earner first and then carry the residual to the secondary earner) was that his suggestion is reasonable, in line with my reading (and Anderson's apparently) of the intent of the proposal, and fairly straightforward.

econ02:
I have updated the code so it does not double the exemption. The exemption is first applied to primary earner's wage and salary income, then self-employment income, before doing the same to the secondary earner.

andersonfrailey:
The policy proposal is very vauge, but based on my interpretation the latest commit looks like the best way to implement it for now. What's the final call on merging this? @MattHJensen @martinholmer

andersonfrailey:
Ignore my last comment. Just talked to @econ02 and got clarification about the status of this PR.

MattHJensen:
Given the lack of clarity from user who requested this change, I am closing the issue until we have a better defined proposal. Otherwise we are likely to model something that isn't valuable to our users.

PR #1383, opened 2017-05-26
Removed benefit cap switch

### LINKED ISSUE #1373: Congressional Staffer Requested Preset

hdoupe:
Close #1373

### DIFF
--- a/taxcalc/reforms/Renacci.json
+++ b/taxcalc/reforms/Renacci.json
@@ -84,9 +84,6 @@
 		"_ID_RealEstate_hc": {
 			"2017": [1]
 		},
-		"_ID_BenefitCap_Switch": {
-			"2017": [[false, false, true, false, false, false, false]]
-		},
 		"_ID_c": {
 			"2017": [[19730.0, 19730.0, 19730.0, 19730.0, 19730.0]]
 		},


econ02:
> OK, the exemption is applied to the sum of the husband's and wife's labor
> income. Note that the FICA tax is not part of the income tax, so a
> taxpayer might have a lot of FICA and no income tax, so unless the
> exemption is refundable it is not very valuable/expensive.
> 
> This isn't really an exemption on the FICA, it is a credit on the income
> tax for FICA tax already paid.
> 
> dan

@feenberg is there any way you can clarify your comment on #1381 ?

feenberg:
On Fri, 26 May 2017, econ02 wrote:

>
>       OK, the exemption is applied to the sum of the husband's and wife's
>       labor
>       income. Note that the FICA tax is not part of the income tax, so a
>       taxpayer might have a lot of FICA and no income tax, so unless the
>       exemption is refundable it is not very valuable/expensive.
>
>       This isn't really an exemption on the FICA, it is a credit on the income
>       tax for FICA tax already paid.
>
>       dan
> 
> @feenberg is there any way you can clarify your comment on #1381 ?

I have to say that the EITC was initially characterized as a way to return 
FICA to low income earners, and was initially scaled to do about that. 
This proposal is a lot like an increase in the EITC if it is refundable.

Is it one of the credits that is refundable on the income tax? If it is, 
then it is very much like the EITC. If it is not refundable then it is 
much less expensive. Recall that there are separate child credits 
(refundable, a Democratic initiative, and non-refundable, a Republican 
initiative) depending on refundability status.

Should it be called an exemption at all? It is really a credit on the 
income tax of an amount calculated to have some relation to the FICA tax 
and some demographics. Shouldn't it be called a credit?

Note that for employees FICA is not collected on the 1040 but is 
calculated and remitted entirely by the employer. We can do the 
calculation for both, but they are separate things.

dan

MattHJensen:
@hdoupe this looks good, and I am going to merge it. I am concerned about a source of potential user confusion, but will address it in a dedicated issue.

Issue #1367, opened 2017-05-16
DataFrame Column Issues


brittainhard:
@martinholmer @MattHJensen @PeterDSteinberg since the API for `dropq` has been refactored, I am getting DataFrame errors with the new function.

The error is `Error: Length mismatch: Expected axis has 219813 elements, new values have 219814 elements`.

It seems like it might be a quick fix, but I'm not sure where to look. Any ideas?

PeterDSteinberg:
@brittainhard  What's the full traceback?  Maybe it is a pandas update thing, like handling edge cases on joins with nulls or something like that?  The B-Tax issue we have resolved in last 24 hours was a pandas internal change that we didn't test before our dev release.  One way to test it out would be to install using `python setup.py develop` on each of the tagged commits from the repos, then install the prereqs modified to have the older numpy / pandas - may help isolate the pandas code issues vs our own.

PeterDSteinberg:
@brittainhard  I think this problem may be related to this `chooser` function: https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/dropq/dropq_utils.py#L198

Or the usage of `transform` with `chooser` as the argument.  I can look into it further after my meeting this morning.

PeterDSteinberg:
This bug may relate to pandas updates and how `pd.cut` works with bins and including which edge of each bin (right edge or not).  Or it may have to do with the `transform` function `chooser` returning an unexpected data type?  (something in `taxcalc.dropq.dropq_utils` related to pandas updates)

```
    2017-05-17 00:15:46,056 DEBG 'celery' stderr output:
    [2017-05-17 00:15:46,056: WARNING/MainProcess] {'start_year': 2017, 'user_mods': {'growdiff_response': {}, 'consumption': {}, 'growdiff_baseline': {}, 'behavior': {}, 'policy': {2017: {u'_NIIT_PT_taxed': [False], u'_FICA_ss_trt': [0.2], u'_ID_BenefitCap_Switch': [[True, True, True, True, 1.0, 1.0, 1.0]], u'_ALD_InvInc_ec_base_RyanBrady': [False], u'_EITC_indiv': [False], u'_ID_BenefitSurtax_Switch': [[True, True, True, True, True, True, True]], u'_CTC_new_refund_limited': [False], u'_CG_nodiff': [False]}}, 'gdp_elasticity': {}}, 'year_n': 1}
    
[...]

PR #1387, opened 2017-06-03
Make dropq test fail like in issue 1367; then fix bug

martinholmer:
Commit 6610a62 in this pull request revises an existing dropq test to have a policy reform like the one supplied by @PeterDSteinberg on June 2, 2017, in issue #1367.  The revised test generates a fatal error indicating a bug somewhere in the Tax-Calculator code base.  A subsequent commit in this pull request will fix this bug.

The policy reform in the new failing test consists of a single reform provision: an increase in 2017 of the FICA tax rate from 0.124 to 0.200.  But other values of the FICA tax rate cause the exact same test to pass as shown below:
```
0.22        PASS
0.21        FAIL
0.20        FAIL
0.19999     PASS
0.19        PASS
```
For the record, here is the error traceback report:
```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

dropq/dropq.py:70: in run_nth_year_tax_calc_model
    itax_sum2, ptax_sum2, comb_sum2) = dropq_summary(rawres1, rawres2, mask)

dropq/dropq_utils.py:293: in dropq_summary
    df1, df2 = drop_records(df1, df2, mask)

dropq/dropq_utils.py:257: in drop_records
    df2['flag_dec'] = gp2_dec['mask'].transform(chooser)

/Users/mrh/anaconda/lib/python2.7/site-packages/pandas/core/groupby.py:2990: in transform
    result.index = self._selected_obj.index

/Users/mrh/anaconda/lib/python2.7/site-packages/pandas/core/generic.py:2983: in __setattr__
    return object.__setattr__(self, name, value)
pandas/_libs/src/properties.pyx:65: in pandas._libs.lib.AxisProperty.__set__ (pandas/_libs/lib.c:45103)
    ???

/Users/mrh/anaconda/lib/python2.7/site-packages/pandas/core/series.py:308: in _set_axis
    self._data.set_axis(axis, labels)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = SingleBlockManager
Items: Int64Index([     0,      1,      2,      3,      4, ...219812],
           dtype='int64', length=219813)
BoolBlock: 219813 dtype: bool
axis = 0
new_labels = Int64Index([   108,     46,     54,     26,     53,     68,     28,     24,
  ..., 172375,
            143487, 214120],
           dtype='int64', length=219814)

    def set_axis(self, axis, new_labels):
        new_labels = _ensure_index(new_labels)
        old_len = len(self.axes[axis])
        new_len = len(new_labels)
    
        if new_len != old_len:
            raise ValueError('Length mismatch: Expected axis has %d elements, '
                             'new values have %d elements' %
>                            (old_len, new_len))
E           ValueError: Length mismatch: Expected axis has 219813 elements, new values have 219814 elements

/Users/mrh/anaconda/lib/python2.7/site-packages/pandas/core/internals.py:2836: ValueError
```


martinholmer:
Commit 7016899 makes the dropq_utils.py dropq_calculate function better able to handle policy reforms that are focused on payroll taxes rather than on income taxes.  But these changes do not fix the bug detected by the revised dropq test implemented in commit 6610a62.

martinholmer:
Commit 5a79fa4 fixes the bug.  This bug has existed for years, back to the beginning of work on utils.py functions.  For example, look at the add_weighted_income_bins function as it existed on February 4, 2016, by following [this link](https://github.com/open-source-economics/Tax-Calculator/blob/5678a026b9e14ce6b6124641bd87641f0c261fa6/taxcalc/utils.py#L104-L128).

So, this bug has nothing to do with the recent refactoring of the dropq code or with the changes in pandas when moving from version 0.18 to 0.20.  It has appeared now simply because @brittainhard happened to specify a very particular policy reform that revealed the bug.

@PeterDSteinberg @brittainhard @MattHJensen

martinholmer:
**More about the bug in the `add_weighted_income_bins` Tax-Calculator utility function.**

The revisions to the taxcalc/utils.py `add_weighted_income_bins` function in #1387 corrected two logic problems.  The first was the fatal error described in the pull request and in issue #1367.  The second was a logic problem that has never caused a fatal error but would have whenever the `add_weighted_income_bins` function was called with `weight_by_income_measure=True` as a parameter.

Both problems would cause the `add_weighted_income_bins` function to fail because not all the filing units are put into histogram bins.  Why does this happen?   The pandas.cut function converts a continuous variable into a categorical (or discrete-valued) variable using the values of the continuous variable and a set of bin edges.  The problem with the old version of the `add_weighted_income_bins` function was that it incorrectly specified the bin edges.

The first problem was caused by the less-than-perfect precision of floating-point arithmetic performed by computers.  The old code counted on perfect precision to get the top edge of the last bin to be precisely equal to the maximum value of the continuous variable.  But the less-than-perfect precision of the computer caused a few filing units to not be placed into the last bin.  You can see that is the case because commenting out just one of the new lines of code (see below) causes the error to reoccur.  The following code works fine:
```
    bin_width = cumsum_range / float(num_bins)
    bin_edges = [-9e99] + list(np.arange(1, (num_bins + 1)) * bin_width)
    bin_edges[-1] += 9e9  # add to top of last bin to include all observations
```
while the following code generates a fatal error in the new test:
```
    bin_width = cumsum_range / float(num_bins)
    bin_edges = [-9e99] + list(np.arange(1, (num_bins + 1)) * bin_width)
    # bin_edges[-1] += 9e9  # add to top of last bin to include all observations
```
The second problem was that the old `add_weighted_income_bins` function code assumed incorrectly that every value of the continuous variable was non-negative.  So, in the old code the bottom of the first bin was specified to be zero.  This is correct whenever the function's `weight_by_income_measure` parameter is False, which it is by default.  But if this parameter is set to True, all the filing units with negative expanded_income will not be placed in the first bin and this situation will generate a fatal error.  If you're thinking there can't be any filing units with negative expanded_income, then think again.  Here is a simple tabulation under current-law policy:
```
$ tc puf.csv 2017 --sqldb
You loaded data for 2009.
Tax-Calculator startup automatically extrapolated your data to 2017.
$ sqlite3 puf-17-#-#.db
SQLite version 3.13.0 2016-05-18 10:57:30
Enter ".help" for usage hints.
sqlite> select count(*),round(sum(s006)*1e-6,3) from dump;
219814|170.42
sqlite> select count(*),round(sum(s006)*1e-6,3) from dump where expanded_income<0;
5211|1.475
sqlite> .quit
$ 
```
So, the `puf.csv` sample has almost 2.4 percent of unweighted filing units with negative expanded_income.  The weighted fraction is slightly less than 0.9 percent.  These are small fractions, but with thousands not being put into the ill-defined first bin, the fatal error would have occurred.  (I'm assuming that the negative expanded_incomes are usually caused by large capital losses or big business losses, but I really don't know the details of these people's situations.)

@MattHJensen @feenberg @Amy-Xu @andersonfrailey @GoFroggyRun @codykallen 

martinholmer:
Note that the Saturday afternoon test failures are not really Tax-Calculator test failures, but the inability of Continuum Analytics web servers to transfer conda packages to Travis-CI.  I don't know which end of that transaction is not working.

martinholmer:
Pull request #1387 changes the basic dropq algorithm in two ways.  I want to emphasize this here so that those involved in the original development of that algorithm can give these changes a careful review.

Here is the relevant part of the `dropq_calculate` function with the two new/revised lines marked:
```
    # optionally compute mask
    if mask_computed:
        # create pre-reform Calculator instance with extra income
        recs1p = Records(data=taxrec_df.copy(deep=True),
                         gfactors=growfactors_pre)
        # add one dollar to total wages and salaries of each filing unit
        recs1p.e00200 += 1.0  # pylint: disable=no-member
        recs1p.e00200p += 1.0  # pylint: disable=no-member       <---- NEW LINE ----
        policy1p = Policy(gfactors=growfactors_pre)
        # create Calculator with recs1p and calculate for start_year
        calc1p = Calculator(policy=policy1p, records=recs1p,
                            consumption=consump)
        while calc1p.current_year < start_year:
            calc1p.increment_year()
        calc1p.calc_all()
        # compute mask that shows which of the calc1 and calc1p results differ
        res1 = results(calc1)
        res1p = results(calc1p)
        mask = (res1.combined != res1p.combined)                  <---- REVISED LINE ----
    else:
        mask = None
```
I added the NEW LINE so that payroll taxes would change as well as income taxes.
Apparently this dropq code was written before there was an appreciation that FICA is an **individual tax**, not an **income-tax filing-unit tax**.

I REVISED the other line so that the mask is defined by changes in the combined tax, not just by changes in the income tax amount.

Does this make sense?  Or has my lack of knowledge about the dropq algorithm led me to make a mistake?

@MattHJensen @feenberg @talumbau

talumbau:
I don't know the meaning of the `e00200p` variable, but I can tell you the original intention behind this portion of the code. The idea, as explained to me by @feenberg and @MattHJensen, is to create "noise" in the final results by dropping tax units in a careful way. The tax units that are eligible to be dropped are only units that are impacted by the reform. In other words, if the unit pays the same amount of tax under current policy and under the reform, then it is not eligible to be dropped. This is the purpose behind creating the `mask` variable. I hope this provides enough detail to help inform whether the changes you make in this PR are the right way to go. I think it will likely be up to your best judgement, as well as @feenberg and @MattHJensen, to determine if this change is best.

feenberg:
I would prefer it if we left the dropq code dependent on the income tax 
only. The reason is that we want to drop a few records that affect the 
tables in every table row. If we drop a record that does not contribute to 
the row, not much confidentiality is obtained. By considering FICA in the 
dropq selection we include records with no positive or negative tax to be 
among those dropped. All they would need is a small amount of earnings but 
no other cells would be affected. The dropq procedure depends on the 
dropping of records having a (hopefully small but not zero) effect on the 
results.

In any case, it isn't important that every record be eligible for 
dropping.

dan

martinholmer:
@feenberg said:
> I would prefer it if we left the dropq code [that computes the `mask` variable] dependent on the income tax only.    ...[gives reasons]...

OK, I've changed back to the original definition of `mask`:
```
        mask = (res1.iitax != res1p.iitax)
```

MattHJensen:
@martinholmer, thanks for identifying and resolving this problem. I agree w Dan about using iitax only for the mask. The other changes look good to me too. Would be helpful to have @amy-xu's review as well.

martinholmer:
@MattHJensen said:
> thanks for identifying and resolving this problem. I agree w Dan about using iitax only for the mask. The other changes look good to me too. Would be helpful to have @Amy-Xu's review as well.

The  definition of the `mask` array has already been switched back to its original definition.
I'll await review by @Amy-Xu before merging #1387.

taxcalc/tests/test_dropq.py
```diff
@@ -215,45 +215,44 @@ def test_dropq_diff_table(groupby, res_column, puf_1991_path):
...
-    reforms['_PT_rt3'] = [0.33]
-    reforms['_II_rt4'] = [0.33]
-    reforms['_PT_rt4'] = [0.33]
+    start_year = 2017
+    reform_year = start_year
+    analysis_year = 2026
+    year_n = analysis_year - start_year
+    reform = {
+        '_FICA_ss_trt': [0.2]
+    }
```
Amy-Xu: Just to confirm: the reason of switching to this reform is that it will change not only iitax but also payroll right?

MattHJensen: Martin got started on this PR because that reform failed: https://github.com/open-source-economics/Tax-Calculator/pull/1387#issue-233337447

martinholmer: @Amy-Xu,  @MattHJensen is exactly correct.  The new reform (with just a payroll tax rate increase) was the reform reported by @PeterDSteinberg as causing a fatal error.  So, I revised the test to be like in his bug report and worked until I found what was causing that test failure.  Does this make sense?

Amy-Xu:
@martinholmer said

>The revisions to the taxcalc/utils.py add_weighted_income_bins function in #1387 corrected two logic problems. The first was the fatal error described in the pull request and in issue #1367. The second was a logic problem that has never caused a fatal error but would have whenever the add_weighted_income_bins function was called with weight_by_income_measure=True as a parameter.

All the changes in this PR look good to me but I guess I'm confused on two things. First is regarding the first problem -- if this line of code is a problem, why is it only a problem for payroll reforms but not iitax reforms? Shouldn't it always result in missing the last record? Second is about what records need to be included in bins. I believe we have ran into the second problem before that negative expanded income people are not included in tables, which was discussed in this [issue](https://github.com/OpenSourcePolicyCenter/webapp-public/issues/214) when Joe Sullivan brought it up. After we checked the TPC table example mentioned in that PR, we concluded it is fine to leave out negative income people. Do we need to refer to TPC tables or we simply want to make everything look consistent? @MattHJensen

taxcalc/tests/test_dropq.py
Amy-Xu: Yep that makes sense.

martinholmer:
@Amy-Xu asked first:
> ... regarding the first problem -- if this line of code is a problem, why is it only a problem for payroll reforms but not iitax reforms? Shouldn't it always result in missing the last record?

I think this a floating-point arithmetic precision issue.  I don't think it has anything to do with payroll tax vs income tax reforms.  My guess is that the bug report happened to involve a FICA-only reform, but this problem could have arisen with any kink of tax reform.

@Amy-Xu also asked:
> Second is about what records need to be included in bins. I believe we have ran into the second problem before that negative expanded income people are not included in tables, which was discussed in [this issue](https://github.com/OpenSourcePolicyCenter/webapp-public/issues/214) when Joe Sullivan brought it up. After we checked the TPC table example linked in that issue, we concluded it was fine to leave out negative income people. Do we need to refer to TPC tables or we simply want to make everything look consistent?

The second logic problem in #1387 has nothing to do with the issue of whether or not TaxBrain tables show people with negative expanded_income.  It has to do with avoiding a fatal error in the calculations.  I don't think the `add_weighted_income_bins` utility function is ever called by TaxBrain with `weight_by_income_measure=True`, which is the only way the second logic problem would have arisen.

Does all this make sense?

Amy-Xu:
@martinholmer You're right this function hasn't been called by TaxBrain with `weighted_by_income_measure=True`. Sorry I wasn't being clear. It appears to me that your revision would include negative-expanded-income people regardless `weighted_by_income_measure` is true or false. Is that correct? If so, is that an intended result of this PR?

martinholmer:
@Amy-Xu said:
> You're right this function hasn't been called by TaxBrain with `weighted_by_income_measure=True`. Sorry I wasn't being clear. It appears to me that your revision would include negative-expanded-income people regardless weighted_by_income_measure is true or false. Is that correct? If so, is that an intended result of this PR?

Yes and Yes.  There is no change in the basic logic of the `add_weighted_income_bins` function when `weighted_by_income_measure=False`, which is the way it is usually called.

taxcalc/utils.py
```diff
@@ -192,9 +192,12 @@ def add_weighted_income_bins(pdf, num_bins=10, labels=None,
...
     else:
         pdf['cumsum_temp'] = np.cumsum(pdf['s006'].values)
+    min_cumsum = pdf['cumsum_temp'].values[0]
     max_cumsum = pdf['cumsum_temp'].values[-1]
-    bin_edges = [0] + list(np.arange(1, (num_bins + 1)) *
-                           (max_cumsum / float(num_bins)))
+    cumsum_range = max_cumsum - min_cumsum
+    bin_width = cumsum_range / float(num_bins)
+    bin_edges = [-9e99] + list(np.arange(1, (num_bins + 1)) * bin_width)
+    bin_edges[-1] = 9e99  # raise top of last bin to include all observations
```
Amy-Xu: Looking at the 4 lines here, I'm not sure whether it will do the job or not. I have not tested it on my local but will do in a few minutes. Use random number for logic walk-through -- suppose max_cumsum is 165 million when it's not weighted by income_measure, then bin_width would be 16.5 million if we have 10 bins. Then all the bin_widths are stacked on top of negative infinity [-9e99] for bin edges. I don't think 165 million would make any bin edge positive before changing the last one to positive infinity [9e99].

Amy-Xu: Ignore the previous comment. This section looks good to me.

Amy-Xu: Hmmmm there's still a minor issue with weighted_by_income_measure option set to True. Since bin_width is stack on top of zero due to the logic of `list(np.arange(1, (num_bins + 1)) * bin_width)` snippet, the first bin would be larger than expected when cumulated sum is weighed by income_measure and push all other bin edges to the right by the amount equal to the absolute value of min_cumsum, doesn't it?

martinholmer: @Amy-Xu said:
> Ignore the previous comment. This section looks good to me.

martinholmer:
@Amy-Xu,  I appreciate you close review of pull request #1387.  Is there anything else your reviewing, or do you think it is OK to merge it into the master branch?

@MattHJensen @feenberg @andersonfrailey

taxcalc/utils.py
Amy-Xu: That comment was a previous one. Sorry about that. Do you think this current comment is valid?

Amy-Xu:
@martinholmer Could you take a look at my latest comment above? Sorry about back and forth on this -- just want to make sure the bins with `weighted_by_income_measure` set to True is the same as you expected.

martinholmer:
@Amy-Xu,  You are commenting at two different levels in GitHub, so I have lost track about what you mean by "my latest comment above".

Amy-Xu:
@martinholmer my bad, this one:
screen shot 2017-06-05 at 1 59 49 pm

martinholmer:
@Amy-Xu said about this code:
> ```
>    cumsum_range = max_cumsum - min_cumsum
>    bin_width = cumsum_range / float(num_bins)
>    bin_edges = [-9e99] + list(np.arange(1, (num_bins + 1)) * bin_width)
>    bin_edges[-1] = 9e99  # raise top of last bin to include all observations
> ```
> Hmmmm there's still a minor issue with weighted_by_income_measure option set to True. Since bin_width is stacked on top of zero due to the logic of `list(np.arange(1, (num_bins + 1)) * bin_width)` snippet, the first bin would be larger than expected when cumulated sum is weighed by income_measure and push all other bin edges to the right by the amount equal to the absolute value of min_cumsum, doesn't it?

Let's review the pandas.cut function.  Part of the documentation says this: 
`bins [1,2,3,4] indicate (1,2], (2,3], (3,4]`
The first item in the bin_edges list is the bottom of the first bin and the last item in the bin_edges list is the top of the last bin.  So, we can set the bottom of the first bin to minus infinity and the top of the last bin to plus infinity, and everything is OK as long as the intermediate values in the bin_edge list are proper.  The expression `list(np.arange(1, (num_bins + 1)) * bin_width)` simply sets the top of the second through the last bin at the appropriate percentile value of the continuous variable being "cut" into bins.  And then we make the top of the last bin be positive infinity in order to get all the observations with high values placed in the top bin.

If this explanation is not an effective response to your concerns, then I'm not understanding exact what your concern is.  The code is exactly the same as before except that the bottom of the first bin has been made very small and the top of the last bin has been made very large.

If I'm not being responsive, my apologies.  Ask the question in a different way and maybe I'll understand better your concerns.

Amy-Xu:
>The code is exactly the same as before except that the bottom of the first bin has been made very small and the top of the last bin has been made very large.

I agree with this part when `weighted_by_income_measure` is False, but it doesn't seem like so when that option is True, the reason being `min_cumsum` is very likely negative and therefore each bin is larger than previous version. If max_cumsum is 160 and min_cumsum is -40, then in previous version the bin size was 16 and in this version it would be 20, given 10 bins in total. 

I would assume when bin_width is calculated as (max_cumsum - min_cumsum)/num_bins, it suggests that people get cumsum_temp between -40 to -20 are thrown into the lowest bin. If that's the expectation, that doesn't seem consistent with reality. As the bin_edges looks like [-9e99, 20, 40, ..., 140, 9e99], the lowest would include people from -40 to 20 instead of -40 to -20. Similarly the bins afterward would be pushed right by about 40. Is this a clearer way to describe my concern? If this is intended, I have finished review. This PR looks good to me.

martinholmer:
@Amy-Xu said:
>> The code is exactly the same as before except that the bottom of the first bin has been made very small and the top of the last bin has been made very large.
>
> I agree with this part when weighted_by_income_measure is False, but it doesn't seem like so when that option is True, the reason being min_cumsum is very likely negative and therefore each bin is larger than previous version. If max_cumsum is 160 and min_cumsum is -40, then in previous version the bin size was 16 and in this version it would be 20, given 10 bins in total.
>
> I would assume when bin_width is calculated as (max_cumsum - min_cumsum)/num_bins, it suggests that people get cumsum_temp between -40 to -20 are thrown into the lowest bin. If that's the expectation, that doesn't seem consistent with reality. As the bin_edges looks like [-9e99, 20, 40, ..., 140, 9e99], the lowest would include people from -40 to 20 instead of -40 to -20. Similarly the bins afterward would be pushed right by about 40. Is this a clearer way to describe my concern? If this is intended, I have finished review. This PR looks good to me.

Thanks for being patient with me in expressing your concerns.  I think you are right about this, so I'll change it.  I'd appreciate your review of the change I make later this afternoon.

martinholmer:
@Amy-Xu,  I've taken what I understand to be you point and fixed the `add_weighted_income_bins` utility function logic when`weighted_by_income_measure=True`.
I would appreciate it if you could look over commit 0523649 to see if I understood your point correctly.

Amy-Xu:
@martinholmer Thank you very much for the modification. I really appreciate your time on this. My point was a minor issue. Since no one has risen any problem with setting up bins in the current form, I'm happy not to worry anything about it. This PR looks good to me.

martinholmer:
@Amy-Xu,  Thanks again for your careful review of pull request #1387.  I'll merge it now.

PR #1400, opened 2017-06-07
Revert needs_sphinx from 1.5.6 to 1.3.5

martinholmer:
Change because 1.3.5 is hardwired somewhere inaccessible to the Tax-Calculator source code.


talumbau:
@martinholmer, it looks like the astropy project hard a similar issue and resolved it with this PR:

https://github.com/astropy/astropy/pull/5681

Don't know much more than that. Could make sense to post a question if you have any about the PR. Hth

Issue #1368, opened 2017-05-16
Where did OG-USA related metadata from taxcalc.dropq go?


PeterDSteinberg:
@martinholmer I found this diff in the github blame report for `taxcalc.dropq`:

https://github.com/open-source-economics/Tax-Calculator/commit/fc06a73e64f2109804bd08c7207e5c9f8993ab88#diff-0fbb9d8bd5e87c63c64db8be5816d69bL38

It shows that `OGUSA_ROW_NAMES` has been deleted, among other OG-USA variables.  
 * Where should webapp-public import `OGUSA_ROW_NAMES` and other former `taxcalc.dropq` OG-USA variables from now?  
 * Was there a corresponding change to OG-USA, i.e. has the underlying Tax-Calculator code changed such that `OGUSA_ROW_NAMES` are no longer descriptive of the model?  webapp-public imports OG-USA stuff here from `taxcalc.dropq`: https://github.com/OpenSourcePolicyCenter/webapp-public/blob/4f5c9cfb30ffec1f2e9b58917a7bb75f352ae6fc/webapp/apps/dynamic/helpers.py#L29

When we make changes to Tax-Calculator, we need to test locally with all the tentative release candidates of:

 * webapp-public - running locally
 * deploy - (will be moved to webapp-public soon)
 * B-Tax
 * Tax-Calculator
 * OG-USA

We have had problems recently due to one of those repos being changed without being tested against the versions of the other repos that will be released at the same time.

MattHJensen:
@martinholmer warned @PeterDSteinberg, @rickecon, and @jdebacker of this issue 15 days ago in this comment https://github.com/open-source-economics/Tax-Calculator/pull/1314#issuecomment-298444018 before the PR was merged.

MattHJensen:
@PeterDSteinberg said:

> When we make changes to Tax-Calculator, we need to test locally with all the tentative release candidates of:
> 
> webapp-public - running locally
> deploy - (will be moved to webapp-public soon)
> B-Tax
> Tax-Calculator
> OG-USA

Webapp-public, deploy, b-tax, and og-usa should not depend on taxcalc developers to check compatibility before releases just like taxcalc devs don't expect Pandas to test new pandas versions against taxcalc. 

> We have had problems recently due to one of those repos being changed without being tested against the versions of the other repos **that will be released at the same time.**

[emphasis added]

This seems like the problem. Why do we need to release versions of all repos at the same time? 

Why isn't it like this: 

[...]

PR #1421, opened 2017-06-09
Minor edits to docstrings in records.py

martinholmer:
This activity is part of the ongoing effort to define a **public API** for Tax-Calculator.


feenberg:
On Fri, 9 Jun 2017, Martin Holmer wrote:

> 
> This activity is part of the ongoing effort to define a public API for
> Tax-Calculator.
>

Shouldn't there be some discussion of the purpose of the API before 
patches start being posted? How does it differ from allowing a user to 
upload a JSON file of tax parameters?

dan feenberg

martinholmer:
@MattHJensen said 24 days ago in the #1368 discussion:
> Something else that will make this easier for everyone is if taxcalc declares a public API and adopts [semantic versioning](http://semver.org/) to identify which releases make incompatible API changes, which releases add backwards-compatible functionality, and which releases make backwards-compatible bug fixes.  More on that later. 

@feenberg, The first step in implementing [semantic versioning](http://semver.org/) is to declare a **public API**, which means thinking through what functions and class methods are private (that is, are just behind-the-scenes helpers) and which are public (that is, are part of what other projects and users can rely on to be relatively stable).

We now have the **public API** defined (with just a few outstanding issues with the Calculator class and some utility functions) and you can read about it at [this link](http://taxcalc.readthedocs.io/en/latest/public_api.html).  At the top of that link is the following:
> **Developers who want to use Tax-Calculator capabilities in their own projects should restrict themselves to using this public API.  All other Tax-Calculator members are private and subject to change without advance notice.**

So, the definition of a public API does not change any capabilities of Tax-Calculator, it just makes it clear what parts of Tax-Calculator other developers can count on as being stable.

Let me know if you have any other questions.

PR #1423, opened 2017-06-09
Add conditional import of bokeh package in utils.py

martinholmer:
Original intention was to add bokeh package to conda.recipe, but still have no answer to question about `slug size` of taxcalc package.  These changes work around this lack of information for the time being.


martinholmer:
Note that the increase in uncovered statements from 8 t0 10 put us back where we were before the code changes in #1419.

Issue #1231, opened 2017-03-09
Validation Tests Error


andersonfrailey:
I'm working on a new branch and when I ran the validation test this traceback error popped up:
```
STARTING WITH VALIDATION TESTS : Thu Mar  9 16:00:52 EST 2017
Traceback (most recent call last):
  File "simtax.py", line 104, in <module>
    sys.exit(main())
  File "simtax.py", line 96, in main
    output_records=args.records)
  File "/Users/andersonfrailey/Tax-Calculator/taxcalc/validation/taxsim/simpletaxio.py", line 117, in __init__
    output_records)
  File "/Users/andersonfrailey/Tax-Calculator/taxcalc/validation/taxsim/simpletaxio.py", line 609, in _calc_object
    start_year=self._policy.start_year)
TypeError: __init__() got an unexpected keyword argument 'gfactors'
ERROR: first-output-file d15.in.out-simtax does not exist
. taxsim/d15
============ ALL EXECUTED TAXSIM VALIDATION TESTS PASS
FINISHED WITH VALIDATION TESTS : Thu Mar  9 16:01:10 EST 2017
```

As you can see, it says the tests passed, but I'm worried about the error that comes up. Also, after I ran the tests, the content of the file `d15.taxdiffs` was erased. Before the tests I checked to make sure the branch was up-to-date.

When I switch over to my master branch, which is also up-to-date, I get a different error and the tests failed. Here is that error:

```
STARTING WITH VALIDATION TESTS : Thu Mar  9 16:15:20 EST 2017
Traceback (most recent call last):
  File "simtax.py", line 104, in <module>
    sys.exit(main())
  File "simtax.py", line 96, in main
    output_records=args.records)
  File "/Users/andersonfrailey/Tax-Calculator/taxcalc/validation/taxsim/simpletaxio.py", line 117, in __init__
    output_records)
  File "/Users/andersonfrailey/Tax-Calculator/taxcalc/validation/taxsim/simpletaxio.py", line 609, in _calc_object
    start_year=self._policy.start_year)
TypeError: __init__() got an unexpected keyword argument 'gfactors'
ERROR: first-output-file d15.in.out-simtax does not exist
F	taxcalc/validation/taxsim/d15.taxdiffs
TEST FAILURE: any red lines starting with F between the
              STARTING... and FINISHED... lines above
              indicate a test failure caused by the actual
              *.taxdiffs results being different from the
              expected *.taxdiffs results.
[...]

PR #1435, opened 2017-06-15
Add new unit test to check for taxcalc package

martinholmer:
This pull request adds a new pytest that will fail whenever the unit tests are run when a taxcalc package is installed.  This test is meant to help alert developers to the confusing situation described in issue #1231.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -11,5 +11,4 @@ dependencies:
 - pylint
 - coverage
 - pytest-pep8
-- conda
 
--- a/environment.yml
+++ b/environment.yml
@@ -11,4 +11,5 @@ dependencies:
 - pylint
 - coverage
 - pytest-pep8
+- conda
 
--- a/taxcalc/tests/test_4package.py
+++ b/taxcalc/tests/test_4package.py
@@ -3,9 +3,12 @@ import subprocess
 import pytest
 
 
+@pytest.mark.requires_pufcsv
 def test_for_package_existence():
     """
-    Ensure that no conda taxcalc package is installed when running pytest
+    Ensure that no conda taxcalc package is installed when running pytest.
+    Primarily to help developers catch mistaken installations of taxcalc;
+    the requires_pufcsv mark prevents test from running on GitHub.
     """
     out = subprocess.check_output(['conda', 'list', 'taxcalc']).decode('ascii')
     envless_out = out.replace('taxcalc-dev', 'environment')
--- a/taxcalc/tests/test_4package.py
+++ b/taxcalc/tests/test_4package.py
@@ -1,9 +1,12 @@
-import pytest
-import subprocess
 import re
+import subprocess
+import pytest
 
 
 def test_for_package_existence():
+    """
+    Ensure that no conda taxcalc package is installed when running pytest
+    """
     out = subprocess.check_output(['conda', 'list', 'taxcalc'])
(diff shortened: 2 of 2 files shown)


martinholmer:
Does anybody have any ideas about why the new test in pull request #1435 works on my computer (OSX, Python 2.7) but does not work on the GitHub checkin run of the test?  Here is the new test:
```
def test_for_package_existence():
    """
    Ensure that no conda taxcalc package is installed when running pytest
    """
    out = subprocess.check_output(['conda', 'list', 'taxcalc'])
    if re.search('taxcalc', out.decode('ascii')) is not None:
        assert 'taxcalc package' == 'installed'
```
My understanding is that when run on GitHub the pytest suite is executed from code in the taxcalc-dev environment and that there is no conda taxcalc package involved.  But since things are not working, my understanding must be wrong.

@MattHJensen @PeterDSteinberg

hdoupe:
@martinholmer, I just ran the test on in my local environment (OSX, python 3.6).  I added a print statement so that I could see the contents of `out.decode('ascii')` and got the following ouput:


```
# packages in environment at /anaconda/envs/taxcalc-dev:
#
```
and traceback: 
```

  File "test_4package.py", line 16, in <module>
    test_for_package_existence()
  File "test_4package.py", line 13, in test_for_package_existence
    assert 'taxcalc package' == 'installed'
AssertionError
```

It looks like re.search is finding taxcalc in the string "packages in environment at /anaconda/envs/taxcalc-dev:"  I hope this helps.

martinholmer:
@hdoupe said about problem with pull request #1435:
> It looks like re.search is finding `taxcalc` in the string "packages in environment at /anaconda/envs/taxcalc-dev:" I hope this helps.

Thanks!  And, yes, I think your insight will help.  Back to the drawing board for me.

hdoupe:
@martinholmer No problem, glad I could help!

martinholmer:
@hdoupe,  I tried to factor in your insight on #1435 in the newest version.
The Windows/Python3.6 test passed on GitHub (so that is progress), but
all three of the Linux tests on GitHub failed.
Does the newest version of #1435 pass on your computer?

hdoupe:
@martinholmer, #1435 passes on my computer for python 2.7 and 3.6.

martinholmer:
@hdoupe said:
> [latest version of] #1435 passes on my computer for python 2.7 and 3.6.
Thanks so much for checking.

hdoupe:
@martinholmer, what do you think about doing something like this?

```
def test_for_package_existence():
    """
    Ensure that no conda taxcalc package is installed when running pytest
    """
    installed = True
    try:
        import taxcalc
    except ImportError:
        installed = False

    if installed:
        assert 'taxcalc package' == 'installed'
```

martinholmer:
@hdoupe suggested using `try import taxcalc` construct.

A good idea, but I couldn't get it to work reliably on my computer,
maybe because we're using an absolute-addressing import style that
involves `from taxcalc import ...` statements.

Anyway, the latest version just by-passes the execution of the new test on GitHub.

The whole point of the new test is to help developers discover a mistakenly installed taxcalc package on their own computer, which is a problem I've had in the past (and I think Anderson was also a victim of this problem as described in issue #1231).  So, there really is no reason to run the test on GitHub.

Issue #1440, opened 2017-06-27
Treatment of weights in Records object constructor, Part 2


martinholmer:
Pull request #844, which was merged on August 10, 2016, included the following code in the Records class constructor:
```
        # weights must be same size as tax record data
        if not self.WT.empty and self.dim != len(self.WT):
            frac = float(self.dim) / len(self.WT)
            self.WT = self.WT.iloc[self.index]
            self.WT = self.WT / frac
```
The calculation of `frac` has already been discussed in issue #1439.

An additional problem with the above code is that it ignores the fact that the weights vary from year to year, and therefore, applying one value of `frac` to weights in all the years creates unnecessary variation in sample weights across years.

@MattHJensen @feenberg @hdoupe

martinholmer:
Issue #1440 has been resolved by the merge of pull request #1441.

Issue #1439, opened 2017-06-20
Treatment of weights in Records object constructor, Part 1


hdoupe:
I'm opening this issue in response to PR #1429.

In the case where a subsample of the PUF file is passed to a `Records` object, I'm a bit confused as to why we scale the weights by the full sample size divided by the subsample size.  It would make more sense to me to scale the weights by the the sum of the weights in the full sample divided by the sum of the weights in the subsample.

```
# we are doing this:
wt = wt * (N_fullsample / N_subsample)
# but why don't we do it this way
wt = wt * (sum(fullsample_weights) / sum(subsample_weights))
```

I feel like I'm missing something.  I'd appreciate any thoughts that you all may have on this. 

cc @martinholmer

martinholmer:
Thanks for opening an issue on this matter. I'll be back from Canada next week and I'll study this.

feenberg:
On Tue, 20 Jun 2017, Henry Doupe wrote:

> 
> I'm opening this issue in response to PR #1429.
> 
> In the case where a subsample of the PUF file is passed to a Records object, I'm a bit confused as to why we scale the
> weights by the full sample size divided by the subsample size. It would make more sense to me to scale the weights by the
> the sum of the weights in the full sample divided by the sum of the weights in the subsample.

Same in expected value terms, but more accurate.

dan

hdoupe:
@feenberg Thanks for the reply.  That makes sense, but more accurate in what way?

martinholmer:
@hdoupe said:
> @feenberg Thanks for the reply.
>
>> Same in expected value terms, but more accurate.
>
> That makes sense, but more accurate in what way?

I think the point @feenberg is making is that when picking an **unweighted** sample, scaling up the  sample weights in these two ways produces the same expected value of the sum of the scaled-up sample weights.  But that using the scale-up factor you suggest (the ratio of the weights) produces a sum of the scaled-up weights that has a lower variance than using a scale-up factor that is simply the ratio of the unweighted counts (as in the current code).  By lower variance, I mean when the random-number seed changes that sum of the scaled-up weights will vary less (when using the ratio of the weights for the scale-up factor).

Independently, scaling up the sample weights using the ratio of the weights (instead of ratio of the raw counts) is the correct approach when picking a **weighted** sample.

Does that make sense?

hdoupe:
[...]

PR #1441, opened 2017-06-27
Improve sample weight adjustment

martinholmer:
This pull request attempts to improve the logic of scaling up sub-sample weights so that they produces results that are closer to those produced using a full sample.  The proposed changes attempt to resolve the issues discussed in #1439 and #1440.

@MattHJensen @feenberg @hdoupe

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -15,7 +15,7 @@ Release 0.9.1 on 2017-??-??
 **New Features**
 - Improve calculation of sub-sample weights
   [[#1441](https://github.com/open-source-economics/Tax-Calculator/pull/1441)
-  by Hank Doupe and Martin Holmer]
+  by Hank Doupe]
 
 **Bug Fixes**
 - None
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -10,11 +10,15 @@ Release 0.9.1 on 2017-??-??
 [#XXXX](https://github.com/open-source-economics/Tax-Calculator/pull/XXXX))
 
 **API Changes**
+- None
 
 **New Features**
+- Improve calculation of sub-sample weights
+  [[#1441](https://github.com/open-source-economics/Tax-Calculator/pull/1441)
+  by Hank Doupe and Martin Holmer]
 
 **Bug Fixes**
-
+- None
 
 Release 0.9.0 on 2017-06-14
 ---------------------------
--- a/taxcalc/records.py
+++ b/taxcalc/records.py
@@ -146,9 +146,16 @@ class Records(object):
         self._read_adjust(adjust_ratios)
         # weights must be same size as tax record data
         if not self.WT.empty and self.dim != len(self.WT):
-            frac = float(self.dim) / len(self.WT)
+            # .. compute sum of full-sample weights for each year
+            sum_full_weights = dict()
+            for colname in list(self.WT):
+                sum_full_weights[colname] = self.WT[colname].sum()
+            # .. make WT index correspond to specified Records sub-sample index
             self.WT = self.WT.iloc[self.index]
-            self.WT = self.WT / frac
+            # .. scale up WT using a differenct factor for each year
+            for colname in list(self.WT):
+                sum_sub_weights = self.WT[colname].sum()
+                self.WT[colname] *= sum_full_weights[colname] / sum_sub_weights
         # specify current_year and FLPDYR values
         if isinstance(start_year, int):
             self._current_year = start_year
--- a/taxcalc/records.py
+++ b/taxcalc/records.py
@@ -5,6 +5,7 @@ Tax-Calculator tax-filing-unit Records class.
 # pep8 --ignore=E402 records.py
 # pylint --disable=locally-disabled records.py
 
+from __future__ import division
 import os
 import json
 import six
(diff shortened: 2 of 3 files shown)


martinholmer:
@hdoupe,  Could you review the changes in pull request #1441 to make sure they are sensible given your recent work on this topic?  Thanks.

hdoupe:
@martinholmer this looks sensible.  What do you think about doing it this way?

```
WT_full_sum = self.WT.sum()
self.WT = self.WT.iloc[self.index]
WT_sub_sum = self.WT.sum()
WT_ratio = WT_full_sum / WT_sub_sum

self.WT = self.WT * WT_ratio
```

These should be equivalent, but this way be a bit faster.  However, I'm not sure if it would matter too much given that we are only dealing with 10 years of data.

martinholmer:
@hdoupe said:
> this [#1441] looks sensible. What do you think about doing it this way?
>```
> WT_full_sum = self.WT.sum()
> self.WT = self.WT.iloc[self.index]
> WT_sub_sum = self.WT.sum()
> WT_ratio = WT_full_sum / WT_sub_sum
> self.WT = self.WT * WT_ratio
>```
> These should be equivalent, but this way may be a bit faster. However, I'm not sure if it would matter too much given that we are only dealing with 10 years of data.

PR#1441 and what you suggest above would be equivalent if the weights were the same in each year.  But the weights are different in each year, so I don't think they are equivalent.  In addition to resolving your issue #1439, pull request #1441 is trying to resolve my issue #1440.

Does this make sense?

hdoupe:
@martinholmer , I think they should be the same.  `WT_full_sum` and `WT_sub_sum` are pandas Series objects of length 18 with each entry being the total sum of weights for 2009, 2010, ..., 2026.  Then, `WT_ratio` is a pandas Series object where the i-th item is the i-th item of `WT_full_sum` divided by the i-th item of  `WT_sub_sum`.  

Finally, `self.WT * WT_ratio` multiplies the i-th column of `self.WT` by the i-th item of `WT_ratio`.

If my understanding is correct, this is just another way to solve the problem but without using loops.  Whichever implementation you choose is probably just a matter of preference in this situation.

martinholmer:
@hdoupe,
Thanks for the tutorial on the magic of Pandas, which is something I've had trouble getting.

I tried to use your code fragment in place of what is in #1441, but did not get the same result.  In fact, the test failed even though it passes with the current code.

The most likely answer is that I made a mistake in applying your code.  Maybe you can take a look.
Here is what I tried:
```
        # weights must be same size as tax record data
        if not self.WT.empty and self.dim != len(self.WT):
            WT_full_sum = self.WT.sum()
            self.WT = self.WT.iloc[self.index]
            WT_sub_sum = self.WT.sum()
            WT_ratio = WT_full_sum / WT_sub_sum
            self.WT = self.WT * WT_ratio
```
But instead of passing the `test_agg` test in the `test_pufcsv.py` file, I get this failure message:
```
E           ValueError: PUFCSV AGG RESULTS DIFFER IN SUB-SAMPLE AND FULL-SAMPLE
E           WHEN subfrac=0.030, rtol=0.0100, seed=180
E            year,sub,full,reldiff= 2013	2199.38	2176.94	0.0103
E            year,sub,full,reldiff= 2017	2821.50	2790.64	0.0111
E            year,sub,full,reldiff= 2018	2931.17	2899.85	0.0108
E            year,sub,full,reldiff= 2020	3136.53	3105.29	0.0101
E            year,sub,full,reldiff= 2021	3266.19	3230.47	0.0111
E            year,sub,full,reldiff= 2022	3416.16	3369.21	0.0139
tests/test_pufcsv.py:111: ValueError
```
Any suggestions would be welcome as I am far from being a Pandas expert.

hdoupe:
@martinholmer No problem, glad I can help.  I ran the test with the first implementation and the second implementation and got the same results for each implementation: 

```
E           ValueError: PUFCSV AGG RESULTS DIFFER IN SUB-SAMPLE AND FULL-SAMPLE
E           WHEN subfrac=0.030, rtol=0.0100, seed=180
E            year,sub,full,reldiff= 2013	2199.38	2176.94	0.0103
E            year,sub,full,reldiff= 2017	2821.50	2790.64	0.0111
E            year,sub,full,reldiff= 2018	2931.17	2899.85	0.0108
E            year,sub,full,reldiff= 2020	3136.53	3105.29	0.0101
E            year,sub,full,reldiff= 2021	3266.19	3230.47	0.0111
E            year,sub,full,reldiff= 2022	3416.16	3369.21	0.0139

tests/test_pufcsv.py:111: ValueError
```

The adjusted weights should sum to the same amount as the full sample weights.  I tested this here: 
```
from taxcalc.records import Records
import pandas as pd

puf = pd.read_csv("taxcalc/puf.csv")
puf_samp = puf.sample(frac=0.03, random_state=180)

rec = Records(data = puf_samp)
wt = pd.read_csv("taxcalc/puf_weights.csv")

print(wt.sum() - rec.WT.sum())
```

Both implementations were off by the same amount: 

```
WT2009    0.000109
WT2010   -0.000067
WT2011   -0.000021
WT2012   -0.000074
WT2013    0.000086
WT2014    0.000078
WT2015    0.000147
WT2016   -0.000093
WT2017   -0.000097
WT2018    0.000175
WT2019   -0.000130
WT2020   -0.000175
WT2021    0.000084
WT2022   -0.000065
WT2023    0.000156
WT2024    0.000072
WT2025    0.000092
WT2026   -0.000191
dtype: float64
```

Are you sure that it passed the test on the first implementation?  My tests indicate that the two implementations are producing the same result.

martinholmer:
@hdoupe,  Thanks again for all the help on Pandas usage.  You are absolutely correct.  I had no idea that Pandas would do sums for all the DataFrame columns when you do dframe.sum().  I discovered we were getting different results with my column-loop code (which should have been logically equivalent but slower) because of a Python 2.7 versus Python 3.x difference: 5/2=2 in Python 2.7, but 5/2=2.5 in Python 3.x.  So my now-removed column-loop code was not accurate because the weight scale-up factor was an interger (since the weights in the `puf_weights.csv` file are integers).

Again, thanks for all the help.  I'm going to merge this very soon.

@MattHJensen @feenberg @Amy-Xu @andersonfrailey

Issue #1449, opened 2017-06-30
positive personal refundable credit raises revenue


MattHJensen:
A TaxBrain user just notified me that increasing the personal refundable credit raises revenue rather than lowers it.  

https://www.ospc.org/taxbrain/edit/14829/?start_year=2017

martinholmer:
Merge of pull request #1450 resolves bug report in issue #1449.

PR #1450, opened 2017-06-30
Add personal (non)refundable credit tests and fix bug reported in issue #1449

martinholmer:
This pull request adds a test that replicates the incorrect personal refundable credit reform results reported in issue #1449.  The new test also identifies a separate bug (which has not been reported by TaxBrain users) in the personal nonrefundable credit logic.  The second bug causes aggregate income tax revenue to never change no matter what policy parameters are used for the personal nonrefundable credit reform.

Subsequent commits will untangle the confusion in the code between personal refundable credits (which were added nine months ago to the AGI function) and personal nonrefundable credits (which were added seventeen days ago in a new PersonalTaxCredit function).  This confusion is the cause of the first bug reported in #1449.

The cause of the second bug is that the recently added PersonalTaxCredit function was never called by the Calculator.calc_all() method.

Early work showed that commenting out the personal nonrefundable credit code makes the first bug go away.  The new test uses a small puf.csv sample, so it is not run on GitHub for data privacy reasons.


martinholmer:
Pull request #1450, which is a fix for the bug reported in issue #1449, and a fix for an unreported bug in personal nonrefundable credit logic.

It should be noted that neither of these two personal credits are part of current-law policy.  This means that these two bug fixes do not cause any change in tax results under current-law policy.  The only results differences caused by these two bug fixes are for reforms that involve adding either a personal refundable credit or a personal nonrefundable credit to tax law.

@MattHJensen @feenberg @Amy-Xu @andersonfrailey @hdoupe @econ02 @GoFroggyRun

taxcalc/tests/test_pufcsv.py
```diff
@@ -239,3 +239,42 @@ def test_mtr(tests_path, puf_path):  # pylint: disable=redefined-outer-name
         msg += '---            and rerun test.                ---\n'
         msg += '-------------------------------------------------\n'
         raise ValueError(msg)
+
+
+@pytest.mark.requires_pufcsv
+def test_credit_reforms(puf_path):  # pylint: disable=redefined-outer-name
```
MattHJensen: @martinholmer, do you envision these tests catching errors that the /comparison results would not, or is this primarily a safety guard for when comparison/reform_results.py is not run?

martinholmer: Some of both: the test of the personal nonrefundable credit is **not** in the `comparison/reform_results.py` suite.

MattHJensen: Thanks.

MattHJensen:
LGTM. Thanks for fixing the bug @martinholmer. Merging.

PR #1488, opened 2017-07-20
Add comments to dropq_utils.py

hdoupe:
I had some trouble understanding certain parts of the dropq algorithm.  So I added some comments in hopes of clarifying the code.


martinholmer:
@hdoupe said in pull request #1488:
> I had some trouble understanding certain parts of the dropq algorithm.
> So I added some comments in hopes of clarifying the code.

Welcome to the club!  The dropq code was originally developed by people who view the source code as the best documentation.  I don't agree with that view.  I think the original developer of code can make a great contribution by adding a few comments that guide others who read the code for the first time months or even years later.

However, there is a real risk in newcomers adding comments.  The code might be so complicated that the newcomer doesn't fully understand what is going on, and therefore, adds a misleading or incorrect comment.

I think there is a real risk of that here.  In particular, I'm not at all sure that the following three added comments are correct:
```
+        # to determine which records are taxpayers
```
```
+        # mask is true if record pays taxes; if it does not change after adding
+        # a dollar then this record does not pay taxes
```
```
+       # select indices of recs that are taxpaying records
```
My guess --- and that is all it is, a guess --- is that what is going on here is that the code is finding filing units whose tax liability **changes**.   My guess is that the comments, which suggest the code is finding filing units who do or don't pay any taxes, are incorrect.

But that is the problem.  The lack of code documentation forces me to **guess**.  So, @hdoupe, I'm highly sympathetic.  I still have memories of the shock I had when I first read the dropq code.

I think we should not add any comments until @MattHJensen, @feenberg and @talumbau have all reviewed them and all three agree that they are accurate and useful descriptions of the dropq algorithm.  Given that view, I'm adding a [WAIT] label on this pull request until we get advice from all three of them.

hdoupe:
@martinholmer said

> However, there is a real risk in newcomers adding comments. The code might be so complicated that the newcomer doesn't fully understand what is going on, and therefore, adds a misleading or incorrect comment.

I totally agree.  I didn't think of it that way.  Thanks for weighing in.

MattHJensen:
@Amy-Xu, could you take a look at this?

taxcalc/dropq/dropq_utils.py
```diff
@@ -91,6 +91,7 @@ def dropq_calculate(year_n, start_year,
         recs1p = Records(data=taxrec_df.copy(deep=True),
                          gfactors=growfactors_pre)
         # add one dollar to total wages and salaries of each filing unit
+        # to determine which records are taxpayers
```
Amy-Xu: Hmmm `taxpayer` might be not be the most accurate term, since some people might be have credits completly offset liabilities, but technically they're still taxpayers. Quoting Dan's original documentation - "for every taxpayer we calculate the effect of adding a dollar to income, then keep track of the number of taxpayers in each bracket for whickh those additions make a difference in the tax liability. This is the count of effective cell size." So maybe rephrase a bit?

Amy-Xu:
I added a comment for one line that I feel needs minor modification, but overall this PR looks good to me. As the TC has evolved quite a bit since drop-q was first born, maybe it's time to add an full documentation that combines Dan's description and current TC capacities?

@MattHJensen @martinholmer @hdoupe

taxcalc/dropq/dropq_utils.py
hdoupe: @Amy-Xu Thanks for reviewing this.  What about this:
```
# add one dollar to the income of each filing unit to determine
# which filing units undergo a resulting change in tax liability
```
We can replace "filing unit" with "taxpayer" or "tax unit" or some other term.

taxcalc/dropq/dropq_utils.py
```diff
@@ -102,6 +103,8 @@ def dropq_calculate(year_n, start_year,
         calc1p.calc_all()
         assert calc1p.current_year == start_year
         # compute mask that shows which of the calc1 and calc1p results differ
+        # mask is true if record pays taxes; if it does not change after adding
+        # a dollar then this record does not pay taxes
```
hdoupe: @Amy-Xu Should this be:

```
# mask is true if a filing unit's tax liability changed after a dollar 
# was added to the filing unit's income
```

taxcalc/dropq/dropq_utils.py
Amy-Xu: Filing unit sounds good to me.

taxcalc/dropq/dropq_utils.py
Amy-Xu: Yep looks better.

martinholmer:
@Amy-Xu said:
> Quoting Dan's original documentation - "for every taxpayer we calculate the effect of adding a dollar to income, then keep track of the number of taxpayers in each bracket for whickh those additions make a difference in the tax liability. 

@Amy-Xu, can you distribute (somehow) "Dan's original documentation"?
That is what is missing if we want to better document the dropq source code.

@MattHJensen @hdoupe

Amy-Xu:
I only have a hard copy on hand and couldn't find the corresponding pdf on my laptop. Maybe @MattHJensen still has it? If not, I could always ask Dan or scan the one I have on hand.

MattHJensen:
I have attached a couple of early files from Dan, including the first pseudocode for dropq in `pc3.pdf`. Proceed with caution, though, because we updated the algorithm during implementation, and so what you read in these historical documents may not be what you find in the code, and the differences may have been introduced for good reason. Sorry for this state of affairs! 

I am also including prep notes from a meeting with SOI to gain approval for the algorithm. Those are pasted below. If I recall correctly, Amy sorted out some bugs in `pc3` as we were preparing these notes, and so these notes are more recent than `pc3`. 

I'll also refer you to the commit history and closed issues in the repository that originally housed dropq: https://github.com/open-source-economics/drop-q

image

[pc3.pdf](https://github.com/open-source-economics/Tax-Calculator/files/1180965/pc3.pdf)
[disclosure avoidance purpose.pdf](https://github.com/open-source-economics/Tax-Calculator/files/1180966/disclosure.avoidance.purpose.pdf)

feenberg:
I have attached a document that may be the source of the quote.

dan

On Thu, 27 Jul 2017, Martin Holmer wrote:

> 
> @Amy-Xu said:
>
>       Quoting Dan's original documentation - "for every taxpayer we
>       calculate the effect of adding a dollar to income, then keep
>       track of the number of taxpayers in each bracket for whickh
>       those additions make a difference in the tax liability.
> 
> @Amy-Xu, can you distribute (somehow) "Dan's original documentation"?
> That is what is missing if we want to better document the dropq source code.
> 
> @MattHJensen @hdoupe
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the
> thread.[AHvQVYvkeEYn53mdT1tsVdic21T1bA7zks5sSNgFgaJpZM4Oen2j.gif]
> 
> 
>
<h3>Disclosure avoidance for an Online Tax Calculator</h3>

<h3>The project</h3>

<p>The AEI wishes to place on a public webpage a calculator that would 
allow anonymous users to specify parameters of a US individual income tax 
law and receive the resulting "distribution table". The calculator would 
simply apply the specified law to a large database of anonymized tax 
returns and summarize the resulting tax liability by income class. The tax 
returns are from the public use file sold by the IRS Statistics of Income 
Division. A large number of tax parameters would be available, 
characterizing current law and many plausible reforms.

<p>Although the SOI believes the returns are sufficiently rounded, blurred 
and aggregated to avoid any disclosure, they are nervous and fear they are 
mistaken. They worry that an intruder would take advantage of the online 
calculator to extract information from the PUF file, and then identify (or 
merely claim to identify) an individual taxpayer. Therefore they have 
asked us to further protect the underlying data from inadvertent 
disclosure through differential analysis. 

<p>We are aware of the inherent contradiction of protecting a public use 
file, but we are dependent on the SOI, and need to satisfy them 
regardless.

<p>
Here are my thoughts on disclosure problems with an online tax calculator, 
and how I believe they can be addressed.

<h3>Mitigating Factors</h3>

<p>Our initial offer might point out certain mitigating factors about an 
online tax calculator:

<ol>

<li>Uses 2009 PUF, no extreme item values
<li>No state identifiers available to users
<li>Distribution table rows are fixed and have many taxpayers 
<li>Limited precison in output
</ol>

These factors make it difficult to extract individual level data from 
the distribution table output.

<h3>The Remaining Problem</h3> <p> 

<p> Nevertheless, SOI may wish for additional protections. In 2008 the 
distribution table row for taxpayers with AGI greater than $5 million has 
2,500 taxpayer records, so at first blush it would seem very resistant to 
intrusion. The possibility of disclosure arises from the ability of the 
user to change dollar thresholds. For example the user could set all the 
tax rates but the top bracket rate to zero, and then try many values for 
the threshold that begins the top bracket. With sufficient trys, they 
could identify the highest taxable income in the sample once they found a 
threshold for the start of the top bracket that just barely generated 
positive tax. That is, if setting the top (and only) bracket threshold to 
200 million dollars, generated no revenue, and 199 million generated some 
revenue, the taxable income of the highest record must be about 200 
million. From there the intruder could identify other characteristics by 
varying other parameters. In the 2009 PUF these particularly 
vulnerable records will be aggregated into "average records", eliminating 
the risk for the highest income records, but more complex parameter 
selections would allow lower income records to be singled out.

<p>We are going to want to include a switch to turn various deductions 
into credits. By manipulating that switch it would be possible to obtain 
the amount of itemized deductions, and if the switch were per deductible 
item, then the amount of individual items such as property tax or state 
income tax would become visible for that top record.  Although this would 
not be an actual violation of the law unless the taxpayer were actually 
identified, it would concern SOI and might be misinterpreted by press or 
public.

<p>All of this could be detected and prevented with some minimal human 
supervision of run logs, as hundreds of odd looking runs would be required 
before any information could be extracted from an individual record with 
techniques of brute force. However users may wish to titrate reforms to a 
specific revenue target, which might justify repeated runs with similar 
parameters, so it would not be desirable to limit users to a small number 
of runs. In any case requests are presumably anonymous, so that it would 
not be possible to limit runs only for the intruder, rather all users 
would be adversely affected. 

<p>The Census Bureau has developed an effective automated technique for 
disclosure avoidance in user-specified tables that I believe can be 
adapted to the case of an online tax calculator. I will quote from <i>The 
Microdata Analysis System at the U.S. Census Bureau</i>, by Freiman, 
Lucero, Singh and others to describe the Drop <i>q</i> rule.

<blockquote> 

<p>The drop <i>q</i> rule works as follows. A user-defined universe that 
passes all of the prvious rules has q records removed at random. To do 
this, the MAS will first draw a random integer value of <i>q</i> such that 
<i>2<=q<=k</i> and such that when the universe is modified by omitting 
<i>q</i> records, the number of remaining records is a multiple of 3. Here 
<i>k</i> is some predetermined number, which may depend on the size of the 
universe. The exact method for determining the maximum possible number 
<i>k</i> of observations to remove is still under consideration. Then, 
given <i>q</i>, the MAS will subsample the universe U(<i>n</i>) by 
removing <i>q</i> records at random from the U(<i>n</i>) to yield a new 
subsampled universe U(<i>n-q</i>).

<p>Within the MAS, all statistical analyses are performed on the 
subsampled universe U(<i>n-q</i>) and not on the original universe 
U(<i>n</i>). Each unique universe U(<i>n</i>) that is defined on the MAS 
will be subsampled independently according to the Drop <i>q</i> Rule. To 
prevent an "averaging of results" attack, the MAS will produce only one 
subsampled universe U(<i>n-q</i>) for each unique universe U(<i>n</i>), 
with this unique subsample persiting for the lifetime of the system. That 
is, all users who select a specific universe U(<i>n</i>) will have all 
analyses performed on exactly the same subsampled universe U(<i>n-q</i>). 
The MAS accomplishes consistent subsampling of universes by using the same 
random seed to perform the subsampling every time a given universe comes 
up. To receive the full disclosure protection offered by the Drop 
<i>q</i> rule, it is necessary that the seed, while constant for a given 
universe, differs across universes, and this can be implemented by having 
the seed be a function of the set of units in the universe. 

<p>
The differencing attacks of most concern require, among other things, that 
two universes are available that differ in size by 1 or 2. However, under 
the drop <i>q</i> rule described above, all subsampled universes have 
sizes that are multiples of 3, and no pair of multiples of 3 (including 
pairs where both numbers are the same) can have a difference of 1 or 2. 
Hence the Drop <i>q</i> rule eliminates the possibility of this sort of 
disclosure, or even of an apparent disclosure where taking the difference 
of the resulting tables gives an answer that is palusible (because it has 
nonnegative numbers in all cells) but is not correct.

</blockquote> 

<h4>How would this apply to a distribution table?</h4>

<p>The rule still depends on reported cells having multiple members. With a 
user-supplied tax law this is not just a matter of counting the taxpayers in a 
cell, since some taxpayers will not be contributing any variation to the result. 
For example, everyone could be subject to a poll tax and a single taxpayer 
subject to additional tax. The intruder could 
subtract the poll tax from the cell aggregate to be left with the value for the 
single taxpayer, even though many taxpayers "contributed" to the cell.

<p>The tax brackets and other tax thresholds are specified by the user, so they 
are the ones that are of concern as they can be used to create implicit cells.  
There are bracket thresholds in the regular and minimum tax, and thresholds for 
the start and end of various credits and clawbacks. A taxpayer may be affected 
by some thresholds and not others. For example, a taxpayer in the first income 
bracket is affected by small changes in that bracket, but not higher brackets. 
If he is an itemizer he is not affected by small changes in the standard 
deduction.

<p>It is also possible to create an implied threshold that wasn't apparent by an 
explicit law paramter. For example, by manipulating the rates in the regular tax 
and the alternative minimum tax, it might be possible to create an incentive for 
taxpayers to switch from one to the other at some very high income, and disclose 
the existence or not of a taxpayer record at the income. I wonder what to do 
about that. We can limit the precision of the tax parameters to 2 or 3 digits, 
which would make it difficult to ensure that lines crossed at particular values, 
and we can limit the precision of returned values, which would make it difficult 
to determine which tax regime was in effect. If we do return the regular and AMT 
in separate columns, then we could make sure the number of taxpayers 
contributing to each cell was sufficient. But there may be other places in the 
tax system that such crossings could be created.

<pRather then attempting to enumerate every possible implicit cell I suggest 
that for every taxpayer we calculate the effect of adding a dollar to AGI or 
itemized deductions, then keep track of the number of taxpayers in each bracket 
for which that addition makes a difference in the tax liability. If the count 
for any cell is a small positive number, that is a potential disclosure that 
drop-<i>q</i> won't prevent. Such cells would have to be suppressed. I don't 
think that would be a common result, for plausible tax reforms.
 
<p>I do not propose that we expend any effort to ensure that columns
aggregate to published totals. This seems to be more of a fetish of 
agencies rather than a useful feature for data users and is a potential 
source of bias.

<h4>Alternatives to Drop <i>q</i></h4>

<p>The PUF file rounds all dollar values to 4 digits. Experiments with TAXSIM 
suggest that rounding to 3 digits affects the estimate of tax liability by only 
.03% (3 parts in 10,000) and rounding to 2 digits affects liability by only .3%. 
Similar differences are observed within the individual income classes. So 
additional rounding would not seriously affect the usability of the calculator.

<p>The PUF file blurs some variables. It is possible to blur more, even all the 
variables, without greatly affecting tax liability by income class. For example, 
blurring all the dollar values used by TAXSIM into groups of 2 affects aggregate 
tax liability by only .2%.

<p>Apparently blurring and rounding are acceptable for a tax calculator in ways 
that would devastate other uses of the data.
   
<p>It is also possible to limit precision in returned aggregates to prevent 
differencing attacks. I suppose it may be because the 
intruder can observe the exact value of a threshold that causes the aggregate to 
jump but that suggests just adding some noise to the returned cells. The noise 
could be scaled to be at least as large in expectation as the largest single 
contributor to the cell. The literature I have seen complicates the process by 
insisting that row and column totals be maintained at published values, but this 
is not necessary in our application. Some control of repeated runs would be 
required to avoid "averaging attacks".

<h4>Summary</h4>

<ol>
<li>We start with the 2009 file - no state id, no extreme values.
<li>We drop k randomly chosen records from each cell.
<li>We track the number of actual contributors to each cell by 
    checking finite differences.
</ol>

<p>
Daniel Feenberg <br>
feenberg@nber.org<br>
617-863-0343

martinholmer:
@feenberg, Thanks for the documentation on the original development of the dropq algorithm used by TaxBrain when calling Tax-Calculator.  Hank has read that documentation and revised his proposed additions to or changes in the comments (see below).  Can you look over these proposed changes to the comments to see if they accurately describe the algorithm?

Here are the changes:
First, in the `dropq_calculate` function:
```
     if mask_computed:
          # create pre-reform Calculator instance with extra income
          recs1p = Records(data=taxrec_df.copy(deep=True),
                           gfactors=growfactors_pre)
 -        # add one dollar to total wages and salaries of each filing unit
 +        # add one dollar to the income of each filing unit to determine
 +        # which filing units undergo a resulting change in tax liability
          recs1p.e00200 += 1.0  # pylint: disable=no-member
          recs1p.e00200p += 1.0  # pylint: disable=no-member
          policy1p = Policy(gfactors=growfactors_pre)
         # create Calculator with recs1p and calculate for start_year
         calc1p = Calculator(policy=policy1p, records=recs1p,
                             consumption=consump)
         while calc1p.current_year < start_year:
             calc1p.increment_year()
          calc1p.calc_all()
          # compute mask that shows which of the calc1 and calc1p results differ
 +        # mask is true if a filing unit's tax liability changed after a dollar
 +        # was added to the filing unit's income
          res1 = results(calc1.records)
          res1p = results(calc1p.records)
          mask = (res1.iitax != res1p.iitax)
     else:
         mask = None
```

And then in the `chooser` function:
```
def chooser(agg):
     """
     This is a transformation function that should be called on each group.
     It is assumed that the chunk 'agg' is a chunk of the 'mask' column.
     This chooser selects three of those mask indices with the output for
      those three indices being zero and the output for all the other indices
      being one.
      """
 +    # select indices of recs with change in tax liability after
 +    # $1 increase in income
      indices = np.where(agg)
      three = 3
      if len(indices[0]) >= three:
         choices = np.random.choice(indices[0],  # pylint: disable=no-member
                                    size=three, replace=False)
     else:
          msg = ('Not enough differences in income tax when adding '
                 'one dollar for chunk with name: {}')
          raise ValueError(msg.format(agg.name))
 +    # drop chosen records
      ans = [1] * len(agg)
      for idx in choices:
          ans[idx] = 0
     return ans
```

@hdoupe @Amy-Xu

feenberg:
I don't know Python, so I can't offer an informed opinion, but I do have 
two comments.

I worry a little bit about the comparison of floats for equality here:

     >           mask = (res1.iitax != res1p.iitax)

I would prefer to see a epsilon of perhaps at least a penny before 
declaring a record eligible.

A style comment - defining

   three = 3

doesn't accomplish much in my view. Does it meet the letter of some 
programming standard? It doesn't meet the spirit. While my view is a 
minority one, I would NOT prefer

   minimum_number_of_records_to_drop_per_cell = 3

I think "3" is a perfectly acceptable literal that needs no indirection. 
It only appears once so there is no chance of things getting out of sync.

dan

martinholmer:
@feenberg,  Thanks for you dropq code substance and style review.  We'll take those recommendations under advisement.

But what we really wanted was your views on the proposed changes in the COMMENTS that are marked with either a leading `-` for deletion or with a leading `+` for addition.  Do these comment changes accurately reflect the dropq algorithm logic?

@hdoupe @Amy-Xu

feenberg:
On Wed, 2 Aug 2017, Martin Holmer wrote:

> 
> @feenberg, Thanks for you dropq code substance and style review. We'll take
> those recommendations under advisement.
> 
> But what we really wanted was your views on the proposed changes in the
> COMMENTS that are marked with either a leading - for deletion or with a
> leading + for addition. Do these comment changes accurately reflect the
> dropq algorithm logic?

The comments conform to my understanding of dropq.

dan

hdoupe:
@feenberg Thanks for reviewing these comments and posting the documentation.  I found it very helpful to learn the origin and motivation for this algorithm.

@martinholmer Thanks for reviewing.  Are you comfortable with these comments?

Also, would you want to take a look at the coding style comments that @feenberg proposed?  I agree that `three = 3` is a little unnecessary.  

Also, if `iitax` is calculated out to 7 or 8 decimal places, then there could be instances where the individual tax liability may have changed only due to numerical stability issues.  So maybe we could do something this instead:

`mask = np.isclose(res1.iitax, res1p.iitax, atol=0.01, rtol=0.0)`

martinholmer:
@hdoupe said:
> Also, would you want to take a look at the coding style comments that @feenberg proposed? I agree that three = 3 is a little unnecessary.

Yes.  Please eliminate the `three` variable as part of this #1488 pull request.

> Also, if iitax is calculated out to 7 or 8 decimal places, then there could be instances where the individual tax liability may have changed only due to numerical stability issues. So maybe we could do something this instead:
>
> `mask = np.isclose(res1.iitax, res1p.iitax, atol=0.01, rtol=0.0)`

Actually, Python floats are "double-precision" as they say in Fortran, which means there are 14 or 15 digits being computed.  Given this observation, I think (despite what Dan @feenberg suggests) that a penny (`atol=0.01`) may be too large.  But we can experiment with different `atol` values.  Because this particular change is a (potential) change in dropq logic, I suggest it be handled in another pull request, leaving the current pull request containing only clarifying documentation changes plus a code simplification that does not change dropq logic.

Does that make sense?

hdoupe:
@martinholmer That makes sense to me.  I'll make the change and then open another PR for the vector comparison issue.

martinholmer:
@feenberg said about the dropq code:
> A style comment - defining
>```
>   three = 3
>```
> doesn't accomplish much in my view. Does it meet the letter of some programming standard? ...
> I think "3" is a perfectly acceptable literal that needs no indirection.
> It only appears once so there is no chance of things getting out of sync.

But that is **exactly** the point.  Three does appear more than once and so there is a "chance of things getting out of sync" as you say.
```
    three = 3
    if len(indices[0]) >= three:
        choices = np.random.choice(indices[0],  # pylint: disable=no-member
                                   size=three, replace=False)
```

talumbau:
> But that is exactly the point. Three does appear more than once and so there is a "chance of things getting out of sync" as you say.
> 
>     three = 3
>     if len(indices[0]) >= three:
>         choices = np.random.choice(indices[0],  # pylint: disable=no-member
>                                    size=three, replace=False)


An alternative would be a more descriptive variable name at function-level scope, such as

`num_records_to_drop = 3`

or a module-level scoped variable at the top of the file if this is used in multiple places:

```
NUM_RECORDS_TO_DROP = 3
```

hdoupe:
@talumbau said 
> An alternative would be a more descriptive variable name at function-level scope, such as
>
>`num_records_to_drop = 3`
>
>or a module-level scoped variable at the top of the file if this is used in multiple places:
>
>`NUM_RECORDS_TO_DROP = 3`

I agree with this approach.  This is more informative than just `three` and `3` and it will keep everything in sync.  It appears that this variable is only needed in the `chooser` function.  So, it would make sense to do:

```
    indices = np.where(agg)
    num_records_to_drop = 3
    if len(indices[0]) >= num_records_to_drop:
        choices = np.random.choice(indices[0],  # pylint: disable=no-member
                                   size=num_records_to_drop, replace=False)
```

martinholmer:
@hdoupe said:
> I agree with this approach [suggested by @talumbau].  ...  So, it would make sense to do:
>```
>    indices = np.where(agg)
>    num_records_to_drop = 3
>    if len(indices[0]) >= num_records_to_drop:
>        choices = np.random.choice(indices[0],  # pylint: disable=no-member
>                                   size=num_records_to_drop, replace=False)
>```

I also agree with TJ's suggestion (your version above).
Why don't you prepare a pull request that implements this version of TJ's suggestion?

martinholmer:
@hdoupe,  When you prepare the dropq PR, can you look though the comments and docstrings to generalize any reference to `three` or `3`.  I guess, putting `NUM_RECORDS_TO_DROP = 3` toward to top of the file would make rewriting the comments/docstrings easier.

martinholmer:
@hdoupe,  I know you're busy with important TaxBrain work that is far more important than following up the suggestions made in the discussion of Tax-Calculator pull request #1488.
Given that situation, I'll handle those dropq follow-up issues.

PR #1498, opened 2017-08-03
Make reform comparison tests use puf.csv subsample

martinholmer:
Convert the 54 reform comparison tests so that instead of using the full `puf.csv` sample they use a five percent subsample.  This change reduces the execution time for the command
```
py.test -m "not notebook" -n4
```
from about 237 seconds to about 143 seconds, which is a reduction of about a minute and a half, or in relative terms, a reduction of about 40 percent in execution time.  Many of the reform test results change, but not by large amounts.

@Amy-Xu @andersonfrailey @hdoupe @GoFroggyRun @yuying27 @codykallen


Amy-Xu:
Looks like a great improvement to me!

PR #1504, opened 2017-08-09
[WIP] Add functionality to save XTR plots as .png files

andersonfrailey:
In release 0.12.6, Bokeh added the ability to save plots as .PNG files, rather than just .HTML. This PR modifies the `write_graph_file` function in `utils.py` to take advantage of this new capability and updates `environment.yml` to include `phantomjs` and Bokeh 0.12.6.

There is no change to the arguments accepted by `write_graph_file`, I have just added a few lines to check if the file ends in `.PNG` and, if so, use `export_png` rather than the usual save feature.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -16,6 +16,9 @@ Release 0.9.3 on 2017-??-??
 - Add validation checking of policy parameters involved in a reform
   [[#1502](https://github.com/open-source-economics/Tax-Calculator/pull/1502)
   by Martin Holmer]
+- Add capability to save XTR graphs as .PNG files
+  [[#1504](https://github.com/open-source-economics/Tax-Calculator/pull/1504)
+  by Anderson Frailey]
 
 **Bug Fixes**
 - None
--- a/environment.yml
+++ b/environment.yml
@@ -12,3 +12,4 @@ dependencies:
 - coverage
 - pytest-pep8
 - pytest-xdist
+- phantomjs
--- a/environment.yml
+++ b/environment.yml
@@ -5,7 +5,7 @@ dependencies:
 - numba
 - toolz
 - six
-- bokeh >=0.12.3
+- bokeh >=0.12.6
 - mock
 - pep8
 - pylint
--- a/environment.yml
+++ b/environment.yml
@@ -13,3 +13,5 @@ dependencies:
 - pytest-pep8
 - pytest-xdist
 - phantomjs
+- pillow
+- selenium
--- a/taxcalc/utils.py
+++ b/taxcalc/utils.py
@@ -1038,7 +1038,7 @@ def write_graph_file(figure, filename, title):
     figure : bokeh.plotting figure object
 
     filename : string
-        name of HTML file to which figure is written;
+        name of file to which figure is written;
         should end in .html or .png
 
     title : string
--- a/taxcalc/utils.py
+++ b/taxcalc/utils.py
@@ -1038,7 +1038,8 @@ def write_graph_file(figure, filename, title):
     figure : bokeh.plotting figure object
 
     filename : string
-        name of HTML file to which figure is written; should end in .html
+        name of HTML file to which figure is written;
+        should end in .html or .png
 
     title : string
         text displayed in browser tab when HTML file is displayed in browser
@@ -1048,8 +1049,12 @@ def write_graph_file(figure, filename, title):
     Nothing
     """
     delete_file(filename)    # work around annoying 'already exists' bokeh msg
-    bio.output_file(filename=filename, title=title)
-    bio.save(figure)
+    # Save file as either PNG or HTML file
+    if filename[-4:].lower() == '.png':
+        bio.export_png(figure, filename)
+    else:
+        bio.output_file(filename=filename, title=title)
+        bio.save(figure)
 
 
 def isoelastic_utility_function(consumption, crra, cmin):


andersonfrailey:
I'm working on getting that Travis build error fixed. Will post when I do.

martinholmer:
@andersonfrailey, What's the purpose of adding `phantomjs` to the taxcalc-dev environment?

martinholmer:
@andersonfrailey,  The new docstring documentation of `filename` should probably be changed:
```
  filename : string
-       name of HTML file to which figure is written;
+       name of file to which figure is written;
        should end in .html or .png
```

martinholmer:
@andersonfrailey, The Travis-CI tests failed because `phantomjs` was not available.
What does `phantomjs` have to do with these graphing enhancements?

andersonfrailey:
@martinholmer, `phantomjs` is used by Bokeh to capture the PNG image. What's essentially happening is Bokeh takes a screenshot of what would be rendered in the browser and saves it. They give some details in their [release announcement](https://bokeh.github.io/blog/2017/6/13/release-0-12-6/).

My understanding of `environment.yml` is that it contains the packages needed to use all of Tax-Calc's features. If that's incorrect I'll update the PR to remove `phantomjs` from it.

martinholmer:
@andersonfrailey said:
> `phantomjs` is used by Bokeh to capture the PNG image. What's essentially happening is Bokeh takes a screenshot of what would be rendered in the browser and saves it. They give some details in their [release announcement](https://bokeh.github.io/blog/2017/6/13/release-0-12-6/).

When I go to the Bokeh release announcement and search for `phantomjs`, I find nothing.
Where does it say that `phantomjs` is **required** to use bokeh 0.12.6?

andersonfrailey:
@martinholmer, sorry, I should have added a second link to my post. It's explicitly state in the [PR that added this feature](https://github.com/bokeh/bokeh/pull/6189).

When you try to use it without phantomjs installed you receive this error:
```
RuntimeError: PhantomJS is not present in PATH. Try "conda install phantomjs" or                            "npm install -g hpantomjs-prebuilt"
```
You also need pillow and selenium. Errors weren't raised for either of those when I was working on this because I forgot to switch to the taxcalc-dev environment and had them installed already in the environment I was working in. I can add them to `environment.yml` in my next commit.

martinholmer:
@andersonfrailey said:
> sorry, I should have added a second link to my post. It's explicitly state in the PR that added this feature.
>
> When you try to use it without phantomjs installed you receive this error:
>
>```
> RuntimeError: PhantomJS is not present in PATH.
> Try "conda install phantomjs" or "npm install -g hpantomjs-prebuilt"
>```
> **You also need pillow and selenium**. Errors weren't raised for either of those when I was working on this because I forgot to switch to the taxcalc-dev environment and had them installed already in the environment I was working in. I can add them to `environment.yml` in my next commit.

Two things:

First, why don't we leave the .png graph output as a Tax-Calculator **option**?  Which means we document the dependencies in the `utils.py` docstring, but omit the three new packages from the `environment.yml` file.  Presumably, if somebody wants to write graphs as .png files, they will add the three extra packages to their environment.  The cost of adding packages is not zero; the taxcalc package gets bigger and the GitHub tests that longer to set up.

Second, I don't see how you can add `phantomjs` to the `environment.yml` file because it appears as if there is no conda `phantomjs` package for linux.  This seems to be why you can't pass the check-in tests.

Taken together, these two things suggest strongly that this .png feature should remain an option.

Oh, and by the way, how are you going to test the new code?  Once you pass the unit tests, your pull request is going to fail the coverage test.

Can we step back to the very beginning?  Why do we need this capability?
More options are better but in this case the extra option entails considerable costs?
What's the problem in having only .html graph output?

andersonfrailey:
@martinholmer the motivating reason behind this PR was to give the user one more option for saving an image because we already do have the `write_graph_file` method. An HTML file would generally be satisfactory, but if the user wanted to include the graph in a paper they would have to go through the process of opening and saving the file themselves.

Will and I ran into this issue when we were working on a paper a few months ago. We were writing it in LaTex, which can read PNG file and include them in the paper automatically, but is not able to open an HTML file. Basically it's just a convenience issue.

I'm seeing that this would take much more leg work than I initially anticipated, and users could always install the necessary packages themselves to take advantage of Bokeh's new features, so I have no problem with tabling this PR indefinitely.

martinholmer:
@andersonfrailey said:
> The motivating reason behind this PR was to give the user one more option for saving an image because we already do have the `write_graph_file` method. An HTML file would generally be satisfactory, but if the user wanted to include the graph in a paper they would have to go through the process of opening and saving the file themselves.
>
> Will and I ran into this issue when we were working on a paper a few months ago. We were writing it in LaTex, which can read PNG file and include them in the paper automatically, but is not able to open an HTML file. Basically it's just a convenience issue.

Yes, I noticed this problem when I first worked with the graphing capabilities in the `utils.py` file.  It's not very elegant, but while viewing the graph in your browser you can do a "screen shot" and crop the "screen" to be the size of the graph itself.  At least on a Mac, the resulting "screen shot" is saved as a .png file, which could then be imported into your document.

@andersonfrailey continued:
> I'm seeing that this would take much more leg work than I initially anticipated, and users could always install the necessary packages themselves to take advantage of Bokeh's new features, so I have no problem with tabling this PR indefinitely.

Besides the size of the extra packages required by bokeh to generate a .png graph file, it seems the key problem is the apparent lack of a phantomjs conda package for Linux.   Have you investigated that problem?  That is what is causing the test failure when you check-in your pull request on GitHub.

Closing this pull request would mean doing creating the .png files by hand (as described above).
It's not convenient.  But are you OK with that?  The real problem here is that bokeh is a very young project and so it doesn't have hardly any of the file-output capabilities that are routine in other graphics programs.  I'm not sure why our project settled on bokeh because I wasn't involved in the early discussions.

hdoupe:
@andersonfrailey While the Bokeh graphs are much nicer the matplotlib graphs, you can save matplotlib graphs as png's pretty easily.  Until Bokeh develops a better graph saving solution, a short-term fix could be to just use matplotlib if the user wants to save a graph.

@martinholmer

martinholmer:
@hdoupe said:
> While the Bokeh graphs are much nicer than matplotlib graphs, you can save matplotlib graphs as png's pretty easily. Until Bokeh develops a better graph saving solution, a short-term fix could be to just use matplotlib if the user wants to save a graph.

I haven't made any visual comparisons.  Are bokeh graphs **much** nicer looking?  In what respects are they nicer?

You're absolutely right about the wide range of graph output formats.  Here is what the docs say:

> Here is a summary of the matplotlib renderers (there is an eponymous backed for each; these are non-interactive backends, capable of writing to a file):
>
> Renderer | Filetypes | Description
> -- | -- | --
> AGG | png | raster graphics – high quality images using the Anti-Grain Geometry engine
> PS | ps eps | vector graphics – Postscript output
> PDF | pdf | vector graphics – Portable Document Format
> SVG | svg | vector graphics – Scalable Vector Graphics
> Cairo | png ps pdf svg ... | vector graphics – Cairo graphics
> GDK | png jpg tiff ... | raster graphics – the Gimp Drawing Kit Deprecated in 2.0

This information makes me wonder even more why the project went with bokeh.

@MattHJensen @andersonfrailey @GoFroggyRun

andersonfrailey:
@martinholmer asked:
> it seems the key problem is the apparent lack of a phantomjs conda package for Linux. Have you investigated that problem? That is what is causing the test failure when you check-in your pull request on GitHub.

A quick search through the [Anaconda Package List](https://docs.continuum.io/anaconda/packages/pkg-docs) confirms conda only has the phantomjs package for mac and windows. I haven't found any information as to why that's the case yet.

> Closing this pull request would mean doing creating the .png files by hand (as described above).
It's not convenient. But are you OK with that?

I'm ok with that for now because it's still possible for a user to automate the .png files if they really wanted. They just have to install the extra packages on their own. Everything is available using conda so it wouldn't require too much extra effort. I would love to eventually add this feature, but figuring out how we could do that and still pass all the tests is a much lower priority than other work that needs to be done so I'm in no rush.

> I haven't made any visual comparisons. Are bokeh graphs much nicer looking? In what respects are they nicer?

I would say they're nicer looking, but that's purely from a personal preference standpoint. I don't have any objective standard by which I'm making that determination.

martinholmer:
@andersonfrailey said about bokeh pull request #1504:
> A quick search through the Anaconda Package List confirms conda only has the `phantomjs` package for mac and windows. I haven't found any information as to why that's the case yet.
>
>> Closing this pull request would mean doing creating the .png files by hand (as described above).  It's not convenient. But are you OK with that?
>
> I'm ok with that for now because it's still possible for a user to automate the .png files if they really wanted. They just have to install the extra packages on their own. Everything is available using conda so it wouldn't require too much extra effort. I would love to eventually add this feature, but figuring out how we could do that and still pass all the tests is a much lower priority than other work that needs to be done so I'm in no rush.

OK.  Sorry this has been such a hassle.

I'm going to close #1504 now, but it will always be available on GitHub from the list of closed pull requests.  I think this would be an improvement, but the bokeh group is a little slow rolling out these enhancements in a self-contained way.  The bokeh package should be able to export graphs in PNG format as a core capability that does not require any extra packages.  Maybe they will get there sometime in the future.

PR #1524, opened 2017-08-22
Add logic to report reform parameter warnings and errors

martinholmer:
This pull request implements the approach to warnings and errors discussed at the end of the discussion in [TaxBrain issue 630](https://github.com/OpenSourcePolicyCenter/webapp-public/issues/630).  As such, it replaces the less flexible approach used in #1502.

The Tax-Calculator CLI, tc, now uses these warning and error messages.  See the changes to the `taxcalcio.py` file for details.

TaxBrain can use this new Tax-Calculator capability by calling the `reform_warnings_errors` function in the `dropq.py` file, which is new to the Tax-Calculator public API.  Read this new function's docstring carefully before using it in TaxBrain.  The consensus in #630 is to show warnings and proceed with tax calculation if there are no errors.  If there are errors, we should show all the warning and error messages and then stop without doing any tax calculations.  That is what the Tax-Calculator CLI, tc, does now.

If anyone sees a better way to provide these capabilities, please make your suggestion here.

@MattHJensen @feenberg @Amy-Xu @andersonfrailey @hdoupe @GoFroggyRun @codykallen 
@jdebacker @rickecon @brittainhard @talumbau

Issue #1485, opened 2017-07-18
EITC phase-in rate, "_EITC_rt"


MattHJensen:
A TaxBrain user is wanting to set an EITC rate (`eitc_rt`) over 100 percent. 

The EITC rate description states that the range of the parameter is between 0 and 100 percent. 

To experiment, I ran two reforms, one with the rate set to 100 percent for all filing statuses and one with the rates at 150 percent, and I got identical results for both.

All EITC rates at 100 percent.
http://www.ospc.org/taxbrain/15189/

All EITC rates at 150 percent. 
http://www.ospc.org/taxbrain/15190/

- can we allow rates over 100 percent?
- was this a silent error, and are there other scenarios where that might be happening?

martinholmer:
The current description of the `_EITC_rt` policy parameter dates back to the merge of pull request #327 on August 12, 2015.  Here are the changes made then with respect to `_EITC_rt`:
```
      "_EITC_rt":{
          "long_name": "Earned income credit rate",
-         "description": "Apply this rate on Income.",
+         "description": "If one taxpayer's AGI is below AGI phaseout start, he can apply this rate to his earned income to calculate the credit amount eligible. ",
          "irs_ref": "Form 1040, line 66a&b, calculation (table: Max_EIC/Max_EIC_base_income).",
-         "notes": "",
+         "notes": "Range of this parameter is between zero and one. ",
          "start_year": 2013,
```
There have been no changes in this `_EITC_rt` entry since then.

I'm going to do several things:
(1) investigate whether or not the `notes` information is enforced any place in Tax-Calculator;
(2) improve the `description` wording; and
(3) drop the `notes` text if Tax-Calculator does not enforce the zero-to-one range.

I will report what I find in my investigations in a pull request, which at a minimum will do item (2).

@MattHJensen

martinholmer:
@MattHJensen,

There is no place in Tax-Calculator that constrains the value of `_EITC_rt`.  It is simply passed unchanged to the `EITCamount` function as its first `phasein_rate` parameter:
```
def EITCamount(phasein_rate, earnings, max_amount,
               phaseout_start, eitc_agi, phaseout_rate):
    """
    Return EITC amount given specified parameters
    """
    eitc = min(phasein_rate * earnings, max_amount)
    if earnings > phaseout_start or eitc_agi > phaseout_start:
        eitcx = max(0., (max_amount - phaseout_rate *
[...]

PR #1529, opened 2017-08-26
permissive validation

MattHJensen:
The new warn/stop capabilities in #1524 and #1527 are great. Thanks @martinholmer for introducing them. 

I do not feel, however, that the extensive use of 'stop' in those PRs is in line with our longstanding goal to provide as much flexibility as possible to our users. For example, there were many places where a rate was bound between 0 and 1. While I certainly understand and support the inclination to warn users who use a rate outside that range as they might have entered 10 for 10%, I don't think it makes sense to return an error as some users will set rates outside [0,1] intentionally. I and others I know of have done so to achieve our objectives with Tax-Calculator. @feenberg has reported TaxSim users doing the same. 

This PR limits the use of "stop" to logical errors only, such as the one described in https://github.com/OpenSourcePolicyCenter/webapp-public/issues/630. 

(Aside from the usability aspects, if we were to adopt the stricter validation introduced in #1524, then I would argue that the next release should be major (increment y in x.y.z since we are below 1.0.0) rather than minor.)

### DIFF
--- a/taxcalc/current_law_policy.json
+++ b/taxcalc/current_law_policy.json
@@ -16,7 +16,7 @@
         "range": {"min": 0, "max": 1},
         "out_of_range_minmsg": "",
         "out_of_range_maxmsg": "",
-        "out_of_range_action": "stop"
+        "out_of_range_action": "warn"
     },
 
     "_SS_Earnings_c": {
@@ -44,7 +44,7 @@
         "range": {"min": 0, "max": 9e99},
         "out_of_range_minmsg": "",
         "out_of_range_maxmsg": "",
-        "out_of_range_action": "stop"
+        "out_of_range_action": "warn"
     },
 
     "_FICA_mc_trt": {
@@ -64,7 +64,7 @@
         "range": {"min": 0, "max": 1},
         "out_of_range_minmsg": "",
         "out_of_range_maxmsg": "",
-        "out_of_range_action": "stop"
+        "out_of_range_action": "warn"
     },
 
     "_AMEDT_ec": {
@@ -84,7 +84,7 @@
         "range": {"min": 0, "max": 9e99},
         "out_of_range_minmsg": "",
         "out_of_range_maxmsg": "",
-        "out_of_range_action": "stop"
+        "out_of_range_action": "warn"
     },
 
     "_AMEDT_rt": {
@@ -104,7 +104,7 @@
         "range": {"min": 0, "max": 1},
         "out_of_range_minmsg": "",
         "out_of_range_maxmsg": "",
-        "out_of_range_action": "stop"
+        "out_of_range_action": "warn"
     },
 
 
@@ -145,7 +145,7 @@
         "range": {"min": 0, "max": 1},
         "out_of_range_minmsg": "",
         "out_of_range_maxmsg": "",
-        "out_of_range_action": "stop"
+        "out_of_range_action": "warn"
     },
 
     "_SS_thd85": {
@@ -185,7 +185,7 @@
         "range": {"min": 0, "max": 1},
         "out_of_range_minmsg": "",
         "out_of_range_maxmsg": "",
-        "out_of_range_action": "stop"
+        "out_of_range_action": "warn"
     },
 
 
@@ -206,7 +206,7 @@
         "range": {"min": 0, "max": 1},
         "out_of_range_minmsg": "",
         "out_of_range_maxmsg": "",
-        "out_of_range_action": "stop"
+        "out_of_range_action": "warn"
     },
 
     "_ALD_SelfEmploymentTax_hc": {
@@ -226,7 +226,7 @@
         "range": {"min": 0, "max": 1},
         "out_of_range_minmsg": "",
         "out_of_range_maxmsg": "",
-        "out_of_range_action": "stop"
+        "out_of_range_action": "warn"
     },
 
     "_ALD_SelfEmp_HealthIns_hc": {
@@ -246,7 +246,7 @@
         "range": {"min": 0, "max": 1},
         "out_of_range_minmsg": "",
         "out_of_range_maxmsg": "",
-        "out_of_range_action": "stop"
+        "out_of_range_action": "warn"
     },
 
     "_ALD_KEOGH_SEP_hc": {
@@ -266,7 +266,7 @@
         "range": {"min": 0, "max": 1},
         "out_of_range_minmsg": "",
         "out_of_range_maxmsg": "",
-        "out_of_range_action": "stop"
+        "out_of_range_action": "warn"
     },
 
     "_ALD_EarlyWithdraw_hc": {
@@ -286,7 +286,7 @@
         "range": {"min": 0, "max": 1},
         "out_of_range_minmsg": "",
         "out_of_range_maxmsg": "",
-        "out_of_range_action": "stop"
+        "out_of_range_action": "warn"
     },
 
     "_ALD_Alimony_hc": {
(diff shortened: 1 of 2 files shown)


martinholmer:
@MattHJensen said:
> I do not feel, however, that the extensive use of 'stop' in [#1524 and #1527] is in line with our longstanding goal to provide as much flexibility as possible to our users. For example, there were many places where a rate was bound between 0 and 1. While I certainly understand and support the inclination to warn users who use a rate outside that range as they might have entered 10 for 10%, I don't think it makes sense to return an error as some users will set rates outside [0,1] intentionally. I and others I know of have done so to achieve our objectives with Tax-Calculator. @feenberg has reported TaxSim users doing the same.

Can you provide examples of reforms you analyzed where you intentionally set a rate outside the [0,1] range?

martinholmer:
@MattHJensen,  There is a problem with your proposed changes in #1529.  Please read the whole discussion in [TaxBrain issue 628](https://github.com/OpenSourcePolicyCenter/webapp-public/issues/628).  Your changes in #1529 are so "permissive" that it will allow the crash reported in 628.  Allowing that to occur doesn't make any sense to me.  Don't we want to help users?  I certainly would have appreciated some help in the 628 situation.

Can you provide a concrete example of why it would be useful to specify a reform that has a payroll tax rate above 1.0 (that is, above 100 percent)?

martinholmer:
@MattHJensen, on further review of #1529, I notice that you want to do tax calculations on reforms that specify "haircut" parameters outside the [0,1] range.  Do you know whether or not Tax-Calculator will produce sensible results when a user does that?  What does a negative "haircut" parameter mean?  What does a "haircut" parameter of two mean?  In what kind of tax reforms would it be useful to be able to specify "haircut" parameters outside the [0,1] range?

MattHJensen:
> There is a problem with your proposed changes in #1529. Please read the whole discussion in TaxBrain issue 628. Your changes in #1529 are so "permissive" that it will allow the crash reported in 628. 

I don't think that necessarily indicates a problem with #1529. 

Over at TaxBrain issue 628, you said: 

> Even with the absurdly high payroll tax rate, I would think the double-precision arithmetic we are using in Tax-Calculator would simply calculate enormous payroll tax revenue beginning in 2020. 

That would be my expectation as well. 

So I think it is more accurate to say that there is a problem somewhere in Tax-Calculator that is leading to a result that confounds both of our expectations than it is to say that there is a problem in #1529. 

That said, I do see how permissive validation is 'at fault' for requiring us to ensure Tax-Calculator's proper functioning under extreme parameter values. I could make arguments both ways for whether it is good or bad to need to understand how Tax-Calculator works under extreme parameter values.  

Putting all of that aside, though, I have been thinking about your questions:

> Can you provide examples of reforms you analyzed where you intentionally set a rate outside the [0,1] range?

Both of the examples I have in mind involved setting phase in rates above 1. See, e.g., https://github.com/open-source-economics/Tax-Calculator/issues/1485. 

> Can you provide a concrete example of why it would be useful to specify a reform that has a payroll tax rate above 1.0 (that is, above 100 percent)?

Not really. Perhaps @feenberg has an example. I do think it is reasonable to specify a payroll tax rate below 0, (and even below -1.)

> In what kind of tax reforms would it be useful to be able to specify "haircut" parameters outside the [0,1] range?

Consider a haircut on an itemized deduction: 

A haircut of -1 means that 200% of the expense is deductible. This seems conceptually reasonable to me and appears to work well in the calculator, although I haven't tested all of the haircuts. 

A haircut of 2 means that the expense is subtracted from total deductions allowed rather than added. This also seems conceptually reasonable to me and appears to work well in Tax-Calculator. 

> Can you provide examples of reforms you analyzed where you intentionally set a rate outside the [0,1] range? 

Both of the examples I have in mind involved setting phase in rates above 1. See, for example, https://github.com/open-source-economics/Tax-Calculator/issues/1485. 

-- 

Given my answer to your first question, especially, I can see that it may make sense to validate more strictly than what is described in #1529. In particular, tax rates could be bound above at 1. For now, that approach is still my preference. 

I am also open, though, to withdrawing this PR and validating as strictly as #1524/#1527, with some added flexibility for phase in rates. If we pursue this approach, I do think the next release should be major as we will be breaking conceptually valid functionality, and I also would want us to be open to relaxing some of the validations in the future if requested by users. So basically, it will just mean that any taxcalc package or TaxBrain user who wants do something weird but conceptually valid will have to wait a couple of weeks for the next release before they can do it.

martinholmer:
@MattHJensen said a number of things in the discussion of #1529:

> So I think it is more accurate to say that there is a problem somewhere in Tax-Calculator that is leading to a result [described in 628] that confounds both @martinholmer and my expectations than it is to say that there is a problem in #1529.

Perhaps so.  It would take a lot of work to get to the bottom of what is causing the error reported in [TaxBrain issue 628](https://github.com/OpenSourcePolicyCenter/webapp-public/issues/628).

> Both of the examples I have in mind [of setting rates outside the zero-one range] involve setting phase in rates above 1.  See, for example, #1485.

Yes, I agree, phase-in rates definitely need to be able to be set above one.  My search of `current_law_policy.json` found two phase-in rates: `_EITC_rt` (which is the focus of #1485) and `_CTC_new_rt` (the phase-in rate as a fraction of AGI for the new child tax credit).  But the `range` for these two parameters is `[0,9e99]` not `[0,1]`.  So, users will have no problem phasing in at a very fast rate either the maximum allowable EITC amount or new child tax credit.

>> Can you provide a concrete example of why it would be useful to specify a reform that has a payroll tax rate above 1.0 (that is, above 100 percent)?
>
> Not really. Perhaps @feenberg has an example. I do think it is reasonable to specify a payroll tax rate below 0, (and even below -1.)

First, let me say I'm confused about your earlier comment in the discussion saying that @feenberg has reported TAXSIM users setting tax rates outside the zero-one range.  [Internet TAXSIM](http://users.nber.org/~taxsim/taxsim-calc9/) provides users complete flexibility in specifying 22 attributes of each filing unit.  But there is no way for Internet TAXSIM users to specify changes in specific tax rates; Internet TAXSIM basically simulates current-law policy.  Now it is true that there are a small number of ["options"](http://users.nber.org/~taxsim/taxsim-calc9/plans.html), but when I scan that list I don't see any tax rates outside the one-one range. 

Second, can you provide some motivation and explanation of a reform that involves a negative payroll tax rate.  Perhaps my imagination is limited, but I'm having trouble imagining what kind of reform would call for negative payroll taxes.

> Consider a haircut on an itemized deduction:
>
> A haircut of -1 means that 200% of the expense is deductible. This seems conceptually reasonable to me and appears to work well in the calculator, although I haven't tested all of the haircuts.
>
> A haircut of 2 means that the expense is subtracted from total deductions allowed rather than added. This also seems conceptually reasonable to me and appears to work well in Tax-Calculator.

So, when the haircut is -1, the taxpayer gets to deduct twice the amount of their actual itemized expense.  I can see the arithmetic of it, but again I'm having a difficult time coming up with a reform that I could describe in English that would involve a -1 haircut.  Same for a haircut value of 2: I can understand the arithmetic, but can't think of a plausible reform that would involve subtracting your, say, charitable contributions from total itemized deductions.  

> I am also open, though, to withdrawing this PR and validating as strictly as #1524/#1527, with some added flexibility for phase in rates. If we pursue this approach, I do think the next release should be major since it will break conceptually valid functionality. I would also want us to be open to relaxing some of the validations in the future if requested by users. So basically, it will just mean that any taxcalc package user or TaxBrain user who wants do something weird but conceptually valid will have to wait a couple of weeks for the next release before they can do it.

To me this seems like a sensible approach.  We don't want to make the validation tests so restrictive that they prevent TaxBrain users from doing something reasonable (albeit unusual).

Unless I've missed something, there is no need to have "added flexibility for phase in rates" because those rates can already be as high as 9e99, which implies a very fast phase-in.

And I fully agree with the notion of "relaxing some of the validations in the future if requested by users".

MattHJensen:
Closing this PR in favor of stricter validation that will be relaxed upon request by users. Thanks again to @martinholmer for the warn/stop capabilities that allow us to have these discussions about the appropriate level of validation.

PR #1521, opened 2017-08-21
[WIP] Add After-Tax Income Percent Change column to dropq results

andersonfrailey:
In PR #1375, I added a feature to include the percent change in after tax income in the difference table. @hdoupe and I are trying to add it to TaxBrain as well and the purpose of this PR is primarily to solicit help. We believe that one of the first steps is to modify `dropq.py` and `dropq_utils.py`, as shown in this PR. Any help progressing from this point would be appreciated.

PR #1534, opened 2017-08-30
Simplify create_difference_table utility and use it in dropq logic

martinholmer:
This pull request was developed after looking at the problems faced by @andersonfrailey in pull request #1521.  The goal of that pull request was simple enough and his skills are substantial, but the complexity of the dropq difference table logic made the task in #1521 quite challenging.  I looked at the issue and couldn't figure out how to reach the goal of #1521.  My conclusion was that the dropq difference table requirements were not that complicated, but that the code was extremely complicated and difficult to follow.

This pull request is an attempt to simplify the dropq difference table code.  The main strategy is to develop the `create_difference_table` utility function code to be general enough to support difference table creation in the `dropq_utils.py` file.

In the course of doing that, it was easy to add two features missing in the old dropq logic.

- FIrst, it is now possible to create difference tables with an income measure that the TaxBrain difference table calls "Adjusted Income", which I assume is AGI.  The high-level dropq logic does not generate difference tables with AGI as the income measure, but doing so would now be trivial.  

- Second, the dropq difference tables now include the tax difference expressed relative to baseline after-tax expanded income.  However, TaxBrain would have to be revised to display that new tax difference statistic.

@MattHJensen @andersonfrailey @hdoupe @GoFroggyRun @brittainhard


hdoupe:
@martinholmer This looks great.  Thanks for cleaning up the dropq code!

I'll take a look at incorporating these changes into TaxBrain.

martinholmer:
Part of pull request #1534 adds a column label at the end of the `DIFF_TABLE_LABELS` list in `utils.py`.  I'm struggling with the wording of the new label.  Right now its too long and not very clear, or at least, I worry about that being so.  Now I have this:
```
DIFF_TABLE_LABELS = ['Tax Units with Tax Cut',
                     'Tax Units with Tax Increase',
                     'Count',
                     'Average Tax Change',
                     'Total Tax Difference',
                     'Percent with Tax Increase',
                     'Percent with Tax Decrease',
                     'Share of Overall Change',
                     'Change relative to Aftertax Income']
```
Can anyone suggest a better label?  I'm open to any suggestions.

@MattHJensen @andersonfrailey @hdoupe

andersonfrailey:
Thanks for working on this @martinholmer!

>Can anyone suggest a better label? I'm open to any suggestions.

What about "Change as % of Aftertax Income"? It's only a couple characters shorter than what you have now, but I believe it's a little more clear.

martinholmer:
@hdoupe said:
> Thanks for cleaning up the dropq code!   
> I'll take a look at incorporating these changes into TaxBrain.

Before you start on that, let me ask you (and the others) about one thing I'm thinking about.

The result that is returned from the dropq `run_nth_year_tax_calc_model` function is a tuple containing 13 (yes, thirteen) objects.  As we go forward, it seems like this is an accident waiting to happen.  In other places in the code, we've replaced tuples with dictionaries.  What do you think about doing that in this case?  This change would make the dropq code less error-prone and should make working with the TaxBrain code easier, but it is a change in the public API of the taxcalc package.  All thoughts are welcome.

@MattHJensen @Amy-Xu @andersonfrailey @GoFroggyRun @brittainhard

martinholmer:
@andersonfrailey said:
> What about "Change as % of Aftertax Income"? It's only a couple characters shorter than what you have now, but I believe it's a little more clear.

Thanks.  Your suggestion is definitely better than what I have now.
I'll make this change and see if anybody else has an even better suggestion.

@MattHJensen @hdoupe

martinholmer:
Pull request #1534 is now complete.
I'll leave it open a few days for people to review and post questions.

@MattHJensen @Amy-Xu @andersonfrailey @hdoupe @GoFroggyRun @brittainhard

PR #1547, opened 2017-09-13
Add bootstrap_se_ci utility function and associated test

martinholmer:
This pull request adds a utility function that computes bootstrap a standard error and confidence interval for the specified data and statistic.  The test of the function uses data and bootstrap results from Efron and Tibshirani's classic 1993 book *An Introduction to the Bootstrap*.

### DIFF
--- a/taxcalc/tests/test_utils.py
+++ b/taxcalc/tests/test_utils.py
@@ -28,6 +28,7 @@ from taxcalc.utils import (STATS_COLUMNS,
                            mtr_graph_data, atr_graph_data,
                            xtr_graph_plot, write_graph_file,
                            read_egg_csv, read_egg_json, delete_file,
+                           bootstrap_se_ci,
                            certainty_equivalent, ce_aftertax_income)
 
 
@@ -704,3 +705,18 @@ def test_create_delete_temp_file():
     assert os.path.isfile(fname) is True
     delete_file(fname)
     assert os.path.isfile(fname) is False
+
+
+def test_bootstrap_se_ci():
+    # Use treated mouse data from Table 2.1 and
+    # results from Table 2.2 and Table 13.1 in
+    # Bradley Efron and Robert Tibshirani,
+    # "An Introduction to the Bootstrap"
+    # (Chapman & Hall, 1993).
+    data = np.array([94, 197, 16, 38, 99, 141, 23], dtype=np.float64)
+    assert abs(np.mean(data) - 86.86) < 0.005  # just rounding error
+    bsd = bootstrap_se_ci(data, 123456789, 1000, np.mean, alpha=0.025)
+    # following comparisons less precise because of rn seed differences
+    assert abs(bsd['se'] / 23.02 - 1) < 0.02
+    assert abs(bsd['cilo'] / 45.9 - 1) < 0.02
+    assert abs(bsd['cihi'] / 135.4 - 1) < 0.03
--- a/taxcalc/utils.py
+++ b/taxcalc/utils.py
@@ -1291,3 +1291,31 @@ def delete_file(filename):
     """
     if os.path.isfile(filename):
         os.remove(filename)
+
+
+def bootstrap_se_ci(data, seed, num_samples, statistic, alpha):
+    """
+    Return bootstrap estimate of standard error of statistic and
+    bootstrap estimate of 100*(1-2*alpha)% confidence interval for statistic
+    in a dictionary along with seed and nun_samples (B) and alpha.
+    """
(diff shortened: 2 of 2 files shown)


feenberg:
There is a nice analytic formula for the standard error at:

   http://www.itl.nist.gov/div898/software/dataplot/refman2/ch2/weightsd.pdf

but Stata (for one) doesn't offer it as an option. I have always wondered 
if there was some reason it wasn't generally applicable, and if that is 
the case, is it applicable to our TB? It would be a substantial economy 
over bootstrapping.

dan

martinholmer:
@feenberg said with respect to pull request #1547:
> There is a nice analytic formula for the standard error at:
>
>   http://www.itl.nist.gov/div898/software/dataplot/refman2/ch2/weightsd.pdf
>
> but Stata (for one) doesn't offer it as an option. I have always wondered if there was some reason it wasn't generally applicable, and if that is the case, is it applicable to our TB? It would be a substantial economy over bootstrapping.

A formula would always be faster and more accurate than a bootstrap estimate, but the problem is that there are no simple standard-error formulas for statistics other than the sample mean.

Here are some relevant quotes from the Efron and Tibshirani book:

From page 10:
> The bootstrap is a computer-based method for assigning measures of accuracy to statistical estimates.  The basic idea behind the bootstrap is very simple, and goes back at least two centuries.  After reviewing some background material, this book describes the bootstrap method, it implementation on the computer, and its application to some real data analysis problems.  First though, this chapter on the one example of a statistical estimator where we really don't need a computer to assess accuracy: the sample mean.

From page 15:
> Standard errors [of the sample mean] are the simplest measures of statistical accuracy.  Later chapters show how bootstrap methods can assess more complicated accuracy measures, like biases, prediction errors, and confidence intervals.  Bootstrap confidence intervals add another factor of 10 [relative to standard error estimation] to the computational burden.  The payoff for all this computation is an increase in the statistical problems that can be analyzed, a reduction in the [distributional] assumptions of the analysis, and the elimination of the routine but tedious theoretical calculations usually associated with accuracy assessment.

feenberg:
On Wed, 13 Sep 2017, Martin Holmer wrote:

> 
> @feenberg said with respect to pull request #1547:
>
>       There is a nice analytic formula for the standard error at:
>
>       http://www.itl.nist.gov/div898/software/dataplot/refman2/ch2/weightsd.pdf
>
>       but Stata (for one) doesn't offer it as an option. I have always wondered
>       if there was some reason it wasn't generally applicable, and if that is the
>       case, is it applicable to our TB? It would be a substantial economy over
>       bootstrapping.
> 
> A formula would always be faster and more accurate than a bootstrap estimate, but the
> problem is that there are no simple standard-error formulas for statistics other than
> the sample mean.
> 
> Here are some relevant quotes from the Efron and Tibshirani book:
> 
> From page 10:
>
>       The bootstrap is a computer-based method for assigning measures of accuracy
>       to statistical estimates. The basic idea behind the bootstrap is very
>       simple, and goes back at least two centuries. After reviewing some
>       background material, this book describes the bootstrap method, it
>       implementation on the computer, and its application to some real data
>       analysis problems. First though, this chapter on the one example of a
>       statistical estimator where we really don't need a computer to assess
>       accuracy: the sample mean.
> 
> From page 15:
>
>       Standard errors [of the sample mean] are the simplest measures of
>       statistical accuracy. Later chapters show how bootstrap methods can assess
>       more complicated accuracy measures, like biases, prediction errors, and
>       confidence intervals. Bootstrap confidence intervals add another factor of
>       10 [relative to standard error estimation] to the computational burden. The
>       payoff for all this computation is an increase in the statistical problems
>       that can be analyzed, a reduction in the [distributional] assumptions of
>       the analysis, and the elimination of the routine but tedious theoretical
>       calculations usually associated with accuracy assessment.

So you think the formula at

        http://www.itl.nist.gov/div898/software/dataplot/refman2/ch2/weightsd.pdf

is wrong? Or not simple? Or are we calculating "more complicated accuracy 
meansures"? I was afraid it was wrong given the way our sample was 
stratified, it doesn't look that complicated to me. It is available in SAS 
proc means, but nowhere in Stata.


dan

martinholmer:
@feenberg said:
> So you think the formula at
>
>  http://www.itl.nist.gov/div898/software/dataplot/refman2/ch2/weightsd.pdf
>
is wrong?  Or not simple? Or are we calculating "more complicated accuracy 
measures"? 

The formula is correct for the simplest case, but we are calculating "more complicated accuracy measures" (like confidence intervals of count statistics).

Issue #1562, opened 2017-09-22
taxcalc/comparison/stats_summary.py script does not work


martinholmer:
The `taxcalc/comparison/variable_stats_summary.csv` file was last generated on May 4, 2017.

The `taxcalc/comparison/correlation.csv` file was last generated on April 20, 2017.

According to the [taxpuf repo](https://github.com/OpenSourcePolicyCenter/taxpuf/releases), the `puf.csv` file has changed twice since then: once on June 16 and again on July 5.

But the `taxcalc/comparison/stats_summary.py` script no longer works.

@Amy-Xu, can you look into this?

Amy-Xu:
@martinholmer I will look into this script shortly.

martinholmer:
Issue #1562 has been resolved by pull request #1566.

PR #1566, opened 2017-09-25
Update stats summary script

Amy-Xu:
This script has been updated to work with the latest version of Tax-Calculator. Output files get updated as well. Thanks to Martin for pointing out in issue #1562.


Amy-Xu:
@martinholmer this PR fixed the `stats_summary.py` in comparison. Could you review?

martinholmer:
@Amy-Xu said in #1566:
> This PR fixed the `stats_summary.py` in comparison. Could you review?

Thanks.  Yes.  I'll try to review it no later than tomorrow morning.

martinholmer:
@Amy-Xu,  When I download pull request #1566 to my computer and do the two runs, I get the same results in the `variable_stats_summary.csv` file, but I don't get the same results in the `correlation.csv` file.  My suspicion is that this is caused by you showing excessive floating-point precision in the `correlation.csv` file.  We already have other evidence that floating point calculations are not exactly the same on Python 2.7 and Python 3.x.  And besides, telling users that the correlation coefficient between two variables has a value of, say, `-0.0050643581807` is not sensible.  Not only is is hard to read a fraction with that many digits, it assigns way too much precision.  Why not round the correlation coefficients to three decimal points, so that instead of `-0.0050643581807`, you show `-0.005`?  Is there any practical difference between those two values?

martinholmer:
@Amy-Xu,  I haven't studied the code in the `stats_summary.py` script, but I was wondering whether it is possible to have that script produce **both** output files at the same time.  Would it be possible to do that without an excessive amount of work?

Amy-Xu:
@martinholmer Thanks for the review. I'll round all results to three decimal points. 

In terms of generating both files at the same time, the original thought was simply that users may just be interested in one of them so it would be the most convenient for users to produce one a time. Since we also need to update the files, it makes sense to me producing both at the same time. I'll revise the code accordingly.

martinholmer:
@Amy-Xu said:
> Thanks for the review. I'll round all results to three decimal points.
>
> In terms of generating both files at the same time, the original thought was simply that users may just be interested in one of them so it would be the most convenient for users to produce one a time. Since we also need to update the files, it makes sense to me producing both at the same time. I'll revise the code accordingly.

Thanks.

martinholmer:
@Amy-Xu, Thanks for this update.  Look good, so I'll merge it now.
I already updated the `RELEASES.md` info to describe your enhancement.

PR #1569, opened 2017-09-27
Specify parameter arrays as having either integer or float data type

martinholmer:
This pull request changes the long-standing practice of specifying **all** parameters as numpy arrays with dtype=numpy.float64.  This old practice worked fine most of the time because most parameters were float values.  But when parameters were boolean or true integers (like age), the old practice produced correct but potentially confusing results.

The new practice is to handle parameter arrays in the same way as Records variables are handled.  To implement this new approach, all four of the JSON parameter files now have a new `integer_value` field (which can be true or false) for each parameter.  And the ParametersBase `_expand_array` method has been revised to create numpy arrays from the specified list that have a data type consistent with the value of that parameter's `integer_value` attribute.

New tests have been added to make sure that the parameter values specified in the four JSON parameter files have values that are consistent with the parameter's `boolean_value` and `integer_value` attributes.

And finally, this pull request reduces the size of integer numpy arrays: the integer Records-variable arrays have been reduced from int64 to int32 and the integer parameter arrays have been reduced from int64 to itt8.  Besides saving memory space during Tax-Calculator execution, this change reduces the execution time of the pytest suite by about five percent.

This pull request does not change any tax-calculating logic and all the test results are unchanged.

@MattHJensen @feenberg @Amy-Xu @andersonfrailey @hdoupe @GoFroggyRun


martinholmer:
Another benefit of the changes in pull request #1569, is that the reform documentation now shows the baseline_value of integer parameters as integers instead of floats.  So, in the example shown in pull request #1567, the following documentation fragment:
```
2017:
 _EITC_MinEligAge : 21
  name: Minimum Age for Childless EITC Eligibility
  desc: For a childless filling unit, at least one individual's age needs to
        be no less than this age (but no greater than the EITC_MaxEligAge) in
        order to be eligible for an earned income tax credit.
  baseline_value: 25.0
```
is now shown as follows:
```
2017:
 _EITC_MinEligAge : 21
  name: Minimum Age for Childless EITC Eligibility
  desc: For a childless filling unit, at least one individual's age needs to
        be no less than this age (but no greater than the EITC_MaxEligAge) in
        order to be eligible for an earned income tax credit.
  baseline_value: 25
```

@MattHJensen @feenberg @Amy-Xu @andersonfrailey @hdoupe @GoFroggyRun

MattHJensen:
> Another benefit of the changes in pull request #1569, is that the reform documentation now shows the baseline_value of integer parameters as integers instead of floats.

That looks much nicer.

PR #1574, opened 2017-09-29
Translate comparison/stats_summary.py script into a new test

martinholmer:
This pull request makes the calculation of weighted means and correlations coefficients a test rather than being a free-standing capability located in the `taxcalc/comparison` directory.  This is being done so that data enhancements and fixes in tax-calculating logic automatically show up in the weighted means and correlation coefficients for each variable.  The new test fails when the new statistics are not exactly the same as the old statistics, which are in the version-controlled files named  as follows:
```
taxcalc/tests/puf_var_wght_means_by_year.csv
taxcalc/tests/puf_var_correl_coeffs_2016.csv
```
These new names mean that HTML links in TaxBrain will need to be updated to point to these new files.

When developing the new test, it became clear that the variables for which statistics were calculated needed to be sorted alphabetically because not sorting left the order of the variables essentially random.  Another comment in this pull request documents how the new ordered results were compared with the old unordered results to confirm that the new test is producing the same results as the old script.

Also, when developing the new test, it become clear that there was a bug in the old script.  That bug caused the weighted means of all variables in the years after 2013 to be incorrect.  For example, the old script generated a 2026 weighted mean for AGI (c00100) that was too high:
```
2026 AGI = 100091.941   in old script
2026 AGI =  85059.871   in new test
```
The bug was placing the statement that computes the sum of all the weights outside of the year loop.  Doing this meant that the sum of the population weights was correct only in 2013 and increasingly too small in years after 2013.  Here is a code fragment from the old script:
```
def gen_sum_stats(table, calc):
    # calculates weighted mean for each variable
    # in total 10 years
    total_pop = calc.records.s006.sum()  # <------- BUG: outside year loop ----------
    for year in range(2013, 2027):
        calc.calc_all()
        stat = []
        for variable in table.index:
            weighted = getattr(calc.records, variable) * calc.records.s006
            this_record = weighted.sum() / total_pop
            stat.append(this_record)

        column = "mean_" + str(year)
        table[column] = stat

        year += 1
        if year != 2027:
            calc.increment_year()
    table.to_csv("variable_stats_summary.csv",
                 header=['description', '2013', '2014', '2015',
                         '2016', '2017', '2018', '2019', '2020',
                         '2021', '2022', '2023', '2024', '2025', '2026'],
                 float_format='%8.3f')
```

@MattHJensen @feenberg @Amy-Xu @andersonfrailey @hdoupe


martinholmer:
The following AWK scripts were used to verify that the new test produces the same results as the old script (apart from the weighted means for years after 2013).

The correlation coefficient results were checked with this script:
```
# check_corr.awk compares correlation coefficent of VAR1 and VAR2 in two files.
# USAGE: awk -f check_corr.awk -v var1=VAR1 -v var2=VAR2 FILE1 FILE2
# EXAMPLE: awk -f check_corr.awk -v var1=eitc -v var2=iitax         \
#                                puf_var_correl_coeffs_2016.csv     \
#                                puf_var_correl_coeffs_2016.csv-old
BEGIN {
    FS = ","
}
ARGIND==1 {
    if (FNR == 1) {
        for (i=0;i<=NF;i++) head1[$i] = i
    } else {
        if ($1 == var1) cc1 = $head1[var2]
    }
}
ARGIND==2 {
    if (FNR == 1) {
        for (i=0;i<=NF;i++) head2[$i] = i
    } else {
        if ($1 == var1) cc2 = $head2[var2]
    }
}
END {
    if (cc1 == cc2) {
        print "OK for", var1, var2
    } else {
        print "cc1=", cc1, "for", var1, var2
        print "cc2=", cc2, "for", var1, var2
    }
}
```
And the weighted mean results were checked with this script:
```
# check_mean.awk compares weights means of VAR in two files.
# USAGE: awk -f check_mean.awk -v var=VAR FILE1 FILE2
# EXAMPLE: awk -f check_mean.awk -v var=iitax                       \
#                                puf_var_wght_means_by_year.csv     \
#                                puf_var_wght_means_by_year.csv-old
BEGIN {
    FS = ","
}
ARGIND==1 && $1==var {
    m1 = $0  # using "m1 = $3" side steps errors in old results for 2014+
}
ARGIND==2 && $1==var {
    m2 = $0  # using "m1 = $3" side steps errors in old results for 2014+
}
END {
    if (m1 == m2) {
        print "OK for", var
    } else {
        print "m1=", m1
        print "m2=", m2
    }
}
```

Amy-Xu:
Thanks you @martinholmer for spotting the bug and refactoring the scripts. The improvement looks great to me. It seems the new tests check whether the stats and correlation results stay exactly the same. Does that mean if there is a difference, then the test will fail? My only concern is if we get a new PUF from a later year, the stats and correlation may vary slightly.

martinholmer:
@Amy-Xu asked about pull request #1574:
> It seems the new tests check whether the stats and correlation results stay exactly the same.  Does that mean if there is a difference, then the test will fail?  My only concern is if we get a new PUF from a later year, the stats and correlation may vary slightly.

The answer to your questions is: YES, the tests are likely to fail with any new `puf.csv` file.
But this is no different than many other tests that have been part of the py.test suite for years.
Don't we want the tests to fail, so that we know the weighted means and correlation coefficients need to be updated?  Otherwise, we'll be showing users incorrect, out-dated information.  We don't want that, do we?

@MattHJensen @andersonfrailey @hdoupe @GoFroggyRun

Amy-Xu:
@martinholmer So a second run will clear the error message?

martinholmer:
@Amy-Xu asked about PR #1574:
> So a second run will clear the error message?

Yes.  Just like a new `puf.csv` will likely trigger changes in the `taxcalc/tests/pufcsv_agg_expect.txt` and `pufcsv_mtr_expect.txt` results files, and after updating those files, the tests pass.

There is nothing different here than what we've been doing for years.

Amy-Xu:
Thanks for the explanation Martin.

PR #1576, opened 2017-09-29
Revise test_reforms.py so that reform info is in a JSON file

martinholmer:
This pull request refactors the `test_reforms.py` code so that information about all of the 55 tested reforms is in one JSON file named `reforms.json`.   These changes are based on the suggestion made by @hdoupe in an recent discussion of how to approach the reform tests.

These changes permit TaxBrain to specify an HTML link to the more readable `reforms.json` file, which contains no Python test code.  There is no change in test logic and there is no change in tax-calculating logic or test results.

This pull request will require a change in one HTML link in the "Transparency and Replicability" section of the "About TaxBrain" page.

@MattHJensen @Amy-Xu @hdoupe @brittainhard


MattHJensen:
This is definitely an improvement. I could see this leading to a system where we write the results into `reforms_actual` and copy them to `reforms_expect.json`

martinholmer:
@MattHJensen said about pull request #1576:
> This is definitely an improvement.  I could see this leading to a system where we write the results into reforms_actual and copy them to `reforms_expect.json`. 

I didn't see how to do that because the 55 tests are not necessarily run in order (because of the use of `py.test -n4` option).  This makes writing an actual file impossible.  Or, if not impossible, more work than it is worth.

MattHJensen:
@martinholmer said:

> I didn't see how to do that because the 55 tests are not necessarily run in order (because of the use of py.test -n4 option). This makes writing an actual file impossible. Or, if not impossible, more work than it is worth.

I see, thanks for the explanation, @martinholmer

hdoupe:
Nice work @martinholmer. This looks good

martinholmer:
Pull request #1576 has been open for seven days and discussion of it seems to be resolved, so it is being merged.

PR #1593, opened 2017-10-19
Include taxcalc/tbi/*.py files in code coverage calculations

martinholmer:
This pull request includes, for the first time, the files in the `taxcalc/tbi` directory in the code test coverage calculations.  In order to maintain complete code coverage, another test was added to the `taxcalc/tests/test_tbi.py` file and one statement in the `tbi_utils.py` file was omitted from the coverage calculations because reading from the taxcalc package "egg" is not testable.

This pull request contains no changes in tax-calculating code or in tax results.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -4,5 +4,6 @@ omit =
     taxcalc/functions.py
     taxcalc/*.json
     taxcalc/cli/*
+    taxcalc/tbi/*
     taxcalc/tests/*
     taxcalc/validation/*
--- a/.coveragerc
+++ b/.coveragerc
@@ -4,6 +4,5 @@ omit =
     taxcalc/functions.py
     taxcalc/*.json
     taxcalc/cli/*
-    taxcalc/tbi/*
     taxcalc/tests/*
     taxcalc/validation/*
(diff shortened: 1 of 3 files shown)


martinholmer:
When I run the code coverage test on my computer, I get complete test coverage.

When it runs on GitHub, it shows that there are 267 untested statements in the two `tbi*py` files. 

In particular, GitHub shows that the `calculate` function in the `tbi_utils.py` file is completely untested.  But that function is called indirectly several times in `test_tbi.py`.   It appears that on my computer those indirect calls to `calculate` "count" as test coverage, but on GitHub they don't.  If this is what is going on, how can this be fixed on GitHub?  If something else is going on, what is it?  I don't understand why the results on my computer would be so different from the GitHub coverage results.

@MattHJensen @talumbau

martinholmer:
My best guess is that code coverage was complete on my computer because the `puf.csv` file was available and all the tests in `test_tbi.py` that are labeled `@pytest.mark.requires_pufcsv` were executed in the code coverage calculations, while they were not on GitHub (because of the privacy of the `puf.csv` file).

Given this (apparent) state of affairs, I'm changing the `.coveragerc` file back to the way it has been for over a year.  In the future, code coverage in the `taxcalc/tbi` directory can be checked locally.

PR #1597, opened 2017-10-20
constrain range for child and dependent care credit maximum

MattHJensen:


martinholmer:
@MattHJensen proposed this change in pull request #1597:
```
 -        "range": {"min": 0, "max": 9e99},
 +        "range": {"min": 0, "max": 3000},
          "out_of_range_minmsg": "",
          "out_of_range_maxmsg": "",
          "out_of_range_action": "stop"     <---------------- ????
```
It's OK to use `3000` rather than `"default"` because this policy parameter is not inflation indexed.
But I was thinking that the `out_of_range_action` would be **warn** not **stop**.
We only warn when users who want to lower the value of `_STD`, so isn't this the same case?

MattHJensen:
> It's OK to use 3000 rather than "default" because this policy parameter is not inflation indexed.

I added a commit to switch to default. 

> We only warn when users who want to lower the value of _STD, so isn't this the same case?

I don't think it is quite the same. In the case of the standard deduction, a user could (and many do) eliminate all itemized deductions, after which they can lower the value of _STD safely.  In the case of the CDCC, the tax form specifically says not to enter a value higher than $3000 for child and dependent care expenses, and there are no provisions that could be combined with this reform to make the parameter work at higher values.

martinholmer:
@MattHJensen said in PR #1597:
>> It's OK to use 3000 rather than "default" because this policy parameter is not inflation indexed.
>
> I added a commit to switch to default.
>
>> We only warn when users who want to lower the value of _STD, so isn't this the same case?
>
> I don't think it is quite the same. In the case of the standard deduction, a user could (and many do) eliminate all itemized deductions, after which they can lower the value of _STD safely. In the case of the CDCC, the tax form specifically says not to enter a value higher than $3000 for child and dependent care expenses, and there are no provisions that could be combined with this reform to make the parameter work at higher values.

OK, this is a good point.  Going all the way down to zero standard deduction is sensible even though our data don't support a lower positive standard deduction.

But if the `out_of_range_action` is going to be `stop`, then you'd better go back to the `"max": 3000` range specification to avoid the unit test error.

MattHJensen:
> But if the out_of_range_action is going to be stop, then you'd better go back to the "max": 3000 range specification to avoid the unit test error.

Just reverted. 

> OK, this is a good point. Going all the way down to zero standard deduction is sensible even though our data don't support a lower positive standard deduction.

I'm not sure we're on the same page about this yet. I'm suggesting that if the user eliminates itemized deductions, then our data will support both a lower positive standard deduction and a zero standard deduction.

martinholmer:
@MattHJensen said in #1597:
> our data will support both a lower positive standard deduction and a zero standard deduction.

I guess I'm a little slow this morning.  I thought that standard deduction amounts below current-law levels were problematic because we don't have itemized expenses for those who use the standard deduction in the `puf.csv` input data.  If you cut the standard deduction, more will itemize.  But using the current version of `puf.csv` will underestimate deductions because everybody who would be a new itemizer under the reform has zeros for their itemized expense items.  Right?   I thought that was why we have an warning on lowering the `_STD` parameter below the "default" (that is, current-law) value.

MattHJensen:
>  I thought that standard deduction amounts below current-law levels were problematic because we don't have itemized expenses for those who use the standard deduction in the puf.csv input data. If you cut the standard deduction, more will itemize. But using the current version of puf.csv will underestimate deductions because everybody who would be a new itemizer under the reform has zeros for their itemized expense items. Right? I thought that was why we have an warning on lowering the _STD parameter below the "default" (that is, current-law) value.

That's all correct under current law itemization rules. But if a user repeals itemized deductions, then she can also safely reduce the standard deduction. The safety of reducing the standard deduction will depend on how completely itemized deductions are repealed.

martinholmer:
@MattHJensen said in #1597:
> That's all correct under current law itemization rules. But if a user repeals itemized deductions, then she can also safely reduce the standard deduction. The safety of reducing the standard deduction will depend on how completely itemized deductions are repealed.

Thanks for the detailed explanation.

Issue #1374, opened 2017-05-19
Expanded income and aftertax income should be options in distribution tables


MattHJensen:
![image](https://cloud.githubusercontent.com/assets/8301092/26267034/114e5d62-3cb6-11e7-8486-e4e11e6460f6.png)

Users may want to see average expanded income by expanded income bin or decile in TaxBrain diagnostic tables. Same with aftertax income.

martinholmer:
I've looked into Tax-Calculator issue #1374, which says:
> Users may want to see average expanded income by expanded income bin or decile in TaxBrain diagnostic tables. Same with aftertax income.

The construction in dropq code of a TaxBrain "diagnostic" table (which is called in Tax-Calculator a "distribution" table) is fairly involved.  The data structures that are created by the dropq code have been carefully designed to meet the needs of TaxBrain developers, which is exactly as it should be.

My conclusion is that the TaxBrain developers need to start out this enhancement process by specifying exactly how they want the thirteen objects returned from the `dropq.run_nth_year_tax_calc_model` function to be changed in order to implement the #1374 enhancement.  Then, with this change specification in hand, Tax-Calculator developers can revise taxcalc code to supply revised versions of those thirteen objects.

Does this approach make sense?

@MattHJensen @feenberg @Amy-Xu @GoFroggyRun @PeterDSteinberg @brittainhard

feenberg:
On Wed, 28 Jun 2017, Martin Holmer wrote:

> 
> I've looked into Tax-Calculator issue #1374, which says:
>
>       Users may want to see average expanded income by expanded income
>       bin or decile in TaxBrain diagnostic tables. Same with aftertax
>       income.
> 
> The construction in dropq code of a TaxBrain "diagnostic" table (which is
> called in Tax-Calculator a "distribution" table) is fairly involved. The
> data structures that are created by the dropq code have been carefully
> designed to meet the needs of TaxBrain developers, which is exactly as it
> should be.

The important thing is that we drop 3 meaningful records from each row 
returned to the user. So if we tab by expanded income we drop 3 records 
from each expanded income class, not each AGI class. Otherwise I think the 
logic should be exactly the same.

dan

martinholmer:
Dan @feenberg said with respect to issue #1374:
 > The important thing is that we drop 3 meaningful records from each row returned to the user. So if we tab[ulate] by expanded income we drop 3 records from each expanded income class, not each AGI class.

[...]

PR #1602, opened 2017-10-23
Add expanded_income and aftertax_income to distribution table

martinholmer:
This pull request adds two new columns to the distribution table as requested in issue #1374.
The pull request also removes several redundant tests in the `test_calculate.py` file.

There is no changes in tax-calculating logic or results.  There are only two more columns in each table produced by the `create_distribution_table` utility function.


MattHJensen:
Looks great. Thanks @martinholmer.

Issue #1603, opened 2017-10-23
Status of read-the-docs/notebooks/Behavioral_example.ipynb


martinholmer:
Since pull request #1582 was merged, the example in the `Behavioral_example.ipynb` notebook does not work correctly.

I'm not sure what the purpose of this notebook is because it is (as far as I can see) not mentioned anywhere in the read-the-docs documentation or anywhere else in the Tax-Calculator repository.

If we want to maintain this notebook, its logic needs to be fixed and it seems as if it should be referred to somewhere in the documentation.

Otherwise, the notebook should be deleted because if a user does stumble onto it that user will be shown an example that does not work.

I don't know how to use notebooks, so somebody who does needs to handle this issue if the decision is to fix and reference the notebook.   If the decision is to delete the notebook, then I'm happy to do that.

@MattHJensen @feenberg @Amy-Xu @andersonfrailey @hdoupe @GoFroggyRun @codykallen

andersonfrailey:
Given that the notebook is both out-dated and not referenced anywhere, I think it should be deleted. It also might be best to update the general tutorial notebook to include a behavioral reform example so that everything is in one place.

MattHJensen:
> Given that the notebook is both out-dated and not referenced anywhere, I think it should be deleted. It also might be best to update the general tutorial notebook to include a behavioral reform example so that everything is in one place.

Makes sense to me too.

MattHJensen:
I do wonder why our [notebook tests ](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/tests/test_notebooks.py) didn't catch this though.

codykallen:
> Given that the notebook is both out-dated and not referenced anywhere, I think it should be deleted. It also might be best to update the general tutorial notebook to include a behavioral reform example so that everything is in one place.

A behavioral response example is useful for those just learning how to use Tax-Calculator. If such an example is included in the general tutorial notebook, then I don't see any reason to keep the behavioral reform notebook.

martinholmer:
@MattHJensen asked in issue #1603:
> I do wonder why our notebook tests didn't catch this though.

Because the notebook tests are **very weak**.  They only detect a fatal Python error.
They have never done any validity checking on the answers the notebooks produce.

All I'm saying is that many of the results produced in the `Behavioral_example.ipynb` notebook are numerically incorrect after the merge of PR#1582.  The notebook runs without a Python crash, but produces many incorrect results.

[...]

PR #1604, opened 2017-10-23
Drop Behavioral_example.ipynb file and simplify test_notebooks.py logic

martinholmer:
This pull request implements part of what was planned in the discussion of issue #1603.

Adding to the one remaining notebook an example of using the Behavior class and strengthening the test in `test_notebooks.py` will be undertaken in another pull request.


MattHJensen:
I opened #1609 and #1610 for followups, so this PR now closes #1603.

Issue #1592, opened 2017-10-18
Cool charts to add to Tax-Calculator


MattHJensen:
Looking for chart ideas and contributors to build the charts into Tax-Calculator. These charts will be available to Tax-Calculator users and would be possibilities for adding to PolicyBrain. 

If you are interested in contributing a visualization to Tax-Calculator, look through the comments on this issue for ideas. 

If you have an idea for a cool chart or feedback (including itty bitty details) on other posted charts, please leave a comment on this issue. 

- [ ] DOCUMENTATION on how to add charts. Where to look in the code base for examples. What technology is recommended. What citation should be included by default. Etc. 
- [ ] [Distribution of tax units by tax bracket and filing status.](https://github.com/open-source-economics/Tax-Calculator/issues/1592#issuecomment-337728251) (NYT)
- [ ] [Percent change in after-tax income.](https://github.com/open-source-economics/Tax-Calculator/issues/1592#issuecomment-337728778) (TPC)
- [ ]
...

MattHJensen:
![image](https://user-images.githubusercontent.com/8301092/31742804-dc224a7a-b426-11e7-8fc6-4485bfcc0c3b.png)

MattHJensen:
![image](https://user-images.githubusercontent.com/8301092/31742904-367766f4-b427-11e7-844d-12503232ec78.png)

MattHJensen:
Change in after tax income by income group. Bonus points for allowing stacked provisions, like [this example](https://www.nytimes.com/interactive/2017/10/17/upshot/tax-reform-winners-losers-child-tax-credit.html) from Ernie Tedeschi in the NY Times using PolicyBrain output. (The NYT version is a cool looping gif) 

![image](https://user-images.githubusercontent.com/8301092/31746440-7243af36-b435-11e7-9882-e24794811bd6.png)

martinholmer:
@MattHJensen said in issue #1592:
> Looking for chart ideas and contributors to build the charts into Tax-Calculator. These charts will be available to Tax-Calculator users and would be possibilities for adding to PolicyBrain.

@MattHJensen then shows a recent example from the *New York Times*:

![screen shot 2017-10-19 at 12 38 48 pm](https://user-images.githubusercontent.com/12170745/31785710-10d1d18e-b4d4-11e7-8858-d2187875dde8.png)

[...]

PR #1605, opened 2017-10-24
Add top-decile detail and new statistic to difference table

martinholmer:
This pull request is built on pull request #1602, which added two new statistics (expanded income and after-tax expanded income) to the distribution table.

This pull request adds a new statistic (percentage change in after-tax expanded income) to the difference table and also adds top-decile details when the `create_difference_table` function is called with `groupby=weighted_deciles`.  These changes will be helpful in developing graphs like those discussed in issue #1592.

This pull request also converts the data type of the percentage columns in the difference table from str to float.


MattHJensen:
Looks great! Thanks @martinholmer.

PR #1615, opened 2017-11-02
Break up DependentCredit

codykallen:
This splits the non-refundable dependent credit into two parts, one for children and one for non-child dependents. The original version was intended to implement the Brady-Ryan proposal from 2016, which increased the  CTC cap by $500 (with the new portion non-refundable) and created a $500 non-refundable credit for non-child dependents. 

The new GOP tax plan appears to expand the CTC by $600 and create a $300 non-refundable credit for non-child dependents. If the CTC expansion is non-refundable then it should be implemented with the _DependentCredit_Child_c variable. Otherwise, it should be done by changing the CTC cap.

### DIFF
--- a/taxcalc/current_law_policy.json
+++ b/taxcalc/current_law_policy.json
@@ -3404,7 +3404,7 @@
 
     "_DependentCredit_c": {
         "long_name": "Nonrefundable credit for dependents",
-        "description": "This nonrefundable credit is applied to dependents and phases out with the Child Tax Credit. The dependent credit's phaseout begins when the CTC's phaseout ends.",
+        "description": "This nonrefundable credit is applied to (child and non-child) dependents and phases out with the Child Tax Credit. The dependent credit's phaseout begins when the CTC's phaseout ends.",
         "section_1": "Nonrefundable Credits",
         "section_2": "Child Tax Credit",
         "irs_ref": "",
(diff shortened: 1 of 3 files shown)


MattHJensen:
We should keep the existing unified `_DependentCredit_c` rather than replacing it for the sake of backwards compatibility.

codykallen:
@MattHJensen said

> We should keep the existing unified _DependentCredit_c rather than replacing it for the sake of backwards compatibility. Add the two new parameters.

Fair enough. Modifying now.

codykallen:
@hdoupe @andersonfrailey 

PR in progress now. 

This PR adds the capability to separately implement non-refundable expansions of the child tax credit and credits for non-child dependents.

andersonfrailey:
@codykallen this looks good. My one concern is that a user may be a little confused when the see a dependent credit, a dependent credit for children, and a dependent credit for non-children. Do you think it would be helpful to a small clarification in the description of `_DependentCredit_c` that it is a credit for all dependents. It could read something like this:

>This nonrefundable credit is applied to all dependents (child and non-child) and phases out with the Child Tax Credit. The dependent credit's phaseout begins when the CTC's phaseout ends.

codykallen:
@martinholmer @Amy-Xu 
Can we quickly review and merge this in? It's very relevant for those trying to model the plan coming out today.

MattHJensen:
This lgtm. Merging.

Issue #1619, opened 2017-11-02
Restructuring pass-through income


codykallen:
The current version of the treatment of separate pass-through income allows it to be taxed at a different rate schedule. It determines pass-through income as the sum of `e00900` and `e26270`, which is technically true under current law. However, not all of this income is necessarily eligible for reduced pass-through tax rates. For example, the new Tax Cuts and Jobs Act would make 30% of business net income eligible for the reduced rates and 70% remain as ordinary income. 

To address this properly, I think we need a new function to determine what amount of income qualifies for the separate pass-through tax rates. A basic (and reasonable) approach would be to add indicators for whether `e00900` (sole proprietorship income), `e26270` (partnership and S corporation income) and `e02000 - e26270` (other Sch E income) would count toward business income, and a separate parameter for what share of "business income" is eligible for the reduced tax rates. 

I think I can build this reasonably easily (with some annoyance to design tests), but I'd like to hear some feedback before doing this. 

@MattHJensen @andersonfrailey @martinholmer @Amy-Xu @feenberg @GoFroggyRun @hdoupe

MattHJensen:
This would allow for a dramatic simplification of the TaxBrain and Tax-Calculator user experiences if I understand it correctly. Users would only need to specify a separate pass thru rate schedule if they set this new percent-eligible parameter above 0. 

+1.

GoFroggyRun:
@MattHJensen said: 
>Users would only need to specify a separate pass thru rate schedule if they set this new percent-eligible parameter above 0.

It seems that, as shown in PR #1620, what @codykallen suggests in #1619 is a more precise composition of PT amount. If there were reforms involving PT brackets, users still have to go through those PT fields on TaxBrain page. 

@codykallen said:
>To address this properly, I think we need a new function to determine what amount of income qualifies for the separate pass-through tax rates. A basic (and reasonable) approach would be to add indicators for whether e00900 (sole proprietorship income), e26270 (partnership and S corporation income) and e02000 - e26270 (other Sch E income) would count toward business income, and a separate parameter for what share of "business income" is eligible for the reduced tax rates.

I agree that this approach would yield a much more precise PT amount under related reforms. + 1

codykallen:
This issue has been addressed by PR #1620. Closing.

PR #1614, opened 2017-11-02
Compatible data fields

MattHJensen:
Resolves #1480.

PR #1620, opened 2017-11-02
Restructure PT income determination

codykallen:
This PR improves the capabilities to determine how much pass-through income is eligible to be taxed at the PT rates instead of the II rates. This PR creates 4 new parameters, which determine whether Sch C income (`e00900`), partnership and S corporation income (`e26270`) and other Sch E income (`e02000 - e26270`) count as business income, with a separate parameter to determine what portion of business income is taxed at the PT rates. 

This PR implements the idea in issue #1619 that can be applied for the Tax Cuts and Jobs Act. Note that this is applicable for the 30/70 capital labor split, although not for their alternative "normal return" method.


MattHJensen:
Moving an email chain back to GH. 

@martinholmer said:

> Why do so many of the create-tables test results change?  Does that seem right to you?  If so, you need explain in #1620 why most of the table results are different. On the other hand, if you didn’t expect these table results to change, then all these changes may indicate a bug. 


@codykallen said:

> The results for that test change because under the current version, setting `_II_rt1` to 0.15 raises the tax on ordinary income but not on pass-through income, but in the new version, raising `_II_rt1` to 0.15 raises the tax rate on both ordinary and pass-through income. Thus the results should change.
> 
> However, when I first went to set the new results, I passed it the wrong results (using the PUF instead of the CPS subsample). I’m fixing that now and will send a new commit shortly.

@martinholmer said:

> I’m missing something. I thought the default values of the three new Boolean parameters were False, meaning that those types of income were not eligible for pass they rates. 
> What am I missing?
> 
> Plus your explanation implies we are making an abrupt change in taxcalc logic. How are all the past results going to be handle?  Are we going to label them incorrect?
> 

@codykallen said:

> Under the current set-up, changing the II_rt* parameters change the tax on ordinary income but not on pass-through income. Thus if one does not want to differentiate by ordinary and pass-through income, the current set-up requires changing the II_rt* and PT_rt* terms identically. Under the new set-up, an individual would have to actively decide to tax pass-through income separately for the PT_rt* terms to come into play.
> 
> This doesn’t seem like a substantial change in logic. This change would make taxing ordinary income and pass-through income identically the default status.

@martinholmer said:

> I appreciate all your general explanation, but I still don’t understand why the test results change. Can you spell out why they change in the test situation?

cc @andersonfrailey @hdoupe

MattHJensen:
> Under the current set-up, changing the II_rt* parameters change the tax on ordinary income but not on pass-through income. Thus if one does not want to differentiate by ordinary and pass-through income, the current set-up requires changing the II_rt* and PT_rt* terms identically. Under the new set-up, an individual would have to actively decide to tax pass-through income separately for the PT_rt* terms to come into play.
> 
> This doesn’t seem like a substantial change in logic. This change would make taxing ordinary income and pass-through income identically the default status.

@codykallen, my understanding is that this would invalidate existing results. While we may want to make this change in the future (and in fact I essentially lauded this outome as an upside of your approach), for the short term, I would suggest setting the defaults here so that existing results don't change. That way we don't need to deal with the break in backwards compatibility (even if it is a desired one) in the middle of everything else that's going on right now.

codykallen:
@martinholmer said

> I appreciate all your general explanation, but I still don’t understand why the test results change. Can you spell out why they change in the test situation?

Under the current tax system, if a policy proposed to raise the tax rate on income in the first bracket from 12% to 15%, we would implement this in the existing version of Tax-Calculator by changing `_II_rt1` and `_PT_rt1`. We have to do this because the tax rates and brackets in legal policy do not distinguish between ordinary and pass-through income, but Tax-Calculator does. Personally, I suspect most users are not aware of this.

codykallen:
@MattHJensen said:

> While we may want to make this change in the future (and in fact I essentially lauded this outome as an upside of your approach), for the short term, I would suggest setting the defaults here so that existing results don't change. 

Fine, that can be done. Perhaps we should open a separate issue for the default method of dealing with tax rates and brackets. 

@MattHJensen and @MattHJensen have said, roughly:

> this would invalidate existing results

I partially disagree. It would invalidate existing results only insofar as people would have applied the results incorrectly. I does change taxcalc logic, but I think the approach here is more reasonable for users.

MattHJensen:
@codykallen said:

> It would invalidate existing results only insofar as people would have applied the results incorrectly. 

That's not what I thought, so I want to follow up on this to make sure I understand what is going on. 

If this PR were merged as is, then wouldn't a pre-existing reform file that sets II_rt1 to 0% and PT_rt1 to 10% re-run in the new version of tax-calculator, end up taxing the pass through income at 0%?

codykallen:
@MattHJensen 

> If this PR were merged as is, then wouldn't a pre-existing reform file that sets II_rt1 to 0% and PT_rt1 to 10% re-run in the new version of tax-calculator, end up taxing the pass through income at 0%?

That's correct. The reason I set the default values I did is that these _are_ the current law policy values. Under current law, pass-through income is taxed on the same schedule as ordinary income. Schedule C , partnership and S corporation income are not eligible for preferential treatment.

MattHJensen:
> That's correct. The reason I set the default values I did is that these are the current law policy values. Under current law, pass-through income is taxed on the same schedule as ordinary income. Schedule C , partnership and S corporation income are not eligible for preferential treatment.

Ok. I agree with you on that point. It's just not how the PT_rt parameters were defined previously, so when we do adopt the 'correct' defaults for the new definition, it will invalidate reforms that worked previously and were specified correctly previously given how PT_rt was defined previously. We'll need to give more thought about how to deal with that break in backwards compatibility.

codykallen:
In response to concerns from @martinholmer and @MattHJensen regarding backwards compatibility, the latest commit restores the defaults in test_utils.py, and it changes the default values for `_PT_eligible_SchC`, `_PT_eligible_partner_scorp`, and `_PT_capitalshare` to maintain backwards compatibility. 

Note that I still disagree with the decision to maintain backwards compatibility in this manner. I am very confident that plenty of users have intended to change tax rates applied to all income but did not know that they would also need to change the PT rates. And I'm not happy that the default values in current_law_policy.json for 3 parameters are explicitly _not_ the default values under current law. But we can continue the discussion of backward compatibility and default treatment of pass-through taxation in a separate issue.

martinholmer:
@codykallen said 
> And I'm not happy that the default values in current_law_policy.json for 3 parameters are explicitly not the default values under current law. 

I don’t understand. You said in #1619 that the definition of passthru income on the master branch under current law is correct. So why did this PR originally change that definition?

MattHJensen:
@codykallen, my understanding is that the share of pass through income eligible for the the pass thru rates parameters doesn't allow for the 70/30 rule to be implemented. I suppose one option would be to make this PR [WIP] for now and then open a separate PR with only one parameter for the 70/30 rule. Then we could have this discussion on a more leisurely period without having to set the parameters to values that you don't think is right. I apologize for my encouragement in https://github.com/open-source-economics/Tax-Calculator/issues/1619#issuecomment-341574236 that didn't think through all of the issues we would encounter going this direction. 

cc @martinholmer

codykallen:
@martinholmer said:

>  You said in #1619 that the definition of passthru income on the master branch under current law is correct. So why did this PR originally change that definition?

On #1619, I said:

> The current version of the treatment of separate pass-through income allows it to be taxed at a different rate schedule. It determines pass-through income as the sum of e00900 and e26270, which is technically true under current law. 

Pass-through income is generally defined as income from a sole proprietorship, partnership or S corporation. Thus total pass-through income is `e00900 + e26270`. 

However, current law taxes pass-through income income on the same schedule as ordinary income. The existing structure in Tax-Calculator taxes pass-through income on a separate schedule than ordinary income, but the default values for the pass-through income rates and brackets are the same as the default values of the ordinary income rates and brackets. Thus as long as you don't change the `_II_*` parameters or the `_PT_*` parameters, the tax on pass-through income is effectively the reality you would get from filing your taxes under current law. 

The problems arise when people want to change income tax rates. I am fairly confident that nearly every new user who just wants to change the income tax brackets is unaware that they also need to change the pass-through tax brackets.

codykallen:
@MattHJensen said 

> my understanding is that the new share of pass through income eligible for the the pass thru rates parameters doesn't allow for the 70/30 rule to be implemented.

That doesn't seem right. It seems like it should work fine, at least on a static basis.

MattHJensen:
> That doesn't seem right. It seems like it should work fine, at least on a static basis.

My understanding is that if someone has $100 wages from their firm and $100 pass thru income, then the 70/30 rule says $60 will be taxed at pass through rates and $140 will be taxed at ordinary rates.

codykallen:
> My understanding is that if someone has $100 wages from their firm and $100 pass thru income, then the 70/30 rule says $60 will be taxed at pass through rates and $140 will be taxed at ordinary rates.

In that case, this PR does not model it correctly, but the existing method of taxing pass-through rates is also incorrect, more so than this PR. 

If they are coming up with more difficult rules, I'll close this PR, and it's on you to model them.

martinholmer:
Sent from my iPhone

> On Nov 3, 2017, at 8:23 AM, codykallen <notifications@github.com> wrote:
> 
> My understanding is that if someone has $100 wages from their firm and $100 pass thru income, then the 70/30 rule says $60 will be taxed at pass through rates and $140 will be taxed at ordinary rates.
> 
> In that case, this PR does not model it correctly, but the existing method of taxing pass-through rates is also incorrect, more so than this PR.
> 

Cody, Why do keep saying this when you’ve already said in #1619 that the master branch has the correct definition of passthru income under current law?


> If they are coming up with more difficult rules, I'll close this PR, and it's on you to model them.
>

MattHJensen:
>  but the existing method of taxing pass-through rates is also incorrect

Could you explain? I don't understand your point. The existing method of modeling pass-through rates is correct for what it says it does, which is, in my view, the only correct definition of correct.

codykallen:
@MattHJensen said

> The existing method of modeling pass-through rates is correct for what it says it does, which is, in my view, the only correct definition of correct.

Yes, the existing method does what it says it does. My point is that a new user would have little to no idea that reducing tax rates on ordinary income does not also change the rates on pass-through income. In the actual tax system, reducing ordinary income tax rates also reduces tax rates on pass-through income, because pass-through income is taxed as ordinary income. But Tax-Calculator taxes ordinary and pass-through income separately. 

That's not a problem for people who know Tax-Calculator, but I'll bet that nearly every journalist and congressional staffer who just wants to change the tax rates in general does not know to also change the pass-through rates.

MattHJensen:
I definitely agree that doing both is a pain, and I agree that many users probably err.

(Although we did make the situation better with clear labelling:)

image

I also still agree that we may want to merge this PR or something like it in the future. 

But the pressing need right now is to add the capability to do the 70/30 rule, and this PR doesn't do that, so I recommend opening a new PR with a single parameter for implementing the 70/30 rule and adding [WIP] to the title of this PR.

MattHJensen:
Here is a description of the 70/30 rule from Daniel Shaviro:

https://danshaviro.blogspot.com/2017/11/all-hail-npid.html

MattHJensen:
I said:
> so I recommend opening a new PR with a single parameter for implementing the 70/30 rule

After speaking more w @codykallen, I understand the desire to provide more flexibility than a single parameter as users are likely to want to experiment with alternatives to this provision of TCJA.

codykallen:
@MattHJensen pointed out that the TCJA's business income definitions are more complex than the method used in this PR. I believe the following is a reasonably accurate description of the bill's provisions in this regard:

> Net income from a passive business activity is entirely business income and fully eligible or the 25% maximum rate.

I think we could best approximate passive business income as `e02000 - e26270`, which is how we define it for the NIIT. Note that the tax definitions of active and passive business income would not change in the bill.

For active business income (including wages from that business), the filer could elect to apply a "capital percentage" of this income for the 25% maximum tax rate. The default capital percentage in general would be 30%. For active business income from a personal services business in the field of law, accounting, consulting, engineering, financial services or performing arts, the default capital percentage would be zero. However, the owner(s) with active business income could alternatively elect to calculate their eligible capital income as their capital investments in the business multiplied by the sum of the federal short-term rate and 7 percent. The use of this alternative determination would be binding for 5 years. 

There are additional rules on the recharacterization of wages and interest income (although other investment income would not be eligible for recharacterization). 

On considering this, I think the closest we could come to approximating this would be to determine the capital income as `(e02000 - e26270) + PT_capincshare * (e00900 + e26270)`. We should consider whether to apply different `PT_capincshare` variables for `e00900`, `e26270`, and `e02000 - e26270`.

The 2009 PUF documentation mentions 2 variables that I don't see in our PUF: `e26170` total passive income, and `e26190` total non-passive income. There are also passive and non-passive loss terms. These could potentially be more useful for modeling the TCJA pass-through provisions.

MattHJensen:
> On considering this, I think the closest we could come to approximating this would be to determine the capital income as (e02000 - e26270) + PT_capincshare * (e00900 + e26270). We should consider whether to apply different PT_capincshare variables for e00900, e26270, and e02000 - e26270.

I'd be interested to hear from @andersonfrailey or @martinholmer or @jdebacker  about this given their experience with these variables.

jdebacker:
I think @codykallen's approach makes sense.

Of course with the "guardrails" in the legislation, this all depends on the line of business the income comes from, which we have no way of knowing from the PUF.  Given that, perhaps just applying this formula to all filers makes sense (especially since it's ripe for noncompliance anyway so knowing the breakdown of capital/labor income by filer under current law might not makes much sense going forward).

codykallen:
I updated the policy parameters to allow a user to implement an alternative definition of what counts as income eligible for the pass through rate schedule (`_PT_rt*` and `_PT_brk*`). If either of the two new parameters `_PT_eligibleRate_active_TCJA` or `_PT_eligibleRate_passive_TCJA` is greater than zero, then `functions.py` determines eligibility for the pass-through schedule based on the TCJA rules. If both parameters are zero, then it uses the default definition of pass-through income as active income from a sole proprietorship `e00900` or from a partnership/S corporation `e26270`. 

To implement the basic rules in the TCJA, one would set `'_PT_eligibleRate_passive_TCJA': [1.0]` and `'_PT_eligibleRate_active_TCJA': [0.3]`. Note that the 0.3 value assumes that "personal services" firms will be able to remain eligible for this rate. If they cannot, then one could attempt to estimate an adjusted eligible share based on IRS SOI data by industry.

Note that this also ignores the alternative capital share definition based on capital investment in the firm, since we do not have the information of model this. 

By the way, allowing different definitions in this manner makes the `SchXYZTax` function a bit messy. It would be useful to consider a more complete restructuring of differential pass-through taxation at a future time.

MattHJensen:
@codykallen, thanks for the update. I will begin to review this. Also, I just removed the WIP label to indicate that this is ready for review by others. (Pls let me know if that was in error).

codykallen:
Another important note: This does not reflect the change in the stacking order. The current version of Tax-Calculator stacks pass-through income on top of ordinary income, but the TCJA proposes to stack ordinary income on top of business income. Adding this change will make the `SchXYZ` function more complicated, but I believe it can be implemented in a way that does not break backwards compatibility. If you think this is worthwhile, I'll implement this aspect of the reform as well.

codykallen:
The existing version of Tax-Calculator stacks `pt_taxinc` on top of `reg_taxinc`, but this commit stacks `reg_taxinc` on top of `pt_taxinc` (per the rules in the TCJA) if either of the two `PT_eligibleRate_*` parameters are non-zero.

MattHJensen:
> The existing version of Tax-Calculator stacks pt_taxinc on top of reg_taxinc, but this commit stacks reg_taxinc on top of pt_taxinc (per the rules in the TCJA) if either of the two PT_eligibleRate_* parameters are non-zero.

@codykallen Thanks for the update to account for the stacking order. Could you either document how the PT_eligibleRate_* params will have this secondary effect or, if possible, create a new parameter to swap change the stacking order?

MattHJensen:
I am currently looking for verification in the bill that 30% of wage income is eligible for the lower rate. I will report back when I have found a citation.

MattHJensen:
>I am currently looking for verification in the bill that 30% of wage income is eligible for the lower rate. I will report back when I have found a citation.

My reading of the Sec. 1004. description starting on page 3 of the [section by section summary](https://waysandmeansforms.house.gov/uploadedfiles/tax_cuts_and_jobs_act_section_by_section_hr1.pdf) corroborates that the wage income should be included.

codykallen:
@MattHJensen said

> I am currently looking for verification in the bill that 30% of wage income is eligible for the lower rate. I will report back when I have found a citation.

The bill counts active business income as net income from the business and wages paid from the business to its owner. The sum of these is multiplied by the applicable capital share to determine the amount of income eligible for the reduced rates. Unfortunately, we cannot distinguish between wages paid from the business to its owner from wages paid for outside activities, so I have assumed that those with business income receive wages from their participation in the business. 

@MattHJensen also said

> Could you either document how the PT_eligibleRate_* params will have this secondary effect or, if possible, create a new parameter to swap change the stacking order?

In this case, if the PT_eligibleRate_* parameters are not zero, then I apply the approach used in the bill (or as close as we can approximate). Since the bill applies a different stacking order than the default (or pt_taxinc stacked on top of reg_taxinc), I change the stacking order accordingly. But it may be appropriate to create a separate stacking order parameter. If you think it's worthwhile, I can add that tomorrow morning.

MattHJensen:
Using Tax-Calculator master, the top payroll tax rate to 35% would cost $16.76 Bin 2017. 

The same reform with the following additions costs $5.66 B. 

        '_PT_eligibleRate_passive_TCJA': [1],
        '_PT_eligibleRate_active_TCJA': [.3]

codykallen:
@MattHJensen, do you mean the "top income tax rate"? And did you just change the `_II_rt7`, or did you also change `_PT_rt7`?

MattHJensen:
@codykallen asked for a better description of the reforms:

Reform 1: - $16.76 B 

        '_PT_rt7': [0.35]

Reform 2: - $5.66 B

        '_PT_rt7': [0.35],
        '_PT_eligibleRate_passive_TCJA': [1],
        '_PT_eligibleRate_active_TCJA': [.3]

MattHJensen:
Reform 3: -41.7

        '_II_rt7': [0.35],
        '_PT_rt7': [0.35]

Reform 4: -41.7B

        '_II_rt7': [0.35],
        '_PT_rt7': [0.35],
        '_PT_eligibleRate_passive_TCJA': [1],
        '_PT_eligibleRate_active_TCJA': [.3]

MattHJensen:
Reforms 1-4 all produce estimates that are generally what I would expect.

codykallen:
@MattHJensen's first reform:

`'_PT_rt7': [0.35]`

This reduces the top tax rate on active business income (`e00900 + e26270`) to 35%. 

@MattHJensen's second reform:

`'_PT_rt7': [0.35],
'_PT_eligibleRate_passive_TCJA': [1],
'_PT_eligibleRate_active_TCJA': [.3]`

This reduces the top tax rate on passive business income and on 30% of active business income (including wages).

I would expect the second reform to lose less revenue than the first one.

taxcalc/current_law_policy.json
```diff
@@ -3005,6 +3005,49 @@
         "out_of_range_action": "stop"
     },
 
+    "_PT_eligibleRate_active_TCJA": {
+        "long_name": "Share of active business income (including wages) eligible for PT rate schedule",
+        "description": "Eligibility rate of active business income for separate pass-through rates, based on rules in the in the Tax Cuts and Jobs Act.",
+        "section_1": "Personal Income",
+        "section_2": "Pass-Through",
+        "irs_ref": "",
+        "notes": "Active business income defined as e00900 + e26270",
```
MattHJensen: Should this be e00900 + e26270 + e00200 ?

codykallen:
@MattHJensen asked

> Should this be e00900 + e26270 + e00200 ?

Not really. If the filer has `e00900 + e26270 != 0`, then his active business income (to which the eligible rate is applied) is `e00900 + e26270 + e00200`. But if `e00900 + e26270 == 0`, then the filer's active business income is zero.

MattHJensen:
@codykallen, said:

> But it may be appropriate to create a separate stacking order parameter. If you think it's worthwhile, I can add that tomorrow morning.

1. Yes, I think that would be worthwhile. 

2. Also, what do you think about adding a new parameter called something like PT_wages_in_active_inc that is default to false. Then you could drop TCJA from the names of PT_eligibleRate_passive_TCJA and PT_eligibleRate_active_TCJA and set both to 1 as default. 

--

As an alternative to doing those two things, I think you could add a new parameter above the two PT_eligibleRate_*_TCJA parameters, called something like _PT_TCJA_taxation, that would activate the PT_eligibleRate parameters and swap the stacking order. The PT_eligibleRate parameters could be set at 0.3 and 1 by default. Based on the way those two parameters are structured currently, I don't see a good way to set them at a default that is consistent with both current law and tax-calculator's current default handling of the pass thru rates, so requiring that they be switched on would be a concession to that. 

Other than this fairly tricky issue of parameter naming/construction, the PR looks good to me. 

cc @martinholmer @andersonfrailey

taxcalc/current_law_policy.json
MattHJensen: What about `e00900 + e26270 ( + e00200 if e00900 + e26270 > 0 )

codykallen:
@MattHJensen suggested

> As an alternative to doing those two things, I think you could add a new parameter above the two PT_eligibleRate_*_TCJA parameters, called something like _PT_TCJA_taxation, that would activate the PT_eligibleRate parameters and swap the stacking order. The PT_eligibleRate parameters could be set at 0.3 and 1 by default. Based on the way those two parameters are structured currently, I don't see a good way to set them at a default that is consistent with both current law and tax-calculator's current default handling of the pass thru rates, so requiring that they be switched on would be a concession to that.

This is actually closer to my preference for restructuring with more flexibility. Such an approach would use separate eligible rate parameters for sole proprietorship income `e00900`, partnership/S corporation income `e26270`, and passive business income `e02000 - e26270`. There could be a separate parameter for whether wage income is included with active business income when applying the eligibility rate, and another parameter for the stacking order (default of `pt_taxinc` on top of `reg_taxinc`. To maintain backwards compatibility, we would need to set the default values as 
`'_PT_eligibleRate_active': [1.0], '_PT_eligibleRate_passive': [0.0], '_PT_wages_active_income': [0.0], '_PT_stacking': [false]`

MattHJensen:
> This is actually closer to my preference for restructuring with more flexibility. Such an approach would use separate eligible rate parameters for sole proprietorship income e00900, partnership/S corporation income e26270, and passive business income e02000 - e26270. There could be a separate parameter for whether wage income is included with active business income when applying the eligibility rate, and another parameter for the stacking order (default of pt_taxinc on top of reg_taxinc. To maintain backwards compatibility, we would need to set the default values as
> '_PT_eligibleRate_active': [1.0], '_PT_eligibleRate_passive': [0.0], '_PT_wages_active_income': [0.0], '_PT_stacking': [false]

- Should _PT_wages_active_income be boolean? 
- Wouldn't '_PT_eligibleRate_passive': [1.0] allow for backwards compatibility under this setup? 
- Perhaps `_PT_top_stacking`?

codykallen:
> Should _PT_wages_active_income be boolean?

Yes, that should be a boolean. Thanks for catching.

> Wouldn't '_PT_eligibleRate_passive': [1.0] allow for backwards compatibility under this setup?

No. The current structure excludes passive business income (`e02000 - e26270`) from eligibility for the pass-through rates, so the backwards compatible default value for `_PT_eligibleRate_passive` is zero. 

> Perhaps `_PT_top_stacking`?

That would be fine by me. I'm not committed to any particular naming convention for this.

MattHJensen:
@codykallen, cool, that approach makes sense to me. Just to clarify, though, isn't this more similar to the steps 1&2 above the line break in https://github.com/open-source-economics/Tax-Calculator/pull/1620#issuecomment-342017801 rather than the piece of that comment that you copied in https://github.com/open-source-economics/Tax-Calculator/pull/1620#issuecomment-342018766?

codykallen:
@MattHJensen, yes, it's actually a bit of both.

andersonfrailey:
Overall, this PR looks good to me. I just have one question for @codykallen. In one of your previous comments, you said that to maintain backwards compatibility `_PT_top_stacking` would need to default to false, but in this PR you have it default to true. Shouldn't that be switched or did I miss something while catching up on the comments?

codykallen:
@martinholmer, I'm encountering a problem in test_utils.py. For some reason, there appears to be some kind of type error or rounding error. @andersonfrailey and @hdoupe have been looking over this with me, and we can't figure out the problem. By the way, all of the tests pass on my local branch and on @hdoupe's computer, but some fail on Travis and AppVeyor. Can you tell what's going wrong?

martinholmer:
GitHub test failures:

FACTS
- 18 hours ago #1614 passed all tests on GitHub when it used 2.7.13
-  now #1620 fails some tests on GitHub when it used Python 2.7.14
-  all tests pass on local computers that are running <=2.7.13

SOLUTION

-  upgrade local computer to newest version of Python (2.7 or 3.6)
-  fix errors that have appeared in a new PR
-  merge that PR into master branch 
-  merge new master into development branch(s)

Above is my best guess.  A new version of Python (or maybe other packages) is revealing code problems that earlier versions didn’t care about.

MattHJensen:
@andersonfrailey said:

> Overall, this PR looks good to me. I just have one question for @codykallen. In one of your previous comments, you said that to maintain backwards compatibility _PT_top_stacking would need to default to false, but in this PR you have it default to true. Shouldn't that be switched or did I miss something while catching up on the comments?

Thanks very much for the review @andersonfrailey. I believe PT_top stacking should be set to true under current law as it is in this PR.

MattHJensen:
@codykallen, if the tests pass, I plan to merge this in about 20 minutes.

hdoupe:
@codykallen you may want to update the `RELESES.md` file.  However, we may have to do a separate PR for this anyways since the `RELEASES.md` was not updated with some of the other PR's.

codykallen:
@hdoupe suggested

> you may want to update the RELESES.md file. However, we may have to do a separate PR for this anyways since the RELEASES.md was not updated with some of the other PR's.

Since I need this merged before pushing the new version of #1625, I plan to update RELEASES.md as part of that PR.

intownllc:
@codykallen see test failure

intownllc:
Never mind. This looks good. Merging.

codykallen:
@MattHJensen, can you tell where the code coverage error is coming from?

MattHJensen:
@codykallen I don't think the code coverage report is a problem right now. Merging. Thanks again for this contribution.

PR #1624, opened 2017-11-05
2027 Extrapolation

andersonfrailey:
This PR updates the weights, growth factors, and adjustment factors to extrapolate to 2027. We were originally going to wait to do this until we updated the PUF.csv file, but given the activity around tax reform and the unexpected road blocks to updating the PUF, a higher priority was placed on getting this out. There is no update to PUF.csv, just the associated weights and growth factors.

More details can be seen in TaxData PR [#122](https://github.com/open-source-economics/taxdata/pull/122). A note book with the changes can be seen [here](http://nbviewer.jupyter.org/github/andersonfrailey/Notebook-Uploads/blob/master/2027%20Extrapolation%20Notebook.ipynb).

There is one issue I was having with the test suite. In `test_policy.py`, there is a function `test_constant_inflation_rate_with_reform`. The test compares implied inflation rates with those in `policy.inflation_rates`. It seems to me that the error has to do with how the `round` function works. Here is the assertion error message that comes up when running py.test:

```
E       assert 0.023099 == 0.0231
E        +  where 0.023099 = round((1.023099324734464 - 1.0), 6)
E        +  and   0.0231 = round(0.0231, 6)
```

I found that 0.0231 is the result of `round(0.023099999999999999, 6)`. So the numbers are the same to the point where if both were rounded to 4 digits, this test would pass. I'd appreciate any insights into how to handle this.

cc. @MattHJensen @codykallen


MattHJensen:
> I found that 0.0231 is the result of round(0.023099999999999999, 6). So the numbers are the same to the point where if both were rounded to 4 digits, this test would pass. I'd appreciate any insights into how to handle this.

What do you think about rounding to fewer decimals. 

For example instead of 

`assert round(grate - 1.0, 6) == round(irate2019, 6)`

use

`assert round(grate - 1.0, 4) == round(irate2019, 4)`

andersonfrailey:
@MattHJensen said:
>What do you think about rounding to fewer decimals.

For example instead of

assert round(grate - 1.0, 6) == round(irate2019, 6)

use

assert round(grate - 1.0, 4) == round(irate2019, 4)

This solution was my first thought as well. Don't know if there's a specific reason why it wouldn't be satisfactory, if not I can make the change to the test in this PR.

MattHJensen:
> This solution was my first thought as well. Don't know if there's a specific reason why it wouldn't be satisfactory, if not I can make the change to the test in this PR.

I can't think of any problem.

MattHJensen:
@andersonfrailey, to what should we attribute the minor changes before 2027? Could you explain the intuition for why adding a year would change results in intervening years?

codykallen:
@MattHJensen asked

> to what should we attribute the minor changes before 2027? Could you explain the intuition for why adding a year would change results in intervening years?

Shouldn't this be caused by switching to the June 2017 CBO forecasts instead of whatever forecasts we were using?

andersonfrailey:
@codykallen is correct. The June 2017 CBO forecasts will be a different from those in the August 2016 report, which is what we are currently using. This will affect the calculated growth rates, targets used during the re-weighting process, and subsequently the final weights. All of these factors combined will result in slightly different calculations for all of the years.

MattHJensen:
> @codykallen is correct. The June 2017 CBO forecasts will be a different from those in the August 2016 report, which is what we are currently using. This will affect the calculated growth rates, targets used during the re-weighting process, and subsequently the final weights. All of these factors combined will result in slightly different calculations for all of the years.

Thanks @codykallen and @andersonfrailey. This PR LGTM as soon as the test error is fixed.

andersonfrailey:
@MattHJensen, latest commit should fix the test error.

MattHJensen:
Thanks @andersonfrailey! Merging.

martinholmer:
Never mind. Answered my own question.

PR #1629, opened 2017-11-06
Pandas update to 0.21.0

hdoupe:
Resolves #1628

PR #1630, opened 2017-11-06
Update RELEASES.md after PR 1629

hdoupe:
Sorry, I forgot to do this with PR #1629 .

@MattHJensen

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -19,7 +19,9 @@ for a complete commit history.
 - None
 
 **Bug Fixes**
-- None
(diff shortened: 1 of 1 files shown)


MattHJensen:
@hdoupe, were these really exposed bugs, or just updates to be compatible with the new pandas version?

hdoupe:
@MattHJensen You're right. It should be more like "Resolves compatibility issues with Pandas 0.21.0"

MattHJensen:
Thanks @hdoupe (and thanks for reminding us all to keep this updated!). Merging.

PR #1631, opened 2017-11-07
Update TCJA json for CTC rules

codykallen:
This PR updates the `_CTC_ps` parameter to include the new CTC phase-out thresholds (which I forgot to include in the previous version), and it updates the `_CTC_c` and `_DependentCredit_Child_c` parameters to include the inflation adjustment rules in the TCJA. These rules can be found in Section 1101 of the bill, beginning on page 68. A quick summary of the rules encoded here:

For any tax year beginning after January 1, 2017 (effective in 2018), the $1,000 refundable portion of the CTC is increased by the percent difference between the chained CPI in that year and the (non-chained) CPI in 2016, which is rounded up to the nearest $100, and that the sum of the refundable portion and the nonrefundable portion may not exceed $1600. 

Therefore, the `_CTC_c` parameter is $1100 in 2018, $1200 beginning in 2019, $1300 beginning in 2023, and $1400 beginning in 2027. The nonrefundable portion `_DependentCredit_Child_c` decreases according with each change in the CTC cap. 

@MattHJensen @andersonfrailey @hdoupe @martinholmer @evtedeschi3


codykallen:
Thanks to @evtedeschi3 for catching the inflation adjustment to the refundable portion of the CTC.

MattHJensen:
@codykallen, could you explain why ctc_c starts at 1100 instead of 1000?

codykallen:
@matt asked

> could you explain why ctc_c starts at 1100 instead of 1000?

Under current law, the child tax credit is refundable up to $1000 per child. The text of the bill (on pages 71-72), adds an inflation adjustment to this $1000 portion (although not to the entire $1600 credit). The increase in the credit from the inflation adjustment is rounded to the next highest multiple of $100, and the total of the refundable and nonrefundable portions may not exceed $1600. 

The inflation adjustment is 

> the percentage (if any) by which the C-CPI-U for the preceding calendar year exceeds the normalized CPI for calendar year 2016.

The normalized CPI in 2016 is defined as the C-CPI-U in 2016 divided by the CPI in 2016. According to the text, the inflation adjustment applied to the $1000 refundable portion of the CTC is effective for any taxable year beginning after January 1, 2017. Therefore, the inflation adjustment actually applies in 2018, so the refundable portion of the CTC is increases accordingly and rounded up to the nearest $100. Hence it has a value in 2018 of $1100.

evtedeschi3:
> Therefore, the _CTC_c parameter is $1100 in 2018, $1200 beginning in 2019, $1300 beginning in 2023, and $1400 beginning in 2027. The nonrefundable portion _DependentCredit_Child_c decreases according with each change in the CTC cap.

The rounding to $100 is clear in the legislative text, but is it right that the next year's benefit is adjusted off of the prior year's _rounded_ benefit? That very quickly compounds. If we assume that chained CPI runs 25bp slower than CBO's headline CPI-U projections, that suggests that $1,000 would growth to ~$1,209 by 2027 if it were continuous / unrounded. Under these assumptions, the benefit grows to $1,400 or _almost double the effective growth rate_. So essentially, the rounding approach here would completely obviate the benefit of using a slower inflation adjustment in the first place.

This could be the right read (Lord knows Congress has occasionally written language without thinking through the implications!) but wanted to raise the issue so that everyone is absolutely sure.

evtedeschi3:
Nevermind, I see what's going on. It's a combination of starting with $1,100 rather than $1,000 which Cody justified, and rounding up to $1,400 in 2027 (rather than the $1,330 continuous result).

codykallen:
Here is the text of the inflation determination for the refundable portion of the CTC (page 71-72):

> (2) INFLATION ADJUSTMENT.—Section 24(d) is amended by inserting after paragraph (2) the following new paragraph: (3) INFLATION ADJUSTMENT.—In the case of any taxable year beginning in a calendar year after 2017, the $1,000 amount in paragraph (1)(A)(iii) shall be increased by an amount equal to—(A) such dollar amount, multiplied by (B) the cost-of-living adjustment under section 1(c)(2)(A) for such calendar year. Any increase determined under the preceding sentence shall be rounded to the next highest multiple of $100 and shall not exceed the amount in effect under subsection (a)(2).’’

The “amount under subsection (a)(2)” is $1600. 
The “cost-of-living adjustment under section 1(c)(2)(A)” is:
> IN GENERAL.—The cost-of-living adjustment for any calendar year is the percentage (if any) by which—(i) the C-CPI-U for the preceding calendar year, exceeds (ii) the normalized CPI for calendar year 2016.

MattHJensen:
Thanks @codykallen for the contribution and thanks @evtedeschi3 for the review. Merging.

Issue #1628, opened 2017-11-06
Master branch fails tests using Python 3.6.3


hdoupe:
In response to the test failures in issue #1620, I ran the tests using Python 3.6.3 and received the same errors as the [travis run](https://travis-ci.org/open-source-economics/Tax-Calculator/jobs/298126321).  Further, I ran the tests on the master branch and received the same errors.  

```
================================================================================================================================== FAILURES ==================================================================================================================================
_____________________________________________________________________________________________________________________________ test_create_tables _____________________________________________________________________________________________________________________________

cps_subsample =         age_head  age_spouse      e00200p      e00900p      e02100p  \
445636        41           0   922,485.99      ...2  
52701       0    0        5  
66990       0    0       16  
428804      0    0        9  

[4565 rows x 54 columns]

    def test_create_tables(cps_subsample):
        # pylint: disable=too-many-statements
        # create a current-law Policy object and Calculator object calc1
        rec = Records.cps_constructor(data=cps_subsample)
        pol = Policy()
        calc1 = Calculator(policy=pol, records=rec)
        calc1.calc_all()
        # create a policy-reform Policy object and Calculator object calc2
        reform = {2013: {'_II_rt1': [0.15]}}
        pol.implement_reform(reform)
        calc2 = Calculator(policy=pol, records=rec)
        calc2.calc_all()
    
        # test creating various difference tables
    
        diff = create_difference_table(calc1.records, calc2.records,
                                       groupby='large_income_bins',
                                       income_measure='expanded_income',
                                       tax_to_diff='combined')
        assert isinstance(diff, pd.DataFrame)
        expected = [0.00,
                    0.01,
                    0.41,
                    0.84,
                    0.92,
                    1.10,
                    1.15,
                    1.04,
                    0.78,
                    0.27,
                    np.nan]
        assert np.allclose(diff['perc_aftertax'], expected,
                           atol=0.005, rtol=0.0, equal_nan=True)
    
        diff = create_difference_table(calc1.records, calc2.records,
                                       groupby='webapp_income_bins',
                                       income_measure='expanded_income',
                                       tax_to_diff='iitax')
        assert isinstance(diff, pd.DataFrame)
        expected = [0.00,
                    0.01,
                    0.41,
                    0.84,
                    0.92,
                    1.10,
                    1.15,
                    1.04,
                    0.78,
                    0.30,
                    0.08,
                    0.07,
                    np.nan]
[...]

PR #1634, opened 2017-11-07
pandas 0.21.0 compatibility part 2

hdoupe:
This implements a cleaner solution to pandas 0.21.0 compatibility issue #1628 that was initially addressed by PR #1629. The approach here was advocated in this [comment](https://github.com/pandas-dev/pandas/issues/9422#issuecomment-342515489).  See pandas-dev/pandas#9422 for context.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -24,6 +24,9 @@ for a complete commit history.
 - Resolve compatibility issues with Pandas 0.21.0
   [[#1629](https://github.com/open-source-economics/Tax-Calculator/pull/1629)
   by Hank Doupe]
+- Cleaner solution to compatibility issues with Pandas 0.21.0
+  [[#1634](https://github.com/open-source-economics/Tax-Calculator/pull/1634)
+  by Hank Doupe]
 
 
 2017-10-20 Release 0.12.0
--- a/taxcalc/__init__.py
+++ b/taxcalc/__init__.py
@@ -11,7 +11,11 @@ from taxcalc.utils import *
 from taxcalc.macro_elasticity import *
 from taxcalc.tbi import *
 from taxcalc.cli import *
+import pandas as pd
 
 from taxcalc._version import get_versions
 __version__ = get_versions()['version']
 del get_versions
+
+# zsum is defined in utils.py
+pd.Series.zsum = zsum
--- a/taxcalc/utils.py
+++ b/taxcalc/utils.py
@@ -126,6 +126,16 @@ SMALL_INCOME_BINS = [-9e99, 0, 4999, 9999, 14999, 19999, 24999, 29999, 39999,
                      1999999, 4999999, 9999999, 9e99]
 
 
+def zsum(self):
+    """
+    pandas 0.21.0 changes sum() behavior so that the result of applying sum
+    over an empty DataFrame is NaN. Since we apply the sum function over
+    grouped DataFrames that may or not be empty, it makes more sense for us
+    to have a sum() function that returns 0 instead of NaN.
+    """
+    return self.sum() if len(self) > 0 else 0
+
+
 def unweighted_sum(pdf, col_name):
     """
     Return unweighted sum of Pandas DataFrame col_name items.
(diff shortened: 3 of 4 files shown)


MattHJensen:
This looks good to me. Thanks @hdoupe. Merging.

martinholmer:
@hdoupe,  I like the approach in #1634.  Thanks for the very good work.

hdoupe:
Thanks @martinholmer

martinholmer:
@hdoupe,
I've been looking more closely at the use in #1634 of the new `numpy.Series.zsum()` function.
You replaced `sum()` with `zsum()` everywhere in the `utilsprvt.py` file, which seems sensible.
But you didn't make that replacement in the `utils.py` file or in any other files where `sum()` is used.
Shouldn't we replace all the `sum()` functions with `zsum()` functions?

Please advise.  If we decide to do a complete replacement, I'll be happy to prepare the pull request.

hdoupe:
@martinholmer I thought about this some, too.  I decided not to do it because it seemed like a pretty invasive change.  However, there are some places where this should be used such as `utils.weighted`.  

The only downside that I can think of is that this might cover up a situation where an error should be thrown or the result really should be `NaN`.  I'm having trouble thinking of a case where this would be a problem.  The `zsum` function only changes the `sum` behavior when the dataframe is empty.

I'm not advocating this approach, but I was curious to see what would happen if we changed all of the sum operations to `zsum`.  So, I implemented a hacky solution on [this branch.](https://github.com/hdoupe/Tax-Calculator/tree/sum_experiment)  It still passes all of the tests.

In general, we should think really hard before we over-write changes made by a package like pandas.  So, my view is that this `zsum` function should be used sparingly.

martinholmer:
@hdoupe said:
>  I thought about this some, too. I decided not to do it because it seemed like a pretty invasive change. However, there are some places where this should be used such as utils.weighted.
>
> The only downside that I can think of is that this might cover up a situation where an error should be thrown or the result really should be NaN. I'm having trouble thinking of a case where this would be a problem. The zsum function only changes the sum behavior when the dataframe is empty.
>
> I'm not advocating this approach, but I was curious to see what would happen if we changed all of the sum operations to zsum. So, I implemented a hacky solution on this branch. It still passes all of the tests.

I agree we do not want to redefine the pandas Series method sum() as you did in your branch.

> In general, we should think really hard before we over-write changes made by a package like pandas. So, my view is that this zsum function should be used sparingly.

Will we just wait until something goes wrong before making any further use of zsum()?
Will we always know when something goes wrong?
Despite these questions, I'm comfortable going with the use-as-needed approach.

Except maybe you think zsum() should be used in weighted() in utils.py, is that right?
Should I make that change?  It's sort of weird that zsum() is defined in utils.py, but never used there.

**A question about the implementation of the zsum() function**:
pylint does not like the construct `len(self) > 0` saying it violates PEP8.
So, I've tentatively changed the code in #1637 to be this:
```
def zsum(self):
    """
    pandas 0.21.0 changes sum() behavior so that the result of applying sum
    over an empty Series is NaN.  Since we apply the sum() function over
    grouped DataFrames that may contain an empty Series, it makes more sense
    for us to have a sum() function that returns zero instead of NaN.
    """
    return self.sum() if self.size > 0 else 0
```
Does this make sense to you?

hdoupe:
@martinholmer said
> Will we just wait until something goes wrong before making any further use of zsum()?
> Will we always know when something goes wrong?

That's the downside of the as-needed approach.  It's difficult to know what situations will call for this.  I think any function that is heavily used after pandas `cut` or group-by type functions should probably use `zsum`.  This type of operation occurs frequently in `utils.py`.  So, I'm sure there are more places in `utils.py` where `zsum` should be used.

> Except maybe you think zsum() should be used in weighted() in utils.py, is that right?
> Should I make that change?

Sounds good. Go for it.

> It's sort of weird that zsum() is defined in utils.py, but never used there.

Right, `utilsprvt.py` isn't one of the modules that's imported in `__init__.py`.  I figured that I could put `zsum` in `utils.py` and not have to add another import to `__init__.py`.  However, I think `zsum` should only be used internally.  Thus, it belongs more in `utilsprvt.py`.  I guess we could add `from utilsprvt import zsum` to `__init__.py`.

Using `self.size` makes sense to me as long as `bool(len(self))` is equivalent to `bool(self.size)`

Issue #1636, opened 2017-11-09
Pass-through income MTR outliers


rickecon:
The Tax Cuts and Jobs Act has some changes to pass through income that cause some huge outlier negative marginal tax rates in self employment income (`e00900p`). These influence the marginal tax rates of self employment income. When the pass through variables in the [`TCJA_Nov3.json`](?) are passed into Tax-Calculator `[mtr_fica_sey, mtr_iit_sey, mtr_combined_sey] = calc1.mtr('e00900p')`. When I describe the variable `mtr_combined_sey`, which I will call `MTR_sey` ("MTR self employment income"), I get the following.
```python
[In] MTR_sey[MTR_sey >= -0.45].describe()
[Out]
count         212111
mean          0.278268
std           0.140092
min          -0.416316
25%           0.250000
50%           0.252818
75%           0.373634
max           1.054188

[In] MTR_sey[MTR_sey < -0.45].describe()
[Out]
count    3414
mean    -5.900011e+06
std      1.152289e+07
min     -2.274409e+08
25%     -6.169098e+06
50%     -2.822819e+06
75%     -9.383364e+05
max     -4.500000e-01
```
The average MTR for all the values with MTR < -0.45 is -5.9e+06 or -59,000 percent. I'm sure these are coming from the marginal taxpayers at some big ledges, but we might want to do some smoothing here. @martinholmer @MattHJensen @jdebacker

martinholmer:
@rickecon [reported](https://github.com/open-source-economics/Tax-Calculator/issues/1636#issue-272491029) 3414 observations in `puf.csv` that have a MTR wrt e00900 that is less than -45% under the `TCJA_Nov3.json` reform.  Then he said:
> The average MTR for all the values with MTR < -0.45 is -5.9e+06 or -59,000 percent. I'm sure these are coming from the marginal taxpayers at some big ledges, but we might want to do some smoothing here. 

It seems to me that we need answers to a few questions (assuming you're using taxcalc 0.13.0):
- What does this MTR distribution look like under current-law policy?
- Can you provide us with the RECID for one or two cases that have an extreme negative MTR?
- Can you, or anybody else, explain what "big ledges" are in the `TCJA_Nov3.json` reform?

@MattHJensen @jdebacker @codykallen

martinholmer:
[...]

PR #1645, opened 2017-11-10
Fix PT mtr issue

codykallen:
This PR resolves the problem identified in issue #1636. In that issue, @rickecon discovered that the TCJA caused some individuals to have absurdly negative MTRs on `e00900p`. Upon investigtion, I realized that this is due to the `_PT_wages_active_income` parameter, which allows filers with active pass-through income (`e00900` or `e26270`) to include their wages as part of the base to which `_PT_EligibleRate_active` is applied. When we calculate the MTR on `e00900p`, we cause filing units that did not have active pass-through income to have active pass-through income, which made 30% (value of `_PT_EligibleRate_active`) of their wages eligible for the lower PT rates and substantially reduced their tax liability. Note that this phenomenon should also occur when calculating the MTRs on `e26270`, 

This PR eliminates the problem by setting the amount of pass-through income eligible for the PT rates to not exceed `e00900 + e26270`. This resolves the issue. 

I also added a test in test_pufcsv.py with assertion statements that the lowest MTRs on `e00900p` and `e26270` may not be less than -1. I tested this for 4 scenarios: 
- Current law (which should obviously pass)
- Set `'_PT_rt7': [0.35]`. The passage of this shows that the problem is not due to changing the PT rates.
- Set `'_PT_wages_active_income': [True]`. The passage of this shows that this parameter does not produce a problem when PT income and regular income face the same tax schedules.
- Set `'_PT_rt7': [0.35], '_PT_wages_active_income': [True]`. This fails on the master branch, but it passes with my fix. 

@rickecon @martinholmer @MattHJensen @feenberg @jdebacker @andersonfrailey @Amy-Xu

### DIFF
--- a/taxcalc/functions.py
+++ b/taxcalc/functions.py
@@ -719,6 +719,7 @@ def SchXYZ(taxable_income, MARS, e00900, e26270, e02000, e00200,
     if (pt_active_gross > 0) and PT_wages_active_income:
         pt_active_gross = pt_active_gross + e00200
     pt_active = PT_EligibleRate_active * pt_active_gross
+    pt_active = min(pt_active, e00900 + e26270)
     pt_taxinc = max(0., pt_passive + pt_active)
     if pt_taxinc >= taxable_income:
         pt_taxinc = taxable_income
--- a/taxcalc/tests/test_pufcsv.py
+++ b/taxcalc/tests/test_pufcsv.py
@@ -235,6 +235,59 @@ def test_mtr(tests_path, puf_path):
 
 
 @pytest.mark.requires_pufcsv
+def test_mtr_pt_active(puf_subsample):
+    """
+    Test whether including wages in active income causes
+    MTRs on e00900p and e26270 to be less than -1
+    """
+    rec = Records(data=puf_subsample)
+    reform_year = 2018
+    # create current-law Calculator object, calc1
+    pol1 = Policy()
+    calc1 = Calculator(policy=pol1, records=rec)
+    calc1.advance_to_year(reform_year)
+    calc1.calc_all()
+    mtr1_e00900p = calc1.mtr('e00900p')[2]
+    mtr1_e26270 = calc1.mtr('e26270')[2]
+    assert min(mtr1_e00900p) > -1
+    assert min(mtr1_e26270) > -1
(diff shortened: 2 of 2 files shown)


codykallen:
By the way, under all 4 of the scenarios described above, the lowest MTR on e00900 is always the same (-0.6). This is the MTR for individuals who have negative tax liability due to the EITC and the ACTC, since the EITC has a phase-in rate of 45% and the ACTC has a phase-in rate of 15%, resulting in the -60% MTR.

martinholmer:
@codykallen,  Your pull request #1645 is very nicely done.
Thanks for the fast and high-quality work on issue #1636.

And thanks to @rickecon who documented the problem in #1636.

@codykallen,  I'm curious about whether or not the legislative language actually contains the logic you added to the code.  Or is your code fix a loophole that the drafters of the bill haven't yet discovered.  Do you know which it is?

@MattHJensen, Do you want to review before merging this?

codykallen:
@martinholmer asked

> I'm curious about whether or not the legislative language actually contains the logic you added to the code. Or is your code fix a loophole that the drafters of the bill haven't yet discovered. Do you know which it is?

Actually, this PR implements the legislative language more precisely (from a legal perspective) than my original implementation. The (House) bill includes a rule that the applicable capital percentage for active business income (`_PT_EligibleRate_active`, applied to active business income plus wages) would be limited if the resulting income eligible for the preferential PT rates exceeds the actual active business income (excluding wages). 

When considering how to implement the bill's pass-through income rules, I originally decided to ignore this provision because it is toothless. Business owners would simply retain their wages as business income. Therefore, I think the original implementation is closer to what would occur in practice. 

There is a difficult trade-off between legal accuracy and realistic accuracy. In a purely static model with no income shifting, this PR implements the House bill's TCJA more precisely than the existing approach, but it is likely less accurate in practice.

MattHJensen:
This looks good to me. Beautiful testing. Thanks very much @codykallen! @martinholmer, could you merge if you are satisfied with @codykallen's answer above?

martinholmer:
@codykallen,  Thanks for the detailed answer to my question.

hdoupe:
@codykallen +1 on the testing

Issue #1641, opened 2017-11-10
_ID_RealEstate_c is not documented


MaxGhenis:
`_ID_RealEstate_c` is in the TCJA reform file (https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/reforms/TCJA_Nov3.json) but not the parameters documentation  (http://open-source-economics.github.io/Tax-Calculator/index.html#pol).

MattHJensen:
I will make sure this gets into the docs. 

In the meantime, you can read about it in the [current_law_policy.json](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/current_law_policy.json#L1158) parameter file from which the documentation is created.

MattHJensen:
Looks like @martinholmer beat me to it!

martinholmer:
@MaxGhenis,  Thanks for your documentation bug report in Tax-Calculator issue #1641.

That bug has now been fixed, so the `_ID_RealEstate_*` policy parameters are now documented in the [itemized deductions policy parameters section of the user documentation](http://open-source-economics.github.io/Tax-Calculator/#pol-itemized-deductions).

PR #1646, opened 2017-11-10
Change index.htmx ID subheading to be same as in clp.json

martinholmer:
Responds to documentation bug reported in issue #1641.
This pull request fixes the [user documentation]() only.
There is no change in tax logic or results.


MattHJensen:
Thanks @martinholmer!

Issue #1650, opened 2017-11-10
Pass-through income from service activities


codykallen:
Both the House and Senate versions of the TCJA have rules determining what types of pass-through income are eligible for preferential tax treatment, with separate rules for personal services businesses. 

The House bill defines personal services businesses as 

> businesses involving the performance of services in the fields of law, accounting, consulting, engineering, financial services, or performing arts.

Under the rules in the House bill, the default capital percentage for these firms would be zero (instead of 30% for other types of pass-through businesses), but they would be eligible to apply an alternative capital percentage as the business's capital investments multiplied by the federal short-term rate plus 7 percent. 

As @jdebacker noted in PR #1620, 

> Perhaps just applying this formula [i.e. 30%] to all filers makes sense (especially since it's ripe for noncompliance anyway so knowing the breakdown of capital/labor income by filer under current law might not makes much sense going forward).

The Senate bill defines personal services businesses as 

>  any trade or business activity involving the performance of services in the fields of health, law, engineering, architecture, accounting, actuarial science, performing arts, consulting, athletics, financial services, brokerage services, or any trade or business where the principal asset of such trade or business is the reputation or skill of one or more of its employees. 

In the Senate bill, a taxpayer may deduct 17.4% of qualified business income from a sole proprietorship, partnership or S corporation. For personal services businesses, the benefit of this deduction is phased out for taxable income in excess of $75,000 ($150k for married joint), with a phase-out length of $25,000 ($50k for married joint). The chairman's mark does not clearly define the "benefit of this deduction"; without the actual bill text or more detail, I'm not certain we can model the phase-out. 

However, the inability to model the phase-out may be irrelevant, since we cannot distinguish personal services business income from other business income. And as @jdebacker noted, this is ripe for noncompliance anyway. 

Given this, I think it would be best to simply model this as an exclusion of 17.4 percent of qualified business income. 

@MattHJensen @martinholmer @jdebacker

MattHJensen:
> Given this, I think it would be best to simply model this as an exclusion of 17.4 percent of qualified business income.

That makes sense to me as well.

codykallen:
Modeling for this has been addressed in PR #1651. Closing.

PR #1651, opened 2017-11-10
Add business income exclusion in Senate TCJA

codykallen:
This PR models two parts of the rules related to the exclusion of 17.4 percent of qualified business income, from the Senate's [Tax Cuts and Jobs Act](https://www.finance.senate.gov/imo/media/doc/11.9.17%20Chairman's%20Mark.pdf), pages 17-18. The proposal would allow 17.4% of qualified business income to be excluded from taxable income. The exclusion would phase out for personal services business income for taxable income over $75000 ($150k for married filing jointly). Since we cannot distinguish personal services business income from other business income, I have not modeled the phase-out. I also discuss this aspect in #1650. 

The exclusion rate for qualified business income is modeled by `_PT_exclusion_rt`. 
I also model an unusual rule from the bill:

> In the case of a taxpayer who has qualified business income from a partnership or S corporation, the amount of the deduction is limited to 50 percent of the W-2 wages of the taxpayer. 

This is modeled by `_PT_exclusion_limit`. 

@MattHJensen @martinholmer @jdebacker @feenberg @andersonfrailey @hdoupe @Amy-Xu @GoFroggyRun

### DIFF
--- a/taxcalc/current_law_policy.json
+++ b/taxcalc/current_law_policy.json
@@ -3093,6 +3093,50 @@
         "out_of_range_action": "stop"
     },
 
+    "_PT_exclusion_rt": {
+        "long_name": "Pass-through income exclusion rate",
+        "description": "Fraction of qualified business income excluded from taxable income.",
+        "section_1": "Personal Income",
+        "section_2": "Pass-Through",
+        "irs_ref": "",
+        "notes": "Applies to e00900 + e26270",
+        "row_var": "FLPDYR",
+        "row_label": ["2013"],
+        "start_year": 2013,
+        "cpi_inflated": false,
+        "col_var": "",
+        "col_label": "",
+        "boolean_value": false,
+        "integer_value": false,
+        "value": [0.0],
+        "range": {"min": 0.0, "max": 1.0},
+        "out_of_range_minmsg": "",
+        "out_of_range_maxmsg": "",
+        "out_of_range_action": "stop"
+    },
+
+    "_PT_exclusion_limit": {
+        "long_name": "Limit on pass-through income exclusion",
+        "description": "If taxpayer has partnership/ S corporation income, the amount of business income excluded from taxable income may not exceed this fraction of wages.",
+        "section_1": "Personal Income",
+        "section_2": "Pass-Through",
+        "irs_ref": "",
+        "notes": "",
+        "row_var": "FLPDYR",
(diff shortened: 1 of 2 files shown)


taxcalc/current_law_policy.json
```diff
@@ -3093,6 +3093,50 @@
...
+        "boolean_value": false,
+        "integer_value": false,
+        "value": [0.0],
+        "range": {"min": 0.0, "max": 1.0},
+        "out_of_range_minmsg": "",
+        "out_of_range_maxmsg": "",
+        "out_of_range_action": "stop"
+    },
+
+    "_PT_exclusion_limit": {
```
MattHJensen: Could this be _PT_exclusion_wage_limit ?

MattHJensen:
@codykallen, should this still have the WIP label (i.e., should it wait for review? )

MattHJensen:
This looks good to me other than the one potential parameter naming issue.

taxcalc/current_law_policy.json
codykallen: That would be fine.

taxcalc/current_law_policy.json
```diff
@@ -3093,6 +3093,50 @@
...
+        "row_var": "FLPDYR",
+        "row_label": ["2013"],
+        "start_year": 2013,
+        "cpi_inflated": false,
+        "col_var": "",
+        "col_label": "",
+        "boolean_value": false,
+        "integer_value": false,
+        "value": [9e99],
+        "range": {"min": 0.0, "max": 9e99},
```
andersonfrailey: I'm a little confused about this parameter. The description says it's a fraction so shouldn't the value range be 0.0-1.0?

MattHJensen: I think it's ok for this to be greater than 1. For example, you could limit PT income to 200% of wages.

codykallen: When applied in the TCJA, it should be applied as a fraction. However, in the default state, it shouldn't apply at all. Thus the default value of 9e99, which makes the restriction inoperative unless deliberately changed.

andersonfrailey: Ah gotcha. Thanks for the clarification.

MattHJensen:
Looks good. Thanks for the contribution @codykallen and review @andersonfrailey.

PR #1653, opened 2017-11-10
Add FIPS State Codes to CPS

andersonfrailey:
This PR complements PR [#126](https://github.com/open-source-economics/taxdata/pull/126) in the TaxData repo. It adds the FIPS state code associated with each tax unit in the CPS file.

cc @MattHJensen


MattHJensen:
LGTM and merging once checks complete. Thanks @andersonfrailey.

Needs caveated documentation in `index.html` @andersonfrailey.

andersonfrailey:
Fixing those tests.

andersonfrailey:
Seems like the test failure is due to rounding error. I looked and each error was .1, which is why it failed on 2.7 but not 3.6. I updated the CPS expected text file so that it would pass both.

MattHJensen:
> Seems like the test failure is due to rounding error. I looked and each error was .1, which is why it failed on 2.7 but not 3.6. I updated the CPS expected text file so that it would pass both.

Do you know why the results changed at all? I would have thought there'd be no change.

MattHJensen:
Merging, but would still like a followup on why the tests changed at all.

PR #1654, opened 2017-11-10
update RELEASES.md for 0.13.1

MattHJensen:
Thanks to @andersonfrailey, @martinholmer, @codykallen, and @rickecon for contributions to 0.13.1! Please let me know if I made any mistakes describing your contributions in RELEASES.md. We plan to release taxcalc 0.13.1 later this evening. I plan to merge this by 6 pm EST today (30 minutes).

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,8 +1,38 @@
 TAX-CALCULATOR RELEASE HISTORY
 ==============================
-Go
-[here](https://github.com/open-source-economics/Tax-Calculator/pulls?q=is%3Apr+is%3Aclosed)
+Go [here](https://github.com/open-source-economics/Tax-Calculator/pulls?q=is%3Apr+is%3Aclosed)
 for a complete commit history.
+
+
+2017-11-10 Release 0.13.1
+-------------------------
+(last merged pull request is
+[#1654](https://github.com/open-source-economics/Tax-Calculator/pull/1654))
+
+**API Changes**
+- None
+
+**New Features**
+- Add household and family identifiers from the CPS for the cps.csv.gz file that ships with taxcalc
+  [[#1635](https://github.com/open-source-economics/Tax-Calculator/pull/1635)
+  by Anderson Frailey]
+- Improved documentation for the cps.csv.gz file that ships with taxcalc
+  [[#1648](https://github.com/open-source-economics/Tax-Calculator/pull/1648)
+  by Martin Holmer]
+- Add parameter for the business income exclusion in the Senate TCJA Chairman's mark
+  [[#1651](https://github.com/open-source-economics/Tax-Calculator/pull/1648)
+  by Cody Kallen]
+- Add TCJA reform file for the Senate Chairman's mark
+  [[#1652](https://github.com/open-source-economics/Tax-Calculator/pull/1652)
+  by Cody Kallen]
+- Add FIPS state codes to the cps.csv.gz file that ships with taxcalc
+  [[#1653](https://github.com/open-source-economics/Tax-Calculator/pull/1653)
+  by Anderson Frailey]
+
+**Bug Fixes**
+- Fix an edge case related to new pass-through parameters that caused some extreme MTRs
+  [[#1645](https://github.com/open-source-economics/Tax-Calculator/pull/1645)
+  by Cody Kallen, reported by Richard Evans]
 
 
 2017-11-07 Release 0.13.0


MattHJensen:
Merging. @martinholmer, release 0.13.1 should be good to go. Thanks in advance for doing it.

martinholmer:
@MattHJensen,  Thanks for updating the `RELEASES.md` file for the 0.13.1 release.

However, there is a problem: when I run the tests on my computer I get one failure, which I'll document below.
Is it possible that in the rush to get all this work done on Friday afternoon that somebody forgot to run the tests on their local computer?

Here is the traceback for the test failure and my guess about what it all means after that:
```
STARTING WITH PY.TEST : Fri Nov 10 18:52:24 EST 2017
============================= test session starts ==============================
platform darwin -- Python 2.7.14, pytest-3.2.1, py-1.4.34, pluggy-0.4.0
rootdir: /Users/mrh/work/OSPC/Tax-Calculator, inifile: setup.cfg
plugins: xdist-1.17.1, pep8-1.0.6
gw0 [373] / gw1 [373] / gw2 [373] / gw3 [373]
scheduling tests via LoadScheduling
.....................................................................................................F...............................................................................................................................................................................................................................................................................
=================================== FAILURES ===================================
______________________________ test_puf_var_stats ______________________________
[gw3] darwin -- Python 2.7.14 /Users/mrh/anaconda2/bin/python
tests_path = '/Users/mrh/work/OSPC/Tax-Calculator/taxcalc/tests'
puf_fullsample =         DSI   e00200   e00300   e00400   e00600   e00650   e00700  e00800  \
0....0      0.0       0.00      0.00        1       0  

[215525 rows x 87 columns]

    @pytest.mark.requires_pufcsv
    def test_puf_var_stats(tests_path, puf_fullsample):
        """
        Main logic of test.
        """
        # create a Calculator object
        rec = Records(data=puf_fullsample)
        calc = Calculator(policy=Policy(), records=rec, verbose=False)
        # create base tables
        table_mean = create_base_table(tests_path)
        table_corr = copy.deepcopy(table_mean)
        # add statistics to tables
        year_headers = ['description']
        for year in range(Policy.JSON_START_YEAR, Policy.LAST_BUDGET_YEAR + 1):
            assert year == calc.policy.current_year
            year_headers.append(str(year))
            calc.calc_all()
            calculate_mean_stats(calc, table_mean, year)
            if year == 2016:
>               calculate_corr_stats(calc, table_corr)

tests/test_puf_var_stats.py:159: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/test_puf_var_stats.py:78: in calculate_corr_stats
    cor = np.corrcoef(var1, var2)[0][1]
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

x = array([ 0.,  0.,  0., ...,  0.,  0.,  0.])
y = array([0, 0, 0, ..., 0, 0, 0], dtype=int32), rowvar = True

    def corrcoef(x, y=None, rowvar=True, bias=np._NoValue, ddof=np._NoValue):
        """
        Return Pearson product-moment correlation coefficients.
    
        Please refer to the documentation for `cov` for more detail.  The
        relationship between the correlation coefficient matrix, `R`, and the
        covariance matrix, `C`, is
    
        .. math:: R_{ij} = \\frac{ C_{ij} } { \\sqrt{ C_{ii} * C_{jj} } }
    
        The values of `R` are between -1 and 1, inclusive.
    
        Parameters
        ----------
        x : array_like
            A 1-D or 2-D array containing multiple variables and observations.
            Each row of `x` represents a variable, and each column a single
            observation of all those variables. Also see `rowvar` below.
        y : array_like, optional
            An additional set of variables and observations. `y` has the same
            shape as `x`.
        rowvar : bool, optional
            If `rowvar` is True (default), then each row represents a
            variable, with observations in the columns. Otherwise, the relationship
            is transposed: each column represents a variable, while the rows
            contain observations.
        bias : _NoValue, optional
            Has no effect, do not use.
    
            .. deprecated:: 1.10.0
        ddof : _NoValue, optional
            Has no effect, do not use.
    
            .. deprecated:: 1.10.0
    
        Returns
        -------
        R : ndarray
            The correlation coefficient matrix of the variables.
    
        See Also
        --------
        cov : Covariance matrix
    
        Notes
        -----
        Due to floating point rounding the resulting array may not be Hermitian,
        the diagonal elements may not be 1, and the elements may not satisfy the
        inequality abs(a) <= 1. The real and imaginary parts are clipped to the
        interval [-1,  1] in an attempt to improve on that situation but is not
        much help in the complex case.
    
        This function accepts but discards arguments `bias` and `ddof`.  This is
        for backwards compatibility with previous versions of this function.  These
        arguments had no effect on the return values of the function and can be
        safely ignored in this and previous versions of numpy.
    
        """
        if bias is not np._NoValue or ddof is not np._NoValue:
            # 2015-03-15, 1.10
            warnings.warn('bias and ddof have no effect and are deprecated',
                          DeprecationWarning, stacklevel=2)
        c = cov(x, y, rowvar)
        try:
            d = diag(c)
        except ValueError:
            # scalar covariance
            # nan if incorrect value (nan, inf, 0), 1 otherwise
            return c / c
        stddev = sqrt(d.real)
>       c /= stddev[:, None]
E       FloatingPointError: invalid value encountered in true_divide
```
I think what numpy is trying to say is that it cannot compute the correlation coefficient between two arrays when the arrays contain all the same value because in that case the stddev is zero and dividing by zero causes a FloatingPointError.

But the big question is why this error did not occur a noon today and it does appear now.

@andersonfrailey @codykallen

MattHJensen:
@martinholmer said:

> Is it possible that in the rush to get all this work done on Friday afternoon that somebody forgot to run the tests on their local computer?

I did not run the tests but should have when I was teeing up the release. 

> But the big question is why this error did not occur a noon today and it does appear now.

It does see like no one ran the full test suite locally and was either counting on the travis-ci tests or `not requires_pufcsv` testing. 

I place the lion's share of the blame on myself for not doing the full testing during the release process, but it is also a good reminder to everyone who is lucky enough to have access to the PUF to run the full test suite before opening PRs.

MattHJensen:
Also, thanks very much to @martinholmer for the fix in #1655

Issue #1665, opened 2017-11-15
Clarity on TC handling of Additional Child Tax Credit


evtedeschi3:
My impression reading the tc documentation is that the model by default treats only those families with 3 or more children as eligible for the Additional Child Tax Credit.

If that impression is wrong, then this thread should hard stop here.

But if that's the way tc treats it, that's not what the law says. See e.g. [here](https://www.hrblock.com/tax-center/filing/credits/child-tax-credit/):

> **Additional child tax credit**
>
>If your child tax credit is limited, you might be able to claim the additional child tax credit. To qualify, **one** of these must apply _[emphasis mine]_:
>
>- Your earned income must be more than $3,000 for 2017.
>- You must have three or more qualifying children.
>
>If you have at least one qualifying child, you can claim a credit of up to 15% of the amount your earned income is more than $3,000. If you have three or more qualifying children, you can either:
>
> - Claim a refundable credit if the net Social Security and Medicare tax you paid is more than your Earned Income Credit (EIC), if any
> - Use the 15% method described earlier

Also, if you walk through the [IRS worksheet for the CTC and ACTC](https://www.irs.gov/pub/irs-pdf/f8812.pdf), it's clear that having 3 or more kids does not disqualify you, it just bumps you into a different eligibility worksheet.

So having 1 or 2 kids simply means your ACTC must be calculated using the 15%*(earned income - $3,000) approach, not the approach that allows a credit for FICA taxes in excess of the EITC.

Wanted to clarify because recent tax reform proposals have tinkered with the ACTC, which can have a meaningful impact particularly on low-income parents.

martinholmer:
@evtedeschi3,  Thanks for your continued interest in Tax-Calculator.

Your question in issue #1665 about the Additional Child Tax Credit (ACTC) is a good one because it reveals either poor documentation or incorrect tax-calculating logic or both.

Let's figure out, in this issue, what needs to be fixed.  

**First**, I'll show you below the ACTC calculation logic (which you will see is dated, and possibly out-of-date if there have been changes in the ACTC over the past few years).  The Python code, which is strongly oriented to using variable names that IRS-SOI staff, but few others, know by heart, is a little hard to follow but the outlines of the logic should be apparent.  I particular, my quick glance at the code, suggests that filing units with fewer than three children are eligible for the ACTC.  But I would appreciate your review of the logic.

**Second**, I hope you can point us to the place(s) in the user documentation that led you to raise this issue.

[...]

PR #1666, opened 2017-11-15
Fix _ACTC_ChildNum documentation and use it more often in the code

martinholmer:
The changes in this pull request were suggested by @evtedeschi3 in issue #1665.

Hopefully the new documentation is less confusing and certainly the code is more accurately parameterized.
These changes have no effect on current-law tax results, but the more thorough parameterization of the ACTC function could cause reforms that change `_ACTC_ChildNum` to produce slightly different (and more accurate) results than they did before these changes.

@MattHJensen @feenberg @Amy-Xu @andersonfrailey @hdoupe @codykallen @evtedeschi3


martinholmer:
No concerns or comments, so merging pull request #1666, which responds to bug report in issue #1665.

PR #1676, opened 2017-11-17
Add param for switching stacking order of dependent credit and ctc

### LINKED ISSUE #1671: Add new param for stacking order of refundable and nonrefundable CTC
A user said:

> 
> Here’s the issue: you’ve got a $2000 regular CTC in the Senate bill and a $1,100 refundable portion a la the ACTC (in 2018). One can interpret these parameters in two ways.
> 
> 1. Up to $1,100 per child of the regular CTC is eligible to spill over into the ACTC in 2018. When the regular CTC claimed is above this threshold, none spills over to the ACTC.
> 
> 2. Up to $1,100 per child of the regular CTC is eligible to spill over into the ACTC in 2018. When the regular CTC claimed is below the $2000 maximum threshold, the difference between the CTC and $2000 (up to $1,100) spills over to the ACTC.
> 
> I’m pretty sure, based on looking at the tc output of families with children getting tax hikes in 2018 using the Senate JSON, that tc is treating the ACTC the first way: the regular CTC goes up but the ACTC goes away completely for many families. My initial read is that [other modeling group] is interpreting it the second way. 

[Other modeling group] said: 

> 
> ·        The bill indexes the refundable child tax credit threshold of $1,000 to C-CPI-U starting in 2018 and rounds up to the nearest $100.  Under the bill, it is expected to be $1,100 refundable in 2018.  You can index going forward by looking at the ratio of chained CPI in the preceding year to chained CPI in the base year.  So inflation adjustment for 2019 would be C-CPI-U 2018 divided by  C-CPI-U 2017 all multiplied by $1,000, and then rounded up to the nearest hundred.  Non-refundable piece is then $1,600 minus the calculated and rounded refundable piece.
> ·        When calculating a credit with a refundable and non-refundable piece, sometimes it is easiest to split it into two lines.  Then you subtract non-refundable piece first, but don’t let taxable income go below zero.  Then subtract the refundable piece to get taxable income. 
> o   1st example: 300 in taxable income before child tax credit and qualifies for $1,600 child tax credit ($500 non-refunadable, $1,100 refundable). 
> §  first calculation is max(0,taxable income – non-refunadable CTC) = max (0, 300 – 500) = 0 = X. 
> §  second calculation is X- refundable CTC = X – 1,100 = -1,100 tax liability.
> o   2nd example: 800 in taxable income before CTC and qualifies for $1,600 CTC ($500 non-refundable, $1,100 refundable).
> §  first calculation is max(0,taxable income – non-refunadable CTC) = max (0, 800 – 500) = 300 = X. 
> §  second calculation is X – refundable CTC = X – 1,100 = -800 tax liability.

The request is to add a boolean parameter to current_law_policy.json that allows to swap the stacking order of refundable and nonrefundable CTC. 

@codykallen suggested via email that the [other modeling group's] interpretation could be modeled by:

> If you want to switch the order here to match their description, go to functions.py, to the NonrefundableCredits function. To have them draw down the non-refundable dependent_credit first, move lines 1413 and 1414 to be executed before line 1398.  

cc @martinholmer @Amy-Xu @hdoupe @andersonfrailey @evtedeschi3 @GoFroggyRun

MattHJensen:
Resolves #1671 following recommendation by @codykallen. 

See results.json which shows that stacking dependent credit above ctc does reduce tax liabilities as is expected based on the discussion in #1671. 

cc @evtedeschi3 @martinholmer


martinholmer:
@MattHJensen said in pull request #1676:
> Resolves #1671 following recommendation by @codykallen.
>
> See results.json which shows that **stacking dependent credit above CTC** does reduce tax liabilities as is expected based on the discussion in #1671.

I haven't been following this issue closely, but I thought #1671 was about how the  CTC and the Additional CTC interacted.  What does the stacking order of the "dependent credit" and the nonrefundable CTC have to do with #1671?  As you can see, I'm confused and need help understanding this issue.   If the pull request does not produce the expected results, maybe we need to think about this issue some more.

MattHJensen:
> I haven't been following this issue closely, but I thought #1671 was about how the CTC and the Additional CTC interacted. What does the stacking order of the "dependent credit" and the nonrefundable CTC have to do with #1671?

Under current law the maximum CTC is $1000 and up to $1000 can flow through to the ACTC for refundability. The TCJA adds to the CTC an amount that can not flow to the ACTC for refundability. We have been modeling that amount with a parameter that we call the `Nonrefundable credit for children, on top of the child tax credit` There is also a nonrefundable credit for non-children dependents that is treated similarly. The naming here is definitely somewhat confusing, and perhaps if this is enacted we should rethink it (or beforehand), but no one yet has suggested a better approach. 

> If the pull request does not produce the expected results, maybe we need to think about this issue some more.

The pull request _does_ produce the expected result.

martinholmer:
@MattHJensen said:
> The pull request does produce the expected result.

Right.  Sorry, I misread your original comment in pull request #1676.

PR #1685, opened 2017-11-19
add cpi_offset reform to reforms.json

MattHJensen:
Just checking the magnitude and pattern of cpi_offset reforms. Does not change results.

### DIFF
--- a/taxcalc/tests/reforms.json
+++ b/taxcalc/tests/reforms.json
@@ -560,27 +560,27 @@
 
   "56": {
       "start_year": 2017,
-      "value": {"_DependentCredit_before_CTC": [true],
-                "_DependentCredit_Child_c": [500]},
-      "name": "Stack dependent credit before CTC",
-      "output_type": "iitax",
-      "compare_with": {},
-      "expected": "Tax-Calculator,-13.2,-12.9,-12.3,-12.1"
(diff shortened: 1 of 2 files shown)


taxcalc/tests/reforms.json
```diff
@@ -575,6 +575,14 @@
...
-  }
+  },
 
+      "58": {
+      "start_year": 2017,
+      "value": {"_cpi_offset": [-0.0025]},
+      "name": "Stack dependent credit before CTC",
+      "output_type": "iitax",
+      "compare_with": {},
+      "expected": "Tax-Calculator,0.0,2.2,4.6,7.2"
```
MattHJensen: @codykallen, notice how the cpi_offset reform doesn't change results in the year it is implemented. In the TCJA reform files, do you think we should implement it in 2016?

MattHJensen: > Do you mean 2017 (so that it first takes effect in 2018)? 

I was looking at this: https://waysandmeansforms.house.gov/uploadedfiles/bill_text.pdf
image

MattHJensen:
Does not affect results and conversation about followups has been moved to #1687. Merging to prevent potential for conflicts with with #1687.

PR #1686, opened 2017-11-20
Test new pt params

MattHJensen:
In my work on #1614, I found that `_PT_wages_active_income` does not affect results, even in the presence of a reduced `_PT_rt7` compared to `_II_rt7`. The tests in this PR confirm that finding. 

@codykallen, could you suggest how I can 'activate' the _PT_wages_active_income parameter so that switching it from `false` to `true` has an effect on results?


MattHJensen:
Merging. Further discussion in #1692.

Issue #1668, opened 2017-11-16
Are Tax-Calculator results too sensitive to substitution effect elasticity?


rickecon:
I ran some tests today using [`TCJA_Senate.json`](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/reforms/TCJA_Senate.json) as my reform and with the following behavioral assumptions `json`:
```json
{
  "behavior": {
        "_BE_sub": {"2018": [0.25]},
        "_BE_cg": {"2018": [-3.45]}
    },
  "consumption": {
    },
  "growdiff_response": { 
  "_ACPIU": {"2017": [-0.0025]}
  },
  "growdiff_baseline": {}
}
```
The difference in government revenue over the 10-year budget window for that simulation is -$530 billion. I ran some other simulations of the [`TCJA_Senate.json`](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/reforms/TCJA_Senate.json) reform in which I varied the substitution effect elasticity (`_BE_sub`) and the capital gains elasticity (`_BE_cg`). These are the results I got.

| description | `_BE_sub` | `_BE_cg` | `growdiff_response` | 10-yr rev. chg. |
| :----------- | :--------: | :---------: | :-------------: | ------------: |
| Original       | 0.250        | -3.450         |   -0.0025          | -$530 B |
| No behavior | 0.000       | 0.000              | 0.0000          | -$1,300 B |
| `_BE_cg` down | 0.250      | - 1.725         | -0.0025            | -$556 B |
| `_BE_sub` down | 0.125      | - 3.450         | -0.0025            | -$906 B |
| `_BE_sub` up | 0.400      | - 3.450         | -0.0025            | -$76 B |

Going from an elasticity of substitution of 0.4 (which is still within a reasonable range from the literature) to a lower elasticity of 0.125 wipes out nearly $1 trillion of lost revenue. That seems like too much sensitivity.

@jdebacker pointed out one possible culprit. The effect of the elasticity on behavior is the following. Let income be `Inc`. Let `MTR_b` be the marginal tax rate in the baseline scenario, and let `MTR_r` be the marginal tax rate in the reform scenario. The behavioral effect of a change in `MTR_r` is approximated as the percent change in `1 - MTR` multiplied by the substitution effect elasticity of taxable income `_BE_sub` times income `Inc`.

change in income = `[(1 - MTR_r) / (1 - MTR_b) - 1] x _BE_sub x Inc`

[...]

PR #1698, opened 2017-11-22
Fix Behavior.response function logic to handle high MTRs

### LINKED ISSUE #1697: Behavior.response logic needs to be fixed
In issue #1668, @rickecon pointed out that there is a bug in the `Behavior.response` function.  The current logic does not correctly handle marginal tax rates that are 100% or above.  As shown in the discussion of #1668, there are a few filing units with marginal tax rates at that high level.

martinholmer:
Resolves #1697, a bug reported originally by @rickecon in issue #1668.

### DIFF
--- a/taxcalc/behavior.py
+++ b/taxcalc/behavior.py
@@ -162,7 +162,10 @@ class Behavior(ParametersBase):
                 sub = np.zeros(calc_x.records.dim)
             else:
                 # proportional change in marginal net-of-tax rates on earnings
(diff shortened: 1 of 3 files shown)


martinholmer:
@rickecon and @jdebacker,  
Does pull request #1698 seem like a reasonable way to handle the `pch` problem you identified in issue #1668.

jdebacker:
@martinholmer, this looks like a good fix to me.  Thanks!

PR #1708, opened 2017-11-28
Both create_di*_table utility functions now take Calculator (not Records) object as argument

martinholmer:
This pull request is part of an ongoing program of making the public Tax-Calculator API more object oriented.

As before, the `create_distribution_table` and `create_difference_table` utility functions accept Pandas DataFrame(s) as their first argument(s).  However, now these two functions accept Calculator object(s) --- rather than Records object(s) --- as the other kind of valid first argument(s).

There is no change in tax-calculating logic or in tax results.

### DIFF
--- a/taxcalc/utils.py
+++ b/taxcalc/utils.py
@@ -287,7 +287,7 @@ def create_distribution_table(obj, groupby, income_measure, result_type):
     Parameters
     ----------
     obj : any object with array-like attributes named as in STATS_COLUMNS list
-        Examples include a Tax-Calculator Records object and a
+        Examples include a Tax-Calculator Calculator object and a
         Pandas DataFrame object.
 
     groupby : String object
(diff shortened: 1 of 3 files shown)


taxcalc/utils.py
```diff
@@ -357,7 +358,10 @@ def add_columns(pdf):
         columns = STATS_COLUMNS + [income_measure]
     else:
         columns = None
-    res = results(obj, cols=columns)
+    if isinstance(obj, pd.DataFrame):
+        res = results(obj, cols=columns)
+    else:
+        res = results(obj.records, cols=columns)
```
hdoupe: @martinholmer do you think you should add an assertion statement: `assert isinstance(obj, Calculator)`?

martinholmer: Yes, I tried that but we have never imported Calculator or Records into the utils.py file.

hdoupe: Ok, I see.

PR #1690, opened 2017-11-20
[DO NOT MERGE] update for Python 3.6

MattHJensen:
As far as I can tell, the only change in this PR was to update from python 2.7 to python 3.6, which I needed to do to handle some bugs in my WIP PR #1614. Yet that Python update appears to have influenced tax-calculator results. Has anyone else run the `requires_pufcsv` tests on master on a machine running Python 3.6?

cc @martinholmer @hdoupe @GoFroggyRun @andersonfrailey 

Here is my conda env:

```
# packages in environment at /Users/Matt.Jensen/anaconda3/envs/taxcalc-dev:
#
astroid                   1.5.3            py36h1333018_0  
bokeh                     0.12.10          py36hfd5be35_0  
ca-certificates           2017.08.26           ha1e5d58_0  
certifi                   2017.7.27.1      py36hd973bb6_0  
coverage                  4.4.1                    py36_0  
execnet                   1.3.0                    py36_1  
intel-openmp              2018.0.0             h68bdfb3_7  
isort                     4.2.15           py36hceb2a01_0  
jinja2                    2.9.6            py36hde4beb4_1  
lazy-object-proxy         1.3.1            py36h2fbbe47_0  
libcxx                    4.0.1                h579ed51_0  
libcxxabi                 4.0.1                hebd6815_0  
libedit                   3.1                  hb4e282d_0  
libffi                    3.2.1                h475c297_4  
libgfortran               3.0.1                h93005f0_2  
llvmlite                  0.20.0                   py36_0  
markupsafe                1.0              py36h3a1e703_1  
mccabe                    0.6.1            py36hdaeb55d_0  
mkl                       2018.0.0             h5ef208c_6  
mock                      2.0.0            py36hd48142b_0  
ncurses                   6.0                  hd04f020_2  
numba                     0.35.0              np113py36_6  
numpy                     1.13.3           py36h2cdce51_0  
openssl                   1.0.2m               h86d3e6a_1  
pandas                    0.21.0           py36hfed917e_1  
pbr                       3.1.1            py36h1bc6f65_0  
pep8                      1.7.0            py36hc268eb1_0  
pip                       9.0.1            py36h1555ced_4  
py                        1.4.34           py36hecf431b_1  
pylint                    1.7.4            py36hdee9077_0  
pytest                    3.2.1            py36h9963153_1  
pytest-cache              1.0                      py36_0  
pytest-pep8               1.0.6                    py36_0  
pytest-xdist              1.17.1                   py36_0  
[...]

PR #1715, opened 2017-11-29
[DO NOT MERGE] update for Python 3.6 

MattHJensen:
Continuation of discussion in #1690.


martinholmer:
@MattHJensen,  I'd like to suggest several things to do in pull request #1715:

1. The display of variable means to the nearest **one-tenth of a cent** seems to be excessive precision (see [this line](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/tests/test_puf_var_stats.py#L167) of code).  Maybe you should try something like this `float_format='%8.1f'` instead of `float_format='%8.3f'` in that line of code in the `test_puf_var_stats.py` file.  Or maybe rounding to the nearest dollar is sufficiently accurate for this purpose.

2. The inclusion of the `agi_bin` variable doesn't seem meaningful (or, at least, I couldn't think of a reason why anybody would be interested in the mean or correlations of that variable).  Maybe you should add `agi_bin` to the list of unused variables [here](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/tests/test_puf_var_stats.py#L38-L51) in the `test_puf_var_stats.py` code.

3. After thinking about the results in this pull request, I remembered that we have run into differences in floating-point results when running Tax-Calculator under Python 2.7 and under Python 3.6.  Look at [this code](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/tests/test_cpscsv.py#L25-L120) in the `test_cpscsv.py` file to see how we've been handling these CPS results differences.  These differences became apparent immediately after the test was added because Travis-CI runs all the non-PUF tests under both Python 2.7 and Python 3.6.  It seems unlikely to me that the 2.7-vs-3.6 differences are caused by a bug in Tax-Calculator.  A more likely explanation is that the floating-point libraries (or perhaps numpy or perhaps pandas) are slightly different between 3.6 and 2.7.   What to do?  Over the next year or two, it seems likely that we'll move all developers and users to Python 3.6.  But in the short term, you might consider using the approach currently being used in `test_cpscsv.py` in the `test_puf_var_stats.py` code.

MattHJensen:
@martinholmer, thank you, these are all very helpful suggestions.

martinholmer:
@MattHJensen,  I've implemented some changes along the lines of the suggestions I made earlier in this conversation regarding pull request #1715.  And @derrickchoe has been kind enough to test my changes on his computer, which is running Python 3.6.  As you can see [here](https://github.com/open-source-economics/Tax-Calculator/pull/1711#issuecomment-353644590) it looks as if there are no longer any test failures when you are using Python 3.6.

Can you confirm this on your computer?  Just update your master branch and run all 390+ unit tests to see if there are no failures.  If there are no failures, you could consider closing (without merging) pull request #1715.

PR #1716, opened 2017-11-30
ctc new switch to refund against full payroll taxes

### LINKED ISSUE #1714: Payroll tax limitation on new Child Tax Credit refundability
When the user chooses to limit the refundability of the new CTC to some percent of payroll taxes, `functions.py` uses the `ptax_oasdi` variable (that is, just the OASDI portion of payroll taxes) as the reference for the limit (see line 1536). If the new CTC were limited to 100% of payroll taxes, then this treatment would be equivalent to a 12.4% phase-in.

But this isn't the only approach being bandied about. The Rubio-Lee amendment introduced today assumes a 15.3% phase-in under a 100% payroll tax limitation. That means they're assuming the cap is OASDI + HI payroll taxes. So in that case, `payrolltax` would be the better reference variable.

The current tc treatment isn't wrong per se but would be helpful to have a switch.

MattHJensen:
Resolves #1714. 

The reform results in reforms.json are as expected (refundability against all payroll taxes costs more than refundability against OASDI)

@evtedeschi3, could you please take a look? 

If you would like to check this out locally, see https://help.github.com/articles/checking-out-pull-requests-locally/. 

cc @martinholmer @codykallen


evtedeschi3:
Yes, this looks exactly right.

MattHJensen:
Thanks @evtedeschi3. Merging.

PR #1718, opened 2017-12-02
Add several Calculator table methods and make table utilities low-level

martinholmer:
This pull request continues the movement of Tax-Calculator towards a more objected-oriented structure that began in pull request #1708.  This pull request makes the create_*_table utility functions low-level by changing their arguments from Calculator objects to DataFrame objects, and adds several new table methods to the Calculator class that make it easier to create results tables: a diagnostic table (that can be for one or more years), a pair of distribution tables (that use consistent bins/deciles in cases where the income measure differs between the baseline and reform), and a difference table.

These changes alter the public API of the Tax-Calculator but only for those doing Python programming by developing their own scripts or notebooks.  These changes have no effect on `tc` users (who are accessing Tax-Calculator via the CLI, Command-Line Interface) or on TaxBrain users (who are accessing Tax-Calculator via the TBI, TaxBrain Interface).

There is no change in tax-calculating logic or the structure of the results tables.  It is just that the methods for generating the results tables are now easier to use and hopefully less error prone.

This pull request is part of the object-oriented enhancements described in issue #1720.


MattHJensen:
@martinholmer, thanks for this PR! For the sake of developing a shared understanding of design objectives, could you describe the benefits you see of adopting these changes? For instance, what is the value of "mov[ing] towards a more object oriented structure"?

RELEASES.md
```diff
@@ -10,14 +10,17 @@ for a complete commit history.
...
-  by Martin Holmer as suggested by Matt Jensen and Hank Doupe]
+- Change arguments for the three create-table utility functions and add easier to use Calculator methods that generate results tables
+  [[#1718](https://github.com/open-source-economics/Tax-Calculator/pull/1718)
+  by Martin Holmer]
 
 **New Features**
 - Add new policy parameter for refunding the new CTC against all payroll taxes
   [[#1716](https://github.com/open-source-economics/Tax-Calculator/pull/1716)
   by Matt Jensen as suggested by Ernie Tedeschi]
+- Remove calculation of AGI tables from the TaxBrain Interface, tbi
```
MattHJensen: @martinholmer, what made you switch this to a new feature from an API change? 

cc @hdoupe

martinholmer: Dropping the AGI tables doesn't change the way you call any of the TBI functions, which is the most common cause of an API change.  If TaxBrain had been using (that is, displaying) the AGI tables, then I would have called #1718 an API change.  But TaxBrain 1.2.0 isn't doing anything with the AGI tables as far as I can see, so dropping them has no effect on TaxBrain.  That was the thinking.  If you would like to have it shown as an API change, just let me know and I'll move its position in the `RELEASES.md` file.

MattHJensen: That makes sense. Thanks. Fine as a feature enhancement.

MattHJensen:
This looks good to me. Thanks for the contribution @martinholmer.

PR #1719, opened 2017-12-04
Add Benefit Data and Associated Capabilities

andersonfrailey:
This PR adds benefit data including C-TAM imputed benefits to the CPS file and a new file that contains extrapolated benefit data for the years 2015-2027.

The new file `cps_benefits_extrap.csv.gz`, contains extrapolated benefits only for those who actually receive benefits. All of the zero values have been dropped to reduce file size. 

It also adds the ability for users to repeal benefits and apply welfare multiples to the C-TAM benefits. This would allow them to measure the perceived welfare from a benefit. Examples of these capabilities, as well as plots of total benefits, benefit participation, and other information can be seen in [this notebook](http://nbviewer.jupyter.org/github/andersonfrailey/Notebook-Uploads/blob/master/TaxCalc%20Benefit%20Functionality.ipynb).

There is one question that needs to be answered before this PR is merged. Which growth factor should be used to blow up the `other_ben` variable? These are the non-modeled benefits. @Amy-Xu and @MattHJensen, do either of you have any thoughts on this?

I'll be opening a corresponding TaxData PR shortly.

@martinholmer, can you take a particularly close look at the modifications I made to `index.htmx`? I believe I did everything correctly and it all looks fine when I look at the resulting `index.html` file, I just want to get a second pair of eyes on it.

cc @hdoupe


andersonfrailey:
I'll fix the coverage issues in my next commit.

Amy-Xu:
Anderson asked:

>Which growth factor should be used to blow up the other_ben variable? These are the non-modeled benefits.

Is it possible to find the gross expenditure of all non-modeled programs during the past 5-10 years and apply the 'average' growth rate? 

cc @MattHJensen

taxcalc/records.py
```diff
@@ -386,6 +392,23 @@ def _adjust(self, year):
...
+        self.snap_ben = np.nan_to_num(ben)
+        ben = self.BEN['vb_benefits_{}'.format(year)][self.RECID].values
+        self.vet_ben = np.nan_to_num(ben)
+        ben = self.BEN['ss_benefits_{}'.format(year)][self.RECID].values
+        self.ss_ben = np.nan_to_num(ben)
+        ben = self.BEN['medicare_benefits_{}'.format(year)][self.RECID].values
+        self.mcare_ben = np.nan_to_num(ben)
+        ben = self.BEN['medicaid_benefits_{}'.format(year)][self.RECID].values
+        self.mcaid_ben = np.nan_to_num(ben)
+
```
hdoupe: @andersonfrailey do you think that we should do apply the `np.nan_to_num` function in the `taxdata` repo?

andersonfrailey: No because `NaN` values don't occur until we're assigning benefits. This line of code matches benefits by RECID:
```python
ben = self.BEN['medicaid_benefits_{}'.format(year)][self.RECID].values
```

Because we drop all of the records without benefits when we create the extrapolated benefits file, those who have no benefits receive a `NaN` value instead and so we have to apply the function here.

andersonfrailey:
I changed the default value of benefits to `None` in my last commit because the default values for a Records object correspond to the PUF file and benefits are specifically for the CPS file.

martinholmer:
@andersonfrailey asked:
> can you take a particularly close look at the modifications I made to `index.htmx`? I believe I did everything correctly and it all looks fine when I look at the resulting `index.html` file.

Yes, everything looks fine with the changes to `index.htmx`.
And, if the generated `index.html` file looks good, then that's the proof.

martinholmer:
@andersonfrailey,  In addition to the need for new-code testing, there are several other problems with #1719 that need your attention:
```
PYLINT records.py
************* Module taxcalc.records
records.py:110: [R0914(too-many-locals), Records.__init__] Too many local variables (16/15)
records.py:410: [W0201(attribute-defined-outside-init), Records._extrapolate_benefits] Attribute 'mcaid_ben' defined outside __init__
records.py:400: [W0201(attribute-defined-outside-init), Records._extrapolate_benefits] Attribute 'ssi_ben' defined outside __init__
records.py:402: [W0201(attribute-defined-outside-init), Records._extrapolate_benefits] Attribute 'snap_ben' defined outside __init__
records.py:406: [W0201(attribute-defined-outside-init), Records._extrapolate_benefits] Attribute 'ss_ben' defined outside __init__
records.py:404: [W0201(attribute-defined-outside-init), Records._extrapolate_benefits] Attribute 'vet_ben' defined outside __init__
records.py:408: [W0201(attribute-defined-outside-init), Records._extrapolate_benefits] Attribute 'mcare_ben' defined outside __init__

PEP8 current_law_policy.json
current_law_policy.json:4864:16: E203 whitespace before ':'
current_law_policy.json:4886:16: E203 whitespace before ':'
current_law_policy.json:4908:16: E203 whitespace before ':'
current_law_policy.json:4930:16: E203 whitespace before ':'
current_law_policy.json:4952:16: E203 whitespace before ':'
current_law_policy.json:4974:16: E203 whitespace before ':'
```
The `attribute-defined-outside-init` pylint warnings are especially worrying.

taxcalc/records.py
hdoupe: Ok, that makes sense to me. Thanks @andersonfrailey

andersonfrailey:
@martinholmer said:
>The attribute-defined-outside-init pylint warnings are especially worrying.

Thanks for pointing these out. Is there a way to add the test you ran to Travis-CI or our test suite? I didn't get any of those errors when I was working on this PR.

martinholmer:
@andersonfrailey said:
>> The attribute-defined-outside-init `pylint` warnings are especially worrying.
>
> Thanks for pointing these out. 
> Is there a way to add the test you ran to Travis-CI or our test suite?
> I didn't get any of those errors when I was working on this PR.

All contributors should be following the coding and testing procedures described [here](https://github.com/open-source-economics/Tax-Calculator/blob/master/docs/contributing.md).

It was decided long ago that it is each contributor's responsibility to run the `pylint` coding tests locally.

hdoupe:
@martinholmer said 
> It was decided long ago that it is each contributor's responsibility to run the pylint coding tests locally.

I'm curious about the rationale for not including this in the travis tests.  Could you point me to the issue where this discussion occurred?

martinholmer:
@andersonfrailey said:
> I found it. See #30.

Great archival work!  So, as you can see it was hard enough getting the project to enforce PEP8 at GitHub check-in.  Leaving the `pylint` checking as a local development activity was the compromise required to get mandatory PEP8 checking at check-in.

Also optional, but useful in my opinion, is running the JSON files through PEP8 like in this example:
```
pep8 --ignore=E501,E121 current_law_policy.json
```
Only some of the Python files comply with `pylint`; others are such a mess, at least according to `pylint`, that I have never had the time or nerve to revise the code so that the `pylint` warnings are eliminated.  The fact that we are doing `pylint` checking only for some files would make it complicated to include this kind of testing in Travis-CI.  To make it easy to do these selective coding tests on my computer, I've constructed a rudimentary bash script that does all the work.  It was the output from that script that I posted earlier in this discussion in [this comment](https://github.com/open-source-economics/Tax-Calculator/pull/1719#issuecomment-349620361).  I'd be happy to share the bash script with you if you are interested.

andersonfrailey:
Ran a PEP8 check on the offending files and they came up clean. Latest commit also excludes the lines needed to use `read_egg_csv` from coverage testing as was done with the other `_read_X` methods and all tests are now passing.

That being said, this PR should not be merged yet because @hdoupe, @Amy-Xu, and I noticed a bug in the scripts used to extrapolate the benefits (see TaxData PR [#133](https://github.com/open-source-economics/taxdata/pull/133)). Once that is fixed this is ready to merge upon review from other contributors. Until then, I have updated the title to note that it shouldn't be merged.

martinholmer:
@andersonfrailey,  See my questions about pull request #1719 in this comment on [taxdata pull request 133](https://github.com/open-source-economics/taxdata/pull/133#issuecomment-356158366).

martinholmer:
@andersonfrailey, as I remember our phone conversation months ago, we agreed that when the benefits data were merged into Tax-Calculator that the "ss" terminology for social security would not be part of Tax-Calculator code.

You are are correctly not using "fs" for food stamps, but rather "snap", which is the acronym of the program's name.

So why not use "oasdi" (instead of "ss") because the program's name is Old-Age, Survivors, and Disability Insurance (OASDI)?

martinholmer:
@andersonfrailey, another question about #1719.

What do these variables (other than "ben_total", which I understand) mean?
```
   },
 +    "ben_total": {
 +      "type": "float",
 +      "desc": "Total benefits recieved by that tax unit",
 +      "form": {}
 +    },
 +    "ben_welfare": {
 +      "type": "float",
 +      "desc": "Perceived benefits received after accounting for deadweight loss",
 +      "form": {}
 +    },
 +    "ssi_ben_welfare": {
 +      "type": "float",
 +      "desc": "Perceived SSI benefits after accounting for deadweight loss",
 +      "form": {}
 +    },
 +    "snap_ben_welfare": {
 +      "type": "float",
 +      "desc": "Perceived SNAP benefits after accounting for deadweight loss",
 +      "form": {}
 +    },
 +    "vet_ben_welfare": {
 +      "type": "float",
 +      "desc": "Perceived veterans benefits after accounting for deadweight loss",
 +      "form": {}
 +    },
 +    "mcare_ben_welfare": {
 +      "type": "float",
 +      "desc": "Perceived Medicare benefits after accounting for deadweight loss",
 +      "form": {}
 +    },
 +    "mcaid_ben_welfare": {
 +      "type": "float",
 +      "desc": "Perceived Medicaid benefits after accounting for deadweight loss",
 +      "form": {}
 +    },
 +    "ss_ben_welfare": {
 +      "type": "float",
 +      "desc": "Perceived Social Security benefits after accounting for deadweight loss",
 +      "form": {}
      }
```
I can't seem to find any code in #1719 that does "deadweight loss" calculations on the benefits.
Did I missing that code?

If I didn't miss the code, can you remove these variables from #1719?
Doing this kind of calculation will be involved (and probably controversial), so let's wait until we get there.

martinholmer:
@andersonfrailey, yet another question about #1719.
The pull request contains a new file called `cps_benefits_extrap.csv.gz`, which contains `RECID` and six positive benefit amounts (VET, SNAP, SSI, MCAID, MCARE, OASDI) for fourteen years (2014-2027), 85 variables in all.   And there is a row for each filing unit that has at least one positive benefit, which results in a file that has less than half as many rows as the `cps.csv` file has.
```
$ awk -F, 'NR>1{n++}END{print n}' taxcalc/cps.csv
456465
$ awk -F, 'NR>1{n++}END{print n}' taxcalc/cps_benefits_extrap.csv
192786
```
My question is why the long name for this file?
Why `cps_benefit_extrap.csv.gz`?
Why not simply `cps_benefits.csv.gz`?

MattHJensen:
> I can't seem to find any code in #1719 that does "deadweight loss" calculations on the benefits.

In the working paper we posted last year, we took some deadweight loss assumptions from the literature and applied them as haircuts to the baseline values of the income transfer programs. 

Based on Tax-Calculator's principal of allowing users to make key assumptions, I believe that the long-run outcome should be for Tax-Calculator users to be able to apply these haircut values as assumption parameters, and the adjustments should not be baked into the input data. Does that make sense to others?  If so, these `_welfare` variables should not be included in this PR, but a Tax-Calculator enhancement will be needed to allow for Tax-Calculator-based adjustments to baseline and reform value of input data variables [1], and a PolicyBrain enhancements will be needed make these parameters available from the TaxBrain GUI. 

I would appreciate feedback from @andersonfrailey, @Amy-Xu, @hdoupe, @martinholmer about whether all of this is practicable if our goal is to have these features on the Partial Equilibrium page of TaxBrain in late February in order to meet our grant deadlines, or if there is some alternative approach. 

[1] Growth factors should be applied to the unadjusted values

martinholmer:
**How to merge benefits data into the basic CPS data**

Development work in the taxdata 133 and Tax-Calculator 1719 pull requests seeks to assign positive benefits (from a small `cps_benefits.csv.gz` file) to the appropriate CPS filing unit (from the `cps.csv.gz` file).  This comment suggest a way to do that.

The approach is to merge two Pandas DataFrame objects using `RECID`, which is present in both, to associate rows from the two differently-dimensioned DataFrame objects.  Here is a simple script that does that using the benefits data file in #1719.
```
from taxcalc import *
calc = Calculator(policy=Policy(), records=Records.cps_constructor())
print 'Calculator embedded-records array length:', calc.array_len

rid = calc.dataframe(['RECID'])
print 'rid dataframe shape:', rid.shape
print 'rid RECID dtype:', rid['RECID'].dtype
print rid.head(n=10)

ben_raw = pd.read_csv('taxcalc/cps_benefits.csv')
# print ben_raw.head()  # shows excessive precision in benefit amounts
ben_full = ben_raw.astype(dtype=np.int32, errors='ignore')
print 'ben_full dataframe shape:', ben_full.shape
ben = ben_full[['RECID', 'snap_benefits_2014', 'medicare_benefits_2014']]
print 'ben RECID dtype:', ben['RECID'].dtype
print ben.head(n=10)

tot = rid.merge(ben, how='left', on='RECID', indicator=True)
print 'tot dataframe shape:', tot.shape
print 'tot RECID dtype:', tot['RECID'].dtype
print tot.head(n=10)
```
The use of the `indicator` arguement in the `merge` call is just for illustration and debugging.

When this script is executed, its output is this:
```
You loaded data for 2014.
Tax-Calculator startup automatically extrapolated your data to 2014.
Calculator embedded-records array length: 456465
rid dataframe shape: (456465, 1)
rid RECID dtype: int32
   RECID
0      1
1      2
2      3
3      4
4      5
5      6
6      7
7      8
8      9
9     10
ben_full dataframe shape: (192786, 85)
ben RECID dtype: int32
   RECID  snap_benefits_2014  medicare_benefits_2014
0      3                1589                    1716
1      4                   0                       0
2      5                   0                    5225
3      6                   0                   43071
4      9                1146                       0
5     10                8495                       0
6     13                   0                       0
7     17                2650                    3410
8     19                   0                     230
9     20                   0                    6695
tot dataframe shape: (456465, 4)
tot RECID dtype: int32
   RECID  snap_benefits_2014  medicare_benefits_2014     _merge
0      1                 NaN                     NaN  left_only
1      2                 NaN                     NaN  left_only
2      3              1589.0                  1716.0       both
3      4                 0.0                     0.0       both
4      5                 0.0                  5225.0       both
5      6                 0.0                 43071.0       both
6      7                 NaN                     NaN  left_only
7      8                 NaN                     NaN  left_only
8      9              1146.0                     0.0       both
9     10              8495.0                     0.0       both
```

I would suggest using this approach in #1719 so that a small benefits file (containing only rows for those with at least one positive benefit amount) can be used in Tax-Calculator.

@andersonfrailey @hdoupe @Amy-Xu @MattHJensen

andersonfrailey:
Thanks for your questions, @martinholmer. I'll address each individually(and in separate comments to avoid having one huge chunk of text).

>So why not use "oasdi" (instead of "ss") because the program's name is Old-Age, Survivors, and Disability Insurance (OASDI)?

`ss_ben` is a carry-over from when we were first working on benefit related projects, and in C-TAM the scripts used to create the benefit imputations all use "SS" as a a starting point. My only objection to using "oasdi" rather than "ss" is breaking that connection. But, if the benefits for Tax-Calculator (such as name clarity/accuracy) outweigh those concerns, I can make the change in TaxData and this PR.

andersonfrailey:
>I can't seem to find any code in #1719 that does "deadweight loss" calculations on the benefits.
Did I missing that code?

>If I didn't miss the code, can you remove these variables from #1719?
Doing this kind of calculation will be involved (and probably controversial), so let's wait until we get there.

The calculations are done in in the new `BenefitPrograms` function. Specifically these lines of code:
```python
# use welfare multiples to determine percieved benefit amounts
ssi_ben_welfare = ssi_ben * BEN_SSI_mult
snap_ben_welfare = snap_ben * BEN_SNAP_mult
vet_ben_welfare = vet_ben * BEN_Vet_mult
mcare_ben_welfare = mcare_ben * BEN_Medicare_mult
mcaid_ben_welfare = mcaid_ben * BEN_Medicaid_mult
ss_ben_welfare = ss_ben * BEN_SS_mult
```

As @MattHJensen said, this is related to work we did in the UBI paper.

I completely agree that making deadweight loss assumptions can be complicated and controversial. I designed this PR so that unless the user specifies a change to the `BEN_xx_mult` variables that are found in `current_law_policy.py`, `xx_ben` will equal `xx_ben_welfare` by setting the default values  of `BEN_xx_mult` to one.

If this is still seen as something that should be handled as a later enhancement I will remove it from this PR, but I'm not sure what a better method would be for allowing a user to conduct deadweight loss calculations using just Tax-Calculator capabilities.

martinholmer:
@andersonfrailey said:
> `ss_ben` is a carry-over from when we were first working on benefit related projects, and in C-TAM the scripts used to create the benefit imputations all use "SS" as a starting point. My only objection to using "oasdi" rather than "ss" is breaking that connection. 

Just because a poor choice was made at the being of the C-TAM work, does not mean that we have to propogate that into Tax-Calculator.  Just change it in the final preparation script that produces the file that is going to be used by Tax-Calculator.  I'm not asking you to change it everywhere in taxdata and C-TAM.

> But, if the benefits for Tax-Calculator (such as name clarity/accuracy) outweigh those concerns, I can make the change in TaxData and this PR.

Thanks for making the name change in the `cps_benefits.csv.gz` file that will be part of the Tax-Calculator repository.

andersonfrailey:
>My question is why the long name for this file?
Why cps_benefits_extrap.csv.gz?
Why not simply cps_benefits.csv.gz?

`cps_benefits_extrap.csv.gz` was what the file was named in the PR to include the extrapolation routine in TaxData. I'll open a PR there to shorten it and update this one accordingly.

martinholmer:
@andersonfrailey said:
> The [deadweight loss] calculations are done in the new BenefitPrograms function. Specifically these lines of code:
```
# use welfare multiples to determine percieved benefit amounts
ssi_ben_welfare = ssi_ben * BEN_SSI_mult
snap_ben_welfare = snap_ben * BEN_SNAP_mult
vet_ben_welfare = vet_ben * BEN_Vet_mult
mcare_ben_welfare = mcare_ben * BEN_Medicare_mult
mcaid_ben_welfare = mcaid_ben * BEN_Medicaid_mult
ss_ben_welfare = ss_ben * BEN_SS_mult
```

But the six `BEN_*_mult` parameters are **not** policy parameters, so they don't belong in the `current_law_policy.json` file.

Why not remove them from this PR and approach deadweight loss calculations the way @MattHJensen suggested (as behavioral response parameters or as parameters in their own class) in [this comment](https://github.com/open-source-economics/Tax-Calculator/pull/1719#issuecomment-356285735)?

andersonfrailey:
@martinholmer, I think that's fair. I'll remove the welfare parameters from this PR and we can circle back and try @MattHJensen's suggestion as an enhancement later.

MattHJensen:
As a note for the future work, I don't believe the `current_law_policy.json` approach is accurate because it would reduce the value of the variables in the reform scenario, but not the baseline.

martinholmer:
@andersonfrailey,  You need to merge recent changes on the master branch into your addbenefitdata branch and then read the new version of [TESTING.md](https://github.com/open-source-economics/Tax-Calculator/blob/master/TESTING.md).   In particular, your pull request should not change the `docs/index.html` file to be different from what it is on the master branch.

andersonfrailey:
Last push is just to get everything up to date. Test failures will be fixed in future commits.

martinholmer:
@andersonfrailey said on 04-Dec-2017:
> This PR adds benefit data including C-TAM imputed benefits to the CPS file and a new file that contains extrapolated benefit data for the years 2015-2027.
>
> The new file `cps_benefits.csv.gz`, contains extrapolated benefits only for those who actually receive benefits. All of the zero values have been dropped to reduce file size.

A couple of questions:

1. Doesn't the `cps_benefits.csv.gz` file contain all the imputed benefits for 2014 through 2027?

2. Why does pull request #1719 have a new, larger `cps.csv.gz` file?  The 2014 benefits values are in the new `cps_benefits.csv.gz` file, right?

andersonfrailey:
@martinholmer asked:
>1. Doesn't the cps_benefits.csv.gz file contain all the imputed benefits for 2014 through 2027?

Yes. I actually need to go back and drop the 2014 values because they are repetitive. 
>2. Why does pull request #1719 have a new, larger cps.csv.gz file? The 2014 benefits values are in the new cps_benefits.csv.gz file, right?

In TaxData PR [#120](https://github.com/open-source-economics/taxdata/pull/120) we discussed whether the 2014 benefit data should be contained in `cps.csv.gz` or `cps_benefits.csv.gz` and seemed to be in general agreement that they should be in `cps.csv.gz` and then all extrapolated values be in `cps_benefits.csv.gz`

andersonfrailey:
Latest commit replaces `cps_benefits.csv.gz` with a file that doesn't include the 2014 benefit values.

A notebook showing benefits throughout the years and a sample reform can be found [here](http://nbviewer.jupyter.org/github/andersonfrailey/Notebook-Uploads/blob/master/TaxCalc%20Benefit%20Functionality.ipynb).

I'm removing the "DO NOT MERGE" label from the title of this PR. I believe it is ready to go upon review and test passage.

martinholmer:
@andersonfrailey said:
> @martinholmer asked:
>
>> Doesn't the `cps_benefits.csv.gz` file contain all the imputed benefits for 2014 through 2027?
>
> Yes. I actually need to go back and drop the 2014 values because they are repetitive.
>
>> Why does pull request #1719 have a new, larger `cps.csv.gz` file? The 2014 benefits values are in the new `cps_benefits.csv.gz` file, right?
>
> In TaxData PR #120 we discussed whether the 2014 benefit data should be contained in `cps.csv.gz` or `cps_benefits.csv.gz` and seemed to be in general agreement that they should be in `cps.csv.gz` and then all extrapolated values be in `cps_benefits.csv.gz`.

I think given the current situation, in which only non-zero benefits are being included in CSV files, the "general agreement" in taxdata 120 is a big mistake.  Why change the `cps.csv.gz` file just to include a lot of zeros?  It just inflates the file size unnecessarily and makes the code in Tax-Calculator more complex.

Much cleaner (and smaller file sizes) to leave the `cps.csv.gz` file as it is on the Tax-Calculator master branch and leave the `cps_benefits.csv.gz` file as it is in Tax-Calculator pull request #1719.

I have many other comments from my initial review of #1719, but those can wait until we get this cleared up.

andersonfrailey:
>I think given the current situation, in which only non-zero benefits are being included in CSV files, the "general agreement" in taxdata 120 is a big mistake. Why change the cps.csv.gz file just to include a lot of zeros? It just inflates the file size unnecessarily and makes the code in Tax-Calculator more complex.
Much cleaner (and smaller file sizes) to leave the cps.csv.gz file as it is on the Tax-Calculator master branch and leave the cps_benefits.csv.gz file as it is in Tax-Calculator pull request #1719.

The overall file size of course increases more than it would if removed all of the zero values, but I don't see how including the benefits on `cps.csv.gz` complicates Tax-Calculator code. If we keep all of the benefits separate, we'll need to set the initial benefit variables outside of `_read_data`, where every other variable is initially set, because they're read in afterwards by necessity. This would only take a few lines of code, but it isn't any more simple than leaving the initial benefits on `cps.csv.gz`.

I would also like to hear more from @MattHJensen about this comment in TaxData PR [#120](https://github.com/open-source-economics/taxdata/pull/120):

>Separating the base benefits data from cps.csv would significantly complicate the idea of a 'compatible_data' field in current_law_policy.json as some parameters would require multiple datasets.

Does that mean we would need to add a third option for compatible data for cps_benefits? Or something else?

martinholmer:
@andersonfrailey,  If you are adding benefits data to `cps.csv.gz`, can you follow the same rounding to interger convention that @hdoupe used in preparing the `cps_benefits.csv.gz` file?  The excessive precision is absurd and the file is unnecessarily large.  Here's some examples of benefit amounts measured to the **one ten-thousandth of a cent**:
```
cpsdata$ ./csvvars cps.csv | grep _ben
24 ssi_ben
25 vet_ben
26 snap_ben
27 mcare_ben
28 mcaid_ben
29 oasdi_ben
64 other_ben

cpsdata$ awk -F, 'NR>1&&$24>m{m=$24}END{print m}' cps.csv
64378.754082
iMac2:cpsdata mrh$ awk -F, 'NR>1&&$25>m{m=$25}END{print m}' cps.csv
169920.56968
iMac2:cpsdata mrh$ awk -F, 'NR>1&&$26>m{m=$26}END{print m}' cps.csv
26569.33018
iMac2:cpsdata mrh$ awk -F, 'NR>1&&$27>m{m=$27}END{print m}' cps.csv
691961.07909
iMac2:cpsdata mrh$ awk -F, 'NR>1&&$28>m{m=$28}END{print m}' cps.csv
692753.71683

```

martinholmer:
@andersonfrailey,  Here are some questions I had after reviewing the proposed code changes in #1719.

1. Why isn't it `CPS_BENEFITS_FILENAME`, which would make it similar to all the other filenames in records.py?
`BENEFIT_FILENAME` seems to suggest there is only one benefit and doesn't clarify that they are associated only with CPS data.

2. Do you have a question about how to revert `doc/index.html` back to what it is on the master branch?

3. In policy.py, why use benefit types that are inconsistent with all the other naming?  This is what I'm looking at:
```
        # benefittype-indexed suffixes and list index numbers
         'ssi': 0,
         'snap': 1,
         'vet': 2,
         'medicare': 3,          <=== WHY NOT mcare
         'medicaid': 4,          <=== WHY NOT mcaid
         'social security': 5,   <=== WHY NOT oasdi
         'other': 6
```
4. In records_variables.json, why the confusing "form" description.  For example, here is the mcare_ben entry:
```
"mcare_ben": {
       "type": "float",
       "desc": "Imputed Medicare benefits",
       "form": {"2013-2016": "imputed using the C-TAM model"},
       "availability": "taxdata_cps"
 },
```
Shouldn't it say "2014-2027": "imputed ..."?
There are no 2013 benefits data, right?  And we have imputed out to 2027, right?

5. Also in records_variables.json, why did you remove the "form" info for the "aftertax_income" entry?
```
    "aftertax_income": {
        "type": "float",
        "desc": "After tax income is equal to expanded_income minus combined",
 -      "form": {"2013-20??": "calculated variable"}
 +      "form": {}
 +    },
 +    "ben_total": {
 +      "type": "float",
 +      "desc": "Total benefits recieved by that tax unit",
 +      "form": {}
      }
```
Matt asked me to add that "form" field info for all the calculated variables.  Please restore it for `aftertax_income` and use it (instead of `{}`) in the `ben_total` entry.

6. And finally in records_variables.json, can you clarify the information for `other_ben`?  Here is what you have now:
```
"other_ben": {
       "type": "float",
       "desc": "Non-imputed benefits",
       "form": {"2013-2016": "determined using government benefit program data"},
       "availability": "taxdata_cps"
 },
```
Shouldn't it say we have only 2014 data and the 2014 values are extrapolated using some growfactor?  If the benefits are not imputed where do they come from?  The wording does not need to be lengthy but it needs to be less vague.

7. In records.py, do you think the logic in the `_read_benefits` method, which both reads and merges the `cps_benefit.csv.gz` data, will work when we use a sample of the CPS records?  It seems to me it will not work, but maybe I'm missing something.   Maybe you should add a unit test to see if works correctly when using a CPS sample.

8. And also in records.py, have you forgotten to extrapolate the `other_ben` variable?  Or did I miss where you are doing that?

andersonfrailey:
@martinholmer,

> If you are adding benefits data to cps.csv.gz, can you follow the same rounding to interger convention that @hdoupe used in preparing the cps_benefits.csv.gz file?

Yep. I'll make this change here and in TaxData.

>Why isn't it CPS_BENEFITS_FILENAME, which would make it similar to all the other filenames in records.py?
BENEFIT_FILENAME seems to suggest there is only one benefit and doesn't clarify that they are associated only with CPS data.

Good point. Will be fixed in the next commit.

>Do you have a question about how to revert doc/index.html back to what it is on the master branch

Yes. I updated my branch to account for changes in master, and then I re-ran make_index.py  to include the "Benefits" section in "Policy Parameters." Can you think of why some of the other information has changed? Do I need to revert index.html again and then try again?

>In policy.py, why use benefit types that are inconsistent with all the other naming? 

Good catch. Fixing in next commit.

>In records_variables.json, why the confusing "form" description. For example, here is the mcare_ben entry:
Shouldn't it say "2014-2027": "imputed ..."?
There are no 2013 benefits data, right? And we have imputed out to 2027, right?

Correct on both ends. Fixed in next commit.

>Also in records_variables.json, why did you remove the "form" info for the "aftertax_income" entry?

I believe that happened while I was fixing merge commits. Fixed in next commit.

>And finally in records_variables.json, can you clarify the information for other_ben? Here is what you have now:

I'll update the information in the next commit.

>In records.py, do you think the logic in the _read_benefits method, which both reads and merges the cps_benefit.csv.gz data, will work when we use a sample of the CPS records? It seems to me it will not work, but maybe I'm missing something. Maybe you should add a unit test to see if works correctly when using a CPS sample.

I believe that it should work. In `_read_benefits`, I take the RECID's for every record that has been provided. So when just a sample is used, the RECID data frame I create will only have that subsection of RECID's. I then use a left merge on the RECID data frame and the benefits data frame. A left merge only uses the keys of the left data frame (the RECID data frame in this case) so only benefits for those records should be included. Here's an example

```python
In [2]: a = pd.DataFrame({'id': [i for i in range(1, 6)]})

In [3]: b = pd.DataFrame({'id': [i for i in range(1, 11)],
   ...:                   'val': [i * 10 for i in range(1, 11)]})

In [4]: c = a.merge(b, on='id', how='left')

In [5]: c
Out[5]: 
   id  val
0   1   10
1   2   20
2   3   30
3   4   40
4   5   50
```
I have one assert statement in there to make sure the length of the RECID data frame is the same as the length of the `BEN` attribute at the end of the function. Do you think adding a second one to make sure only the RECID's in the sample are included in the `BEN` attribute would be a sufficient test?

> And also in records.py, have you forgotten to extrapolate the other_ben variable? Or did I miss where you are doing that?

Thanks for bringing this up. I wanted to open a discussion about this but it slipped my mind. @Amy-Xu suggested going back to find the average rate that government outlays on benefits had grown over the past 4 or 5 years and using that to extrapolate `other_ben`. I'd love to hear other people's opinion on this and what the best approach would be.

martinholmer:
@andersonfrailey said:
>> Do you have a question about how to revert doc/index.html back to what it is on the master branch
>
> Yes. I updated my branch to account for changes in master, and then I re-ran make_index.py to include the "Benefits" section in "Policy Parameters." Can you think of why some of the other information has changed? Do I need to revert index.html again and then try again?

Try this to fix #1719:
```
git checkout master
cp docs/index.html docs/index.html-master
git checkout addbenefitdata
mv docs/index.html-master docs/index.html
git commit -a -m "Revert docs/index.html"
git push origin addbenefitdata
```
Going forward, read carefully the new [TESTING procedures](https://github.com/open-source-economics/Tax-Calculator/blob/master/TESTING.md).

martinholmer:
@andersonfrailey said:
>> In records.py, do you think the logic in the _read_benefits method, which both reads and merges the cps_benefit.csv.gz data, will work when we use a sample of the CPS records? It seems to me it will not work, but maybe I'm missing something. Maybe you should add a unit test to see if works correctly when using a CPS sample.
>
> I believe that it should work. In _read_benefits, I take the RECID's for every record that has been provided. ...

Yes, you are correct.  When I asked this question I was confused about where the CPS sample is drawn: its before the Records class constructor is called, not in the midst of the constructor (which is where the weights are adjusted for the sample).

martinholmer:
@andersonfrailey said:
>> And also in records.py, have you forgotten to extrapolate the other_ben variable? Or did I miss where you are doing that?
>
> Thanks for bringing this up. I wanted to open a discussion about this but it slipped my mind. @Amy-Xu suggested going back to find the average rate that government outlays on benefits had grown over the past 4 or 5 years and using that to extrapolate other_ben. I'd love to hear other people's opinion on this and what the best approach would be.

But `other_ben` does not include the big six (ssi, oasdi, snap, vet, mcaid, mcare).

Exactly which benefits are included in `other_ben`?   You need to list them in the `records_variables.json` documentation.

@Amy-Xu's suggestion sounds like a lot of work for not much benefit.  So what if the `other_ben` amount is not extrapolated quite right?  Aren't the `other_ben` amounts very small relative to the total of the six benefits?

There are tradeoffs here and so you are right that there should be a discussion of this matter.

martinholmer:
@andersonfrailey said:
>> Why isn't it CPS_BENEFITS_FILENAME, which would make it similar to all the other filenames in records.py?
BENEFIT_FILENAME seems to suggest there is only one benefit and doesn't clarify that they are associated only with CPS data.
>
> Good point. Will be fixed in the next commit.

But you forgot the S at the end of BENEFIT.

andersonfrailey:
>Going forward, read carefully the new TESTING procedures.

Thanks for your help. I must've missed the update to the new testing procedures.

`other_ben` includes everything in [this CSV file](https://github.com/andersonfrailey/taxdata/blob/dd99321bee7c51ab3e4b9e90855eb5c0d01b176a/cps_data/benefitprograms.csv) that will be added to the TaxData repo. They were pulled from tables 3.2 and 11.3 of these [OMB documents](https://obamawhitehouse.archives.gov/omb/budget/Historicals). The sum of all the listed benefits was then distributed among the CPS based on the distribution of  Medicaid, SSI, SNAP, and Veteran's Benefits, as @MattHJensen suggested when we were working on the UBI paper last spring/summer. This is all done during final prep in TaxData. If I added each variable that `other_ben` included, wouldn't the result be a bunch of zero arrays in Tax-Calculator? Would it be more prudent to simply provide a link to the CSV containing each of the individual programs in the description of `other_ben`?

martinholmer:
@andersonfrailey said:
> `other_ben` includes everything in [this CSV file](https://github.com/andersonfrailey/taxdata/blob/dd99321bee7c51ab3e4b9e90855eb5c0d01b176a/cps_data/benefitprograms.csv) that will be added to the TaxData repo. They were pulled from tables 3.2 and 11.3 of these [OMB documents](https://obamawhitehouse.archives.gov/omb/budget/Historicals). The sum of all the listed benefits was then distributed among the CPS based on the distribution of  Medicaid, SSI, SNAP, and Veteran's Benefits, as @MattHJensen suggested when we were working on the UBI paper last spring/summer. This is all done during final prep in TaxData. If I added each variable that `other_ben` included, wouldn't the result be a bunch of zero arrays in Tax-Calculator? Would it be more prudent to simply provide a link to the CSV containing each of the individual programs in the description of `other_ben`?

I'm not asking for each of the 34 other benefits to be separately included in the data file.  
What I am asking for is better documentation of what is in the `other_ben` variable.

Providing the URL of [this CSV file](https://github.com/andersonfrailey/taxdata/blob/dd99321bee7c51ab3e4b9e90855eb5c0d01b176a/cps_data/benefitprograms.csv) would be a good thing to add to the `records_variables.json` documentation for the `other_ben` variable.

But now that the secrecy over the contents of the `other_ben` variable has been lifted, I have a few substantive questions that all relate to possible double counting of benefits.  Here are my questions:

1. Isn't item 1 "Railroad retirement (excl. social security)" already included in `e02400`?

2. Isn't item 2 "Unemployment Assistance" exactly the same as `e02300`?

3. What exactly is item 34 "Earned income tax credit"?  If it is the federal EITC then, why is it included in the list when we already calculate the federal EITC in Tax-Calculator?  If this variable includes only **state* EITC benefits, then you need to fix the CSV file to emphasize that item 34 includes only **state** EITC benefits.

Why do we want to get into double counting?  If there is a cogent reason, you need to add it to the documentation.  If there is not a cogent reason, you need to drop some items from the `other_ben` variable.

martinholmer:
@andersonfrailey,  I have more questions/suggestions about the `taxdata/cps_data/benefitprograms.csv` file.

1. The file has two fields "Program" and "Cost", but what are the units and year of the Cost?  Is Cost expressed as billions of dollars per annum?  For which year(s)?

2. While adding this file to the taxdata repo will be helpful, why not make it good documentation by adding three new fields?  Here is is the extra information that would be helpful:
   a. `table`, URL of the OMB table from which the Cost is taken.
   b. `table_line`, table line number or line label.
   c. `table_year`, the year column from which the Cost number is taken.

If this info is not added, users are not going to have any idea where the `other_ben` number comes from.
If you don't think anybody will ask about this, you should think again.  My bet is that Max Ghenis will ask about this because he has been watching the UBI/benefits-related features of Tax-Calculator like a hawk.

feenberg:
On Fri, 19 Jan 2018, Martin Holmer wrote:

>
>  2.
>
>     While adding this file to the taxdata repo will be helpful, why not make
>     it good documentation by adding three new fields? Here is is the extra
>     information that would be helpful:
>     a. table, URL of the OMB table from which the Cost is taken.
>     b. table_line, table line number or line label.
>     c. table_year, the year column from which the Cost number is taken.

Labeling the source of numbers is a great service. Are these tables that 
can change under the same URL? If so, we might prefer to link to the 
document at:

     https://archive-it.org/

if it was available there.

dan

andersonfrailey:
>Isn't item 1 "Railroad retirement (excl. social security)" already included in e02400?

Looking at page 12 of the [1040 instructions](https://www.irs.gov/pub/irs-pdf/i1040gi.pdf), you are correct. I'll remove those from the file.

>Isn't item 2 "Unemployment Assistance" exactly the same as e02300?

This I'm not sure about. When I look at e02300 in the PUF for 2014, I get $34 billion, but here we have $43 billion. It's possible that the discrepancy comes from non-filers and we're under-reporting unemployment compensation in the PUF, but I don't know how we could prove this.

>What exactly is item 34 "Earned income tax credit"? If it is the federal EITC then, why is it included in the list when we already calculate the federal EITC in Tax-Calculator? If this variable includes only *state EITC benefits, then you need to fix the CSV file to emphasize that item 34 includes only state EITC benefits.

You are correct that it is federal.

>The file has two fields "Program" and "Cost", but what are the units and year of the Cost? Is Cost expressed as billions of dollars per annum? For which year(s)?

I'll clarify that it is cost in millions for 2014 in the CSV file.

> While adding this file to the taxdata repo will be helpful, why not make it good documentation by adding three new fields? Here is is the extra information that would be helpful:
a. table, URL of the OMB table from which the Cost is taken.
b. table_line, table line number or line label.
c. table_year, the year column from which the Cost number is taken.

I can add all of those columns.

@feenberg, all of the tables are found on the archived version of the Obama White House website so the data linked shouldn't change under that URL.

I also had an idea about extrapolating `other_ben`. The tables we pulled the 2014 number from also include projections out to 2021. What if I created a new growth factor that uses the change in sum of the programs we've selected to determine growth rates for 2015 to 2021 and then we assumed it stays at 2021 levels until 2027?

andersonfrailey:
I talked with @MattHJensen about devectorizing the benefit repeal variable and also using `_repeal` rather than `_switch`. Latest commit takes care of that. There are still some failing tests that I'll be fixing soon.

andersonfrailey:
Latest commit updates this PR to include all of the changes in Tax-Calculator now that @martinholmer has merged #1853, #1854, and #1855, which implemented changes that were originally part of this PR. 

Most of the features in this PR have been discussed throughout this thread. Two recent additions to this PR are an additional growth factor to be used with `other_ben` and including unemployment assistance (`e02300`) to the list of benefit programs that can be repealed.

martinholmer:
@andersonfrailey,  I don't believe the placement of the `BenefitPrograms` function is correct.
That function needs to be at the **very start** of the tax-calculation logic because we are possibly setting `e02300` and `e02400` equal to zero, which will affect the calculation of income taxes.  I guess its OK to tally the `total_benefit` variable as part of that function, right?

andersonfrailey:
>@andersonfrailey, why did you add these lines?

Ah sorry. I was playing with the tests and forgot to delete that before pushing. Thanks for catching it.

>That function needs to be at the very start of the tax-calculation logic because we are possibly setting e02300 and e02400 equal to zero, which will affect the calculation of income taxes.

You're right. I'll move it up

>I guess its OK to tally the total_benefit variable as part of that function, right?

Are you referring to `ben_total` (going to change that to `benefit_total` for clarity) here? If so then yes we can determine it first still.

martinholmer:
@andersonfrailey,  Would you say PR #1719 is ready for a (hopefully) final review?
If so, I'll try to download the PR and do a complete set of tests on my computer, just so we can say that all the tests passed on two different local computers.

andersonfrailey:
@martinholmer asked:
>Would you say PR #1719 is ready for a (hopefully) final review?

Yes, it's ready for final review.

martinholmer:
@andersonfrailey,  Thanks for the addition of the benefits extrapolation data and logic and the benefits repeal logic.  I appreciate your patience with all my questions.  Good work!  Merging now.

Issue #1720, opened 2017-12-04
Finishing the object-oriented design of Tax-Calculator


martinholmer:
In pull request #1718, @MattHJensen asked:
> For the sake of developing a shared understanding of design objectives, could you describe the benefits you see of adopting these changes?  For instance, what is the value of "mov[ing] towards a more object oriented structure"?

Tax-Calculator has from the very beginning been developed using many of the principles of object-oriented programming (OOP).  But there are a few areas in which the current code does not follow OOP principles.  In recent weeks there has been a number of incidents that indicate a need to make Tax-Calculator comply more completely with OOP design principles.  In this issue, I describe the recent incidents and outline the changes that need to be made to complete the original OOP design of Tax-Calculator.

If you are not familiar with OOP, then you should do some background reading before continuing the read this issue.  A brief search of the Internet for "OOP in Python" produced [this document](https://www.python-course.eu/object_oriented_programming.php) as a reasonable first background document.

Tax-Calculator uses classes, class methods, data encapsulation, and even class inheritance (because the Policy, Behavior, Consumption, and Growdiff classes are all derived from the ParametersBase class), which are key concepts in OOP.  However, Tax-Calculator is weak on data hiding.   

Here is an excerpt from [another tutorial](https://www.python-course.eu/python3_object_oriented_programming.php) on OOP in Python:
> Data Abstraction, Data Encapsulation and Information Hiding are often synonymously used in books and tutorials on OOP. But there is a difference. Encapsulation is seen as the bundling of data with the methods that operate on that data. Information hiding on the other hand is the principle that some internal information or data is "hidden", so that it can't be accidentally changed. Data encapsulation via methods doesn't necessarily mean that the data is hidden. You might be capable of accessing and seeing the data anyway, but using the methods is recommended. Finally, data abstraction is present, if both data hiding and data encapsulation is used. This means data abstraction is the broader term: 
>
> Data Abstraction = Data Encapsulation + Data Hiding  

The OOP design principle of data abstraction has not been achieved in Tax-Calculator because we have no data hiding.  For additional discussion on the preferred approach to implementing data hiding in Python, see [this tutorial](https://www.python-course.eu/python3_properties.php).

And the current version of Tax-Calculator also deviates from OOP design principles when it defines many high-level utility functions that should be methods of the class the utility function is manipulating.

[...]

PR #1722, opened 2017-12-04
Add Calculator class graph methods and make graph utilities low-level

martinholmer:
This pull request adds three graph methods to the Calculator class, leaving all the low-level details in utility functions.  This pull request is part of the process of finishing the object-oriented design of Tax-Calculator described in issue #1720.


hdoupe:
@martinholmer What do you think about splitting this Calculator class that you are building into two classes: `Calculator` and `CalculatorFormatter`?  If we keep the calculating methods and the formatting methods together in one class, I think that over time this could grow into an overly large and unruly class.

I'm not sure what the exact relation between these two classes should be.  Perhaps, Calculator should inherit methods such as `distribution_tables`.

martinholmer:
@hdoupe said in #1722:
> What do you think about splitting this Calculator class that you are building into two classes: Calculator and CalculatorFormatter? If we keep the calculating methods and the formatting methods together in one class, I think that over time this could grow into an overly large and unruly class.
>
> I'm not sure what the exact relation between these two classes should be. Perhaps, Calculator should inherit methods such as distribution_tables.

I think this is worth thinking about, but I see it as a possible follow-on to the current work.  The current changes are already very extensive.  I don't think we want to make those changes even more extensive in this set of pull requests.   Thanks for the suggestion; let's keep it in mind.

hdoupe:
@martinholmer Ok, that makes sense to me. No problem. Thanks for the thoughtful response.

PR #1725, opened 2017-12-06
Reform File for Final Senate TCJA Passed 12-01-2017

evtedeschi3:
Adds a file containing parameters in the final Senate TCJA.


codykallen:
@evtedeschi3, could you fix the indentation on `_DependentCredit_before_CTC` and move it down to the other `DependentCredit_*` parameters? And could you update the `Reform_Description` and `Reform_Parameter_Map` to reflect that the additional standard deduction for the aged and blind (`_STD_Aged`) isn't being repealed and that the AMT threshold is raised instead of being repealed? Also, please add the CPI offset reform to the `Reform_Description` and `Reform_Parameter_Map`. 

This may go beyond just your work here, but we need to modify the `Reform_Parameter_Map` to replace `_ID_rt*` with `_ID_*` and `_PT_rt*` with `_PT_*`. 

For the AMT thresholds, married couples filing separately would have an AMT threshold of $54,700, not $70,300, and their phase-out threshold would be $104,200, not $156,300. 

For the `_ID_RealEstate_c` term, the cap for married couples filing separately should be $5,000, not $10,000. 

Finally, I think we should (debatable) eliminate the itemized deduction for personal casualty losses. The bill actually limits the deduction only to areas subject to a federally declared disaster, and it eliminates the deductions of casualty losses due to all other reasons (fire, storm, theft, etc.).

codykallen:
@evtedeschi3, this looks good to me.

MattHJensen:
This looks good to me. I'll merge as soon as @GoFroggyRun gives a thumbs up. 

Separately, if someone has the opportunity/inclination, reconstructing this chart from Rosenberg and Burman could be a good check on the AMT.

image

http://www.taxpolicycenter.org/taxvox/senate-would-keep-individual-amt-turn-it-very-different-tax

GoFroggyRun:
@evtedeschi3 the only question I have is how `2026` values are obtained for `_AMT_em` and `_AMT_em_ps`? 

When I set `_cpi_offset` to `-0.0025` in 2017, I got the following values for the two parameters in 2026:
```
_AMT_em: [  65461.84,  101869.72,   50934.86,   65461.84,  101869.72];
_AMT_em_ps: [ 145510.95,  193974.4 ,   96987.2 ,  145510.95,  193974.4 ],
```
which seem to be different from yours.

evtedeschi3:
Let’s take the single phase-out threshold as an example: $120,700 in 2017 under current law.

Using ACPIU minus 0.0025 to grow this beginning in 2018, I get $145,731 in 2026.

To be clear, I’m using the same-year ACPIU, so e.g. the 2018 level is based on 2017 times ( ACPIU[2018] - 0.0025). If the timing is off here let me know.

martinholmer:
@evtedeschi3 said in pull request #1725:
> To be clear, I’m using the same-year ACPIU, so e.g. the 2018 level is based on [the] 2017 [level] times ( ACPIU[2018] - 0.0025). If the timing is off here let me know.

For better or worse, the way Tax-Calculator has always done parameter indexing is not like that.
See [this comment](https://github.com/open-source-economics/Tax-Calculator/issues/1687#issuecomment-345720640) from several weeks ago about what is going on under the hood.

evtedeschi3:
I was afraid of that. Will update using Sean’s values.

MattHJensen:
@evtedeschi3, thanks very much for this contribution to Tax-Calculator! 

@GoFroggyRun and @codykallen, thanks for your review. 

I just confirmed that the latest commit satisfies @GoFroggyRun's suggestion and am merging this PR now.

Issue #1737, opened 2017-12-09
TaxBrain: Tax liabilities uniformly *increase* when adding a basic income


MaxGhenis:
I tried a basic reform in the TaxBrain GUI in which the only change is entering $2,000 in the "UBI benefit for those under 18" box: http://www.ospc.org/taxbrain/25246/

This produced a tax liability increase for 40,697 tax units, and a tax cut for 0 tax units. Am I doing something wrong?

The "All" row also shows 0 for "Percent with Tax Increase", despite the 40k in "Tax Units with Tax Increase". The percent is calculated correctly for each income bin.

martinholmer:
@MaxGhenis said:
> I tried a basic reform in the TaxBrain GUI in which the only change is entering $2,000 in the "UBI benefit for those under 18" box: http://www.ospc.org/taxbrain/25246/
>
> This produced a tax liability increase for 40,697 tax units, and a tax cut for 0 tax units. Am I doing something wrong?

If you are surprised by these results, maybe you were thinking the UBI benefit would be non-taxable.
But as the following screen shot shows, you (perhaps inadvertently) assumed that **none** of the UBI benefit is excluded from AGI.

![screen shot 2017-12-08 at 9 11 01 pm](https://user-images.githubusercontent.com/12170745/33791664-587869a2-dc5d-11e7-9168-6e279d615557.png)

Do the UBI results correspond more to what you were expecting if you exclude all of the UBI benefits from AGI?

@MaxGhenis also said:
> The "All" row also shows 0 for "Percent with Tax Increase", despite the 40k in "Tax Units with Tax Increase". The percent is calculated correctly for each income bin.

Yes, this is a known bug that we are working on.  In fact, three columns in the difference table have `All` row values that are incorrect: "Percent with Tax Cut", "Percent with Tax Increase", and "Average Tax Change".
Hopefully, these three incorrect `All` row values will be correct in the next version of Tax-Calculator and TaxBrain.

Thanks for your valuable feedback on using Tax-Calculator and TaxBrain.

MaxGhenis:
I see, I thought tax liability would be net tax liability, with UBI being a negative liability. For example, is this not how TaxBrain communicates EITC refunds?

MaxGhenis:
[...]

PR #1741, opened 2017-12-11
Add calculation of ALL values for perc_cut and perc_inc in difference table

martinholmer:
This pull request completes the process of calculating every one of the values in the difference table.  Since the beginning of the project, Tax-Calculator had left several values on the ALL (whole-sample) table row unspecified, setting them to NaN.  Now every value in the difference table is calculated in Tax-Calculator and all TaxBrain needs to do is display the calculated values.

This pull request resolves issue #1737 on the Tax-Calculator side, but there needs to be difference-table enhancements on the TaxBrain side, as described in OpenSourcePolicyCenter/PolicyBrain#73 and OpenSourcePolicyCenter/PolicyBrain#770.


martinholmer:
@hdoupe,  Thanks for testing this pull request with the development version of TaxBrain.

PR #136, opened 2015-02-21
separate description attr into description, irs_ref, and notes.

MattHJensen:
Deals with https://github.com/OpenSourcePolicyCenter/Tax-Calculator/issues/115#issuecomment-75260898

Ready to merge upon review by @Amy-Xu

PR #1750, opened 2017-12-12
Add FHEADIDX to CPS

andersonfrailey:
This PR complements TaxData PR [#136](https://github.com/open-source-economics/taxdata/pull/136). It adds the `fheadidx` variable to the CPS file, as requested by @evtedeschi3 and @MattHJensen.


martinholmer:
@andersonfrailey,  I'm not very knowledgeable about the CPS, but shouldn't this description
```
"desc": "Index to persons record of family reference person",
```
be like this?
```
"desc": "Index to person record of family reference person",
```
It's the family reference person's `person record` your pointing to, right?

andersonfrailey:
@martinholmer, you're correct that it is pointing to the person record of the family reference person. The CPS documentation uses both "index to persons record of family reference person" and "index to person record of family head." I'm indifferent to which we use, but I think we should stick to one of the conventions used in the official CPS documentation.

martinholmer:
@andersonfrailey said:
> you're correct that it is pointing to the person record of the family reference person. The CPS documentation uses both "index to persons record of family reference person" and "index to person record of family head." I'm indifferent to which we use, but I think we should stick to one of the conventions used in the official CPS documentation.

It seems to me that "index to person record of family head" is better English.  It's a record for a person, not a record for persons (that's the family record, right?).

andersonfrailey:
@martinholmer, I agree that it's better English. The CPS documentation uses both "person" and "persons" so I don't think which one we choose will make much of a difference. I can change the spelling in my next commit.

MattHJensen:
@evtedeschi3, does this look like what you need?

evtedeschi3:
I’m still not clear on why PULINENO isn’t possible, but if this is the most detail available it will certainly help with CPS matching.

martinholmer:
@andersonfrailey, Why are you adding a variable that was not requested?

andersonfrailey:
`PULINEO` is not listed in the [CPS documentation](http://www.nber.org/cps/cpsmar2014.pdf).

@evtedeschi3 is it possible that `A_LINENO` is the variable you're looking for? It's listed under the person record variables as simply described as "line number". I looked at the raw data and it ranges from 1 to 16. It's possible that it refers to what line in the family or household a record is from.

martinholmer:
There has been an upgrade in both the CPS file and the PUF file used as Tax-Calculator input data since this pull request was opened.  So, I'm assuming that the initiative in #1750 is now moot.  If not, then a new pull request can be opened.

Issue #1755, opened 2017-12-14
Adding a choice between the SALT Real Estate and Income tax deductions


evtedeschi3:
Based on media reports, there's a possibility that the final TCJA conference tax bill will modify the SALT treatment so that filers will first have to choose between either the real estate or income / sales tax deductions, and then face a cap (e.g. $10K) on their choice.

The way I was mulling over implementing this in Tax-Calculator would be to have the user specify the cap through both `ID_RealEstate_c` and `ID_StateLocalTax_c` (presumably these would be identical under a conference-style SALT compromise), and then add a boolean switch called e.g. `ID_SALT_Limited` or similar to control the calculation of `c18300`. So in the ItemDed function:

```
    c18400 = min((1. - ID_StateLocalTax_hc) * max(e18400_capped, 0.),
                 ID_StateLocalTax_c[MARS - 1])
    c18500 = min((1. - ID_RealEstate_hc) * e18500_capped,
                 ID_RealEstate_c[MARS - 1])
    if ID_SALT_Limited:
    	c18300 = max(c18400, c18500)
    else:
    	c18300 = c18400 + c18500
```

But I wanted to raise this with folks more familiar with Tax-Calculator's structure to see both whether this is conceptually sound and what other changes would need to be made.

MattHJensen:
@evtedeschi3, do you know what the treatment would be if you have $5000 of one and $3000 of another? Would you get to deduct $8000 or $5000?

evtedeschi3:
My impression is that you have to choose one, so you’d get to deduction only $5000 in this case. But I’d emphasize that’s my impression talking with several folks who have only indirect familiarity with what’s being negotiated, and of course it could change.

MattHJensen:
Got it. The reason I ask is that if the taxpayer would get to deduct $8000 in that situation, then I wonder if adding a `ID_SALT_c` parameter (`c18300 = min(ID_SALT_c[MARS - 1], c18400 + c18500)`) would be more clear and would allow the user to change only one parameter rather than three.

MattHJensen:
[...]

PR #1756, opened 2017-12-15
Add cap on state and local income, sales and real estate tax deduction

codykallen:
This PR adds the ability to place a cap on the combined state and local income/sales and real estate deductions. This is consistent with the news descriptions of the reconciliation version of the TCJA, and the description from Kevin Brady as mentioned in #1755. 

Note that this simply caps the total amount of the deduction available.

### DIFF
--- a/taxcalc/current_law_policy.json
+++ b/taxcalc/current_law_policy.json
@@ -1207,12 +1207,11 @@
         "out_of_range_action": "stop"
     },
 
-
     "_ID_AllTaxes_c": {
         "long_name": "Ceiling on the amount of state and local income, sales and real estate tax deductions allowed (dollars)",
         "description": "The amount of state and local income, sales and real estate tax deductions is limited to this dollar amount.",
         "section_1": "Itemized Deductions",
-        "section_2": "State and Local, And Foreign Real Estate Taxes",
+        "section_2": "State And Local Income And Sales Taxes",
         "irs_ref": "",
         "notes": "",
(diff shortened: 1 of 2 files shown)


codykallen:
@MattHJensen, @evtedeschi3, this PR implements the cap on the total state and local income/sales/real estate tax deduction, as described in #1755. I think we should leave the WIP label on until the bill is actually released, but it would be useful to review it in advance so we can quickly merge the changes once the bill is released.

evtedeschi3:
As an implementation of the "and" approach -- filers allowed to allocate the $10K cap between property *and* S&L income / sales -- this makes a lot of sense. Thanks Cody!

MattHJensen:
Thanks for opening this PR @codykallen and thanks for the review @evtedeschi3! This also looks good to me. I'm inclined to merge this as it seems like a useful feature even outside of the context of TCJA. @codykallen, if you agree and remove the WIP label, I'll do that. It would also allow someone to open a WIP reform file PR (which I would wait to merge until the bill is released).

codykallen:
@MattHJensen, I've removed the WIP label. I'm inclined to wait before opening a WIP reform file PR because we don't actually have the final details of the bill, especially regarding the CTC.

MattHJensen:
> I'm inclined to wait before opening a WIP reform file PR because we don't actually have the final details of the bill, especially regarding the CTC.

Ok. We're also missing details on pay fors stills. Heard some talk they might move the sunset date forward a year...

PR #1618, opened 2017-11-02
Add filer credit

codykallen:
This PR adds a new non-refundable credit for the filer and their spouse as structured in the Tax Cuts and Jobs Act. The relevant portions of the act increase the child tax credit by $600, but this increased amount is nonrefundable. This is implemented with existing capabilities by setting `'_DependentCredit_Child_c': [600]`

It also creates a $300 nonrefundable credit for non-child dependents. This is implemented using existing capabilities by setting `'_DependentCredit_Nonchild_c': [300]`. It phases out as though part of the child tax credit. 

It also creates a $300 nonrefundable credit for the filer and their spouse, which phases out as though part of the child tax credit. This phase-out structure means that it cannot be properly implemented using the personal nonrefundable credit. This PR adds the capability to accurately model this, by setting `'FilerCredit_c': [[300, 600, 300, 300, 300]].

PR #1765, opened 2017-12-16
Update TCJA references to reconciliation version

MaxGhenis:
The reconciliation TCJA:
1) Raises _DependentCredit_Nonchild_c to $500, not $300 (https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/reforms/TCJA_Reconciliation.json#L131)
2) Does away with _FilerCredit_c


taxcalc/current_law_policy.json
```diff
@@ -3679,7 +3679,6 @@
         "section_1": "Nonrefundable Credits",
         "section_2": "Child Tax Credit",
         "irs_ref": "",
-        "notes": "The Tax Cuts and Jobs Act sets this at $600 for joint and $300 for others",
```
martinholmer: @MaxGhenis,  Can't remove the "notes", but can set to this: `"notes": "",`

martinholmer:
@codykallen,  Can you review PR #1765?  
In that PR @MaxGhenis is suggesting some updates to the "notes" for the `_DependentCredit_Nonchild_c` and `_FilerCredit_c` policy parameters that you added in early November as part of #1615 and #1618, respectively.

taxcalc/current_law_policy.json
MaxGhenis: Oops, fixed.

martinholmer: Thanks @MaxGhenis.  I've been pretty much out of the loop on the details of specifying the iterations of the TCJA reforms, so I think it is best to wait for Cody's review.  Thanks for your careful review of our work.

MaxGhenis:
Sure, it might make more sense to note either all TCJA changes or none, but
thought I'd make the minimal change for the previously noted ones.

MattHJensen:
The primary value of having these TCJA references in `current_law_policy.json` -- unless I'm missing something -- is that they give the user a contextual reference for the parameters that don't appear in current law. I don't think it would undermine this objective to remove the dollar amount and only cite the bill ("TCJA") in the notes section. It would also reduce the maintenance load.

martinholmer:
@MattHJensen said:
> The primary value of having these TCJA references in `current_law_policy.json` -- unless I'm missing something -- is that they give the user a contextual reference for the parameters that don't appear in current law. I don't think it would undermine this objective to remove the dollar amount and only cite the bill ("TCJA") in the notes section. It would also reduce the maintenance load.

@MaxGhenis,  These are important points.  How about dropping the dollar amounts and rewording the remaining "notes" fields to be something like `First introduced in TCJA bills`?  Saying these parameters are not part of current-law policy is risky because the probability is high that they will be part of current-law policy in a matter of days.

MaxGhenis:
> How about dropping the dollar amounts and rewording the remaining "notes" fields to be something like First introduced in TCJA bills?

Done, for the two that currently reference TCJA. 

I also filed #1777 as a longer-term feature request to understand the origin and evolution of each parameter.

martinholmer:
Thanks @MaxGhenis for the enhancement in pull request #1765.

Issue #1771, opened 2017-12-18
500 Internal Server Error when uploading TCJA_Reconciliation.json to TaxBrain


MaxGhenis:
I uploaded [TCJA_Reconciliation.json](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/reforms/TCJA_Reconciliation.json) to https://www.ospc.org/taxbrain/file/, which takes me to https://www.ospc.org/taxbrain/file/?start_year=2017 where it says

> 500 Internal Server Error!

This error arises both when providing and not providing a behavior file, and when choosing *Show me the results!* and *Quick calculation!*.

I don't get the error when uploading the older [TCJA_Senate.json](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/reforms/TCJA_Senate.json), though the difference table is all zeros (https://www.ospc.org/taxbrain/26994/).

MaxGhenis:
I see [here](https://github.com/open-source-economics/Tax-Calculator/issues/1767#issuecomment-352218735) that TCJA_Reconciliation requires Tax-Calculator version 0.14.1, while TaxBrain is on version 0.14.0 per my just-produced Senate result. 

Is there an issue for updating TaxBrain to 0.14.1 I can track?

If possible, a more informative error message would be nice here, e.g. saying the reform includes parameters not available in the current version.

martinholmer:
@MaxGhenis,  Thanks for your reports on problems using TaxBrain to analyze the conference committee's TCJA bill.  I think TaxBrain will soon be using Tax-Calculator release 0.14.1, which should eliminate the problems you're having.  You can follow the progress of TaxBrain at [this GitHub repository](https://github.com/OpenSourcePolicyCenter/PolicyBrain).

@MaxGhenis also said:
> If possible, a more informative error message would be nice here, e.g. saying the reform includes parameters not available in the current version.

Yes, that would be better than the uninformative `500 Internal Server Error!`.
I'll suggest that to the TaxBrain developers.  @hdoupe and/or @GoFroggyRun, can you open an enhancement issue on this topic?

martinholmer:
I'm going to close issue #1771 now that it is on the radar scope of the TaxBrain developers.

MaxGhenis:
[...]

PR #1775, opened 2017-12-19
Try to standardize format of Policy.implement_reform error messages

martinholmer:
This pull request responds to #1771, a bug report by @MaxGhenis, and to the analysis of that bug report by @hdoupe in [PolicyBrain issue 779](https://github.com/OpenSourcePolicyCenter/PolicyBrain/issues/779), by changing several ValueError messages to have the format expected by PolicyBrain.


martinholmer:
@hdoupe,  Do you have a spare moment to review Tax-Calculator pull request #1775?
I'd like to merge it so that the `implement_reform` message format is more standard.

hdoupe:
@martinholmer Thanks for quickly taking care of this.  This resolves the formatting issue in OpenSourcePolicyCenter/PolicyBrain#779.  However, when an invalid parameter is used, only that error is displayed.  If another parameter causes warnings or errors, then those messages may not be displayed.

For example:
```
import taxcalc

policy_dict = {
    u'policy': {
        2018: {
            u'_STD': [[500, 24000, 12000, 18000, 24000]],
        },
    },
    u'growdiff_response': {},
    u'consumption': {},
    u'behavior': {},
    u'growdiff_baseline': {}
}

ew  = taxcalc.tbi.reform_warnings_errors(policy_dict)

print("errors warnings for: ")
print(policy_dict)
print('\n')
print(ew)

policy_dict["policy"][2018]["_not_a_param"] = [1000.0]
ew = taxcalc.tbi.reform_warnings_errors(policy_dict)
print("\nerrors warnings for: ")
print(policy_dict)
print('\n')
print(ew)
```
output:
```
(aei_dropq) HDoupe-MacBook-Pro:PolicyBrain henrydoupe$ python tc_pr_1775.py 
errors warnings for: 
{u'policy': {2018: {u'_STD': [[500, 24000, 12000, 18000, 24000]]}}, u'growdiff_response': {}, u'behavior': {}, u'consumption': {}, u'growdiff_baseline': {}}


{'errors': '', 'warnings': 'WARNING: 2018 _STD_0 value 500.0 < min value 6392.88\nWARNING: 2019 _STD_0 value 511.2 < min value 6538.32\nWARNING: 2020 _STD_0 value 523.01 < min value 6691.67\nWARNING: 2021 _STD_0 value 535.46 < min value 6853.31\nWARNING: 2022 _STD_0 value 548.47 < min value 7022.28\nWARNING: 2023 _STD_0 value 561.47 < min value 7191.08\nWARNING: 2024 _STD_0 value 574.95 < min value 7366.07\nWARNING: 2025 _STD_0 value 588.63 < min value 7543.76\nWARNING: 2026 _STD_0 value 602.64 < min value 7725.68\nWARNING: 2027 _STD_0 value 617.1 < min value 7913.5\n'}

errors warnings for: 
{u'policy': {2018: {'_not_a_param': [1000.0], u'_STD': [[500, 24000, 12000, 18000, 24000]]}}, u'growdiff_response': {}, u'behavior': {}, u'consumption': {}, u'growdiff_baseline': {}}


{'errors': 'ERROR: 2018 _not_a_param unknown parameter name\n', 'warnings': ''}
```

I think it would be better to display all of the warnings and errors caused by a reform in one pass.  That way the user doesn't submit a reform, see the invalid parameter error, fix the error, and then see even more warnings/errors to resolve.  Does this seem sensible?

martinholmer:
@hdoupe said about pull request #1775:
> Thanks for quickly taking care of this. This resolves the formatting issue in OpenSourcePolicyCenter/PolicyBrain#779. However, when an invalid parameter is used, only that error is displayed. If another parameter causes warnings or errors, then those messages may not be displayed.
>
> For example: ...
>
> I think it would be better to display all of the warnings and errors caused by a reform in one pass. That way the user doesn't submit a reform, see the invalid parameter error, fix the error, and then see even more warnings/errors to resolve. Does this seem sensible?

Yes, it seems desirable from the users point of view, but I'm not sure the collection of subsequent errors and warnings can continue if there are unknown parameter names.  That's a pretty drastic error and its hard (impossible?) to continue processing the reform when there are invalid parameters involved in the specification.  For example, how do we look up the range of valid values for an unknown variable?

martinholmer:
@hdoupe,  I've looked again at the `implement_reform` code and I don't see how the logic can proceed after finding unknown parameter names or improper data types for parameters.  The `Policy._validate_parameter_names_types` function does produce error messages for **all** the unknown parameters and known parameters with an improper data type in the JSON reform file.  So users can make all those basic corrections after the first failed file upload.  And then **all** the other parameter value errors/warnings would be produced by the second file upload.  That doesn't seem too inconvenient for users.

My inclination is to merge this fix now (so TaxBrain doesn't crash because of a poorly formatted error message) and others can look in to how to do the one-pass processing you propose.  Maybe one-pass processing is possible and I'm just suffering from a failure of the imagination.

hdoupe:
@martinholmer said
> Yes, it seems desirable from the user's point of view, but I'm not sure the collection of subsequent errors and warnings can continue if there are unknown parameter names. That's a pretty drastic error and its hard (impossible?) to continue processing the reform when there are invalid parameters involved in the specification. For example, how do we look up the range of valid values for an unknown variable?

>  So users can make all those basic corrections after the first failed file upload. And then all the other parameter value errors/warnings would be produced by the second file upload. That doesn't seem too inconvenient for users.

Ok, I understand.  It makes sense to display worse/more obvious errors first such as a parameter name contains a typo, etc.  I think this is probably a better solution than showing everything at once.

Let's merge #1775.  Thanks for quickly taking care of this.  I'll upgrade PolicyBrain to 0.14.2 as soon as you release it.  We should be able to get this in production tonight or tomorrow.

PR #1766, opened 2017-12-16
Expand scope of testing of the JSON reform files in the taxcalc/reforms directory

martinholmer:
This pull request is an attempt to resolve issue #1688 in a way that (1) does not add too much total run time to the unit test suite and (2) does not use either the `puf.csv` or `cps.csv` files as Tax-Calculator records input.

This pull request just adds more tests; if does not change tax-calculating logic or model results.

The expanded test generates a 2020 static difference table for each JSON reform file in the `taxcalc/reforms` directory using as Tax-Calculator input the nine stylized filing units in the `taxcalc/reforms/cases.csv` file.  It also generates for 2020 a static distribution table for current-law policy using the same `cases.csv` input.

PR #1781, opened 2017-12-20
Change minimum required pandas version to 0.21.0

martinholmer:
**This pull request increases from 0.20.1 to 0.21.0 the minimum required version of pandas.**

This change is required for the new reforms test (see revisions in the `test_reform_json_and_output` function that are part of pull request #1766), which use the mapper argument of the [`pandas.rename` method](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.rename.html) that was first added in pandas version 0.21.  Thanks to @derrickchoe, who was using an older version of pandas, for pointing this out.

The unit tests will fail unless you upgrade using this command: `conda update pandas`

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -10,7 +10,7 @@ requirements:
   build:
     - python
     - numpy >=1.12.1
-    - pandas >=0.20.1
+    - pandas >=0.21.0
     - numba
     - toolz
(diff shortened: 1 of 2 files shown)


derrickchoe:
@martinholmer 

Thanks for letting me know-- I've updated my version of pandas.

PR #1805, opened 2018-01-02
Require pandas >=0.22.0 and remove zsum() hack

martinholmer:
The newest version of pandas eliminates a problem that many had with the `sum()` method.
@hdoupe developed a new `zsum()` utility function that worked around this pandas problem.
Now the pandas developers have made the `sum()` method work like almost everybody expects.
So, we can dispense with the work-around by requiring pandas version 0.22.0 or higher.

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -10,7 +10,7 @@ requirements:
   build:
     - python
     - numpy >=1.12.1
-    - pandas >=0.21.0
+    - pandas >=0.22.0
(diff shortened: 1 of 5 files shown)


martinholmer:
Using the term `hack` in the title of pull request #1805 is meant as a compliment.
A "hack" is a creative approach to solving a problem.

hdoupe:
Looks good to me.  Thanks @martinholmer for updating with the new pandas package.

Issue #1816, opened 2018-01-10
Modify TCJA logic with respect to PT_exclusion_wage_limit


martinholmer:
In #1803 @codykallen said:
> We need to modify the wage limit in functions.py:TaxInc. The wage limitation, `PT_exclusion_wage_limit`, should actually be phased in.  The following code would replace the TaxInc function with a corrected specification of the wage limitation:
>
> PT_excl_rt: TCJA sets to 0.2
> PT_excl_wagelim_rt: TCJA sets to 0.5
> PT_excl_wagelim_thd: TCJA sets to [157500, 315000, 157500, 157500, 315000]
> PT_excl_wagelim_prt: TCJA sets to [0.00002, 0.00001, 0.00002, 0.00002, 0.00001]
>```
> def TaxInc(c00100, standard, c04470, c04600, c04800,
>            PT_excl_rt, PT_excl_wagelim_rt, PT_excl_wagelim_thd,
>            PT_excl_wagelim_prt, e00900, e26270, e00200):
>    """
>    TaxInc function: ...
>    """
>
>    pt_excl_pre = max(0., PT_excl_rt * (e00900 + e26270))
>    wagelim_pre = e00200 * PT_excl_wagelim_rt
>    taxinc_pre = max(0., c00100 - max(c04470, standard) - c04600)
>    # calculate business income exclusion
>    excess = max(taxinc_pre - PT_excl_wagelim_thd[MARS - 1], 0.)
>    wagelim_rt = min(excess * PT_excl_wagelim_prt[MARS - 1], 1.)
>    limit = wagelim_rt * max(pt_excl_pre - wage_limit_pre, 0.)
>    pt_excl = pt_excl_pre - limit
>    c04800 = max(0., taxinc_pre - pt_excl)
>    return c04800
>```
>
> Note that this implementation of the wage limitation is only based on the data we have and can model. I have not included the option for a tax filer to use an alternative to PT_excl_wagelim_rt, which the TCJA sets to 50% of wages, to instead use 25% of wages plus 2.5% of tangible depreciable property.

codykallen:
There is actually another related piece that needs to be modified, although it would apply separately (I believe in the CapGains function or the AGI function). The TCJA also limits pass-through losses (`e00900 + e26270`) to $500,000 for married filers and $250,000 for other filers, although I doubt this provision will affect many filers.

martinholmer:
[...]

PR #1819, opened 2018-01-10
Modify pass-through exclusion calculation

codykallen:
This PR modifies the calculation of the business income exclusion for the TCJA. This renames two parameters: `PT_exclusion_rt` becomes `PT_excl_rt`, and `PT_exclusion_wage_limit` becomes `PT_excl_wagelim_rt`. This PR also adds two parameters, `PT_excl_wagelim_thd` and `PT_excl_wagelim_prt`, which are used to correctly implement the rules for the wage limitation on the business income exclusion in the final version of the TCJA. This PR also changes the names of the parameters in the TCJA json files and adds the two new parameters to the `TCJA_Reconciliation.json` (although I did not correct the TCJA Senate versions). 

The default values for the two new parameters are set to maintain backward compatibility, both with pre-TCJA law and with the TCJA Senate json files. 

This PR fulfills the suggestions in #1816.


martinholmer:
@codykallen, thanks for PR #1819.  
That was quick, probably took you less time than helping me do it.
I'll begin a review now.

martinholmer:
@codykallen, one thing would help.  I know the tests are telling you to update the `docs/index.html` file (because you revised the contents of the `current_law_policy.json` file, but that's not the best thing to do.  I'm still trying to figure out how to handle this better, maybe by making the test that checks whether or not the `docs/index.html` file is up-to-date be marked `pre_release`.

The problem with the way it is now is that if you follow its directions (which you did) then, immediately after this pull request is merged, the documentation will be updated even though there is no taxcalc package available with the new PT limitation parameters.

The way to "fix" this for now is on your modify_pt_exclusion branch run these git commands:
```
git checkout -- docs/index.html
git push origin modify_pt_exclusion
```
Sorry about the hassle.

Now I'll start my review of #1819.

codykallen:
@martinholmer, I just ran the two commands you suggested, and they don't appear to have done anything. Do you have a different suggestion?

martinholmer:
@codykallen said:
>  I just ran the two commands you suggested, and they don't appear to have done anything. 
> Do you have a different suggestion?

Right.  I forgot you've already committed the new `docs/index.html` file.
Just let things be and I'll try to fix this (which is my fault, after all) later.

martinholmer:
@codykallen,  Everything look good to me.

BUT you said in issue #1816 this:
> There is actually another related piece that needs to be modified, although it would apply separately (I believe in the CapGains function or the AGI function). The TCJA also limits pass-through losses (e00900 + e26270) to $500,000 for married filers and $250,000 for other filers, although I doubt this provision will affect many filers.

Do you want  to include this enhancement in pull request #1819?

Does this enhancement require any new policy parameters or is it just a change in logic that uses existing policy parameters?

codykallen:
@martinholmer, this enhancement would require a one new policy parameter, as well as changing the `TCJA_Reconciliation.json` file. However, the logic would be relatively straightforward, if implemented carefully. Do you want me to add this as well?

martinholmer:
@codykallen said:
> this [second] enhancement [mentioned in #1816] would require one new policy parameter, as well as changing the `TCJA_Reconciliation.json` file.  However, the logic would be relatively straightforward, if implemented carefully.  Do you want me to add this as well?

Yes, that would be great!   Just add one or more commits to your `modify_pt_exclusion` branch.

codykallen:
@martinholmer, this adds the parameter an an above-the-line deduction cap.

martinholmer:
@codykallen,  Thanks for adding commit 75d1439 that represents the business loss limitation policy parameter and logic under TCJA.

It all looks good except for one thing.  If TCJA limits business losses in the following code:
```
    # compute ymod1 variable that is included in AGI
    ymod1 = (e00200 + e00700 + e00800 + e01400 + e01700 +
             invinc - invinc_agi_ec + e02100 + e02300 +
             max(e00900 + e02000, -ALD_BusinessLosses_c[MARS - 1]))
```
and `ymod1` is added into AGI, we are going to get a different `expanded_income` for those with limited business losses.  I don't think we want a policy reform changing people's `expanded_income`.  

There is an easy way to fix this, I think.  We already have a variable called `c02900_in_ei`, which keeps track of the income that is included in `expanded_income` but is excluded from AGI.  I think you should modify the CapGains function so that `c02900_in_ei` is a function argument and is returned by the CapGains function.  And then inside the CapGains function compute the amount of the business loss limitation (as a non-negative number) and then add that limitation to the `c02900_in_ei` variable.

Does this make sense?

codykallen:
@martinholmer, I believe this should be ready to go now.

martinholmer:
@codykallen, thanks for commit 9f4cd03, but there seems to be one thing missing.
You need to add `c02900_in_ei` to the tuple of variables **returned** from the CapGains function.

martinholmer:
@codykallen,  Thanks for all the help on these TCJA enhancements!

Issue #1827, opened 2018-01-14
Are tbi behavioral-response results for TCJA reform different from non-tbi results?


MattHJensen:
I am finding these on TaxBrain. 

![image](https://user-images.githubusercontent.com/8301092/34921955-1f4bdbca-f957-11e7-8d03-82dbe182259d.png)

https://www.ospc.org/dynamic/behavior_results/820/

cc @codykallen @martinholmer @hdoupe @siervicul

martinholmer:
When I click on the link to dynamic run 820 in this URL:

https://www.ospc.org/dynamic/behavior_results/820/

I get the following page:

![screen shot 2018-01-14 at 9 33 53 pm](https://user-images.githubusercontent.com/12170745/34924678-d4b63512-f972-11e7-90f6-6f8673bf81ac.png)

I don't understand why I can't view the page.

martinholmer:
@MattHJensen,

Do you get the same "very odd" results using the `tc` command-line tool?

If so, please post your `tc` results here and explain what about them is "very odd".

If not, then this issue should be moved to the PolicyBrain repository.

MattHJensen:
I have not yet replicated the results with the `tc` command line tool. 

I opened this issue here because the TaxBrain PE page is an interface to Tax-Calculator and my intuition is that this is a problem in Tax-Calculator rather than TaxBrain. I could be wrong, but if I'm right, I thought other Tax-Calculator devs would want to know about this quickly. 

The results are very odd because they say that the Tax Cuts and Jobs Act raises $6 Trillion of revenue over 10 years.

GoFroggyRun:
To provide more information regarding this PR, I created a TCJA reform that resembles the current TCJA preset on taxBrain, upon which the result [here](https://www.ospc.org/dynamic/behavior_results/820/) is generated, and got the following result based on the current TC master as of Jan.22, 2018: 

![screen shot 2018-01-22 at 3 16 54 pm](https://user-images.githubusercontent.com/13324931/35242578-f34b19f0-ff87-11e7-9e1a-e123ce2ed56c.png)

Here is its counterpart on TB:
![screen shot 2018-01-22 at 3 22 31 pm](https://user-images.githubusercontent.com/13324931/35242639-188f366a-ff88-11e7-9205-8ca4d57cb46f.png)

Based on the results in TC, the total combined liability over 10 years is $34,662.52 B, which yields a loss of $0.55 T over the period. 

[...]

PR #1840, opened 2018-01-23
Add test that compares tbi and std behavioral-response estimates

martinholmer:
This pull request contains a new test that has been added to the `test_tbi.py` file and is marked `pre_release` and `tbi_vs_std_behavior` and `requires_pufcsv`.  

This test assumes `_BE_sub` equals 0.25 and compares the aggregate tax revenues generated by standard Python Tax-Calculator programming with those generated by calling the tbi.run_nth_year_tax_calc_model() function.  The motivation for adding this test is the discussion in issue #1827.

Because of the results generated by the tbi function call are "fuzzed" for PUF privacy reasons, there is no expectation that those results would be identical to the results generated by standard Tax-Calculator calls. 

The new test simulates a massive tax reductions caused by reducing the regular-income and pass-through tax rates no higher than 25 percent and raising the personal exemption from zero to $1000 beginning in 2019.  This reform causes a substantial reduction in marginal tax rates and hence a substantial behavioral response.

However, despite the discussion in #1827, there are no significant differences --- meaning more than a 0.2 percent difference in aggregate tax revenues --- in the results generated using the tbi function call and the results generated using standard (that is, non-tbi) function calls.   Here is how I ran the test (using code at the tip of the master branch) and here is what I got:
```
$ cd taxcalc
$ py.test -m tbi_vs_std_behavior
============================= test session starts ==============================
platform darwin -- Python 2.7.14, pytest-3.2.1, py-1.4.34, pluggy-0.4.0
rootdir: /Users/mrh/work/OSPC/tax-calculator, inifile: setup.cfg
plugins: xdist-1.17.1
collected 436 items                                                             

tests/test_tbi.py .

============================= 435 tests deselected =============================
================== 1 passed, 435 deselected in 120.94 seconds ==================
```
Why do these test results seem different from the results reported in #1827?   
Perhaps I've made a mistake in writing the test.  
Or perhaps the code in 0.14.2 is different from the code at the tip of the master branch.  
Or perhaps mistakes were made in the work reported in #1827.

Does anybody have any ideas about this?

@MattHJensen @GoFroggyRun

### DIFF
--- a/taxcalc/tests/test_tbi.py
+++ b/taxcalc/tests/test_tbi.py
@@ -1,6 +1,7 @@
 """
 Test functions in taxcalc/tbi directory using both puf.csv and cps.csv input.
 """
+from __future__ import print_function
 import numpy as np
 import pandas as pd
 import pytest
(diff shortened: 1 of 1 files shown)


GoFroggyRun:
@martinholmer If this test is based on current master, then things have changed a lot since version 0.14.2. Most importantly, the current law in current master is TCJA, for 0.14.2 version on the other hand, its current law was still pre-TCJA. If I were not mistaken, the **tbi** interface is only capable of calculating assumption based on current law.

martinholmer:
@GoFroggyRun said about pull request #1840:
> If this test is based on current master, then things have changed a lot since version 0.14.2. Most importantly, the current law in current master is TCJA, for 0.14.2 version on the other hand, its current law was still pre-TCJA. If I were not mistaken, the tbi interface is only capable of calculating assumptions based on current law.

So what?  The main point is that you reported in #1827 that Tax-Calculator produced different results depending on whether you used standard (non-tbi) function calls or a call to the `tbi.run_nth_year_tax_calc_model` function.  In the new #1840 test, the results generated from using Tax-Calculator in those two different ways are the same.

The questions I posed in my original #1840 comment are still unanswered.   Can you help by answering them?

GoFroggyRun:
@martinholmer asked:
>Why do these test results seem different from the results reported in #1827?

Why my answer doesn't answer your question? 

1. The current master is essentially different from 0.14.2 in terms of current law, and thus would affect how behavioral assumptions work via **tbi**. 
2. You are not using the exact reform file in #1827.

Given these, why would you expect your test in #1840 could replicate the bug report in #1827?

MattHJensen:
I would be interested to know whether replicating the exact same baseline and reform from #1827 on the tip of the master branch also produces nonsensical results. That seems more relevant to knowing whether there was/is a bug in 0.14.1.

Issue #1845, opened 2018-01-25
The diagnostic_table method gives inconsistent results in presence of behavior responses


GoFroggyRun:
This issue is related to #1840, #1843 and #1827. 

My question is that how should behavioral assumptions be implemented inTC. I used two different ways to implement behavioral assumptions (with the reform being `2017_law.json`) and got inconsistent results. In particular, the choices of years where calculators choose to advance to, via `advance_to_year` option, seem to affect the output. 

I used the following script to do a simple test:
```
import pandas as pd
import numpy as np
import sys
sys.path.append("../..")
import taxcalc
from taxcalc import Policy, Records, Calculator, Behavior

def diag_table_under_beh(year):
    assump_json = """
        {"behavior": {"_BE_sub": {"2018": [0.25]}},
         "growdiff_baseline": {},
         "growdiff_response": {},
         "consumption": {}
        }
    """
    params = Calculator.read_json_param_objects("../reforms/2017_law.json", assump_json)

    puf_fullsample = pd.read_csv("../../puf.csv")

    policy_base = Policy()
    records_base = Records(data=puf_fullsample)

    policy_reform = Policy()
    records_reform = Records(data=puf_fullsample)
    policy_reform.implement_reform(params['policy'])

    behv0 = Behavior()
    behv0.update_behavior(params['behavior'])

    calcbase = Calculator(policy = policy_base, records = records_base )
    calcreform = Calculator(policy = policy_reform, records = records_reform, behavior=behv0)

    calcbase.advance_to_year(year)
    calcreform.advance_to_year(year)
    calcreform = Behavior.response(calcbase, calcreform)

    return calcreform.diagnostic_table(num_years=(2026 - year))

tab_2018 = diag_table_under_beh(2018)
tab_2020 = diag_table_under_beh(2020)

if not (tab_2018[2020]['Combined Liability ($b)'] == tab_2020[2020]['Combined Liability ($b)']):
    print(tab_2018[2020]['Combined Liability ($b)'])
    print(tab_2020[2020]['Combined Liability ($b)'])
```
and got the following result when executed:
```
(taxcalc-dev) Sean-Wangs-MBP:tests seanwang$ python behav.py 
You loaded data for 2009.
Tax-Calculator startup automatically extrapolated your data to 2013.
You loaded data for 2009.
Tax-Calculator startup automatically extrapolated your data to 2013.
You loaded data for 2009.
Tax-Calculator startup automatically extrapolated your data to 2013.
[...]

PR #1847, opened 2018-01-26
Add test that compares behavior results using different base years

GoFroggyRun:
This PR is based on discussion in #1845, where @martinholmer suggested adding a cps-based, not-free-standing test that uses a simple reform. 

This PR adds a test, namely `test_behavioral_response_using_different_years`, that compares reform results, in presence of behavioral assumptions, of two calculators when advanced to different interim years. The test fails to pass, indicating that the interim years where calculator advances to would affect aggregate results, which is not expected. Note that, unlike the script in #1845, this PR does not use `diagnostic_table` method whatsoever, so there's likely a bug somewhere in tax-calculator, rather than a bug within `diagnostic_table` method.

PS: Resolving this would determine whether #1827 is a bug or not.  

cc @MattHJensen

### DIFF
--- a/taxcalc/tests/test_behavior.py
+++ b/taxcalc/tests/test_behavior.py
@@ -197,3 +197,59 @@ def test_boolean_value_infomation(tests_path):
                   val,
                   val_is_boolean)
             assert beh[param]['boolean_value'] == val_is_boolean
+
+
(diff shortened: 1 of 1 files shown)


GoFroggyRun:
Closing in favor of #1848.

PR #1886, opened 2018-02-16
Change "value" to "cost" in BenefitPrograms

andersonfrailey:
In the `BenefitPrograms` function, the sum of the non consumption-adjusted benefit values are assigned to `benefit_value_total`, rather than `benefit_cost_total`. This PR changes that assignment.


martinholmer:
@andersonfrailey,  Thanks for catching my mistake!

Issue #1904, opened 2018-03-08
Why is e00400 included in eitc_agi?


MaxGhenis:
`eitc_agi` is [defined as](https://github.com/open-source-economics/Tax-Calculator/blob/5b20973bbfe4834420a71f892976ec3555c3b52d/taxcalc/functions.py#L1158) `c00100 + e00400`, where `c00100` is AGI and `e00400` is "tax-exempt interest income" from 1040 line 8b, according to the [documentation](http://open-source-economics.github.io/Tax-Calculator/).

Is something off here? The [2017 EIC publication](https://www.irs.gov/pub/irs-pdf/p596.pdf) refers to 8b as combat pay, which aligns with the [2016 1040EZ](https://www.irs.gov/pub/irs-prior/i1040ez--2016.pdf), and there it seems like it should be counted for EITC. But the [2017 1040](https://www.irs.gov/pub/irs-pdf/i1040gi.pdf) refers to 8b as tax-exempt interest income.

The [`EITC` function](https://github.com/open-source-economics/Tax-Calculator/blob/5b20973bbfe4834420a71f892976ec3555c3b52d/taxcalc/functions.py#L1150) also includes `e00400` in `invinc`, so I wonder if this field is being used for two different things?

The above refers to tax units with `EITC_indiv = False`. For others, which use [`earned_p` and `earned_s`](https://github.com/open-source-economics/Tax-Calculator/blob/5b20973bbfe4834420a71f892976ec3555c3b52d/taxcalc/functions.py#L123), `e00400` does not appear to be part of the equation.

martinholmer:
@MaxGhenis asked in issue #1904:
> `eitc_agi` is defined as c00100 + e00400, where c00100 is AGI and e00400 is "tax-exempt interest income" from 1040 line 8b, according to the documentation.
>
> Is something off here? The 2017 EIC publication refers to 8b as combat pay, which aligns with the 2016 1040EZ, and there it seems like it should be counted for EITC. But the 2017 1040 refers to 8b as tax-exempt interest income.

When you say "the 2017 EIC publication refers to 8b as combat pay", where exactly in that 41 page publication does it "refer to 8b as combat pay"?

MaxGhenis:
> When you say "the 2017 EIC publication refers to 8b as combat pay", where exactly in that 41 page publication does it "refer to 8b as combat pay"?

[...]

PR #1907, opened 2018-03-08
Fix AGI concept used in EITC phase-out logic

martinholmer:
This pull request fixes a bug in EITC phase-out logic.  This issue was raised by @MaxGhenis in #1904.  Thanks, Max, for pointing this out.  Prior to this pull request the EITC amount was phased out using the maximum of earnings and a concept called `eitc_agi`, which was defined as the sum of AGI (`c00100`) and non-taxable interest income (`e00400`).  As Max pointed out `e00400` is included in the investment income EITC eligibility test, but he couldn't find an mention of it being added to AGI and used in the phase-out formula.  After an extensive Google search, I found an IRS document that said in a discussion of the EITC that the AGI concept used was from line 38 of Form 1040, which is standard AGI (`c00100`).  My search also found an [online EITC calculator](https://www.bankrate.com/calculators/tax-planning/earned-income-tax-credit-calculator.aspx) that implied the AGI concept used in EITC calculation was standard (that is, does not include non-taxable interest income).

The results of fixing this bug are quite small.  In some years, total refundable credits (in which the EITC is included) changes by $0.1 billion and in other years it is unchanged.  When it does change, the aggregate individual income tax total and the combined tax total also changes by $0.1 billion.

It appears that this bug went undetected in our validation efforts because Internet TAXSIM does not have non-taxable interest income as one of its input variables, and because our efforts to validate Tax-Calculator against other tax calculators were never completed.

@MaxGhenis, thanks again for raising and pressing this issue.


taxcalc/functions.py
```diff
@@ -1123,14 +1123,14 @@ def F2441(MARS, earned_p, earned_s, f2441, CDCC_c, e32800,
 
 @jit(nopython=True)
 def EITCamount(phasein_rate, earnings, max_amount,
-               phaseout_start, eitc_agi, phaseout_rate):
+               phaseout_start, agi, phaseout_rate):
```
MaxGhenis: Would it be clearer to use `c00100` here instead of introducing a new variable name of `agi`, which might be interpreted as a different concept? functions.py currently only uses the name `c00100` to refer to AGI in code.

PR #1910, opened 2018-03-09
Improve docstrings in functions.py and calculate.py

martinholmer:
Documentation changes only.

### DIFF
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -769,9 +769,8 @@ class Calculator(object):
 
             - 'agi': adjusted gross income, AGI (c00100)
 
-            - 'expanded_income': sum of AGI, non-taxable interest income,
-              non-taxable social security benefits, and employer share of
-              FICA taxes.
+            - 'expanded_income': broader than AGI (see definition in
+                                 functions.py file).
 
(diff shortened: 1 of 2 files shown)


MaxGhenis:
This is in the weeds, but does this PR establish the descriptive rather than imperative mood for docstrings in taxcalc? i.e. "Computes" rather than "Compute".

This aligns with the [Google style guide](https://google.github.io/styleguide/pyguide.html#Comments) and seems to be popular based on this [top SO question](https://stackoverflow.com/q/3898572/1840471). Yet it does deviate from examples in [PEP 257](https://www.python.org/dev/peps/pep-0257/) and [PEP 8](https://www.python.org/dev/peps/pep-0008/#documentation-strings) which use the imperative mood.

I have no preference other than knowing for consistency.

martinholmer:
@MaxGhenis said in PR #1910:
> I have no preference [between docstrings using "the descriptive rather than imperative mode"] other than knowing for consistency.

I'm afraid the project is not yet advanced enough to have a consistent docstring style.  But we do suggest developers read the Google Python Style Guide, so I guess we should strive for consistently using the descriptive mode.

PR #1923, opened 2018-03-14
Add Python 2.7 sunset message to Calculator constructor

martinholmer:
This pull request notifies Tax-Calculator users that beginning in 2019 Python 2.7 will not be supported: the source code will be Python 3.6 and conda `taxcalc` packages will be available only for Python 3.6.

The user documentation already instructs users to install Python 3.6 (and not 2.7) and the contributor guide already says the same thing.  And the developers of projects that are known to use Tax-Calculator have been directly notified of this future change.  So, we have given a comprehensive early warning of this major change.

### DIFF
--- a/taxcalc/calculate.py
+++ b/taxcalc/calculate.py
@@ -8,6 +8,7 @@ Tax-Calculator federal tax Calculator class.
 # pylint: disable=invalid-name,no-value-for-parameter,too-many-lines
 
 import os
+import sys
 import json
 import re
 import copy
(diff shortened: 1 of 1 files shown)


MattHJensen:
LGTM. Thanks @martinholmer.

MaxGhenis:
Do you know if it's possible to warn upon loading the package? I couldn't find a way with a cursory search, but this would be nice for users who create many `Calculator` objects in the same session (like me...) and get warned multiple times.

martinholmer:
@MaxGhenis asked:
> Do you know if it's possible to warn upon loading the package?  I couldn't find a way with a cursory search, but this would be nice for users who create many Calculator objects in the same session (like me...) and get warned multiple times.

No I don't know how to do that.  But if you are getting tired of reading the warning and don't want to get rid of it by upgrading to Python 3.6, then simply set `verbose=False` in the Calculator class constructor.  That should suppress the warning.

MaxGhenis:
Thanks!

PR #1931, opened 2018-03-19
Clarify age requirement for CTC-eligible children

martinholmer:
Change in documentation only.  
But it would appear that the current `cps.csv` file incorrectly contains 17-year olds in the `n24` count as described by @MaxGhenis in [taxdata issue 164](https://github.com/open-source-economics/taxdata/issues/164).

### DIFF
--- a/taxcalc/records_variables.json
+++ b/taxcalc/records_variables.json
@@ -511,7 +511,7 @@
     },
     "n24": {
       "type": "int",
-      "desc": "Number of children eligible for Child Tax Credit",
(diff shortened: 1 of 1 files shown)


MaxGhenis:
This might create more confusion, since it suggests that being under 17 is a separate requirement imposed which isn't already part of the CTC eligibility, and/or that it's the sole condition of CTC eligibility (there are others, though I don't know if they're modeled in taxdata). How about something like this?

> Number of children eligible for Child Tax Credit, of which one condition is being under age 17.

PR #1942, opened 2018-03-29
Begin managing memory usage

martinholmer:
Since the beginning of the Tax-Calculator project, the Python code has never included any memory management logic.  It was possible to ignore this issue because the size of the large data objects were never big enough to exhaust available memory.  But with the use of the much larger `cps.csv` input file and associated benefits data, memory usage became large enough to exhaust physical memory on r4-large AWS instances (with roughly 15GB of total memory) as reported in [PolicyBrain issue 861](https://github.com/OpenSourcePolicyCenter/PolicyBrain/issues/861).  As described in 861, the temporary solution is to run Tax-Calculator on r4-xlarge AWS instances (with roughly 30GB of total memory).  But that solution to the problem doubles the cost of providing Tax-Calculator access via TaxBrain.

Working on a local computer with only 6GM of available memory, I have added some simple memory management logic to Tax-Calculator and I have reduced the internal precision of weights, adjustment ratios, and benefit amount by storing them as 32-bit values instead of 64-bit values.

Before making these changes, the `run_nth_year_tax_calc_mode` function exhausted the 6GB of available memory and started using virtual memory (that is, swapping blocks of memory to disk).  These disk operations are extremely slow relative to memory access, so the execution of the `run_nth_year_tax_calc_mode` function was **very** slow, but in the end the correct results were returned.

After making these changes, the `run_nth_year_tax_calc_mode` function hits the limit of 6GB of available memory only once, and as a result, executes much faster.

I'm hoping that these changes will make it possible for PolicyBrain to run Tax-Calculator on r4-large AWS instances, but to be sure about that PolicyBrain developers will have to test the new version.

@hdoupe @MattHJensen


hdoupe:
Nice work @martinholmer. Thanks for taking care of this quickly.

PR #1952, opened 2018-04-05
Apply Policy parameter processing/validation logic to Behavior class

### LINKED ISSUE #1951: Add Tax-Calculator behavior parameters validation
The `taxcalc.tbi.reform_warnings_errors` function only validates the policy parameters and not the behavior parameters. Can this be augmented to validate behavior parameters, too?  I thought that it validated the behavior parameters and, thus, removed the parameter validation for them from PolicyBrain in 1.4.0.

hdoupe:
This PR tries to resolve #1951 by applying the `Policy` parameter validation logic to the `Behavior` class. Fortunately, the validation code in the `Policy` "just worked" in the `Behavior` class. The attributes and methods that were copied over are:

- `implement_reform` method was implemented as `update_behavior`
- `_validate_parameter_names_types`
  - I found an interesting case where the user could pass in a boolean value for a parameter when the JSON document specified `boolean_type` and `int_type` as `false`. It turns out `isinstance(True, (int, float))` is `True`. I'm not sure if this was desired behavior or not
  - The CPI logic in this method doesn't apply to the `Behavior` class. So, it could be removed.
- `_validate_parameter_values`
- `current_law_version` as `baseline_version`
- `reform_errors` as `behavior_errors`
- `reform_warnings` as `behavior_warnings`
- `_ignore_errors`

The `behaviors.json` file had to be updated so that each parameter had the range and out_of_range messages and action. Please double check my work. I'm not very familiar with these parameters. Also, the [`behaviors.json` document indicates](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/behavior.json#L53-L54) that `_BE_subinc_wrt_earnings` is a boolean parameter, but the [current validation logic](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/behavior.py#L388-L390) seems to indicate that it could be a float or boolean or integer:
```
elif param == '_BE_subinc_wrt_earnings':
    if val < 0 or val > 1:
    raise ValueError(msg.format(param, nob, cyr, val))
```

The `tbi.reform_warnings_errors` function was updated to validate the `behavior` parameters in the `user_mods` dictionary. This function now returns a dictionary that looks like: 
```
{
    'policy': {'warnings': '...', 'errors': '...'}, 
    'behavior': {'warnings': '...', 'errors': '...'}
}
```

I'll make the changes for this part on the PolicyBrain side.

I have a few more questions that I will ask in code comments.

@martinholmer I'm interested in your thoughts on this PR. If you agree with this approach, what do you think about refactoring the `_validate_parameter_names_types` and `_validate_parameter_values` methods into the `ParametersBase` class? If we do that, perhaps, it wouldn't be too hard to apply it to the other classes such as `GrowFactors`.


hdoupe:
I forgot to start on a clean branch. Thus, the first commit "Adds ability to use operator when validating w.r.t. other parameters" is an experiment I did a couple weeks ago and is unrelated to this PR.

taxcalc/behavior.json
```diff
@@ -53,7 +59,10 @@
         "boolean_value": true,
         "integer_value": true,
         "value": [false],
-        "validations": {"min": false, "max": true}
+        "range": {"min": false, "max": true},
+        "out_of_range_minmsg": "",
+        "out_of_range_maxmsg": "",
+        "out_of_range_action": "stop"
     },
```
hdoupe: Should this parameter be boolean?

taxcalc/behavior.py
```diff
@@ -83,6 +90,72 @@ def update_behavior(self, revisions):
...
+        if first_reform_year < self.start_year:
+            msg = 'ERROR: {} YEAR reform provision in YEAR < start_year={}'
+            raise ValueError(msg.format(first_reform_year, self.start_year))
+        if first_reform_year < self.current_year:
+            msg = 'ERROR: {} YEAR reform provision in YEAR < current_year={}'
+            raise ValueError(msg.format(first_reform_year, self.current_year))
+        last_reform_year = max(reform_years)
+        if last_reform_year > self.end_year:
+            msg = 'ERROR: {} YEAR reform provision in YEAR > end_year={}'
+            raise ValueError(msg.format(last_reform_year, self.end_year))
```
hdoupe: All of the year validation logic may be overkill for the behavior class. However, I thought it would be interesting to show how much of the `Policy` code could be copied over.

taxcalc/behavior.py
```diff
@@ -83,6 +90,72 @@ def update_behavior(self, revisions):
...
+        self.set_year(precall_current_year)
+        # validate reform parameter values
+        self._validate_parameter_values(reform_parameters)
+        ####################
+        # Additional Behavior specific logic
+        # check behavior_errors and _ignore_errors after values validation, too
+        ###################
+        if not self._ignore_errors and self.behavior_errors:
+            raise ValueError(self.behavior_errors)
+
```
hdoupe: `Policy` doesn't check whether there are error messages after doing value validation in `_validate_parameter_values`. Why doesn't `Policy` do this?

taxcalc/behavior.py
```diff
@@ -407,6 +480,179 @@ def _validate_parameter_values(self):
...
+                                self.behavior_errors += (
+                                    'ERROR: ' + msg.format(year, pname) + '\n'
+                                )
+                    else:
+                        msg = '{} {} parameter is not true or false'
+                        self.behavior_errors += (
+                            'ERROR: ' + msg.format(year, name) + '\n'
+                        )
+                ########################
+
```
hdoupe: This won't be used by the `Behavior` class, but again, I thought it would be interesting to show how much code from the `Policy` class could simply be copied over.

taxcalc/behavior.py
```diff
@@ -407,6 +480,179 @@ def _validate_parameter_values(self):
...
+                                ################################
+
+                                if (not isinstance(pvalue[idx], (float, int)) or
+                                    isinstance(pvalue[idx], bool)):
+                                    msg = '{} {} value {} is not a number'
+                                    self.behavior_errors += (
+                                        'ERROR: ' +
+                                        msg.format(year, pname, pvalue[idx]) +
+                                        '\n'
+                                    )
```
hdoupe: The test case:
```
# test_behavior.py
def test_validate_param_names_types_errors():
...
+    behv1 = Behavior()
+    specs1 = {2019: {'_BE_inc': [True]}} # _BE_inc should be a float (int_type = boolean_type = False)
+    with pytest.raises(ValueError):
+        behv1.update_behavior(specs1) 
```
did not throw a `ValueError` before the second part of the conditional statement was added. See [policy.py](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/policy.py#L482-L488) for original conditional statement.

taxcalc/behavior.json
martinholmer: Which parameter are you talking about?

taxcalc/behavior.py
martinholmer: Checking is the responsibility of the calling script.
See, for example, the basic recipe in the Cookbook.

taxcalc/behavior.json
hdoupe: Sorry about that. I'm referring to `_BE_subinc_wrt_earnings`.

taxcalc/behavior.py
hdoupe: Ah, I see. Thanks.

martinholmer:
@hdoupe said:
> I found an interesting case where the user could pass in a boolean value for a parameter when the JSON document specified boolean_type and int_type as false. It turns out `isinstance(True, (int, float))` is `True`. I'm not sure if this was desired behavior or not.

Thanks for the bug report, @hdoupe.  Can you provide detail in a new issue?

martinholmer:
@hdoupe said:
> `reform_errors` as ``behavior_errors`

I'm not convinced about this method name.  This would require usage like this: `if behavior.behavior_errors:`, which seems redundant.  The analogue to policy `reform` is behavior `update`.  Why not name the Behavior method `update_errors`?

martinholmer:
@hdoupe said:
> `reform_warnings` as `behavior_warnings`

But there will never be any warnings for Behavior parameters, right?  
Can't you eliminate this method?  And also, eliminate the JSON fields: 
And

martinholmer:
@hdoupe said:
> Also, the `behavior.json` [file] indicates that `_BE_subinc_wrt_earnings` is a boolean parameter, but the current validation logic seems to indicate that it could be a float or boolean or integer:

I'm working to fix this in pull request #1955.  Thanks for pointing this out.

hdoupe:
@martinholmer asked:
> Thanks for the bug report, @hdoupe. Can you provide detail in a new issue?

Yep, see issue #1956

hdoupe:
@martinholmer asked about the `behavior_errors` attribute:
> I'm not convinced about this method name. This would require usage like this:
> 
> `if behavior.behavior_errors:`
> 
> which seems redundant. The analogue to policy `reform` is behavior `update`. Why not name the Behavior method `update_errors`?

I agree that this is redundant. I also agree that `update` would is better. What do you think about changing the attribute name from `reform_errors` or `update_errors` to just `parameter_errors` or simply just `errors`? Then the above statement becomes:

`if behavior.parameter_errors`

Then, it would be obvious that the `Policy` analog to this can be found at `policy.parameter_errors`. It would also make the API a little more predictable. That is, you wouldn't have to look up the specific modifier every time you want to access the attribute in a new class. For example, PolicyBrain wouldn't have to keep track of what object it's working with and map that to a specific error attribute name.

hdoupe:
@martinholmer commented:


> >  `reform_warnings` as `behavior_warnings`
> 
> But there will never be any warnings for Behavior parameters, right?
> Can't you eliminate this method? And also, eliminate the JSON fields:
> ```
> "out_of_range_minmsg": "",
> "out_of_range_maxmsg": "",
> "out_of_range_action": "stop"
> ```
> There is no variation in the values of these fields across non-policy parameters, right?

Yes, you are correct. Right now, there is no need for a warnings attribute on the Behavior class and there is no variation in the `behavior.json` file. My only hesitation to doing this is that it breaks with the Policy API. However, I don't know if this is a good enough reason to have extra code/JSON fields that aren't used. I could see how that could cause some confusion and misunderstanding.

martinholmer:
@hdoupe said:
> Yes, you are correct. Right now, there is no need for a warnings attribute on the Behavior class and there is no variation in the `behavior.json` file. My only hesitation to doing this is that it breaks with the Policy API. However, I don't know if this is a good enough reason to have extra code/JSON fields that aren't used. I could see how that could cause some confusion and misunderstanding.

The Policy class is very different from the other classes derived from ParameterBase.
There is no need to make the non-Policy parameter validation logic exactly mirror that for Policy parameters.
Let's streamline things in an attempt to reduce the likelihood of "some confusion and misunderstanding."

hdoupe:
@martinholmer said
> I think this is an excellent idea! Go ahead and make those changes to the Policy methods.
Be sure to merge recent master changes into your development branch.

Great! I will do that and then we can see how it looks.

hdoupe:
@martinholmer said:
>The Policy class is very different from the other classes derived from ParameterBase.
> There is no need to make the non-Policy parameter validation logic exactly mirror that for Policy parameters.
> Let's streamline things in an attempt to reduce the likelihood of "some confusion and misunderstanding."

Sounds good to me. I will remove the warnings attribute and the extra JSON fields from the `behaviors.json` file.

taxcalc/behavior.py
```diff
@@ -407,6 +476,151 @@ def _validate_parameter_values(self):
...
+        # above handles non-rounding of inflation-indexed parameter values
+        clp = self.baseline_version()
+        parameters = sorted(parameters_set)
+        syr = Behavior.JSON_START_YEAR
+        for pname in parameters:
+            if pname.endswith('_cpi'):
+                continue  # *_cpi parameter values validated elsewhere
+            pvalue = getattr(self, pname)
+            for vop, vval in self._vals[pname]['range'].items():
+                if isinstance(vval, six.string_types):
```
hdoupe: @martinholmer this condition is true if you have a range field in `behavior.json` that looks something like `"range": {"min": "param name", "max": ...}`, right?

If this is correct, do you think we should remove this case? Then if we need it some time in the future, we could just add it back.

martinholmer: OK, remove the code.

hdoupe:
@martinholmer I removed [this call](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/behavior.py#L63) from the `Behavior.__init__` method because it is not in the `Policy.__init__` method. I'm not really sure why it is called. What is it validating?

One place where we may want to do some kind of validation in the `__init__` methods is when the `behavior_dict` [argument is used](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/behavior.py#L56-L57). This applies to the [Policy class](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/policy.py#L74-L75), too.

An [existing behavior test](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/tests/test_behavior.py#L14-L18) exposes this issue, and I added a [policy test](https://github.com/open-source-economics/Tax-Calculator/pull/1952/files#diff-dc74f969d76006abbdf3bcc5b95240e4R45) that exposes the issue in that class, too: 

```
# def test_incorrect_Policy_instantiation():
# ...
+    # similar test to test_behavior.test_incorrect_behavior_instantiation-2
+    # fails because this dict is not validated
+    bad_pol_dict = {
+        '_SS_Earnings_c': {'start_year': 2013, 'value': [0.1]}
+    }
+    with pytest.raises(ValueError): # raises KeyError instead
+        p = Policy(parameter_dict=bad_pol_dict)
```

hdoupe:
@martinholmer I tried to translate the `Policy.implement_reform` docstring to apply to the `Behavior.update_behavior` method. Commit [`9219114`](https://github.com/open-source-economics/Tax-Calculator/pull/1952/commits/9219114950082fa4f88a400f8047582b48003259)  makes revisions to the Policy docstring. Can you review?

martinholmer:
@hdoupe proposed this code change in PR#1952:
```
      def update_behavior(self, reform, raise_errors=True):
         """
-        Update behavior for given revisions, a dictionary consisting
-        of one or more year:modification dictionaries.
-        For example: {2014: {'_BE_sub': [0.4, 0.3]}}
-        Also checks for valid elasticity values in revisions dictionary.
-
-        Note that this method uses the specified revisions to update the
-        default elasticity values, so use this method just once
-        rather than calling it sequentially in an attempt to update
-        elasticities in several steps.
+        Implement multi-year behavior reform and leave current_year unchanged.
```
I think the use of "reform" to describe "revisions" in behavioral response assumptions is very misleading.
The term "reform" is universally used to describe a change in **policy**.
Can we go back to using the "revisions" terminology when referring to changes in non-policy parameters?

taxcalc/tests/test_policy.py
```diff
@@ -36,6 +36,13 @@ def test_incorrect_Policy_instantiation():
...
     with pytest.raises(ValueError):
         p = Policy(parameter_dict=list())
+    # similar test to test_behavior.test_incorrect_behavior_instantiation-2
+    # fails because this dict is not validated
+    bad_pol_dict = {
+        '_SS_Earnings_c': {'start_year': 2013, 'value': [0.1]}
+    }
+    with pytest.raises(ValueError):
+        p = Policy(parameter_dict=bad_pol_dict)
     with pytest.raises(ValueError):
```
hdoupe: @martinholmer sorry for the confusion. I am referring to the [second test case](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/tests/test_behavior.py#L14-L18) in `test_behavior.test_incorrect_behavior_instantiation`:

```
    bad_behv_dict = {
        '_BE_bad': {'start_year': 2013, 'value': [0.0]}
    }
    with pytest.raises(ValueError):
        Behavior(behavior_dict=bad_behv_dict)
```

martinholmer:
@hdoupe,  Is it clear to you why the tests are failing?

hdoupe:
@martinholmer asked:
> Is it clear to you why the tests are failing?

I just fixed the issue with the `tbi.reform_warnings_errors` function. I will fix the pep-8 errors too.

I think I see why the remaining tests are failing. An incorrectly formatted dictionary is passed in to the `__init__` methods of the `Behavior` and `Policy` classes. I think my main confusion is:

1. What is the `parameter_dict` keyword used for in the `Policy` class? What is the `behavior_dict` keyword used for in the `Behavior` class?
2. What is the desired behavior on receiving a poorly formatted dictionary in the respective keyword arguments? There is a [behavior test](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/tests/test_behavior.py#L14-L18) that indicates that this should cause a value error:
```
    bad_behv_dict = {
        '_BE_bad': {'start_year': 2013, 'value': [0.0]}
    }
    with pytest.raises(ValueError):
        Behavior(behavior_dict=bad_behv_dict)
```

However, I do not see a corresponding test case in the Policy test suite. This is why I added the following test to the Policy test suite:
```

+    # similar test to test_behavior.test_incorrect_behavior_instantiation-2
+    # fails because this dict is not validated
+    bad_pol_dict = {
+        '_SS_Earnings_c': {'start_year': 2013, 'value': [0.1]}
+    }
+    with pytest.raises(ValueError):
+        p = Policy(parameter_dict=bad_pol_dict) 
```

This test throws a key error. 

To go forward from this point, I think it would be helpful for me to understand how these keywords are used and then, for us to decide on how they should be validated.

I hope this is clear. Please correct me if I have misunderstood some aspect of this.

martinholmer:
@hdoupe said:
>  I think my main confusion is:
>
> 1. What is the parameter_dict keyword used for in the Policy class [constructor]? What is the behavior_dict keyword used for in the Behavior class [constructor]?

Yes, I agree it's pretty confusing.  Typically, the Policy constructor and the Behavior constructor are called with their `_dict` parameter at the default value of `None`, which means the parameter values are read from a JSON file (either the `current_law_policy.json` file in the case of the Policy constructor or the `behavior.json` file in the case of the Behavior constructor).  But early in the project's history (before I started working on Tax-Calculator) it was thought to be a good idea to allow the constructors to be called with a dictionary to specify the default parameter values, in which case no JSON file would be read.

Given the current state of Tax-Calculator development I'm not aware of any practical use for this "feature".  And your experience in preparing PR #1952 clearly shows that this "feature" creates confusion and generates extra code-maintenance and testing costs.  

So, I propose we drop the `_dict` parameter from the constructor of each of the classes derived from the `ParametersBase` abstract class.  What do you think about this proposal?

@MattHJensen @feenberg @feenberg

hdoupe:
@martinholmer said:
> So, I propose we drop the _dict parameter from the constructor of each of the classes derived from the ParametersBase abstract class. What do you think about this proposal?

This proposal makes sense to me as long as @MattHJensen and @feenberg don't have any objections.

hdoupe:
The latest commits remove the `_dict` arguments from the `Behavior` and `Policy` constructors. They can still be reverted if anyone has any objections to removing this argument.

cc @martinholmer @feenberg @MattHJensen

hdoupe:
All tests are passing except for the test coverage checks. There are three spots that are no longer hit by the test suite. It looks like there is not a path to them now that the `_dict` keywords have been removed from the Behavior/Policy constructors. I'll leave comments on my reasoning in the changes in commit [`5ee4353`](https://github.com/open-source-economics/Tax-Calculator/pull/1952/commits/5ee4353746eace5609886ad657b7bd495dbd2b85).

taxcalc/parameters.py
```diff
@@ -94,9 +94,6 @@ def set_default_vals(self, known_years=999999):
         """
         if hasattr(self, '_vals'):
             for name, data in self._vals.items():
-                if not isinstance(name, six.string_types):
-                    msg = 'parameter name {} is not a string'
-                    raise ValueError(msg.format(name))
```
hdoupe: Paths to this section of code:

`Policy.__init__` --> `ParametersBase.initialize` --> `Parameters.set_default_vals`
`Behavior.__init__` --> `ParametersBase.initialize` --> `Parameters.set_default_vals`

In both of these cases, the values are read directly from the defaults specified in `current_law_policy.json` and `behavior.json`. Thus, it is impossible to have an incorrect name unless the name specified in the defaults files is a number or something like that. A test could potentially be added for this possibility.

`Policy.implement_reform` --> 
```
        self._validate_parameter_names_types(reform)
        ....
        # optionally apply cpi_offset to inflation_rates and re-initialize
        if Policy._cpi_offset_in_reform(reform):
            known_years = self._apply_reform_cpi_offset(reform)
            self.set_default_vals(known_years=known_years)
```
--> `ParametersBase.set_default_vals`

However, `Policy._validate_parameter_names_types` was called directly before `set_default_vals`. Thus, we can be certain that the names are strings and valid otherwise.

taxcalc/parameters.py
```diff
@@ -358,12 +355,8 @@ def _update(self, year_mods):
...
                 continue  # handle elsewhere in this method
-            if name in self._vals:
-                vals_indexed = self._vals[name].get('cpi_inflated', False)
-                integer_values = self._vals[name].get('integer_value')
-            else:
-                msg = 'parameter name {} not in parameter values dictionary'
-                raise ValueError(msg.format(name))
+            vals_indexed = self._vals[name].get('cpi_inflated', False)
+            integer_values = self._vals[name].get('integer_value')
             name_plus_cpi = name + '_cpi'
```
hdoupe: Paths to this section of code:

`Policy.implement_reform` --> `ParametersBase._update`
`Behavior.update_behavior` --> `ParametersBase._update`

These are both called after calling the Policy and Behavior classes' respective `_validate_param_names_types` methods. Thus, we can be certain that the name is in `self._vals`. Also, I think an equivalent condition `pname not in self._vals` is already in `Policy._validate_param_names_types`[lines 446-450](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/policy.py#L446-L450)  and was added to `Behavior._validate_param_names_types`.

taxcalc/parameters.py
```diff
@@ -387,9 +380,6 @@ def _update(self, year_mods):
         for name in unused_names:
             used_names.add(name)
             pname = name[:-4]  # root parameter name
-            if pname not in self._vals:
-                msg = 'root parameter name {} not in values dictionary'
-                raise ValueError(msg.format(pname))
             pindexed = year_mods[year][name]
```
hdoupe: Paths to this section of code:

`Policy.implement_reform` --> `ParametersBase._update`

`Policy._validate_param_names_values` is called before `ParametersBase._update`. `Policy._validate_param_names_values` has a similar check: 

```
pname = name[:-4]  # root parameter name
if pname not in param_names:
    msg = '{} {} unknown parameter name'
    self.reform_errors += (
        'ERROR: ' + msg.format(year, name) + '\n'
)
```

hdoupe:
I said:

> All tests are passing except for the test coverage checks. There are three spots that are no longer hit by the test suite. It looks like there is not a path to them now that the _dict keywords have been removed from the Behavior/Policy constructors. I'll leave comments on my reasoning in the changes in commit 5ee4353.

After further consideration, it looks like these lines of code could still be hit by `GrowFactors` or `GrowDiff` since they have a `_dict` argument in their constructors and do not have any validation code.

@martinholmer what do you think about this?

martinholmer:
@hdoupe said:
> In both of these cases, the values are read directly from the defaults specified in `current_law_policy.json` and `behavior.json`. Thus, it is impossible to have an incorrect name unless the name specified in the defaults files is a number or something like that.

I agree.  Now that the parameters are being input only by reading files, it seems as there is no need for this kind of run-time checking.  

However, we do need to be sure that the unit tests include checks of the content of the JSON files, so that a developer does not mistakenly specify invalid data in the JSON files.  Do we now have that kind of testing of the content of the `current_law_policy.json` and `behavior.json` files?  Or do we need to add some testing?

martinholmer:
@hdoupe said:
> After further consideration, it looks like these lines of code [in the ParametersBase class] could still be hit by Growfactors or Growdiff since they have a `_dict` argument in their constructors and do not have any validation code.

Actually, I think the constructor for the Growfactors class does **not** have a dictionary argument.
But the Growdiff class does have one and does very much need one.

Does it make sense to proceed this way?
Go ahead with deleting the ParametersBase code.
This will have no effect on the Growfactors class.
And soon I'll add whatever validation checking is needed by the Growdiff class to the Growdiff class.

hdoupe:
@martinholmer asked:
> However, we do need to be sure that the unit tests include checks of the content of the JSON files, so that a developer does not mistakenly specify invalid data in the JSON files. Do we now have that kind of testing of the content of the current_law_policy.json and behavior.json files? Or do we need to add some testing?

I do not see any testing that does this. So we should probably add it. I can go ahead and add this. We just want to read in the defaults files and make sure that the keys are all strings, correct?

hdoupe:
@martinholmer  said:
> Actually, I think the constructor for the Growfactors class does not have a dictionary argument.

Yep, this is correct. Thanks for pointing this out.

> Does it make sense to proceed this way?
> Go ahead with deleting the ParametersBase code.
> This will have no effect on the Growfactors class.
> And soon I'll add whatever validation checking is needed by the Growdiff class to the Growdiff class.

Sounds sensible to me.

martinholmer:
@hdoupe said:
> I do not see any testing that does this. So we should probably add it. I can go ahead and add this. We just want to read in the defaults files and make sure that the keys are all strings, correct?

You could easily add the test that each key is a string to this `test_parameters.py` test:
```
@pytest.mark.parametrize("fname",
                         [("behavior.json"),
                          ("consumption.json"),
                          ("current_law_policy.json"),
                          ("growdiff.json")])
def test_json_file_contents(tests_path, fname):
    """
    Check contents of JSON parameter files.
    """
```

hdoupe:
@martinholmer said:
> We don't want to drop these three lines, do we?
> 
> If a JSON reform file includes a parameter named bogus_cpi, we want to catch that error at runtime.
> 
> Unless I'm confused, we want to keep these three lines and add a test that raises a ValueError when the reform includes something like bogus_cpi (because bogus is not a valid policy parameter name).

So, you are saying that a JSON reform file with an incorrect CPI parameter name needs this validation? Something like:

```
{
    "policy": {
        "_Bogus_cpi": {
            2017: true
        }
    }
}
```
should trigger those lines of code?

hdoupe:
@martinholmer I tried the following test:

```
bogus = """
{
    "policy": {
        "_Bogus_cpi": {
            "2017": true
        }
    }
}"""
bogus = Calculator.read_json_param_objects(bogus, None)
pol7 = Policy()
pol7.implement_reform(bogus['policy'])
```

And, I get a `ValueError` that is raised in `Policy.implement_reform` regardless of whether the lines in parameters.py are commented out or not. I think `Policy._validate_param_names_types` is called before `Parameters._update`. `Policy._validate_param_names_types` does a similar check to the lines you mentioned: https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/policy.py#L445-L450

However, I may be missing something.

martinholmer:
@hdoupe said:
> I get a ValueError that is raised in Policy.implement_reform regardless of whether the lines in parameters.py are commented out or not. I think Policy._validate_param_names_types is called before Parameters._update. Policy._validate_param_names_types does a similar check to the lines you mentioned: https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/policy.py#L445-L450
>
> However, I may be missing something.

OK, I see.  You were correct from the beginning.  The Policy class implement_reform method catches this error before the ParametersBase _update method is called.  Sorry, about the confusion.  The way you had it before seems correct.

hdoupe:
Ok, thanks for thoroughly reviewing that change. This is a complex process and it's easy for either one (or both) of us to miss something.

martinholmer:
@hdoupe,  The last code you've added in #1952 that is not tested is as follows:
```
+                        elif int_param_type:
+                            if not pval_is_int:
+                                msg = '{} {} value {} is not integer'
+                                self.parameter_errors += (
+                                    'ERROR: ' +
+                                    msg.format(year, pname, pval) +
+                                    '\n'
+                                )
```
The `elif` statement is tested, but the three following statements are not tested.  They are not tested because there are no behavior parameters with `integer_value` being true.   So, maybe just `#` comment out those three statement explaining that the commented statements can be made active if ever an integer behavioral response parameter is added to the `behavior.json` file.

Does that make sense?

martinholmer:
@hdoupe, pull request #1952 looks great!  Thanks for all the work involved in adding validity checking for revisions in Behavior class parameters.  And also, thanks for finding several needed improvements in the validity checking for Policy parameters.

martinholmer:
@hdoupe,  I notice one of the improvements I've made based on your efforts is causing conflicts in the `parameters.py` file.   After you update your branch to be like the newest version of the master branch, those conflict should go away.  At that stage, all the tests should pass and the code coverage report would say there are three untested statements in the `behavior.py` file.  Your can just skip those lines in the coverage test because there are currently no parameters in `behavior.json` that are integer parameters.  So, the lines would look something like this:
```
+                        elif int_param_type:
+                            if not pval_is_int:  # pragma: no cover
+                                msg = '{} {} value {} is not integer'  # pragma: no cover
+                                self.parameter_errors += (
+                                    'ERROR: ' +
+                                    msg.format(year, pname, pval) +
+                                    '\n'
+                                )  # pragma: no cover
+                        else:  # param is float type
```
Then I'll be happy to merge #1952.  Thanks again for all the great work.

hdoupe:
@martinholmer thanks! And thanks for thoroughly reviewing these changes and walking me through this part of Tax-Calculator. I enjoyed learning more about the parameter processing logic in Tax-Calculator. I think I have a much better grasp about how it works now. Also, a second set of eyes to review complex code such as this makes a big difference.

PR #1954, opened 2018-04-06
Add two optional arguments to Policy implement_reform method

martinholmer:
The need for this enhancement has been pointed out by @hdoupe in #1951.
Now the default behavior of the `implement_reform` method is to print any reform warnings and raise any reform errors as a ValueError.  This is desirable for Tax-Calculator users who are programming in Python on their local computers.  The new method arguments allow the cli and tbi code to defer handling of warnings and errors until after the `implement_method` is called.


hdoupe:
@martinholmer This looks good to me.

PR #1955, opened 2018-04-06
Clarify validation of _BE_subinc_wrt_earnings parameter value

martinholmer:
The need for this enhancement has been pointed out by @hdoupe in #1951.
By way of background, the ParameterBase class stores boolean parameters as numpy arrays with data type `numpy.int8`.


hdoupe:
@martinholmer this is more clear. Thanks for quickly updating check.

Issue #1956, opened 2018-04-06
Parameter validation allows bool value for float parameters


hdoupe:
Cosnider `_AMEDT_rt`. It should be a float:

```
>>> dd = taxcalc.Policy.default_data(start_year=2018, metadata=True)
>>> dd['_AMEDT_rt']['integer_value']
False
>>> dd['_AMEDT_rt']['boolean_value']
False
```

But you can do:

```
>>> import taxcalc
>>> taxcalc.__version__
'0.18.0'
>>> pol = taxcalc.Policy()
>>> ref = {2019: {'_AMEDT_rt': [True]}}
>>> pol.implement_reform(ref)
>>> pol.reform_errors
''
>>> pol.reform_warnings
''
>>> pol._AMEDT_rt
array([0.009, 0.009, 0.009, 0.009, 0.009, 0.009, 1.   , 1.   , 1.   ,
       1.   , 1.   , 1.   , 1.   , 1.   , 1.   ])
```

Thus, there are no errors or warnings if you try to do a reform with this value as a boolean type.

martinholmer:
Issue #1956 has been resolved by merge of pull request #1960.

PR #1960, opened 2018-04-08
Strengthen type checking of reform parameters

martinholmer:
This pull request attempts to resolve the bug report by @hdoupe in issue #1956.
It does so in a way that strengthens the type checking of all parameters (boolean, integer and float) in a reform dictionary passed to the Policy class `implement_reform` method.


hdoupe:
@martinholmer this looks good to me.

martinholmer:
@hdoupe,  Thanks for the review.  Merging now.

PR #1968, opened 2018-04-12
Add copy constructor option to Growfactors and Records classes

martinholmer:
This pull request adds an optional argument to both the Growfactors and Records class constructor that permits the constructor to operate like a C++ copy constructor.

These changes, which have no effect on tax results, are part of an ongoing effort to improve Tax-Calculator memory-management logic.


martinholmer:
The changes in pull request #1968 were developed under the mistaken belief that the memory-usage consequences of using `copy.deepcopy` in the Calculator constructor were contributing to the large size of Calculator class instances.  So, this pull request is being closed.

PR #1986, opened 2018-04-27
Use new CPS data

martinholmer:
This pull request uses new CPS data files generated by [taxdata pull request 183](https://github.com/open-source-economics/taxdata/pull/183) and [taxdata pull request 184](https://github.com/open-source-economics/taxdata/pull/184).


martinholmer:
The new CPS data that was being introduced in pull request #1986 was found to be faulty in several ways.

Issue #1981, opened 2018-04-24
Using behavior class to compare calculators under different growth assumptions/reforms


derrickchoe:
@MattHJensen @martinholmer @andersonfrailey @hdoupe @codykallen 
Hello all,

I'm currently working on a project that involves modelling the effects of additional real growth on charitable giving. I'm making a few modifications to one of my local branches to allow for an income effect on charity in behavior.py, and I am using the code for the income elasticity of taxable income as a guide. 
```
                if calc2.behavior('BE_subinc_wrt_earnings'):
                    # proportional change in after-tax income
                    with np.errstate(invalid='ignore'):
                        pch = np.where(calc1.array('aftertax_income') > 0.,
                                       (calc2.array('aftertax_income') /
                                        calc1.array('aftertax_income')) - 1.,
                                       0.)
                    inc = (calc2.behavior('BE_inc') *
                           pch * calc1.array('e00200'))
                else:
                    # dollar change in after-tax income
                    # Note: combined is f.unit's income+payroll tax liability
                    dch = calc1.array('combined') - calc2.array('combined')
                    inc = calc2.behavior('BE_inc') * dch
```

In the second to last line above, you'll find that the dollar change in after-tax income is simplified as the difference between one's total tax liabilities. However, in cases where one calculator has additional income growth assumptions, this difference should be:
```
calc2.array('aftertax_income') - calc1.array('aftertax_income')
```
Is there any reason that the difference between combined tax liabilities is preferred?

I'm wondering if the behavior class is meant to be used only with calculators under identical growth conditions. If so, I'm happy to model the income effect on charity using different methods. I'm currently using the `behavior.response` function to model the way tax filers respond to both new tax law and income growth. If these two effects are best modeled separately, please let me know! Thanks as always for your advice.

martinholmer:
@derrickchoe said in issue #1981:
> In the second to last line above [in the `Behavior.response` function], you'll find that the dollar change in after-tax income is simplified as the difference between one's total tax liabilities. However, in cases where one calculator has additional income growth assumptions, this difference should be:
```
calc2.array('aftertax_income') - calc1.array('aftertax_income')
```
Is there any reason that the difference between combined tax liabilities is preferred?

Yes, because using the Behavior response parameters is incompatible with using the growdiff_response parameters to simulate "additional income growth".

@derrickchoe continued:
[...]

PR #1992, opened 2018-05-07
Add validity checking for non-behavior assumption parameters

martinholmer:
This pull request moves the two `Behavior._validate_*` methods to the ParametersBase class and then uses them to add validity checking for consumption, growdiff_baseline, and growdiff_response parameters.  It also copies the behavior/growdiff_response consistency checking logic from the TaxCalcIO class to the Calculator class `read_json_param_objects` function, which now can return an ERROR message regarding an improper user specification.


taxcalc/calculate.py
```diff
@@ -1174,6 +1174,10 @@ def read_json_param_objects(reform, assump):
             )
         else:
             raise ValueError('reform is neither None nor string')
+        # raise error if specifying both behavior and growdiff_response
+        if behv_dict and gdiff_resp_dict:
+            msg = 'both behavior and growdiff_response are specified'
+            raise ValueError('ERROR: ' + msg + '\n')
```
hdoupe: @martinholmer do you think this type of integrated test is a rare situation? That is, do you think there will be more tests or validation checks that need to know information about the parameters specified for different TC classes (Policy, Behavior, Consumption, etc.)?

martinholmer: No, I don't think it is that rare a situation.  This seems to be what came up in Tax-Calculator issue #1981.  There will be a few more when I add the Growthmodel class and its parameters in the Growmod class.
Seems as if TaxBrain needs to put its call to `Calculator.read_json_param_objects` in a try statement.

hdoupe: > No, I don't think it is that rare a situation. This seems to be what came up in Tax-Calculator issue #1981. There will be a few more when I add the Growthmodel class and its parameters in the Growmod class.

Ok, thanks.

>  Seems as if TaxBrain needs to put its call to Calculator.read_json_param_objects in a try statement.

Sure, that would work. 

What do you think about adding a method to the Calculator class that does a validation test before `increment_year` is called or on initialization? That way, the `read_json_params_objects` focuses on reading the JSON files and the validation is mostly done elsewhere. This method could look something like:

```
@staticmethod
def integrated_validation(pol, behv, growdiff_resp, growdiff_base):
    assert not (growdiff_resp.has_any_response() and behv.has_any_response())

    ... do some more tests on the objects here
```

Note: you can find a similar check in `tbi.tbi_utils.calculate`: 
```
    # always prevent both behavioral response and growdiff response
    if behv.has_any_response() and growdiff_response.has_any_response():
        msg = 'BOTH behavior AND growdiff_response HAVE RESPONSE'
        raise ValueError(msg)
```

martinholmer: @hdoupe said:
> What do you think about adding a method to the Calculator class that does a validation test before `increment_year` is called or on initialization? That way, the `read_json_params_objects` [function] focuses on reading the JSON files and the validation is mostly done elsewhere. 

That was my first thought, but the Calculator object does not include any growdiff information.
I couldn't see an easy way around that problem.  Did I miss an easy solution?

hdoupe: Ahh, I see. So, the Growdiff information is applied to Growfactors, which is used by Policy and Records. Then, by the time, the Calculator class is initialized, the growdiff information is impossible to access. Is that right?

martinholmer: @hdoupe said:
> Ahh, I see. So, the Growdiff information is applied to Growfactors, which is used by Policy and Records. Then, by the time, the Calculator class is initialized, the growdiff information is impossible to access. Is that right?

Exactly.  Well put.  I can see a very expensive solution, but I view its benefits as being way less than its costs.  Can you see an easy solution to the problem you stated so well?

hdoupe: Well, I have one idea. What do you think about propagating the `Growdiff.has_any_response` up to Growfactors when the `Growdiff.apply_to` method is called? Then, it could be accessed from `self.__policy._gfactors.has_growdiff_response`. It's not the most elegant solution but it could work.

Plus, `integrated_validation` could be called from `reform_warnings_errors` since `Growfactors`, `Policy`, `Consump`, and `Behavior` are all available. 

This approach could look like:
```
# growdiff class
    def update(self, name, year, diff):
        """
        Add to self.gfdf[name][year] the specified diff amount.
        """
        if self.used:
            msg = 'cannot update growfactors after they have been used'
            raise ValueError(msg)
       
        if diff != 0: # logically equivalent to Growdiff.has_any_response
            self.has_growdiff_response = True

        self.gfdf[name][year] += diff
```

```
# Calculator class
    def calc_all(self, zero_out_calc_vars=False):
        """
        Call all tax-calculation functions for the current_year.
        """
        Calculator.integrated_validation(self.__policy, self.__behavior, self.__consumption,
                                         self.__policy._gfactors)
        # conducts static analysis of Calculator object for current_year
       
        ...

    @staticmethod
    def integrated_validation(pol, behv, consump, gfactors):
        assert not (gfactors.has_growdiff_response
                    and behv.has_any_response())
```

hdoupe: @martinholmer what was your expensive approach? Or, was this it?

martinholmer: @hdoupe said:
> What do you think about propagating the `Growdiff.has_any_response` up to `Growfactors` when the `Growdiff.apply_to` method is called? Then, it could be accessed from `self.__policy._gfactors.has_growdiff_response`. 

I see two sorts of problems with this approach:

1. It defers the parameter validity checking until the calculation stage.  I thought the whole point of the parameter validity checking capability we have built for TaxBrain is to catch problematic parameter values **before** calculation starts.

2. It relies on the user calling the **two** `apply_to` calls correctly and differently.  It is only the `growdiff_response.apply_to` call that needs "propagating"; not the `growdiff_baseline.apply_to` call.  If a user is knowledgeable enough to do this correctly, then that user doesn't need the parameter validity checking.  It's only the clueless Python programmers that need to help and I don't think we can rely on them to make the two `apply_to` calls correctly.

The only foolproof way to do this (as far as I can see) is to make the growdiff_baseline and growdiff_response objects two different classes.  But, as I said before, that's a lot of work.

hdoupe: Ok, I see. That was the only idea I had. I'll move the `read_json_param_objects` call into a try block.

hdoupe: Thank you for providing careful and clear explanation for why this approach wouldn't work so well.

hdoupe:
@martinholmer This PR looks good to me.

martinholmer:
@hdoupe,  Thanks for the careful review of pull request #1992.

hdoupe:
Great, thanks for the update @martinholmer.

PR #1997, opened 2018-05-11
Rename charitable_giving_response as quantity_response

martinholmer:
This pull request renames the `charitable_giving_response` utility function added in pull request #1995 as `quantity_response` and enhances its documentation.  The new name reflects the broad applicability of the function.  Also, as part of this pull request, the `_BE_subinc_wrt_earnings` Behavior parameter has been removed because the analysis it enabled is more easily done with the `quantity_response` utility function.

### DIFF
--- a/taxcalc/behavior.json
+++ b/taxcalc/behavior.json
@@ -37,25 +37,6 @@
         "range": {"min": -9e99, "max": 0.0}
     },
 
-    "_BE_subinc_wrt_earnings": {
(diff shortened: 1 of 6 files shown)


martinholmer:
Note that the decline in code coverage is inescapable because there are now no boolean assumption parameters in Tax-Calculator.  A subsequent pull request will add a boolean assumption parameter, which will provide an opportunity to test the three currently-untested statements.

PR #1998, opened 2018-05-12
Add empty shell of a GrowModel class

martinholmer:
This pull request is the first of a series of pull requests that will add a simple macroeconomic growth model to Tax-Calculator.  This model is not an OG-USA replacement.  It is a much simpler model that supports the kind of [macroeconomic growth analysis conducted by CBO](https://www.cbo.gov/sites/default/files/113th-congress-2013-2014/reports/49494-FiscalPolicies.pdf).  And because it is embedded in Tax-Calculator, it is easy to have the macro effects feed back to the micro simulation, providing an analysis that is dynamic and less partial equilibrium than the analysis provided by the Tax-Calculator Behavior class.

**The addition of the GrowModel class requires a change in the public API of Tax-Calculator.**
Now every JSON assumption file must contain a `"growmodel": {}` key-value pair in addition to the four key-value pairs that have been required for some time.  The documentation has been changed to reflect this, but old JSON assumption files will need to be revised before they can be used with the new version of Tax-Calculator.


martinholmer:
Note that one statement in the new `GrowModel` class is unavoidably untested given that currently there are no GrowModel parameters in the `growmodel.json` file that can have a value out of range.  When more parameters are added, it will be possible to write a test that exercises that statement.

Issue #2004, opened 2018-05-18
Confusing warning logic


martinholmer:
In [PolicyBrain issue 890](https://github.com/OpenSourcePolicyCenter/PolicyBrain/issues/890) @andersonfrailey relayed this comment from a TaxBrain user:
> For 2018 with a doubled standard deduction, I'd lower the threshold before a warning is triggered. it's not clear why there's a warning on making changes to decrease the standard deduction more than a fraction of a percent.  If the deduction was just doubled, it seems like a totally fair policy to both estimate if the deduction was scaled back and if was expanded.

feenberg:
On Fri, 18 May 2018, Martin Holmer wrote:

> 
> In PolicyBrain issue 890 @andersonfrailey relayed this comment from a TaxBrain user:
>
>       For 2018 with a doubled standard deduction, I'd lower the threshold before a warning is
>       triggered. it's not clear why there's a warning on making changes to decrease the
>       standard deduction more than a fraction of a percent. If the deduction was just doubled,
>       it seems like a totally fair policy to both estimate if the deduction was scaled back and
>       if was expanded.

Presumably the warning is to remind users that we don't have PUF data for 
itemized deductions that are lower than the standard deduction amount in 
the PUF year, so the revenue estimate is optimistic.

dan

martinholmer:
Pull request #2005, which is included in Tax-Calculator version 0.20.1, eliminates some incorrect standard-deduction warnings.  However, as mentioned in #2005 the most direct way to resolve this issue is to fix the shortcomings of the PUF itemizeable expense variables, and then remove all the warning logic.  There are two different kinds of shortcomings in the PUF data:

1. PUF data include positive itemizeable expense amounts **only** for those who actually itemize.

2. Several positive itemizeable expense variables are capped at some percent of AGI so policy reforms that change those percentage caps may produce unexpected results.

PR #2005, opened 2018-05-21
Fix policy-parameter-value warning logic

martinholmer:
This pull request fixes problems that were revealed when investigating issue #2004.

The warning logic is complex and difficult to maintain for both Tax-Calculator and TaxBrain.  The warning logic exists **ONLY** because the flaws in the PUF data related to itemized expense amounts have never been fixed (see, for example, [taxdata issue 32](https://github.com/open-source-economics/taxdata/issues/32), which is open but has seen no activity for a year).  The fact that TaxBrain users are (rightfully) complaining about the warning messages suggests that fixing the itemized expense amounts in the PUF data should be a high priority in the taxdata repository.

@MattHJensen @andersonfrailey @hdoupe


hdoupe:
@martinholmer it's taking me some time to get going this morning. I'm having trouble figuring out what the `_default_value` method is doing. Is it only for the `_STD` or `_STD_dep` variables? If so, what special function is it providing for these parameters?

martinholmer:
@hdoupe asked:
>  I'm having trouble figuring out what the `_default_value` method is doing.  Is it only for the `_STD` or `_STD_dep` variables?

Yes, because those are the only two parameters that we want to warn about when their value falls below what it is in the PUF data, which has (at the moment) no information about itemizable expenses for those who do not itemize.

> If so, what special function is it providing for these parameters?

It produces the indexed 2013 values of either `_STD` or `_STD_Dep` for each year after 2013.  Only if users of PUF data who try to specify a lower value than the 2013 indexed value will experience a warning.

hdoupe:
@martinholmer said

> Yes, because those are the only two parameters that we want to warn about when their value falls below what it is in the PUF data, which has (at the moment) no information about itemizable expenses for those who do not itemize.

That makes sense. Thanks for explaining this. Do you think the `_default_value` method should contain this information in the docstring? 

> It produces the indexed 2013 values of either _STD or _STD_Dep for each year after 2013. Only if users of PUF data who try to specify a lower value than the 2013 indexed value will experience a warning.

Ah, I see. That's the key part that I was missing.

I think I understand what's going on now. Thanks for breaking this down for me.

hdoupe:
@martinholmer [`b6c81cb`](https://github.com/open-source-economics/Tax-Calculator/pull/2005/commits/b6c81cb491c6c1fe74fbffc22817b4b534c81e1b) looks good to me. Thanks for updating this.

Issue #2042, opened 2018-08-06
About changes to taxcalc/utils.py on tbi-outputs branch


martinholmer:
@hdoupe, I notice that you're adding much needed descriptions of the output tables created in the `taxcalc/tbi/` code: the `RESULTS_TABLE_LABELS` and `RESULTS_TABLE_TAGS` dictionaries.

I have two questions about these changes:

1.  What does `_TAGS` refer to?  When I look at the contents, it seems as if something like `_INFO` would be a more meaningful name.  How would you describe in words the contents of this dictionary?

2.  Why are these two new dictionaries being added to `taxcalc/utils.py`?   They don't seem to be used anywhere but in the `taxcalc/tbi/` code, so why not put them there?

@lucassz

lucassz:
I'm glad to be able to clarify this. I'll start by saying that there is nothing new about this content; it was present in TaxBrain in one form or another, but we think it would be more consistent to have it in Tax-Calculator.

1. The tags refer to characteristics of the table that can be filtered on from a list of options.
For instance, [the table selected by default](https://user-images.githubusercontent.com/4793515/43721372-954f3386-9960-11e8-81df-7c3574b07175.png) in the current version of TaxBrain would be tagged as a 'difference' table, one displaying 'combined' tax and income 'bins'. I don't know if 'tags' is the clearest way to describe this; would you prefer 'toggles'?

2. I don't think there would be an issue putting them in `tbi`, I think they were added to `utils.py` because that's where the other constants used by `tbi` were stored.

hdoupe:
@martinholmer we haven't opened up a PR for this work yet because it's been somewhat exploratory in nature. Now, we are getting a better idea of what we think the API and output structure should look like. So, I think it's time to open a PR to get feedback and allow others to review. Per off-github discussion, @lucassz told me that he will be opening a Tax-Calculator WIP PR with these changes shortly.

martinholmer:
@lucassz said:
> The tags refer to characteristics of the table that can be filtered on from a list of options.
For instance, [the table selected by default](https://user-images.githubusercontent.com/4793515/43721372-954f3386-9960-11e8-81df-7c3574b07175.png) in the current version of TaxBrain would be tagged as a 'difference' table, one displaying 'combined' tax and income 'bins'. I don't know if 'tags' is the clearest way to describe this; would you prefer 'toggles'?

[...]

PR #2043, opened 2018-08-06
Implementation of new PolicyBrain outputs API

lucassz:
This PR is to discuss the implementation of the new PolicyBrain outputs API (OpenSourcePolicyCenter/PolicyBrain#921), continuing the discussion from #2042.


lucassz:
@martinholmer said:

> At a minimum you should preface the dictionary with a comment that describes its contents the way you explained them above.

I will make sure to do that soon.

> I still don't get where the term _TAGS comes from. At a minimum you should preface the dictionary with a comment that describes its contents the way you explained them above. The contents of the dictionary sounds like miscellaneous information about each table, so _INFO sounds more appropriate to me

I understand the confusion, and it's hard to tell what would be a good term. I think at a minimum, the name needs to convey that there can be several pieces of information associated with the table, and ideally that each one is part of a group among which only one attribute can be chosen. Maybe "label" and "label group" would be most evocative.

> @hdoupe and @lucassz, Now that I look more carefully at the contents of the RESULTS_TABLE_LABELS dictionary, it seems as if these are no row or column labels, but actually table TITLES, right? If so, why not name the dictionary RESULTS_TABLE_TITLES?

That sounds good to me, I can make this change.

lucassz:
@martinholmer said:

> Yes, but those constants are used elsewhere in Tax-Calculator or by users writing Python scripts that import the taxcalc package (or at lease that is the intention). But as far as I can see, the two new dictionaries are intended for `tbi` use only, right? If there are any other table constants in `utils.py` that are used **only** by `tbi` code, then they should also be moved to an appropriate location in the `taxcalc/tbi` directory.

It is of course true that these just-added constants won't be used by any other code in Tax-Calculator, though I could see the table titles being useful to outside packages. This would not, though, be true of the labels/tags which should be moved to `tbi` per your suggestion.

martinholmer:
@hdoupe and @lucassz,  Now that you've started revising the tbi output results, I'd like to add something that has been on my to-do list for a while.  If we can agree on the details of this change, you can include these changes in your pending pull request #2043.

Right now, the results that are returned from the `run_nth_year_taxcalc_model` function are expressed in dollars and filing units, and then TaxBrain has to scale and round each result.  Why not do the scaling and rounding in the `tbi` code when each table is a Pandas DataFrame?  In TaxBrain, the dollar results are scaled to billions of dollars and rounded to one significant decimal digit, while filing unit results are scaled to thousands of filing units and rounded to the nearest thousand.

I think it makes sense to express dollar variables as now (for example, 34.7 billion), but that it would be more sensible to express filing unit variables in millions rounded to two significant decimal digits (for example, 1.23 million instead of 1234 thousand).  I think this is more sensible for filing units because of the results in #1550.

I also think that eliminating all commas (so that 1,345.8 is replaced with 1345.8) is a good idea because then TaxBrain will not have to strip out all the commas when a user asks to download the results.  Removing the commas, makes it easier to construct a downloadable CSV file because TaxBrain won't have to quote each result (the quoted numbers are the biggest problem faced by users who want to download the tables and do anything other than load them into a spreadsheet).

What do you think about these changes?  

They would certainly clean up [these Tax-Calculator test results for CPS data](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/tests/tbi_cps_expect.txt), and clean up [these Tax-Calculator test results for PUF data](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/tests/tbi_puf_expect.txt), and should also simplify TaxBrain logic.

lucassz:
@martinholmer Thanks a lot for the input. Actually those changes are not only desirable but also necessary in order to retain the current functionality, since there is no post-processing of results by TaxBrain at all in the new development version. In other words, as it is in the `outputs` branch, results are not rounded or scaled, so we would need to add that back into TC.

> Why not do the scaling and rounding in the tbi code when each table is a Pandas DataFrame? In TaxBrain, the dollar results are scaled to billions of dollars and rounded to one significant decimal digit, while filing unit results are scaled to thousands of filing units and rounded to the nearest thousand.

Yes, it would make sense to do this right before converting the table to HTML.

> I also think that eliminating all commas (so that 1,345.8 is replaced with 1345.8) is a good idea because then TaxBrain will not have to strip out all the commas when a user asks to download the results. Removing the commas, makes it easier to construct a downloadable CSV file because TaxBrain won't have to quote each result (the quoted numbers are the biggest problem faced by users who want to download the tables and do anything other than load them into a spreadsheet).

I believe this part is actually not necessary under the new system, since TaxBrain does not generate CSV files from the HTML tables in which the results are formatted with commas, but rather serves the CSV file generated by the upstream project, in this case, Tax-Calculator. With the `pandas.DataFrame.to_csv` method that is used, the results are already formatted correctly, i.e. without commas (they perhaps should be rounded though, which they aren't).

hdoupe:
@martinholmer made some excellent points and at the end asked ([comment](https://github.com/open-source-economics/Tax-Calculator/pull/2043#issuecomment-410769280)):
> What do you think about these changes?

I agree with these changes. I am happy that they can now be made on the Tax-Calculator side. One of the upsides of the new approach to displaying the outputs is that the upstream project has much more control and thus responsibility over which outputs are displayed and how they are displayed. That is that PolicyBrain no longer needs to makes any decisions about how numbers are formatted, how many significant digits are displayed, etc. All of that can simply be set in Tax-Calculator.

One promising result from the new changes is that I was able to add a completely new table to PolicyBrain in about an hour and a half last week. To test out @lucassz's work so far, I added tables that show the MTR by income percentiles with respect to the primary earner (`e00200p`) and with respect to the spouse (`e00200s`). See taxcalc changes [here](https://github.com/hdoupe/Tax-Calculator/commit/a24b8190ed4a666044e7bf08e9e5b63a28c97c5c#diff-e5824163691bf741cda1621eb1f9d207) and PolicyBrain changes [here](https://github.com/OpenSourcePolicyCenter/PolicyBrain/pull/921/commits/24b90d1b4281f2b18b639961963ac383478f498a).

Note: I made the changes related to the MTR tables purely as a test of functionality, flexibility, and ease of use for the proposed outputs API and data structures.

martinholmer:
@hdoupe said:
> @martinholmer made some excellent points and at the end asked ([comment](https://github.com/open-source-economics/Tax-Calculator/pull/2043#issuecomment-410769280)):
>> What do you think about these changes?
>
>I agree with these changes. I am happy that they can now be made on the Tax-Calculator side. One of the upsides of the new approach to displaying the outputs is that the upstream project has much more control and thus responsibility over which outputs are displayed and how they are displayed. That is that PolicyBrain no longer needs to makes any decisions about how numbers are formatted, how many significant digits are displayed, etc. All of that can simply be set in Tax-Calculator.
>
> One promising result from the new changes is that I was able to add a completely new table to PolicyBrain in about an hour and a half last week. To test out @lucassz's work so far, I added tables that show the MTR by income percentiles with respect to the primary earner (`e00200p`) and with respect to the spouse (`e00200s`). See taxcalc changes [here](https://github.com/hdoupe/Tax-Calculator/commit/a24b8190ed4a666044e7bf08e9e5b63a28c97c5c#diff-e5824163691bf741cda1621eb1f9d207) and PolicyBrain changes [here](https://github.com/OpenSourcePolicyCenter/PolicyBrain/pull/921/commits/24b90d1b4281f2b18b639961963ac383478f498a).
>
> Note: I made the changes related to the MTR tables purely as a test of functionality, flexibility, and ease of use for the proposed outputs API and data structures.

Great to hear that your experiment with adding a new results table under the planned revision went so smoothly and quickly.

And I'm glad to hear you're comfortable with the scaling and rounding changes to be done in the `tbi` code as part of pull request #2043.  The one thing I forgot to mention is that some results tables contain percentage changes (like in aftertax expanded income).  I think now TaxBrain shows those variables rounded to the nearest one-tenth of a percent (like -3.4 percent).  I'm fine with that.  Does that make sense to you and @lucassz?

hdoupe:
@martinholmer asked
> The one thing I forgot to mention is that some results tables contain percentage changes (like in aftertax expanded income). I think now TaxBrain shows those variables rounded to the nearest one-tenth of a percent (like -3.4 percent). I'm fine with that. Does that make sense to you and @lucassz?

That sounds good to me.

martinholmer:
@hdoupe,  What's the work plan for Tax-Calculator pull request #2043?

hdoupe:
@martinholmer thanks for checking in. I'm still exploring backwards compatibility issues on the PolicyBrain side that could have an impact on how the functionality in this PR is implemented.

martinholmer:
@hdoupe, pull request #2043 contains this new utility function:
```
def pdf_to_clean_html(pdf):
    """Takes a PDF and returns an HTML table without any deprecated tags or
    irrelevant styling"""
    return (pdf.to_html()
            .replace(' border="1"', '')
            .replace(' style="text-align: right;"', ''))
```
I know I've used pdf (for Pandas DataFrame) in some code, but I'm increasingly worried about the potential confusion with Adobe's Portable Document Format (pdf) acronym.  I'm planning on changing my own code to replace `pdf` with something like `dframe`.  Given this, I'm hoping the name of the new function being proposed in #2043 can be improved.  Instead of `pdf_to_clean_html` how about `dframe2html`?  Would there ever be a case when we want to convert a Pandas DataFrame into `dirty` HTML?

MaxGhenis:
`df` is very common for both R and pandas dataframes, would that be an option?

martinholmer:
@MaxGhenis said in #2043:
> df is very common for both R and pandas dataframes, would that be an option?

You're right that the use of `df` as a name for a Pandas DataFrame object is quite common.
However, it causes problems in Tax-Calculator code because most of the code is run through pylint.
And pylint doesn't like a short variable name like `df`; it produces an invalid-name error.

MaxGhenis:
>pylint doesn't like a short variable name like df; it produces an invalid-name error.

That's unfortunate/silly, given 2-letter variable names conform to PEP8. Here's a way to make it OK for pylint: https://stackoverflow.com/a/35182977/1840471.

martinholmer:
@hdoupe,  Shall we just close Tax-Calculator PR #2043?
Given the pending switch to the new webapp framework, it there a need for the changes in #2043?

hdoupe:
@martinholmer I'm fine with closing this PR. The changes will be very similar to the ones proposed here. However, I think it would be helpful to start with a fresh PR.

As an update, on my local branch, I've synced the changes proposed in this PR with the master branch of Tax-Calculator and things are running smoothly with the new webapp framework. I've almost finished refactoring the parameter processing library in the webapp. I'm excited to share these changes and get some feedback.

martinholmer:
@hdoupe said about Tax-Calculator pull request #2043:
> I'm fine with closing this PR. The changes will be very similar to the ones proposed here. However, I think it would be helpful to start with a fresh PR.
>
> As an update, on my local branch, I've synced the changes proposed in this PR with the master branch of Tax-Calculator and things are running smoothly with the new webapp framework. I've almost finished refactoring the parameter processing library in the webapp. I'm excited to share these changes and get some feedback.

OK.  Sounds great.   I'll close pull request #2043 and await the new one.

PR #2075, opened 2018-10-09
Move code out of calculator.py & parameters.py into new json2dict utility function

martinholmer:
This pull request adds a new utility function, `json2dict`, that is used to read all JSON text files.  These changes not only consolidate duplicate code, but also provide built-in JSON syntax error detection when reading the six `taxcalc/*.json` files.


hdoupe:
#2075 looks good to me. This makes the `_read_json_x_text` methods much more readable.

Issue #2070, opened 2018-10-04
Add ability to read JSON reform and assumption files on web


martinholmer:
@andersonfrailey has brought to my attention that Tax-Calculator users would like the ability to specify the URL of a JSON reform or assumption file rather than downloading it from some website and specifying its file name on the local computer).  This seems like a reasonable enhancement request.

It seems as if the easiest way to implement this enhancement is to generalize the logic of the `Calculator.read_json_param_objects(reform, assump)` method so that the two arguments can be strings containing the URL of an online JSON file.  So, for example, the URL of the Brown-Khanna Grow American Incomes Now (GAIN) Act of 2017 that is on the Tax-Calculator GitHub website is as follows:
```
https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/reforms/BrownKhanna.json
```
The objective of the enhancement would be to make reading the JSON at a given URL something the `Calculator.read_json_param_objects(reform, assump)` method can do.  This enhancement would be available both to those who develop Python scripts using the taxcalc package and to those who use the `tc` command-line interface to the taxcalc package.

Does this approach make sense as a way to provide this enhancement?  Are there other easier approaches?  The approach described here provides the flexibility to read any URL; is that ability important?

MaxGhenis:
See https://github.com/open-source-economics/Tax-Calculator/issues/1644#issuecomment-352926693 where I provided code to do this:
```
import urllib2

REFORM_PATH = 'https://raw.githubusercontent.com/open-source-economics/Tax-Calculator/master/taxcalc/reforms/'
url = REFORM_PATH + 'TCJA_Reconciliation.json'

reform_text = urllib2.urlopen(url).read()

reform = Calculator.read_json_param_objects(reform_text, 'JCT_Behavior.json')
```

feenberg:
On Wed, 3 Oct 2018, Martin Holmer wrote:

> 
> @andersonfrailey has brought to my attention that Tax-Calculator users would like the ability to
> specify the URL of a JSON reform or assumption file rather than downloading it from some website and
> specifying its file name on the local computer). This seems like a reasonable enhancement request.

This is a nice feature. The only packages I have seen that can do this are 
SAS and Stata - the input statements can take a file name or http address. 
I expect there is some "generalized" input procedure in Python that will 
do this - every language and package should have this ability.

Dan

martinholmer:
@MaxGhenis said in issue #2070:
[...]

PR #2079, opened 2018-10-10
Allow user to specify a reform file from a URL

andersonfrailey:
This PR addresses issue #2070. I've added a few lines of code that will allow a user to enter a URL for either a JSON policy reform file or a JSON behavioral assumptions file rather than having to input a file on their local machine.

I originally planned on only trying to pull from online if the string was a URL, but URL validation proved to be a rather difficult task with a lot of different ways to do it, each with their own shortcomings. Instead, I'm relying on the errors produced by the `urllib` library. If we get an error from `urllib` we'll assume that the string passed into the function is a string representation of a reform and not a URL. I'm not totally in love with this implementation because if a user passes in an invalid URL they'll get this error:
```
ValueError: Policy reform text below contains invalid JSON:
Expecting value: line 1 column 1 (char 0)
Above location of the first error may be approximate.
The invalid JSON reform text is between the lines:
XX----.----1----.----2----.----3----.----4----.----5----.----6----.----7
01www.test.com
XX----.----1----.----2----.----3----.----4----.----5----.----6----.----7
```
I'd prefer they got an error telling them it was a bad URL. If anyone has ideas for URL validation I'd love to hear them. Alternatively, we could include additional information to the above error message saying they passed either an invalid URL or invalid JSON reform text. Downside of that would be we risk the error message becoming confusing.

For testing I picked a short reform file we have in the `reforms` folder on GitHub and compared the parameters value we get from a string representation of a JSON reform to ensure they match up. @MattHJensen mentioned moving the reforms files out of this repo at some point in the future. If this happens we'll need to update the test.

I wasn't able to write a test for reading an assumptions file at this time because I'm not aware of any JSON assumptions files that are hosted online. The closest we have is [here](https://github.com/open-source-economics/Tax-Calculator/blob/master/docs/assumpA.json.html), but that is actually an HTML file with JSON formatted text. I can add a sample assumptions JSON file, but the question becomes where to store it (do we put it in the reforms folder with all of the policy reforms or somewhere else). If at some point we do have an assumptions file online, we can add that extra test.

cc @martinholmer @MaxGhenis @hdoupe

### DIFF
--- a/taxcalc/calculator.py
+++ b/taxcalc/calculator.py
@@ -1140,7 +1140,7 @@ class Calculator(object):
         read_json_*_text method is called.
 
         Either of the two function arguments may also be valid URL strings that
-        point to valid JSON files hosted online.
+        begin with http and point to valid JSON files hosted online.
 
         The reform file contents or JSON string must be like this:
         {"policy": {...}}


martinholmer:
@andersonfrailey,  Thanks for PR #2079.  I'm going to look into the URL-validation issue that you raise:
> I'd prefer they got an error telling them it was a bad URL.
> If anyone has ideas for URL validation I'd love to hear them.

But for now, when you discussed adding tests, you said:
> I wasn't able to write a test for reading an assumptions file at this time because I'm not aware of any JSON assumptions files that are hosted online. 

What about [this JSON assumptions file](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/assumptions/economic_assumptions_template.json)?  Won't that one be sufficient for the new test?

martinholmer:
@andersonfrailey said in #2079:
> I originally planned on only trying to pull from online if the string was a URL, but URL validation proved to be a rather difficult task with a lot of different ways to do it, each with their own shortcomings.

Consider using the `requests` package (which seems to be part of the Anaconda distribution) to check if the strings passed into the `Calculator.read_json_param_objects` function are URLs.

hdoupe:
The `requests` package is pretty nice. If you go that route, consider the [`requests_mock`](https://requests-mock.readthedocs.io/en/latest/mocker.html) package for testing without having to actually do http requests. Then you could do something like:

```

def test_json_reform_url():
    """
    Test reading a JSON reform from a URL. Results from the URL are expected
    to match the results from the string.
    """
    reform_str = """
    {
    "policy": {
        "_FICA_ss_trt": {
            "2018": [0.130],
            "2020": [0.140]
        },
        "_FICA_mc_trt": {
            "2019": [0.030],
            "2021": [0.032]
        }
      }
    }
    """
    with requests_mock.Mocker() as mock:
        reform_url = ('https://raw.githubusercontent.com/open-source-economics/' +
                      'Tax-Calculator/master/taxcalc/reforms/ptaxes0.json')
        mock.get(url, text=reform_str, status=200)
        params_str = Calculator.read_json_param_objects(reform_str, None)
        params_url = Calculator.read_json_param_objects(reform_url, None)
    assert params_str['policy'] == params_url['policy']
```

This allows you to test out different status codes (200, 404, etc) and response structures.

andersonfrailey:
@martinholmer, good points. I'll use the assumptions file you linked to for testing.

After playing a bit with the `requests` package, I'm going to try and update this PR to use that rather than `urllib`. It looks like the types of error messages we get from passing a bad URL and those from passing in a random string are different so we could ignore errors from passing JSON formatted strings while still catching bad URLs. After I experiment a little I'll report back.

martinholmer:
@andersonfrailey said in PR #2079:
> After playing a bit with the `requests` package, I'm going to try and update this PR to use that rather than `urllib`. It looks like the types of error messages we get from passing a bad URL and those from passing in a random string are different so we could ignore errors from passing JSON formatted strings while still catching bad URLs. After I experiment a little I'll report back.

On further thought, perhaps our situation is so simple that we can check to see if we should consider the input string as an URL like this:
```
        elif isinstance(assump, str):
            if os.path.isfile(assump):
                txt = open(assump, 'r').read()
            elif assump.startswith('http'):
                txt = <code to read URL>
            else:
                txt = assump
```
What do you think?

@hdoupe

andersonfrailey:
You know, sometimes the simple solution is best. Thanks, @martinholmer. Using that method we'll get a 404 error when a user puts in a bad URL and a JSON error when they put in a bad string. Updating and running tests now.

martinholmer:
@andersonfrailey,  The code changes in the `calculator.py` file look OK now, but you have unused imports in lines 17-18:

screen shot 2018-10-13 at 9 44 03 am

martinholmer:
@andersonfrailey, the two new tests you added as part of PR #2079 look OK, but the `+` characters you used when specifying the `*_url` strings are completely unnecessary.

andersonfrailey:
Thanks, @martinholmer. Latest commit should fix all the issues.

martinholmer:
@andersonfrailey,  Thanks for the updates to PR #2079.  As far as I can see, there are just two remaining problems with this PR.

1.  You've spent all this time and effort to make `Calculator.read_json_param_objects(reform, assump)` accept string arguments that are URLs beginning with `http`, but you haven't told anybody about this feature.  You need to revise the `Calculator.read_json_param_objects(reform, assump)` docstring so that it explains this new option.

2. When I run the proposed `calculator.py` code through pylint, I get this error:
```
************* Module taxcalc.calculator
taxcalc/calculator.py:16: [C0411(wrong-import-order), ] standard import "import urllib" should be placed before "import numpy as np"
```
Can you fix that?

Once these two things are fixed, I'll be happy to merge PR #2079.

andersonfrailey:
@martinholmer, latest commit changes the oder of the import statements and adds a short line to the doc string about how users can pass a URL as an argument to the function.

martinholmer:
@andersonfrailey,  The documentation you added is generally good but not quite correct.  We cut corners (to make the code simpler) on identifying a string that was likely to be an URL.  We tried to get the JSON text from a URL if the function argument was a string (as opposed to `None`) and began with `http` (indicating hyper text transfer protocol).  But there are perfectly valid URLs that don't begin with `http` (such as `ftp` for file transfer protocol).  So, to not over-promise in the documentation, how about this modification?  Instead of
```
        Either of the two function arguments may also be valid URL strings that
        point to valid JSON files hosted online.
```
you document the new feature this way:
```
        Either of the two function arguments may also be valid URL strings that
        begin with http and point to valid JSON files hosted online.
```

andersonfrailey:
@martinholmer updated in latest commit.

PR #1925, opened 2018-03-15
Add Benefit Information to Tables

andersonfrailey:
This PR adds columns for total cost of benefits and total benefit value to the distribution and difference tables.

cc @martinholmer @MattHJensen @Kpinkelman

PR #1919, opened 2018-03-12
Universal Basic Income Addition - Updated

Kpinkelman:
Had an issue with the branch opened for the first PR, so opening a new PR with both the distribution and difference tables edited.

PR #2087, opened 2018-10-17
Refactor functions in the taxcalc.tbi subpackage and elsewhere

martinholmer:
This pull request refactors functions in the taxcalc/tbi directory so that other models in the Policy Simulation Library (PSL) collection of USA tax models can easily produce the aggregate and distributional tables expected by the TaxBrain web application.  It also adds several new tests; the old test results are unchanged.  It contains no changes in tax logic or tax results (but see the bottom of this comment).

In the course of this work, the logic of the functions that create the (distribution, difference, aggregate, and diagnostic) results tables has been modified so that the table entries are scaled.  Count variables are expressed in millions of filing units and amount variables are expressed in billions of dollars.  These scaling changes, which were agreed to in the discussion of #2043, relieves TaxBrain developers of the tedious task of scaling each column of each table in TaxBrain code.

Also, a bug was fixed in commit 425a38f.  Ever since the `ubi`, `benefit_cost_total`, and `benefit_value_total` variables were added to the difference table (in PR #1919, which was merged on 2018-03-14, and in PR #1925, which was merged on 2018-03-15), they have been incorrect.  Since March they have not been **reform-baseline differences** (which is what should be in the difference table), but rather **reform levels** (which are appropriate only for the distribution table).  This problem, which has never been reported by any users of Tax-Calculator or TaxBrain, has now been fixed.  The fact that no users of Tax-Calculator or TaxBrain reported this bug over a seven month period of time is an interesting observation about how people use the model and what results they look at.


MattHJensen:
Thanks for these changes @martinholmer.

Issue #2099, opened 2018-11-03
Different PUF weighted policy changes between API and tc CLI


evtedeschi3:
My policy reform is a very simple call to `_EITC_basic_frac` (thanks again for implementing):

```
// 	EBITC
{
   "policy": {
        "_EITC_basic_frac":
            {"2019": [0.5]}
  }
}
```

First, my CLI procedure:

```
python 'tc.py' 'puf.csv' 2019 --dump
python 'tc.py' 'puf.csv' 2019 --dump --reform 'ebitc.json'
```

And here's the weighted difference in `combined` that results using Stata. I'm not sure how the API handles non-filers when summing a policy change that involves a UBI, so I'm showing it both without and with nonfilers:

```
. table FLPDYR [iw=s006] if filer == 1, con(sum dtax) format(%20.0fc)

--------------------------------
   FLPDYR |            sum(dtax)
----------+---------------------
     2019 |       -4,620,714,171
--------------------------------


. table FLPDYR [iw=s006], con(sum dtax) format(%20.0fc)

--------------------------------
   FLPDYR |            sum(dtax)
----------+---------------------
     2019 |       -6,229,368,996
--------------------------------
```

So either $4.6B or $6.2B in 2019. Now for the API. Here's my code:

```
import pandas as pd
from taxcalc import *

recs = Records()

pol = Policy(start_year=2019)
calc1 = Calculator(policy=pol, records=recs)

cyr = 2019

calc1.advance_to_year(cyr)
calc1.calc_all()
itax_rev1 = calc1.weighted_total('combined')

reform_filename = 'ebitc.json'

params = Calculator.read_json_param_objects(reform=reform_filename, assump=None)

pol.implement_reform(params['policy'])
calc2 = Calculator(policy=pol, records=recs)

calc2.advance_to_year(cyr)
calc2.calc_all()
itax_rev2 = calc2.weighted_total('combined')
itax_diff = itax_rev2 - itax_rev1

print('{}_CLP_itax_rev($B)= {:.3f}'.format(cyr, itax_rev1 * 1e-9))
print('{}_REF_itax_rev($B)= {:.3f}'.format(cyr, itax_rev2 * 1e-9))
print('{}_DIFF_itax_rev($B)= {:.3f}'.format(cyr, itax_diff * 1e-9))
```

And here's the output:

```
2019_CLP_itax_rev($B)= 3031.060
2019_REF_itax_rev($B)= 3025.865
[...]

PR #2102, opened 2018-11-05
Clarify documentation of filer input variable

martinholmer:
The name of the input variable, `filer`, is misleading because some might think a variable with that name is an output variable that indicates whether or not the unit filed an income tax return in the Tax-Calculator simulation.  This would appear to be what happened in issue #2099, which was raised by one of our most advanced users.

Do people think we should change this variable name in the taxdata and Tax-Calculator repositories?  If so, what should the new variable name be?  Something like `data_source`?  Other ideas?

Can people suggest better `records_variables.json` documentation text than what is proposed in this pull request?

The focus of this pull request has changed as described [here](https://github.com/open-source-economics/Tax-Calculator/pull/2102#issuecomment-436361073) and [here](https://github.com/open-source-economics/taxdata/issues/288).

@MattHJensen @andersonfrailey @hdoupe @evtedeschi3

### DIFF
--- a/read-the-docs/source/public_api.rst
+++ b/read-the-docs/source/public_api.rst
@@ -70,16 +70,10 @@ taxcalc.Growfactors
 .. autoclass:: taxcalc.Growfactors
    :members:
 
-taxcalc.macro_elasticity
-------------------------
+taxcalc.Parameters
+------------------
 
-.. automodule:: taxcalc.macro_elasticity
-   :members:
-
-taxcalc.ParametersBase
-----------------------
-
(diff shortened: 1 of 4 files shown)


martinholmer:
Tax-Calculator pull request #2102 has been changed to eliminate the `filer` variable from the Records.USABLE_READ_VARS set.  This is an interim state of affairs lasting only until [taxdata issue 288](https://github.com/open-source-economics/taxdata/issues/288) is resolved, in which case the renamed and redefined variable will be reintroduced into the Records.USABLE_READ_VARS set.

PR #2107, opened 2018-11-07
Revise Cookbook recipe 2 to show use of Behavioral-Responses behresp package

martinholmer:
This pull request revises the Python code in Cookbook recipe 2 so that it uses the `response` function in the new Behavioral-Response `behresp` package (instead of the old Tax-Calculator `Behavior` class, which is scheduled for removal from Tax-Calculator).  The results of the recipe are exactly the same as before.


MattHJensen:
Very nice! Seems like it will be a smooth transition for users.

PR #2137, opened 2018-11-26
More revisions to www.PSLmodels.org card catalog information

martinholmer:
Documentation changes only.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,5 +1,6 @@
-This project is in the
-public domain within the United States.
+## LICENSE
+
+This project is in the public domain within the United States.
 
 Additionally, we waive copyright and related rights in the work
 worldwide through the CC0 1.0 Universal public domain dedication.
--- a/PSL_catalog.json
+++ b/PSL_catalog.json
@@ -7,25 +7,25 @@
         "data": "<p>USA federal individual income and payroll tax microsimulation model</p>"
     },
     "project_overview": {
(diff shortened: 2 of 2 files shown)


martinholmer:
@MattHJensen asked in merged pull request #2137:
> did you see that these changes removed all content but links from the PSL catalog card? 

Yes, I did realize that.  I would rather have had the project description (rather than a link to it) at the top of the Tax-Calculator card.  I submitted a request to PSL developers to make that possible, but have not received a reply.

> I thought the information on the card was quite useful -- it was very nice to be able to see an overview, citation guidelines, license information, maintainers, etc. in one place. Could you tell us about your goals for this PR?

These changes were attempts to correct problems in the old card.  For example, the old entry for the License was misleading.  It said it was Creative Commons, but actually when you read the `LICENSE.md` page it's more complicated.  That pages starts out saying: `This project is in the public domain within the United States.`  And there are many more details (and they seemed to be important because they were mentioned).  So, it seemed like saying in the link that Tax-Calculator had an open-source license, and letting an interested party click through for the details, was an improvement.

Also, the old entry under the ChangeLog was misleading because the entry was not the changelog, but just the first (that is, most recent) item in the changelog.  Given that the PSL title (which modelers have no control over) says "here is the changelog", it seemed sensible to me to provide a link to the changelog.   Showing just the first item in the changelog is not what the title promised and could often be misleading because it has no context.

And the list goes on from there.

Issue #2193, opened 2019-01-14
Double-counting of expanded_income


martinholmer:
The `expanded_income` variable is not used in the calculation of income or payroll tax liabilities.  However, it is used to sort filing units into decile or dollar bins in the distribution and difference tables.  The current definition of `expanded_income` includes two types of pension-related components:

1. The amount of employee salary-reduction contributions to tax-favored employer-sponsored defined-contribution (DC) pension plans; this amount is not included in the IRS-SOI PUF earnings (`e00200`) variable, and therefore, needs to be included in `expanded_income`.

2. The amount of distributions from pension plans; in the IRS-SOI PUF this is the sum of `e01400`, taxable IRA distributions, and `e01500`, total "pensions and annuities," which includes DC pension account distributions (from 401(k) plans, for example) **plus** defined-benefit (DB) pension benefits.

Many filing units have zeros for both these amounts in a given year.  Some have a positive amount for item 1 and zero for item 2 (working-age people not yet receiving pension distributions) and others have zero for item 1 and a positive amount for item 2 (older people who are no longer working and are receiving pension distributions).  However, there is a non-trivial group that have positive amounts for both item 1 and item 2 in the same year.

Regarding this latter group, Alan Viard (AEI) and his research associate, Erin Melly, have pointed out a double-counting problem.  If a filing unit contributes $2000 to a DC pension plan (item 1) and in the same year withdraws $5000 from that (or another) DC pension plan (item 2), the inclusion of the sum of the two items in `expanded_income` is incorrect.  What should be included in `expanded_income` in this example is not $7000 (as in the current code) but rather $5000 (which eliminates the double-counting of the pension contribution).

This bug in the construction of `expanded_income` has been in the code since the introduction of imputed amounts for item 1 in pull request #2056, which was merged into the master branch on September 5, 2018.

Many thanks to Alan Viard and Erin Melly for pointing out this bug in Tax-Calculator.

martinholmer:
Issue #2193 has been resolved by pull request #2194.

PR #2194, opened 2019-01-14
Fix pension-related double-counting in the expanded_income variable

martinholmer:
This pull request fixes the logic bug described in issue #2193.  The elimination of pension-related double-counting in `expanded_income` has no effect on income or payroll tax liabilities.  But, because `expanded_income` is used to sort filing units into deciles or bins in the distribution and difference tables, eliminating the double-counting will cause some filing units to fall into different deciles or bins than they did under the current logic.  Because the double-counting occurs only among filing units that have positive amounts for both pension contributions and pension distributions in the same year, the effects of the bug fix on the distributional tables is modest.  Below is a comparison of the 2018 expanded-income-decile distribution table for current-law policy with the old (master branch) version on the left and the new (bug-fix branch) version on the right:

screen shot 2019-01-14 at 8 24 22 am

As you can see, the bug fix reduces total `expanded_income` by about $45.1 billion, which is a reduction of about three-tenths of one percent.  The AGI and tax liability amounts on the ALL row are the same, but the decile amounts are somewhat different because the filing units are now being sorted with `expanded_income` that has the pension-related double-counting removed.


martinholmer:
The change in the definition of `expanded_income` implemented in #2194 was backed out in #2229 for the reasons described in the #2229 conversation.

PR #2229, opened 2019-02-09
Simplify calculation of expanded_income

martinholmer:
After reviewing how JCT, CBO, and TPC define expanded income, the changes in #2194 seem undesirable for a couple of reasons: (1) they make it more difficult for users who want to define their own expanded-income concept by adding or subtracting items from the Tax-Calculator's concept of expanded income, and (2) the limited data available to Tax-Calculator make the #2194 changes incorrect because of the inclusion of defined-benefit pension benefits in `e01500`.


MattHJensen:
> they make it more difficult for users who want to define their own expanded-income concept by adding or subtracting items from the Tax-Calculator's concept of expanded income,

Could you explain why this is? I don't understand based on looking at the commits. 

> the limited data available to Tax-Calculator make the #2194 changes incorrect because of the inclusion of defined-benefit pension benefits in `e01500`.

Aren't the changes in #2194 still an improvement, even if they aren't optimal?

martinholmer:
@MattHJensen asked in PR #2229:
>> they make it more difficult for users who want to define their own expanded-income concept by adding or subtracting items from the Tax-Calculator's concept of expanded income,
>
> Could you explain why this is? I don't understand based on looking at the commits.

Because there is no calculated variable for the double-count amount.  So, if you subtract `pencon_p` and `pencon_s` from Tax-Calculator's `expanded_income`, you would be subtracting the double-counted amount twice.

martinholmer:
@MattHJensen asked in PR #2229:
>> the limited data available to Tax-Calculator make the #2194 changes incorrect because of the inclusion of defined-benefit pension benefits in `e01500`.
>
> Aren't the changes in #2194 still an improvement, even if they aren't optimal?

No, I don't think they are an improvement.  They make it more difficult for users to define their own expanded-income concept.  

And after more review of the JCT, CBO, and TPC concepts, they are all different, sometimes different in substantial ways.  This will all be documented in our future document on expanded income.

The bottom line here, is that there is no such thing as a "line" along which expanded income can be said to "improve" and eventual get to the end of the "line" where expanded income is defined optimally.

martinholmer:
@MattHJensen said in PR #2229:
>> > the limited data available to Tax-Calculator make the #2194 changes incorrect because of the inclusion of defined-benefit pension benefits in `e01500`.
>
> Aren't the changes in #2194 still an improvement, even if they aren't optimal?

I said no before, but I want to explain in more detail why I don't think this is an improvement.

First, as I said before, the #2194 changes make it more difficult for users to specify variants of the default `expanded_income`, which is an important ease-of-use issue.

Second, I don't think much of the "double-counting" is really double-counting.  There are millions of people who served in the military, or worked in state/local or federal government, who are receiving defined-benefit pension benefits well before they are sixty years old.  Most of those people get another job until they want to actually retire from the work force.  In many cases those "bridge" jobs offer 401(k) plans.  So, many of these people are making tax-favored pension contributions (that is, `pencon_?` is positive) and also have positive pension income (`e01500`).  It doesn't seem to me that their `pencon_?` amount should be removed from their `expanded_income`: certainly they are earning that money and paying payroll taxes on it.  And furthermore, because we have only the filing unit's `e01500` amount even if the `e01500` amount was a withdrawal from a tax-favored pension account, we don't know whose withdrawal it was.  

All things considered, I think the #2194 changes were counter-productive.

MattHJensen:
My preferred approach is the same as Viard's, which is to include all withdrawals and exclude all contributions. As Alan has described via email

> this approach is self-consistent on a lifetime basis, but it diverges significantly from the correct income measure in each year. Inclusion of both the contribution and the withdrawal results in double counting (which would make it indefensible as  a tax base) and also diverges significantly from the correct income measure in many years, but it may diverge less than my preferred approach and it is a defensible annual income classifier.

This approach should not make it any harder for users to define their own expanded_income measure. 

The additional difficulty that #2194 introduced for users to define their own income measures was a result of trying to deal with annual "double counting", but that wasn't the problem that Viard initially highlighted or the approach that Viard suggested. It is also not used by other modeling groups to my knowledge. 

I think the best way forward is to merge this PR and open a new PR to remove all contributions from expanded income. I would be happy to do the latter if it would be helpful.

martinholmer:
@MattHJensen said in #2229:
> The additional difficulty that #2194 introduced for users to define their own income measures was a result of trying to deal with annual "double counting", but that wasn't the problem that Viard initially highlighted or the approach that Viard suggested.

Actually, if you look back at the private email exchanges, Viard explained his full views **after** (not before) #2194.  His full view on this matter was offered only after he saw #2194.

PR #2254, opened 2019-03-08
Simplify the EITC_indiv parameter

### LINKED ISSUE #2199: Individualizing EITC removes joint filing phaseout bonus w/o notifying user
Setting the `_EITC_indiv` parameter to `True` deactivates the `_EITC_ps_MarriedJ_*kids` parameter, significantly reducing the phaseout start for affected taxpayers. The user is not notified by the `_EITC_indiv` documentation, other than the note, "the phaseout of the credit works slightly differently between the two." 

For context, I estimate that deactivating the phaseout start bonus for joint filers reduces the cost of individualizing the EITC by about $3B in the first year. 

See the two reforms below. The first sets `_EITC_indiv` to True and the second sets `_EITC_indiv` to True **and** raises the `_EITC_ps_MarriedJ_*kids` parameters dramatically. Both have the exact same effects. 

https://www.ospc.org/taxbrain/38925/
https://www.ospc.org/taxbrain/38926/

One fix would be to add documentation to `_EITC_indiv` explaining that it turns off the other parameter. Another fix would be to keep the` _EITC_ps_MarriedJ_*kids` parameters active when `_EITC_indiv` is active (with docs). 

I prefer the latter approach, because I think it is counterintuitive for `EITC_indiv` to automatically reduce the phaseout start for affected taxpayers, but it would break backwards compatibility by redefining `EITC_indiv`. 

Based on how I have observed users approach the individualized EITC option, I think it is best to keep the definition for the individualized EITC quite close to the definition for the current law EITC. Users expect to be able to make `EITC_indiv` `True` without it affecting their other parameter choices. 

Based on this observation, I also think we should add back childless EITC age eligibility rules (which are currently turned off by `EITC_indiv`). 

I do not think we should add back the investment income eligibility rules for now because we don't an investment income breakdown between the joint filers. 

I am happy to make these changes if others think they are reasonable. To summarize the changes that I'm proposing: 

- activate `_EITC_ps_MarriedJ_*kids` when `EITC_indiv` is True
- activate `EITC_*EligAge` when `EITC_indiv` is True
- add documentation to `EITC_indiv` to reflect the above two changes.

MattHJensen:
Resolves #2199. 

@martinholmer, could you please look this over?


martinholmer:
@MattHJensen,  Please don't include the changes to the `docs/index.html` file in this pull request.
We make those changes only with each release.  If you were to include this file in PR #2254, the documentation for the current release of Tax-Calculator would be wrong as soon as PR #2254 was merged.

I'll look at the substance of the proposed changes after you fix this problem.

martinholmer:
@MattHJensen,  You need to add some motivation and clarification for this PR.  For example, why is adding a new policy parameter "simplify the EITC_indiv parameter"?

MattHJensen:
> Please don't include the changes to the docs/index.html file in this pull request.
We make those changes only with each release. If you were to include this file in PR #2254, the documentation for the current release of Tax-Calculator would be wrong as soon as PR #2254 was merged.

The latest commit resets docs/index.html to master. (Thanks to this [stack overflow thread](https://stackoverflow.com/questions/39459467/remove-a-modified-file-from-pull-request))

> You need to add some motivation and clarification for this PR.

Good point. This PR simplifies the `EITC_indiv` parameter by narrowing its scope. 

After these commits, turning on `EITC_indiv` means that, for tax units filing jointly, the EITC is calculated for each spouse based on individual earnings. 

Previously, the `EITC_indiv` had this effect and also changed the investment income eligibility rules, changed the married filing jointly phaseout bonus (without actually changing the married filing jointly phaseout bonus parameter values), and changed the age eligibility rules. 

This PR adds a new parameter that turns on and off whether the EITC can be claimed by tax units that are married filing separately, `EITC_sep_filers_elig`. Under current law, `EITC_sep_filers_elig` is false. I expect that many users of `EITC_indiv` will want `EITC_sep_filers_elig` to be true.

martinholmer:
@MattHJensen,  Thanks for the big improvements in structural EITC reform logic in pull request #2254.

MattHJensen:
Thanks for the discussion and review @martinholmer!

Issue #2259, opened 2019-03-15
recipe00's reformA.json has comments that make it invalid


MaxGhenis:
https://pslmodels.github.io/Tax-Calculator/recipe00.py.html points to https://pslmodels.github.io/Tax-Calculator/reformA.json, which has comments in between the lines. Loading it produced this error:

```
reform_filename = './ingredients/reformA.json'
# This points to the same:
# reform_filename = 'https://pslmodels.github.io/Tax-Calculator/reformA.json'
params = tc.Calculator.read_json_param_objects(reform_filename, None)
# ... Error ...
ValueError: Text below contains invalid JSON:
Expecting value: line 1 column 1 (char 0)
Above location of the first error may be approximate.
The invalid JSON text is between the lines:
XXXX----.----1----.----2----.----3----.----4----.----5----.----6----.----7
0001<title>reformA.json</title><pre>
0002 
0003 
0004{
0005    "policy": {
0006	 
0007	"_II_em": {"2020": [1000]},
0008	 
0009	 
0010	"_II_rt5": {"2020": [0.36]},
0011	"_II_rt6": {"2020": [0.39]},
0012	"_II_rt7": {"2020": [0.41]},
0013	 
0014	"_PT_rt5": {"2020": [0.36]},
0015	"_PT_rt6": {"2020": [0.39]},
0016	"_PT_rt7": {"2020": [0.41]}
0017    }
0018}
0019</pre>
0020
XXXX----.----1----.----2----.----3----.----4----.----5----.----6----.----7
```

Removing the comments in between made it run OK. The comments at the top are safely ignored.

martinholmer:
@MaxGhenis,  Thanks for sharing your experience with cookbook recipe00 in [this comment](https://github.com/PSLmodels/Tax-Calculator/issues/2259#issue-421687320).  I'm sorry you're having difficulty following the recipe.

Let me try to explain.  You start out saying this:
> https://pslmodels.github.io/Tax-Calculator/recipe00.py.html points to https://pslmodels.github.io/Tax-Calculator/reformA.json, which ...

[...]

PR #2261, opened 2019-03-16
Ensure that file/http names passed to read_json_param_objects end in '.json'

martinholmer:
This pull request adds checks to the `Calculator.read_json_param_objects` static method that ensure any filename or string beginning with `http` ends with `.json`.  This error checking has been added in order to reduce the probability of the kind of confusion described in issue #2259.

### DIFF
--- a/taxcalc/calculator.py
+++ b/taxcalc/calculator.py
@@ -1054,7 +1054,7 @@ class Calculator():
     def read_json_param_objects(reform, assump):
         """
         Read JSON reform and assump objects and
-        return a single dictionary containing 6 key:dict pairs:
+        return a single dictionary containing four key:dict pairs:
         'policy':dict, 'consumption':dict,
         'growdiff_baseline':dict, and 'growdiff_response':dict.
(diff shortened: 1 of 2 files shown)


MaxGhenis:
More discussion in #2259 but just to note this doesn't resolve the issue described there, since I was pulling a `.json` file which was actually an HTML file:

```
In [1]: import taxcalc as tc
In [2]: tc.__version__
Out[2]: '1.1.0'
In [3]: tc.Calculator.read_json_param_objects('https://pslmodels.github.io/Tax-Calculator/reformA.json', None)
...
ValueError: Text below contains invalid JSON:
Expecting value: line 1 column 1 (char 0)
Above location of the first error may be approximate.
The invalid JSON text is between the lines:
XXXX----.----1----.----2----.----3----.----4----.----5----.----6----.----7
0001<title>reformA.json</title><pre>
```

i.e. this doesn't produce the new error message `assump does not end with '.json'` added in this PR.

Issue #2291, opened 2019-04-08
Why is specifying a policy reform so complicated?


martinholmer:
The way Tax-Calculator users are asked to specify a policy reform has remained the same since the beginning of the project.  Meanwhile, there have been several important ease-of-use enhancements.  Several years ago, a command-line interface to Tax-Calculator was introduced so that no computer programming was required.  And the new Tax-Brain interface to Tax-Calculator and associated USA tax models makes it much easier for Python programmers to use Tax-Calculator.  Given these improvements, the arcane syntax of specifying policy reforms presents new users with an unnecessary challenge in using Tax-Calculator.

In this comment, I describe a simple tax reform in words, and then show how this reform is specified using the current syntax.  After pointing out the  complexities (and inconsistencies) of the current syntax, I show how this reform could be specified in an easier way.

Consider a two-provision reform: (1) raise the FICA maximum taxable earnings to $300,000 in 2020 and leave it at that level for twelve years before resuming indexing; (2) raise the standard deduction in 2021 and continue to index it in subsequent years.

Using the current version of Tax-Calculator, this reform would be specified using the following Python dictionary:
```
reform = {
   2020: {
       '_SS_Earnings_c': [300000],
       '_SS_Earnings_c_cpi': False
   },
   2021: {
       '_STD': [[14000, 28000, 14000, 21000, 28000]]
   }
}
```

When first encountering this syntax, new users probably have the same reactions I had years ago when I started using Tax-Calculator.  Why the need for the leading underscore character at the start of each parameter name?  Why the need for brackets around the value of the maximum taxable earnings when its value is simply a number?  (After all, the boolean value of `_SS_Earnings_c_cpi` doesn't need brackets.)  Why the need for **two** sets of brackets to enclose the vector-valued `_STD` parameter values?  And finally, why is there a `_cpi` suffix on the end of the `_SS_Earnings_c` parameter (to turn off indexing) when that parameter is indexed by wages not prices?

It is not productive to speculate about how this reform syntax came to be.  It is far more productive to think about how it could be improved.

I think ease of use would be improved if users could specify this reform using the following Python dictionary:
```
reform = {
   2020: {
       'SS_Earnings_c': 300000,
       'SS_Earnings_c-indexed': False
   },
   2021: {
       'STD': [14000, 28000, 14000, 21000, 28000]
   }
}
```

[...]

PR #2292, opened 2019-04-08
Provide simpler syntax to specify policy reforms and assumption updates

martinholmer:
This pull request resolves issue #2291 by providing Tax-Calculator users with a less complicated syntax with which to specify policy reforms or updates to assumption parameters.

This pull request is the culmination of a series of enhancements to the Parameters class in recent pull requests  #2281, #2282, #2283, #2284, #2285, #2289, and #2290.  Only this #2292 pull request includes changes that make new versions of Tax-Calculator incompatible with earlier versions of Tax-Calculator.  One of the major changes is the one discussed in issue #2291: changing the format of Python dictionaries supplied to the `Policy.implement_reform`, `Consumption.update_consumption`, and `GrowDiff.update_growdiff` methods so that the new `param : year : value` format is the same as used in JSON reform/assumption files.  If you are specifying reforms with dictionaries (rather than in JSON files), you will have to edit the format of the dictionaries so that the old `year : param : value` format is transformed into the new `param : year : value` format.  See the new `Tax-Calculator/new_json.py` file for a script that may help you change your personal JSON reform files.

Taken together, all these enhancements to the Parameters class mean that any other model that is already importing the Tax-Calculator package can automatically use this new syntax by creating its own classes derived from the Tax-Calculator Parameters class.  An example, of the resulting simplification can be seen in [Business-Taxation pull request 88](https://github.com/PSLmodels/Business-Taxation/blob/0685fa5afe2c2a74c1dd5dc6230be687a570c62a/biztax/policy.py) that specifies a Business-Taxation Policy class derived from the Tax-Calculator Parameters class in just a few lines of code.

This pull request also brings the name of the CPI offset policy parameter into line with other parameter names by changing it from `cpi_offset` to `CPI_offset`.  A new parameter name-change capability will notify users who forget to use the new name.

The few lines of untested code cannot be tested until this pull request is merged.  When it is merged, the `@pytest.mark.skip` decorators will be removed and those lines of code will be tested.


martinholmer:
There has been a substantial discussion of issue #2291, but that conversation has widened to include the issue of whether or not to make Python reform dictionaries and JSON reform files have the same parameter, year, value order, and if so, which ordering should become the standard.

I agree that this is an important topic, but I think it needs more discussion.  Also, this pull request (#2292) is already quite large.  So, I propose that #2292 be merged and the additional issue still being discussed can be handled in a follow-up pull request.  Does that make sense?

@MattHJensen @andersonfrailey @hdoupe @codykallen @feenberg

martinholmer:
I said yesterday:
> I propose that #2292 be merged and the [parameter-year-value ordering] issue still being discussed can be handled in a follow-up pull request.

Actually, I think we have quickly reached a consensus on the ordering issue, so I'm going to implement the changes to the structure of Python reform/revision dictionaries in this pull request.

taxcalc/policy_current_law.json
```diff
@@ -56,16 +56,16 @@
...
+                      2015,
+                      2016,
+                      2017,
+                      2018],
         "indexable": true,
         "indexed": true,
-        "col_var": "",
-        "col_label": [],
+        "vi_name": "",
+        "vi_vals": [],
```
hdoupe: @martinholmer, what does "vi" stand for?

martinholmer: `vi` stands for `vector index`.

hdoupe: Ah, I see. Thanks.

hdoupe:
@martinholmer would you mind leaving #2292 open through Monday so that people have a chance to test it out?

taxcalc/policy_current_law.json
```diff
@@ -133,7 +133,7 @@
         "indexable": true,
         "indexed": false,
         "vi_name": "MARS",
-        "vi_vals": ["single", "joint", "separate", "headhousehold", "widow"],
+        "vi_vals": ["single", "mjoint", "marsep", "headhh", "widow"],
```
hdoupe: @martinholmer These names are displayed over the parameter boxes on the webapp:
Screen Shot 2019-04-12 at 11 38 54 AM

What do you think about choosing names that are more human readable like:

`["single", "married joint", "married separate", "head household", "widow"]`

The difference is only a few characters, and they are readable to those who are not familiar with the naming scheme.

martinholmer: I think the new values for the vector indexes are more succinct and perfectly understandable to those using Tax-Calculator (or Tax-Brain) via Python programming or the `tc` command-line tool.  If you think  users of the TaxBrain webapp need longer labels on the boxes, then by all means you should supply them with longer box labels.  That's your call.

martinholmer:
@hdoupe asked:
> would you mind leaving #2292 open through Monday so that people have a chance to test it out?

OK.

hdoupe:
@martinholmer `Policy.implement_reform` does not accept years that are strings:

```python
import taxcalc
import json

# 2.0 styled reform
ref = {
    "policy": {
        "STD": {"2018": [10001, 10000, 10000, 10000, 10000]},
        "STD-indexed": {"2018": False},
    }
}
ref_json = json.dumps(ref)

# Use read_json_param_objects to parse JSON reform
pol = taxcalc.Calculator.read_json_param_objects(ref_json, None)

# Note that keys are integers now.
print("parsed params: ", pol["policy"])

print("Parsed parameters load without errors.")
taxcalc.Policy().implement_reform(pol["policy"])

print("Parameters with string years throw an error.")
taxcalc.Policy().implement_reform(ref["policy"])

```

outputs:

```
(taxcalc-dev) HDoupe-MacBook-Pro:Tax-Calculator henrydoupe$ python test.py 
parsed params:  {'STD': {2018: [10001, 10000, 10000, 10000, 10000]}, 'STD-indexed': {2018: False}}
Parsed parameters load without errors.
Parameters with string years throw an error.
  File "test.py", line 23, in <module>
    taxcalc.Policy().implement_reform(ref["policy"])
  File "/Users/henrydoupe/Documents/Tax-Calculator/taxcalc/policy.py", line 108, in implement_reform
    self._update(reform, print_warnings, raise_errors)
  File "/Users/henrydoupe/Documents/Tax-Calculator/taxcalc/parameters.py", line 292, in _update
    raise ValueError(msg.format(name, year))
ValueError: ERROR: KEY STD YEAR 2018 is not an integer year
(taxcalc-dev) HDoupe-MacBook-Pro:Tax-Calculator henrydoupe$ 
```

It seems like `implement_reform` should accept years that are integers or strings. Users may not understand why the JSON files require the years to be strings, but the `implement_reform` method requires them to be integers. This would also make it easier to serialize and deserialize reform parameters that are accepted by `implement_reform`.

martinholmer:
@hdoupe reported in PR #2292 that:
> `Policy.implement_reform` does not accept years that are strings

Yes, and that is by design.  Thanks for calling attention to the need to improve various method docstrings to clarify how things work.  I've improved the docstrings in commit 75deaa3.

hdoupe:
Ok, I see. Glad my question was helpful.

PR #2319, opened 2019-05-10
Set Tuition and Fees Haircut to One

andersonfrailey:
The tuition and fees deduction is no longer available from 2018 onward (see [here](https://www.irs.gov/newsroom/tax-benefits-for-education-information-center)), but `ALD_Tuition_hc` in `policy_current_law.json` is set to zero for all years. This PR updates `policy_current_law.json` so that `ALD_Tuition_hc` is set to 1 for 2018 and beyond.

There are of course small changes to aggregated test results, but they all passed locally after copying over the new results.

This bug was originally reported by Alan Viard and @erinmelly


martinholmer:
@andersonfrailey said in pull request #2319:
> The tuition and fees deduction is no longer available from 2018 onward (see [here](https://www.irs.gov/newsroom/tax-benefits-for-education-information-center)), but `ALD_Tuition_hc` in `policy_current_law.json` is set to zero for all years. This PR updates `policy_current_law.json` so that `ALD_Tuition_hc` is set to 1 for 2018 and beyond.
>
> There are of course small changes to aggregated test results, but they all passed locally after copying over the new results.
>
> This bug was originally reported by Alan Viard and @erinmelly

Thank you Alan Viard and @erinmelly for reporting this bug.  
It seems as if the original specification of the TCJA missed this change.

We definitely need to fix this, but I'm not sure how best to fix this problem.  It is clear that this pull request is incomplete (in that the value of `ALD_Tuition_hc` has not been added to the `taxcalc/reform/TCJA.json` and `taxcalc/reforms/2017_law.json` reform files), and it may be inaccurate in that the appropriate value of the `ALD_Tuition_hc` parameter in 2026 and subsequent years is not clear from the cited IRS document.  Other haircut parameters changed by TCJA are (as best as we understand) slated to revert to their 2017 values beginning in 2026.  Is our understanding of the 2026 reversion in other haircut parameters correct?  Does the value of `ALD_Tuition_hc` revert in 2026?

andersonfrailey:
@martinholmer, after talking with Alan and doing some research myself, it's my understanding that this is unrelated to the TCJA. Rather, the tuition and fees deduction expires every year or two and must be expanded by Congress.

The tuition and fees deduction expired in 2017 – separate from the TCJA – but was retroactively extended by the [Bipartisan Budget Act of 2018](https://www.congress.gov/115/plaws/publ123/PLAW-115publ123.pdf#page=83) for the 2017 tax year. So the only thing that would cause this deduction to be renewed would be another act of Congress to do so. The IRS includes tuition and fees in a 2019 list of expired provisions [here](https://www.irs.gov/forms-pubs/legislative-impact-on-tax-forms). And a note on it's renewal for the 2017 tax year can be found [here](https://www.irs.gov/newsroom/three-popular-tax-benefits-retroactively-renewed-for-2017-irs-ready-to-accept-returns-claiming-these-benefits-e-file-for-fastest-refunds).

martinholmer:
@andersonfrailey,  Thanks for the [clarification](https://github.com/PSLmodels/Tax-Calculator/pull/2319#issuecomment-491391702) on the 2018 change in the above-the-line deductibility of tuition expenses, which was a change in current-law policy having nothing to do with TCJA.

Given this information, the details of your pull request seem to be correct.  So, I'll merge this bug-fix into the master branch now.   Thanks for making this contribution and doing the research of the legislative history of this change in current-law policy.

Issue #2363, opened 2019-07-23
Indexing Variable and Using CPI Offset causing errors


andersonfrailey:
Some Tax-Brain runs have been failing when a user attempts to change `CPI_offset` and toggle the `indexed` variable on another parameter. I have been able to recreate this error when working directly with taxcalc:

```
In [11]: pol = tc.Policy()                                                                                                                                                                                                                                                      

In [12]: reform = {"CTC_c-indexed": {2020: True}}                                                                                                                                                                                                                               

In [13]: pol.implement_reform(reform)                                                                                                                                                                                                                                           

In [14]: pol2 = tc.Policy()                                                                                                                                                                                                                                                     

In [15]: reform2 = {"CPI_offset": {2020: -0.0030}}                                                                                                                                                                                                                              

In [16]: pol2.implement_reform(reform2)                                                                                                                                                                                                                                         

In [17]: pol3 = tc.Policy()                                                                                                                                                                                                                                                     

In [18]: reform3 = {"CPI_offset": {2020: -0.0030}, 
    ...:            "CTC_c-indexed": {2020: True}}                                                                                                                                                                                                                              

In [19]: pol3.implement_reform(reform3)                                                                                                                                                                                                                                         
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-19-9fca5d29e042> in <module>
----> 1 pol3.implement_reform(reform3)

~/anaconda/envs/taxbrain-dev/lib/python3.6/site-packages/taxcalc/policy.py in implement_reform(self, reform, print_warnings, raise_errors)
    126         the expected structure of the reform dictionary.
    127         """
--> 128         self._update(reform, print_warnings, raise_errors)
    129 
    130     @staticmethod

~/anaconda/envs/taxbrain-dev/lib/python3.6/site-packages/taxcalc/parameters.py in _update(self, revision_, print_warnings, raise_errors)
    333             raise ValueError(self.parameter_errors)
    334         # optionally apply CPI_offset to inflation_rates and re-initialize
--> 335         known_years = self._apply_cpi_offset_in_revision(revision)
    336         if known_years is not None:
    337             self._set_default_vals(known_years=known_years)

~/anaconda/envs/taxbrain-dev/lib/python3.6/site-packages/taxcalc/parameters.py in _apply_cpi_offset_in_revision(self, revision)
    848         for year in sorted(revision.keys()):
    849             for name in revision[year]:
[...]

PR #2364, opened 2019-07-24
Modify how taxcalc handles CPI index toggling

andersonfrailey:
This PR addressed issue #2363. I've slightly modified how taxcalc handles reform parameters ending in `-indexed` to avoid getting an error like the one defined in the linked issue. I also added a test for this type of reform. I'm happy to update this PR if @martinholmer or @MattHJensen decided on a different method of fixing the bug.

cc @hdoupe

### DIFF
--- a/taxcalc/parameters.py
+++ b/taxcalc/parameters.py
@@ -847,8 +847,6 @@ class Parameters():
         )
         for year in sorted(revision.keys()):
             for name in revision[year]:
-                if name.endswith('-indexed'):
-                    name = name[:-8]
                 if self._vals[name]['indexed']:
                     if name not in known_years:
                         known_years[name] = kyrs_in_revision
--- a/taxcalc/parameters.py
+++ b/taxcalc/parameters.py
@@ -847,6 +847,8 @@ class Parameters():
         )
         for year in sorted(revision.keys()):
             for name in revision[year]:
+                if name.endswith('-indexed'):
+                    name = name[:-8]
                 if self._vals[name]['indexed']:
                     if name not in known_years:
                         known_years[name] = kyrs_in_revision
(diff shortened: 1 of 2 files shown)


MattHJensen:
Thanks very much, @andersonfrailey.

martinholmer:
@andersonfrailey,  Thanks for the bug report in issue #2363 and the proposed code change in #2364 .

This bug remained undetected for a long time because the pytest suite did not include a reform that changed the value of `CPI_offset` in the same year that the indexing status of a tax policy variable was being changed.  Your code change and the new test you added demonstrate that changing `CPI_offset` in the same year as a policy variable's indexing status is changed no longer causes Tax-Calculator to crash.  That's a big improvement.  But the new test should be stronger: we would like some assurance that your proposed code change not only avoids a crash, but also produces the correct result.

Can you revise the new test to do that?   

One way to do this is to simulate two reforms:
```
reform1 = {'CTC_c-indexed': {2020: True}}
offset = -0.005
reform2 = {'CTC_c-indexed': {2020: True}, 'CPI_offset': {2020: offset}}
```
Use `reform1` to create a `policy1` object and `reform2` to create a `policy2` object.  Then extract the 2019, 2020, and 2021 values of `CTC_c` from the `policy1` object, and use them and the  value of `offset` to calculate the **expected** values of `CTC_c` for those three years under `reform2`.   Next extract the **actual** 2019, 2020, and 2021 values of `CTC_c` from the `policy2` object.  And finally, assert that the three actual values are equal to the three expected values using `np.allclose`.

Does this make sense?  If not, please ask questions.  It would be nice to strengthen the test and merge this pull request quickly, so that we can issue, in the next few days, a new version of Tax-Calculator that fixes this bug.

andersonfrailey:
@martinholmer yes, that all makes sense. I'll jump on that.

PR #2372, opened 2019-09-03
Modify section headings in policy_current_law.json

Peter-Metz:
This PR modifies certain `section_1` and `section_2` headings in `policy_current_law.json`. The need arose because there is a group of unorganized/unsectioned parameters in Tax-Brain and Tax-Cruncher because they are not assigned a `section_1` (see the first screenshot for an example). In the process, I also modified a few of the `section_2` headings because they were the same as the parameters' descriptions (see the second screenshot for an example).

I modified a test in `test_policy.py` but it still fails because the section headings do not match `uguide.htmx`. What is the process for updating `uguide.htmx`?

**Un-sectioned parameters:**

Screen Shot 2019-09-03 at 4 31 50 PM

**Example of section_2 mirroring parameter description:**

Screen Shot 2019-09-03 at 4 32 02 PM


MattHJensen:
@Peter-Metz, I added a WIP label for now since the tests aren't passing. Could you ping me again when they are, and then I'll be happy to review?

Peter-Metz:
@MattHJensen tests now pass after some modifications to `uguide.htmx` and `test_policy.py`

MattHJensen:
@Peter-Metz, thanks again for opening this PR. After reviewing this more closely, I remember some context that might be helpful:

The previous version of Tax-Brain at OSPC.org did not display parameters that were not assigned section headers. You can confirm at ospc.org/taxbrain. This is a relic of when the Tax-Calculator and Tax-Brain teams were more tightly integrated, and the Tax-Calculator team would use the absence of section headers to indicate that a parameter shouldn't appear on Tax-Brain for some reason of user safety. 

I strongly suggest that apps using Tax-Calculator adopt the same rule for now -- i.e., **don't display parameters that don't have Tax-Calculator section headers.** (cc @andersonfrailey @hdoupe @jdebacker)

The Tax-Calculator project will also review how we indicate a parameter's safety for users and whether any of the un-section-headered parameters are actually safe. 

@Peter-Metz, if all of this makes sense to you, then I think it would be helpful to narrow the scope of this PR to:

>  I also modified a few of the section_2 headings because they were the same as the parameters' descriptions (see the second screenshot for an example).

hdoupe:
> I strongly suggest that apps using Tax-Calculator adopt the same rule for now -- i.e., don't display parameters that don't have Tax-Calculator section headers. (cc @andersonfrailey @hdoupe @jdebacker)

You can choose which parameters are available on compute.studio by doing something like this:

```python
def get_inputs(meta_params_dict):
    """
    Return default parameters for Tax-Brain
    """
    metaparams = MetaParameters()
    metaparams.adjust(meta_params_dict)

    policy_params = TCParams()
    policy_params.set_state(
        year=metaparams.year.tolist(),
        data_source=metaparams.data_source.tolist(),
    )
    
    # Only keep "schema" object and parameters that have a section_1
    filtered_pol_params = OrderedDict()
    for k, v in policy_params.dump().items():
        if k == "schema" or v.get("section_1", False):
            filtered_pol_params[k] = v
    

    behavior_params = BehaviorParams()

    default_params = {
        "policy": filtered_pol_params,
        "behavior": behavior_params.dump()
    }
    meta = metaparams.dump()

    return {"meta_parameters": meta, "model_parameters": default_params}
```

Peter-Metz:
Thanks @MattHJensen, I'll update this PR in the next couple of days. Also thanks @hdoupe for the suggestion re: compute studio

Peter-Metz:
@MattHJensen: the latest commits reinstate the un-sectioned params (keeping the modified `section_2` headings that were the same as the parameter description)

taxcalc/tests/test_policy.py
```diff
@@ -699,26 +699,18 @@ def generate_section_dictionary(html_text):
             'Additional Medicare FICA': 0
         },
         'Social Security Taxability': {
```
MattHJensen: @Peter-Metz, could you help me understand what is going on in test_policy.py? 

https://github.com/PSLmodels/Tax-Calculator/pull/2372/files#diff-dc74f969d76006abbdf3bcc5b95240e4L702-L721

Peter-Metz: @MattHJensen sure thing. The gist is that `test_policy` creates a dictionary called `valid_dict` and compares it to `policy_current_law` to check `section_1` and `section_2` titles. For example:

```python
valid_dict = {
       # corresponds to PCL section_1
        'Payroll Taxes': {
            # corresponds to PCL section_2
            'Social Security FICA': 0
        }
```

Since this PR removes some of the `section_2` titles, we modify `valid_dict` with:

```python
valid_dict = {
        'Payroll Taxes': {
            '': 0
        }
```

Does that clear things up?

MattHJensen: Thanks, @Peter-Metz

MattHJensen:
Thanks for this PR, @Peter-Metz!

MattHJensen:
@Peter-Metz, I'm going to have to revert this PR for now. Sorry for undoing your work, at least temporarily. The trouble appears to be that `make_uguide.py` requires the `section_2` headers to create or save the html for the parameter documentation. If you still have the motivation to remove the section_2 headers, I would be very happy to review a new PR that handles this additional wrinkle.

PR #2373, opened 2019-09-05
Update recipe04 to use vectorized behresp.quantity_response

MaxGhenis:
See discussion in https://github.com/PSLmodels/Behavioral-Responses/pull/58.


MattHJensen:
Thanks a lot for opening this PR, @MaxGhenis. I will review soon.

MattHJensen:
@MaxGhenis, thanks a great deal for opening this PR to Tax-Calculator. 

Given the context:
- The Tax-Calculator project is relative mature and desires to provide stable documentation and code
- Your contribution presents an alternative take on Recipe04, not a correction of it. 
- It would definitely be valuable for Tax-Calculator users to have access to your version of the Recipe

What do you think about adding this as a new recipe to the cookbook, rather than as a replacement for Recipe04, perhaps Recipe04-vectorized-behavior.py (welcome better names!). 

If you think this is a good idea and have the bandwidth to take it the extra step, that would be wonderful. If you think it is a good idea but would rather I add it to the cookbook and make sure testing/docs are working ok, I am open to that, too. If you think it is a bad idea, please let me know why and we can do something better.

MaxGhenis:
I reverted the old `recipe04` files and created the new ones as `recipe04_pandas` - I think this is the general difference. Note I didn't left recipe04 as-is, not addressing some stylistic issues like lowercase constants.

`make_cookbook.py` also needs to be updated, could you advise on the regex? This skips `recipe04_pandas`:

    RECIPES = glob.glob('recipe[0-9][0-9].py')

MattHJensen:
> could you advise on the regex?

What do you think about adding a wildcard at the end?

MaxGhenis:
Done

MattHJensen:
Thanks for the contribution, @MaxGhenis. Merging.

PR #2378, opened 2019-10-16
Create conda env from file directly

MaxGhenis:
Mirrors https://github.com/PSLmodels/Tax-Cruncher/pull/19, saves a line and might speed up the build a bit


MaxGhenis:
Huh, this is causing a build error below the modified line:
```
python -c "import taxcalc"; coverage run -m pytest -v -m "not requires_pufcsv and not pre_release and not local" --pep8
```
>AttributeError: 'module' object has no attribute 'getfullargspec'
This isn't an important change so I'm closing, but if anyone wants to suggest a fix feel free to comment and I can reopen.

PR #2379, opened 2019-10-31
release 2-6-0

MattHJensen:


MattHJensen:
Something doesn't look right with the Social Security Taxability section in the docs. Investigating.

MattHJensen:
The latest commits account for #2380 which reverted #2372 to clear up the trouble with the user guide.

PR #2392, opened 2019-12-13
Specify where agg csv files are in tests

MaxGhenis:
This is effectively documentation only. Adds a note in the CPS and PUF agg csv tests stating where the files are, to help the user know where to copy from when the expected aggs change.


MattHJensen:
Thanks, @MaxGhenis. Merging now.

PR #2397, opened 2019-12-20
Expand bounds on growdiff values validation

Peter-Metz:
This PR expands the range of valid values for parameters in `growdiff.json` from `{"min": -1, "max": 1}` to `{"min": -10, "max": 10}`. This will allow users to blow up extrapolated input variables by more than 100%.


MattHJensen:
Thanks @Peter-Metz! Merging.

PR #2401, opened 2020-03-02
Use ParamTools

hdoupe:
This PR shows how ParamTools could be used to handle Tax-Calculator's parameters. I used a script to convert the configuration files from the Tax-Calculator format to the ParamTools format, but besides this difference, there are only a couple public-facing API changes:

- ~~The `inflation_rates()` and `growth_rates()` methods now return dictionaries where the keys are years and the values are the rates. It was easier for me to work with this format since most operations require you to get a value for a certain year at a time. This can easily be reverted.~~

- The parameter values that are accessible as instance attributes (e.g. `pol.II_em`) are not casted down into a lower dimension. So, if the year is 2017, then `pol.II_em` returns `[4050.0]` instead of `4050.0`. ParamTools does this to keep the results consistent for when you have `pol.set_year(2017)` or `pol.set_year([2017, 2018])`. The `Calculator.policy_param` method automatically casts the value into a lower dimension to preserve compatibility:

  ```python
  In [2]: pol = Policy()                                                                                                                                                                        
  
  In [3]: pol.set_year(2017)                                                                                                                                                                    
  
  In [4]: pol.II_em                                                                                                                                                                             
  Out[4]: array([4050.])
  
  In [5]: pol.set_year([2017, 2018])                                                                                                                                                            
  
  In [6]: pol.II_em                                                                                                                                                                             
  Out[6]: array([4050.,    0.])
  
  In [7]: from taxcalc import Policy, Calculator, Records                                                                                                                                       
  
  In [8]: calc = Calculator(policy=pol, records=Records.cps_constructor())                                                                                                                      
  
  In [9]: calc.policy_param("II_em")                                                                                                                                                            
  Out[9]: 4050.0
  
  ```

There may be some other API changes that I have missed. I'll be going over the code to make sure that I have not missed anything else. If anyone is interested, I'd really appreciate them running their downstream projects or notebooks with the changes in this branch to make sure I haven't broken anything.

There are few notable places where I maintained consistency:

- The `_update` method still takes reforms that are in the Tax-Calculator format. The reforms are reformatted and passed to the ParamTools `adjust` method. All classes that inherit from `taxcalc.Parameters` will have this behavior. That's why none of the reforms in this repo needed to be reformatted in this PR.

- Getting the parameter value using a leading underscore still returns its values for all years:

  ```python
  In [10]: pol._II_em                                                                                                                                                                           
  Out[10]: 
  array([3900.  , 3950.  , 4000.  , 4050.  , 4050.  ,    0.  ,    0.  ,
            0.  ,    0.  ,    0.  ,    0.  ,    0.  ,    0.  , 4901.  ,
         5004.41, 5108.5 , 5215.78])
  
  ```

  This is done with a custom attribute getter which checks if the attribute meets this pattern and then does the necessary operations to get the values for all years.

- Other attributes and methods like `start_year`, `current_year`, and `set_year` were either copied over to the new `Parameters` class or re-written to wrap the ParamTools API to maintain consistency.

I opened this PR as a draft because I'm not sure if the Tax-Calculator project is interested in adopting ParamTools. After all, Tax-Calculator has a perfectly good parameters processing backend that is fast, reliable, and well-tested. This exercise was very helpful for ParamTools. It prompted me to improve ParamTools's performance, make it more extendable, add better error messages, add warnings, and the list goes on. I even spotted a small bug in Tax-Calculator (#2381) while I was re-implementing the inflation indexing algorithm with ParamTools. I also included [detailed documentation](https://github.com/hdoupe/Tax-Calculator/blob/use-paramtools/taxcalc/parameters.py#L95-L134) on the indexing algorithm in the new `adjust` method, which could be helpful to have in the Tax-Calculator docs. So, I think this exercise was helpful for Tax-Calculator, too. I'm excited to hear people's feedback on this PR.


hdoupe:
The test errors look like pycodestyle errors. I'll push some fixes soon.

hdoupe:
@MattHJensen asked yesterday on the TC community call for a TODO list for reviewers:

- [ ] Make sure all of my test changes are sound.
  - I'll leave a comment on GitHub on places where I'd especially appreciate advice/review.
- [ ] If you have any notebooks, workflows, analyses, etc. where you use TC, please run them and make sure your code isn't broken.
- [ ] Check the recipes to make sure those are still good-to-go. (I'm happy to help)
  - [ ] One in particular is [`recipe06.py`](https://github.com/PSLmodels/Tax-Calculator/blob/master/docs/cookbook/recipe06.py)
- [ ] Maybe @MattHJensen  or @Peter-Metz  would be interested in stepping through the `adjust` method with me to make sure it's clear and readable.
- [ ] Yes/No on the 2 API changes I mentioned in the initial comment for this PR. 
  - `inflation_rates()` / `growth_rates()` return dicts
  - Instance attributes on `taxcalc.Parameters` are not casted down to lower dimension.

My TODO list is:
- [ ] Go back over code to make sure I'm not missing API changes.
  - [ ] Bring back `initialize` method.
- [ ] Comment on any test changes.
- [ ] Add equivalent version of tests that I found helpful in the early stages of developing this PR.
  - Located here: https://github.com/hdoupe/TaxParams/blob/master/taxparams/tests/test.py
- [ ] Upstream the `select_lt` function into ParamTools.
- [ ] Remove warnings in ParamTools that are shown when adjusting the value of a parameter for a year other than current year.
- [ ] Take another pass through `adjust` code to see if I can make it clearer. (I went a little crazy with list and dict comprehension here.)
- [ ] Make updates to parameter documentation code (i.e. code that creates HTML files for the parameters).

@MattHJensen @Peter-Metz Let me know if I'm missing anything here.

taxcalc/tests/test_reforms.py
```diff
@@ -27,12 +27,13 @@ def test_2017_law_reform(tests_path):
     with open(reform_file, 'r') as rfile:
         rtext = rfile.read()
     pol.implement_reform(Policy.read_json_reform(rtext))
+    # TODO: I'm unclear on how to handle the undefined warning behavior
+    # since the parameter is still defined in the JSON.
```
hdoupe: I'm not sure how to handle the `CTC_c` parameter warning since it's still defined in `policy_current_law.json`, but according to this map, it's been redefined: 

https://github.com/PSLmodels/Tax-Calculator/blob/0e6dee1f5556b0da01a2141d39a77d8ca82b1146/taxcalc/policy.py#L65-L69

@MattHJensen  or @Peter-Metz  do you have any guidance for how this should be handled?

hdoupe:
Commit https://github.com/PSLmodels/Tax-Calculator/pull/2401/commits/395af1c8c0b5986b3c0d2b167a649c7c1454b6cc reverts `_inflation_rates` and `_wage_growth_rates` back to lists. I added an optional `year` key-word argument to the `wage_growth_rates` and `inflation_rates` methods to make it easy to grab the rate for a given year without having to subtract the year away first:

```python
In [1]: from taxcalc import Policy                                                                                                                                                            

In [2]: pol = Policy()                                                                                                                                                                        

In [3]: pol.inflation_rates()                                                                                                                                                                 
Out[3]: 
[0.0148,
 0.0159,
 0.0012,
 0.0127,
 0.0187,
 0.0224,
 0.0186,
 0.0233,
 0.0229,
 0.0228,
 0.0221,
 0.0211,
 0.0209,
 0.0211,
 0.0208,
 0.021,
 0.021]

In [4]: pol.inflation_rates(year=2022)                                                                                                                                                        
Out[4]: 0.0228

In [5]:                                                                                                                                                                                       
```

hdoupe:
Another API change that I've been thinking about is Tax-Calculator's use of `print_warnings` vs ParamTools's use of `ignore_warnings` in the `implement_reform`/`adjust` methods. 

If `print_warnings` is `True` Tax-Calculator prints the warnings and continues to run. If `ignore_warnings` is `False`, then ParamTools raises a `ValidationError` when the warnings are triggered. 

If the warnings are off (`print_warnings=False`, `ignore_warnings=True`), then warnings are handled similarly in that no errors are raised or warnings are printed by either package. However, unlike Tax-Calculator, ParamTools does not store the warnings (since warning validators are ignored in this case).

My question is: Should `print_warnings` be left as an optional argument and mapped to `ignore_warnings` like this:

```python
    def implement_reform(self, reform,
                         print_warnings=True, raise_errors=True):
        return self._update(
            reform,
            # negate print_warnings to map definition
            ignore_warnings=not print_warnings,
            raise_errors=raise_errors
        )
```

Or, should it be deprecated in favor of `ignore_warnings`?

Or, do I need to think about customizing ParamTools's handling of warnings to more closely match how Tax-Calculator handles warnings?

hdoupe:
This commit https://github.com/PSLmodels/Tax-Calculator/pull/2401/commits/55f6e636ccd4ca4601d97d9051e32314d1950e9e ports over a bunch of tests that were helpful in development stages to test interactions between the indexed status of parameters, updates to their values, and the `CPI_offset` parameter.

hdoupe:
@MattHJensen @Peter-Metz this is ready for review.

Peter-Metz:
@hdoupe awesome, look forward to diving in!

hdoupe:
Thanks @Peter-Metz. I'm looking forward to  your review. I think things are looking pretty solid right now, but I'm still a little unclear about the api changes I outlined here https://github.com/PSLmodels/Tax-Calculator/pull/2401#issuecomment-602078224.

hdoupe:
Also, here's the script that I used to convert the tax-calculator values to the paramtools format:

```python
import json

import paramtools


POLICY_SCHEMA = {
    "labels": {
        "year": {
            "type": "int",
            "validators": {
                "range": {"min": 2013, "max": 2030}
            }
        },
        "MARS": {
            "type": "str",
            "validators": {"choice": {"choices": ["single", "mjoint",
                                                  "mseparate", "headhh",
                                                  "widow"]}}
        },
        "idedtype": {
            "type": "str",
            "validators": {"choice": {"choices": ["med", "sltx", "retx", "cas",
                                                  "misc", "int", "char"]}}
        },
        "EIC": {
            "type": "str",
            "validators": {"choice": {"choices": ["0kids", "1kid",
                                                  "2kids", "3+kids"]}}
        },
    },
    "additional_members": {
        "section_1": {"type": "str"},
        "section_2": {"type": "str"},
        "start_year": {"type": "int"},
        "indexable": {"type": "bool"},
        "indexed": {"type": "bool"},
        "compatible_data": {"type": "compatible_data"}
    }
}

def convert_defaults(pcl):

    type_map = {
        "real": "float",
        "boolean": "bool",
        "integer": "int",
        "string": "str",
    }

    new_pcl = defaultdict(dict)
    new_pcl["schema"] = POLICY_SCHEMA
    for param, item in pcl.items():
        values = []
        pol_val = item["value"]
        min_year = min(item["value_yrs"])
        if isinstance(pol_val[0], list):
            for year in range(len(pol_val)):
                for dim1 in range(len(pol_val[0])):
                    values.append({
                        "year": min_year + year,
                        item["vi_name"]: item["vi_vals"][dim1],
                        "value": pol_val[year][dim1],
                    })
        else:
            for year in range(len(pol_val)):
                values.append({
                    "year": min_year + year,
                    "value": pol_val[year],
                })

        new_pcl[param]['value'] = values
        new_pcl[param]['title'] = pcl[param]["long_name"]
        new_pcl[param]['type'] = type_map[pcl[param]["value_type"]]

        if pcl[param].get("invalid_action") == "warn":
            pcl[param]["valid_values"]["level"] = "warn"
        new_pcl[param]["validators"] = {"range": pcl[param]["valid_values"]}

        to_keep = list(POLICY_SCHEMA["additional_members"].keys()) + [
            "description", "notes", "compatible_data"
        ]
        for k in to_keep:
            if k in pcl[param]:
                new_pcl[param][k] = pcl[param][k]

    return new_pcl


if __name__ == "__main__":
    to_convert = [
        "taxcalc/policy_current_law.json",
        "taxcalc/growdiff.json",
        "taxcalc/consumption.json",
    ]

    for path in to_convert:
        print('converting', path)
        with open(path, "r") as f:
            converted = convert_defaults(json.loads(f.read()))

        with open(path, "w") as f:
            f.write(json.dumps(converted, indent=4))

    import marshmallow as ma

    class CompatibleDataSchema(ma.Schema):
        """
        Schema for Compatible data object
        {
            "compatible_data": {"data1": bool, "data2": bool, ...}
        }
        """

        puf = ma.fields.Boolean()
        cps = ma.fields.Boolean()

    paramtools.register_custom_type("compatible_data", ma.fields.Nested(CompatibleDataSchema))

    for path in to_convert:
        print('testing', path)

        class Params(paramtools.Parameters):
            defaults = path

        assert Params()
```

taxcalc/tests/test_reforms.py
Peter-Metz: Before TCJA, `CTC_c` represented the whole (non-refundable) credit and since TCJA, `CTC_c` represents the non-refundable portion of the credit. In my view, the warning message doesn't add much, especially since Tax-Calc 1.0.0 was released over a year ago.

cc @MattHJensen @hdoupe

Peter-Metz:
@hdoupe and I talked about the question raised in [#2401 (comment)](https://github.com/PSLmodels/Tax-Calculator/pull/2401#issuecomment-602078224) -- Tax-Calculator's use of `print_warnings` vs ParamTools's use of `ignore_warnings`. Specifically, like Hank says, Tax-Calculator will continue to run if `print_warnings` is triggered, but ParamTools raises an error if `ignore_warnings` is False and is triggered.

In thinking about what the correct behavior should be, it would be helpful to know why certain parameters trigger warnings during validation as opposed to errors in Tax-Calc (specifically, `ID_Medical_frt`, `ID_Medical_frt_add4aged`, `ID_Charity_crt_all`, `ID_Charity_crt_noncash`, `ID_Miscellaneous_frt`). All five variables are floors/ceilings as a decimal fraction of AGI for deductible expenses and have narrow `valid_values` ranges.

I personally don't see an issue with ParamTools throwing an error in this scenario, but we were interested to hear if @MattHJensen or anyone else had some context for why certain variables trigger warnings during validation.

taxcalc/parameters.py
```diff
@@ -1,876 +1,523 @@
...
+                    yearval = yearval.reshape(
+                        (*(1, ) * short_dims, *yearval.shape)
+                    )
+                    self.set_state(year=year)
+                    try:
+                        yearval = self.from_array(param, yearval)
+                    except IndexError:
+                        msg = (
+                            f"Pameter {param} does not have the correct "
+                            f"array dimensions for year {year}."
```
Peter-Metz, line 461: This error isn't thrown if too many dims are provided. i.e.:

```
ref = {"II_rt1": {2029: [.2, .3]}}
pol_ref.implement_reform(ref)
```

taxcalc/policy.py
```diff
@@ -119,20 +105,55 @@ def read_json_reform(obj):
         return Parameters._read_json_revision(obj, 'policy')
 
     def implement_reform(self, reform,
```
Peter-Metz, line 107: Not totally sure where to leave this comment, and I suspect this may be an issue in Paramtools, but when I try:

```python
salt_ref = {"ID_AllTaxes_c": {2032: [9e99, 9e99, 9e99, 9e99, 9e99]},
pol.implement_reform(salt_ref)
```

The error message is:

```
ValidationError: {
    "errors": {
        "year": "['Must be greater than {max}.']"
    }
}
```

The message should say something to the effect of "Must be <= 2030"

hdoupe: Thanks @Peter-Metz  with PR https://github.com/PSLmodels/ParamTools/pull/100 this will be:

```python
ValidationError: {
    "errors": {
        "year": "['Input 2032 must be less than 2030.']"
    }
}
```

taxcalc/parameters.py
```diff
@@ -1,876 +1,523 @@
...
+                    short_dims = ndims - yearval.ndim
+                    yearval = yearval.reshape(
+                        (*(1, ) * short_dims, *yearval.shape)
+                    )
+                    self.set_state(year=year)
+                    try:
+                        yearval = self.from_array(param, yearval)
+                    except IndexError:
+                        msg = (
+                            f"Pameter {param} does not have the correct "
```
Peter-Metz, line 460: Typo ("Pameter")

hdoupe: Thanks, will have a fix up in a few.

hdoupe:
@Peter-Metz after the recent commits, the examples that you pointed out now look like this:

```python
In [1]: from taxcalc import Policy                                                        

In [2]: pol = Policy()                                                                    

In [3]: pol.implement_reform({"II_rt1": {2029: [.2, .3]}})                                

# ValidationError: {
#     "errors": {
#         "schema": "Parameter II_rt1 has 2 elements but it should have 1 element(s)."
#     }
# }

In [4]: pol = Policy()                                                                    

In [5]: salt_ref = {"ID_AllTaxes_c": {2032: [9e99, 9e99, 9e99, 9e99, 9e99]}}              

In [6]: pol.implement_reform(salt_ref)                                                    

# ValidationError: {
#     "errors": {
#         "year": "['Input 2032 must be less than 2030.']"
#     }
# }

In [7]: pol = Policy()                                                                    

In [8]: ref = {"II_brk1": {2020: [9700, 19400, 9700, 13850, 19400, 19400]}}               

In [9]: pol.implement_reform(ref)                                                         

# ValidationError: {
#     "errors": {
#         "schema": "Parameter II_brk1 has 6 elements but it should have 5 element(s)."
#     }
# }

In [10]: pol = Policy()                                                                   

In [11]: pol.implement_reform({"EITC_MaxEligAge": {2017: 27.5}})                          

# ValidationError: {
#     "errors": {
#         "EITC_MaxEligAge": [
#             "Not a valid integer: 27.5."
#         ]
#     }
# }

In [12]:  
```

hdoupe:
I removed the magic from the code that does array validation and makes error messages. This improves the messages and simplifies the logic. I updated the results in my comment here with the new messages: https://github.com/PSLmodels/Tax-Calculator/pull/2401#issuecomment-615199995

taxcalc/tests/test_parameters.py
```diff
@@ -116,23 +149,24 @@ def update_params(self, revision,
...
-        assert prms.parameter_errors
+        with pytest.raises(paramtools.ValidationError):
+            prms.update_params(revision)
+        assert prms.errors
     elif expect == 'warn':
-        prms.update_params(revision, print_warnings=False, raise_errors=False)
-        assert prms.parameter_warnings
+        with pytest.raises(paramtools.ValidationError):
+            prms.update_params(revision)
+        assert prms.warnings
```
hdoupe, line 169: @Peter-Metz Changes in this section are related to differences in error handling:
- `print_warning` vs `ignore_warning`
- run warning level validators vs not running them

taxcalc/tests/test_parameters.py
```diff
@@ -164,41 +189,17 @@ def test_json_file_contents(tests_path, fname):
...
-        # check for non-empty long_name and description strings
-        assert isinstance(param['long_name'], str)
-        if not param['long_name']:
-            assert '{} long_name'.format(pname) == 'empty string'
-        assert isinstance(param['description'], str)
-        if not param['description']:
-            assert '{} description'.format(pname) == 'empty string'
-        # check that indexable and indexed are boolean
-        assert isinstance(param.get('indexable', False), bool)
-        assert isinstance(param.get('indexed', False), bool)
```
hdoupe, line 201: @Peter-Metz All of these checks are enforced by ParamTools when [loading the default parameters](https://github.com/PSLmodels/ParamTools/blob/0b11d235b58c3e8d2847382437201703a4e433f8/paramtools/schema_factory.py#L94).

taxcalc/tests/test_parameters.py
```diff
@@ -210,75 +211,40 @@ def test_json_file_contents(tests_path, fname):
...
-                assert len(vivals) == 4
-            elif viname == 'idedtype':
-                assert len(vivals) == 7
-            elif viname == 'c00100':
-                pass
-            else:
-                assert viname == 'UNKNOWN vi_name VALUE'
-            # check length of each value row
-            for valuerow in value:
-                assert len(valuerow) == len(vivals)
```
hdoupe, line 264: These checks are enforced by ParamTools on load and when converting the list of dicts to numpy arrays.

taxcalc/tests/test_parameters.py
```diff
@@ -319,161 +285,14 @@ def test_parameters_mentioned(tests_path, jfname, pfname):
...
-    exp5 = 74900. * 1.03
-    exp6 = 37450. * 1.03
-    exp = [[36000.0, 72250.0, 36500.0, 48600.0, 72500.0, 36250.0],
-           [38000.0, 74000.0, 36900.0, 49400.0, 73800.0, 36900.0],
-           [40000.0, 74900.0, 37450.0, 50200.0, 74900.0, 37450.0],
-           [exp1, exp2, exp3, exp4, exp5, exp6]]
-    res = Parameters._expand_2d(np.array(_II_brk2),
-                                inflate=True, inflation_rates=inf_rates,
-                                num_years=4)
-    assert np.allclose(res, exp, atol=0.01, rtol=0.0)
```
hdoupe, line 435: ParamTools has its own array extension tests [here](https://github.com/PSLmodels/ParamTools/blob/master/paramtools/tests/test_parameters.py#L1506-L1910).

taxcalc/tests/test_parameters.py
```diff
@@ -319,161 +285,14 @@ def test_parameters_mentioned(tests_path, jfname, pfname):
...
-                             valstr)
-            assert msg == 'ERROR: integer_value param has non-integer value'
-        # check that val_is_boolean is consistent with boolean_value
-        boolean_type = pdict[param]['value_type'] == 'boolean'
-        if val_is_boolean != boolean_type:
-            msg = 'param,value_type,valstr= {} {} {}'
-            msg = msg.format(str(param),
-                             pdict[param]['value_type'],
-                             valstr)
-            assert msg == 'ERROR: boolean_value param has non-boolean value'
```
hdoupe, line 474: ParamTools does boolean validation as well.

taxcalc/tests/test_policy.py
```diff
@@ -30,45 +53,17 @@ def test_correct_class_instantiation():
...
-        assert isinstance(value_yrs, list)
-        value = data['value']
-        expected_value_yrs = [(start_year + i) for i in range(len(value))]
-        if value_yrs != expected_value_yrs:
-            msg = 'name,value_yrs,expected_value_yrs: {}\n{}\n{}'
-            raise ValueError(msg.format(name, value_yrs, expected_value_yrs))
-        # check entries in vi_vals list
-        vivals = data['vi_vals']
-        if vivals:
-            assert vivals == expected_vi_vals[data['vi_name']]
```
hdoupe, line 69: `value_yrs` is no longer used, and label validation on the `vi_vals` is done when the defaults are loaded.

taxcalc/tests/test_policy.py
```diff
@@ -248,51 +243,6 @@ def test_multi_year_reform():
...
-    assert np.allclose(getattr(pol, '_SS_Earnings_c'),
-                       Policy._expand_array(
-                           np.array([113700, 117000, 118500, 118500, 127200,
-                                     128400, 132900],
-                                    dtype=np.float64),
-                           'real',
-                           inflate=True,
-                           inflation_rates=wratelist,
-                           num_years=nyrs),
-                       atol=0.01, rtol=0.0)
```
hdoupe, line 295: `_expand_array` is no longer used, and loading values correctly is already tested extensively by ParamTools.

taxcalc/tests/test_policy.py
```diff
@@ -441,15 +391,7 @@ def test_policy_metadata():
...
     mdata = clp.metadata()
-    assert isinstance(mdata['STD']['value'], list)
-    assert np.allclose(mdata['STD']['value'],
-                       [6100, 12200, 6100, 8950, 12200])
-    assert isinstance(mdata['CDCC_ps']['value'], float)
-    assert mdata['CDCC_ps']['value'] == 15000
-    dump = False
-    if dump:
-        print(mdata)
-        assert 1 == 2
```
hdoupe, line 452: `metadata()` is now just a wrapper for ParamTools' `Parameters.specification` method:

https://github.com/hdoupe/Tax-Calculator/blob/use-paramtools/taxcalc/parameters.py#L597

taxcalc/tests/test_policy.py
```diff
@@ -837,77 +780,15 @@ def test_description_punctuation(tests_path):
...
-    unmatched = parameters ^ json_range_params
-    if unmatched:
-        assert unmatched == 'UNMATCHED RANGE PARAMETERS'
-    # check all current-law-policy parameters for range validity
-    clp = Policy()
-    clp._validate_values(parameters)
-    # eventually activate: assert not clp.parameter_warnings
-    ctc_c_warning = 'CTC_c was redefined in release 1.0.0\n'
-    assert clp.parameter_warnings == ctc_c_warning
-    assert not clp.parameter_errors
```
hdoupe, line 900: This test is covered by ParamTools schema validation.

I'm also taking a pass on the `CTC_c` redefined warnings until I get more info.

taxcalc/tests/test_policy.py
```diff
@@ -593,6 +535,7 @@ def test_pop_the_cap_reform():
     # specify a "pop the cap" reform that eliminates MTE cap in 2016
     reform = {'SS_Earnings_c': {2016: 9e99}}
     ppo.implement_reform(reform)
+    mte = ppo._SS_Earnings_c
```
hdoupe, line 538: PT does not modify the same copy of the `_SS_Earnings_c` array. A completely new object is created. Thus, we need to re-initialize the `mte` variable.

taxcalc/tests/test_policy.py
```diff
@@ -953,8 +843,20 @@ def test_reform_with_warning():
...
+
+    # pol.implement_reform(reform)
+    # assert pol.parameter_warnings
+    with pytest.raises(pt.ValidationError):
+        pol.implement_reform(reform)
+
+    pol = Policy()
+    pol.implement_reform(reform, ignore_warnings=True)
+    pol.set_state(year=2020)
+    assert pol.ID_Medical_frt == np.array([0.05])
```
hdoupe, line 859: See comment about warnings: https://github.com/PSLmodels/Tax-Calculator/pull/2401#issuecomment-602078224

taxcalc/tests/test_policy.py
```diff
@@ -988,9 +907,9 @@ def test_index_offset_reform():
     pvalue2 = dict()
     for cyr in [2019, 2020, 2021]:
         policy1.set_year(cyr)
-        pvalue1[cyr] = policy1.CTC_c
+        pvalue1[cyr] = policy1.CTC_c[0]
         policy2.set_year(cyr)
-        pvalue2[cyr] = policy2.CTC_c
+        pvalue2[cyr] = policy2.CTC_c[0]
```
hdoupe, line 912: This just brings the parameter down a dimension. This is related to the API change I mentioned in the initial comment of this PR.

taxcalc/tests/test_policy.py
```diff
@@ -837,77 +782,15 @@ def test_description_punctuation(tests_path):
...
     Check private _indexing_rates_for_update method.
     """
     pol = Policy()
-    wgrates = pol._indexing_rates_for_update('_SS_Earnings_c', 2017, 10)
-    pirates = pol._indexing_rates_for_update('_II_em', 2017, 10)
-    assert len(wgrates) == len(pirates)
+    wgrates = pol.get_index_rate('_SS_Earnings_c', 2017)
+    pirates = pol.get_index_rate('_II_em', 2017)
+    assert isinstance(wgrates, np.float64)
+    assert isinstance(pirates, np.float64)
```
hdoupe, line 793: `Parameters._indexing_rates_for_update` is no longer used. A public method for getting the index rate of a paramter at a given year is used instead: [`Parameters.get_index_rate`](https://github.com/hdoupe/Tax-Calculator/blob/use-paramtools/taxcalc/parameters.py#L315-L327).

taxcalc/tests/test_reforms.py
```diff
@@ -91,26 +92,29 @@ def test_round_trip_tcja_reform(tests_path):
...
     pol.implement_reform(Policy.read_json_reform(rtext))
+
+    # TODO: see comments in test_2017_law_reform
     # eventually activate: assert not clp.parameter_warnings
-    ctc_c_warning = 'CTC_c was redefined in release 1.0.0\n'
-    assert pol.parameter_warnings == ctc_c_warning
-    assert not pol.parameter_errors
+    # ctc_c_warning = 'CTC_c was redefined in release 1.0.0\n'
+    # assert pol.parameter_warnings == ctc_c_warning
+    assert not pol.errors
```
hdoupe, line 107: See comment on warnings api change: https://github.com/PSLmodels/Tax-Calculator/pull/2401#issuecomment-602078224

taxcalc/tests/test_reforms.py
```diff
@@ -27,12 +27,13 @@ def test_2017_law_reform(tests_path):
...
+    # since the parameter is still defined in the JSON.
     # eventually activate: assert not clp.parameter_warnings
-    ctc_c_warning = 'CTC_c was redefined in release 1.0.0\n'
-    assert pol.parameter_warnings == ctc_c_warning
-    assert not pol.parameter_errors
+    # ctc_c_warning = 'CTC_c was redefined in release 1.0.0\n'
+    # assert pol.parameter_warnings == ctc_c_warning
     pol.set_year(2018)
-    pre_mdata = pol.metadata()
+    pre_mdata = dict(pol.items())
```
hdoupe, line 36: Using the `items()` method is an easier way to get just the values for all parameters.

taxcalc/parameters.py
```diff
@@ -1,876 +1,541 @@
...
+            After the 'unknown' values have been deleted, the last known value
+            is extrapolated through the budget window. If there are indexed
+            parameters in the adjustment, they will be included in the final
+            adjustment call (unless their indexed status is changed).
+        2. If the "indexed" status is updated for any parameter:
+            a. If a parameter has values that are being adjusted before
+                the indexed status is adjusted, update those parameters first.
+            b. Extend the values of that parameter to the year in which
+                the status is changed.
+            c. Change the the indexed status for the parameter.
```
Peter-Metz, line 93: typo

Peter-Metz:
Many users of Tax-Calculator (myself included) use `policy_current_law.json` as a parameter documentation resource. The new pcl schema makes the file harder for a person to read, and while there is good reason for this, I think there are a couple of easy things we can do to make the file more readable: 

1) We could change the order of the keys for a given parameter to be closer to current pcl layout. Instead of:

```
"FICA_ss_trt": {
        "value": [
            {
                "year": 2013,
                "value": 0.124
            }
        ],
        "title": "Social Security payroll tax rate",
        "type": "float",
        "validators": {
            "range": {
                "min": 0,
                "max": 1
            }
        },
        "section_1": "Payroll Taxes",
        "section_2": "Social Security FICA",
        "indexable": false,
        "indexed": false,
        "compatible_data": {
            "puf": true,
            "cps": true
        },
        "description": "Social Security FICA rate, including both employer and employee.",
        "notes": ""
    }
```
we could have:

```
"FICA_ss_trt": {
        "title": "Social Security payroll tax rate",
        "description": "Social Security FICA rate, including both employer and employee.",
        "section_1": "Payroll Taxes",
        "section_2": "Social Security FICA",
        "value": [
            {
                "year": 2013,
                "value": 0.124
            }
        ],

        ...
```

2) We could also remove the values for years 2020-2025 for parameters that revert to their pre-TCJA values in 2026. This will also help with a forthcoming bug fix for `CPI_offset`. With the new Paramtools functionality, I don't think these intermediary years serve a purpose.

cc @hdoupe @MattHJensen

Peter-Metz:
@hdoupe is there a mechanism for updating parameter values that revert to their pre-TCJA values in 2026 when the inflation rate changes?

Previously, we used `ppp.py` for these updates. (side note: I believe `ppp.py` can now be removed from the repo)

hdoupe:
@peter-metz, I agree with you that the read-ability of the JSON file is not as good in the ParamTools format. The way JSON is printed by default adds a lot of dead space to the parameter values, making it difficult to view all of the values for a parameter at once. Your two suggestions would improve the readability some, but I hope to improve it further with more ParamTools tooling. Right now, I'm working on a [JavaScript version of ParamTools](https://github.com/PSLmodels/ParamTools/issues/88) that will be used to make a GUI for creating, viewing, and updating the `defaults.json` file. This is basically a re-write of the Compute Studio [code](https://github.com/compute-tooling/compute-studio/tree/master/src/ParamTools) for building the model inputs forms. This is also my next ParamTools project for after this PR.

> 1.  We could change the order of the keys for a given parameter to be closer to current pcl layout. Instead of:

Changing the key order isn't a problem. I think they got shuffled in the script I used to automate the process of converting them from the old format to the new format. I can update this with no problems.

> 2. We could also remove the values for years 2020-2025 for parameters that revert to their pre-TCJA values in 2026. This will also help with a forthcoming bug fix for `CPI_offset`. With the new Paramtools functionality, I don't think these intermediary years serve a purpose.

Right, ParamTools will fill in the gaps between defined parameters using the indexing logic. So, there's no need to specify values that are not user-defined. The [extend](https://paramtools.dev/api/extend/) and [indexing](https://paramtools.dev/api/indexing/) examples in the ParamTools docs show how this works.

> Is there a mechanism for updating parameter values that revert to their pre-TCJA values in 2026 when the inflation rate changes?

If the values in `policy_current_law.json` are only values from policy changes (i.e. not the result of inflation), then they will be defined with whatever inflation rate is in place at the time. So, if there is an update to `growfactors.csv` or when the `CPI_offset` changes in 2026, the new inflation rate will be used to fill in the gaps between the user-defined parameters.

Peter-Metz:
@hdoupe that all sounds great and I'm excited to see where ParamToolsJS goes.

To illustrate my last point about updating 2026 values, the standard deduction looks something like this:

```
"STD": {
    "value": [
            { "year": 2019, "MARS": "single", "value": 12200.0},
            { "year": 2019, "MARS": "mjoint", "value": 24400.0},
            ...
            { "year": 2026, "MARS": "single", "value": 7651.0},
            { "year": 2026, "MARS": "mjoint", "value": 15303.0}
```

In 2026, there is both a policy change (revert back to pre-TCJA values) and an inflation component that is updated periodically. Next time we update inflation rates from Tax Data, the 2026 values above will be affected. Maybe we will need something like `ppp.py` after all.

taxcalc/tests/test_reforms.py
MattHJensen: >  In my view, the warning message doesn't add much, especially since Tax-Calc 1.0.0 was released over a year ago.

Agreed that this warning can go given the time elapsed since the 1.0 release.

MattHJensen:
From @Peter-Metz's https://github.com/PSLmodels/Tax-Calculator/pull/2401#issuecomment-614648632: 

> @hdoupe and I talked about the question raised in https://github.com/PSLmodels/Tax-Calculator/pull/2401#issuecomment-602078224 -- Tax-Calculator's use of print_warnings vs ParamTools's use of ignore_warnings. Specifically, like Hank says, Tax-Calculator will continue to run if print_warnings is triggered, but ParamTools raises an error if ignore_warnings is False and is triggered.
> 
> In thinking about what the correct behavior should be, it would be helpful to know why certain parameters trigger warnings during validation as opposed to errors in Tax-Calc (specifically, ID_Medical_frt, ID_Medical_frt_add4aged, ID_Charity_crt_all, ID_Charity_crt_noncash, ID_Miscellaneous_frt). All five variables are floors/ceilings as a decimal fraction of AGI for deductible expenses and have narrow valid_values ranges.
> 
> I personally don't see an issue with ParamTools throwing an error in this scenario, but we were interested to hear if @MattHJensen or anyone else had some context for why certain variables trigger warnings during validation.

Currently, all of our warnings occur when a reform provision is incompatible with puf.csv  but may be compatible with other datasets. 

For instance, the warning for the medical floor rate is  "When using PUF data, lowering this parameter value may produce unexpected results because PUF e17500 variable is zero below the floor."

Tax-Cruncher, for example, could safely ignore all of these warnings since it only uses non-puf data. 

That all of these warnings are situationally alike (relevant when using the PUF) does seem to imply that an unsuspecting user, hitting an irrelevant error initially, could safely turn on ignore_warnings and then run their program. 

But this assumes that the user knows all of the warnings that they will get after the first one will also be irrelevant to them. I don't think this is the case, so I think it would be helpful for the user to be able to see all of the warnings and still get their results. 

I think this means that the answer to Hank's question.... 

"do I need to think about customizing ParamTools's handling of warnings to more closely match how Tax-Calculator handles warnings?"

....is, unfortunately, "yes". 

In particular, I don't think the current ParamTools implementation is very friendly to a user who sees one warning and decides it isn't relevant. 

I would be very happy to be mistaken about the problem, or for there to be some simpler solution!

taxcalc/tests/test_parameters.py
```diff
@@ -210,75 +212,40 @@ def test_json_file_contents(tests_path, fname):
                               param.get('indexed', False),
                               param.get('indexable', False))
             failures += fail + '\n'
-        # check that value_type is correct string
```
Peter-Metz, line 213: Does this need to be ported over?

hdoupe: This just checks that `value_type` is one of `['boolean', 'integer', 'real', 'string']`. `value_type` was removed in favor of `type`. ParamTools uses `marshmallow` to validate that only correct types are used: https://github.com/PSLmodels/ParamTools/blob/master/paramtools/schema.py#L100

hdoupe: No I don't think so.

taxcalc/tests/test_policy.py
```diff
@@ -837,77 +782,15 @@ def test_description_punctuation(tests_path):
...
-    clp._validate_values(parameters)
-    # eventually activate: assert not clp.parameter_warnings
-    ctc_c_warning = 'CTC_c was redefined in release 1.0.0\n'
-    assert clp.parameter_warnings == ctc_c_warning
-    assert not clp.parameter_errors
-
-
 def test_indexing_rates_for_update():
     """
     Check private _indexing_rates_for_update method.
```
Peter-Metz, line 787: Might edit docstring

Peter-Metz:
@hdoupe I reviewed the changes to Tax-Calc tests and they look good to me! (I didn't check the tests that are now handled by ParamTools) 

@MattHJensen

hdoupe:
Thanks @Peter-Metz. Per @MattHJensen's suggestion I'm going to add a ParamTools version of the array tests.

I'm going to have time in about an hour to try to knock out these tasks per your feedback:
- [ ] Fix key order in JSON files
- [ ] Add back `print_warnings` functionality
- [ ] Add array tests

Let me know if I'm missing anything.

hdoupe:
@MattHJensen thanks for your feedback on how warnings should work in Tax-Calculator. I'm going to implement your suggestion in this PR.

Peter-Metz:
@hdoupe the only other thing i can think of is fixing the user guide. Let me know if I can help with any of these tasks.

hdoupe:
> @hdoupe the only other thing i can think of is fixing the user guide. Let me know if I can help with any of these tasks.

Do you have time to look at the user guide? I modified the `make_uguide.py` file so that it runs, but I haven't looked at the output closely yet.

Peter-Metz:
@hdoupe Yep, I'll take a look tomorrow.

jdebacker:
I installed TC from this branch and successfully ran all tests of the latest version of Cost of Capital Calculator with it.

@rickecon or I will try with OG-USA soon.

cc @hdoupe @MattHJensen @Peter-Metz

hdoupe:
Thanks @jdebacker! This is much appreciated!

hdoupe:
> Thanks @Peter-Metz. Per @MattHJensen's suggestion I'm going to add a ParamTools version of the array tests.
> 
> I'm going to have time in about an hour to try to knock out these tasks per your feedback:
> 
>     * [x]  Fix key order in JSON files
> 
>     * [x]  Add back `print_warnings` functionality
> 
>     * [x]  Add array tests
> 
> 
> Let me know if I'm missing anything.

These tasks are complete. @MattHJensen @Peter-Metz let me know how things look.

Here's the script I used to fix the key order:

```python
import json

if __name__ == "__main__":
    to_convert = [
        "taxcalc/policy_current_law.json",
        "taxcalc/growdiff.json",
        "taxcalc/consumption.json",
    ]

    keys = [
        "title",
        "description",
        "notes",
        "section_1",
        "section_2",
        "section_3",
        "start_year",
        "indexable",
        "indexed",
        "type",
        "value",
        "validators",
        "compatible_data",
    ]

    for path in to_convert:
        print('converting', path)
        with open(path, "r") as f:
            defaults = json.loads(f.read())

        sorted_defaults = {}
        if "schema" in defaults:
            sorted_defaults["schema"] = defaults["schema"]

        for param, data in defaults.items():
            if param == "schema":
                continue
            assert not (data.keys() - set(keys)), f"{param} has extra keys: {data.keys()}"
            sorted_defaults[param] = {}
            for key in keys:
                if key in data:
                    sorted_defaults[param][key] = data[key]

        with open(path, "w") as f:
            f.write(json.dumps(sorted_defaults, indent=4))
```

jdebacker:
@hdoupe I ran through OG-USA with TC built from this branch after commit `f2c1dbd` and has no issues.

cc @MattHJensen @Peter-Metz

hdoupe:
Glad to hear it @jdebacker. Thanks again for testing out these changes.

hdoupe:
@Peter-Metz @MattHJensen #2401 is ready for another round of review. I think we should be pretty close to getting this merged.

@Peter-Metz are you still good with re-working the handling of the `CPI_offset` parameter in a follow up PR?

taxcalc/tests/test_parameters.py
hdoupe: I added these tests back with the PT approach of extending parameter values.

taxcalc/tests/test_policy.py
hdoupe: Thanks @Peter-Metz. Just updated the test name and docstring. I also removed the parameter name underscores and added tests for the underlying `inflation_rates` and `wage_growth_rates` methods.

Peter-Metz:
@hdoupe @MattHJensen changes look good to me and `print_warnings` functionality works as desired!

hdoupe:
@Peter-Metz sounds great. I'm excited to see your `CPI_offset` work.

MattHJensen:
@hdoupe, thanks very much for this PR and for so generously accommodating Peter and my requests through the review process. I am merging this PR now and expect it to appear in a new release within the month.

MattHJensen:
@Peter-Metz, thanks for all of your help with the review!

hdoupe:
> thanks very much for this PR and for so generously accommodating Peter and my requests through the review process

It was my pleasure. @MattHJensen and @Peter-Metz thanks for the thorough review!

PR #2404, opened 2020-03-16
Update test_calculator in preparation for release

Peter-Metz:
Minor changes to make sure all pre-release tests (other than documentation) pass before release

@MattHJensen


Peter-Metz:
After looking into the reason for the need for this PR (i.e. why there is a small difference between some calculated components of itemized deductions and the total amount of itemized deductions), I haven't been able to find a good explanation. However here are some things to consider: 
- In the past, the margin or error (or "tolerance") has been adjusted when economic projections are updated (as they are in this case).
- The margin of error disappears for years after 2018.
- In 2017, the margin of error represents a 0.005% difference between the ID component (`c19700`) and total ID

cc @MattHJensen

MattHJensen:
Thanks @Peter-Metz for the PR and for investigating the tolerance increase. I am going to merge this so we can get a release out because people are waiting on it. Could you open up an issue in Tax-Calculator about how this sensitivity test needs documentation? I will tackle it when I have an opportunity.

PR #2444, opened 2020-07-29
Update Tax-Calculator with latest CPS file and CPS weights

Peter-Metz:
This PR brings Tax-Calculator up to speed to the recent updates to the CPS file and CPS weights in PSLmodels/taxdata#314. In addition to changes to the files themselves, `nu05` was replaced with `nu06`.

A few notes:
- The aggregate benefits and tax information statistics change significantly. I'm not totally apprised of the changes made to the CPS, so I'd appreciate @andersonfrailey input about whether this is expected.
- I've changed some testing thresholds [here](https://github.com/PSLmodels/Tax-Calculator/blob/master/taxcalc/tests/test_calculator.py#L720) and [here](https://github.com/PSLmodels/Tax-Calculator/blob/master/taxcalc/tests/test_cpscsv.py#L76:L79) for the tests to pass. The thresholds change significantly but seem arbitrary in the first place. 

cc @MattHJensen 

Note: the `puf.csv` file (needed to pass tests locally) used for this PR was generated with the taxdata Makefile (i.e. `make puf-files`). As of 8/6, the updated `puf.csv` has not been distributed to users.


MattHJensen:
Also cc @MaxGhenis as a frequent user of the file.

MattHJensen:
I'd like to wait on merging this until @andersonfrailey has an opportunity to take a look at the change in results.

andersonfrailey:
Missed when this was opened. I'll take a look later tonight

andersonfrailey:
@Peter-Metz I wouldn't worry about the changes in benefit and tax information primarily because we have validation scripts in taxdata to ensure that all income, benefit, and tax unit composition information is accurately represented in the final CPS file. Also, one of the steps in producing the old CPS file was splitting up any tax units that had information that was top-coded into 10 or 15 separate units and randomly changed each tax unit's value for the top coded variable, which would affect unweighted totals. We no longer do this because now instead of putting a hard cap on the variables that are subject to top coding, Census just shuffles them between similar people.

As for any changes the weighted totals, we switched to a new solver for creating the weights, which would explain the differences there.

MattHJensen:
@Peter-Metz, thanks for this PR and sorry that it now has a minor conflict. I added a WIP tag while we wait for that to be resolved. Afterwards, based on @andersonfrailey's feedback, it sounds like this is good to go.

Peter-Metz:
One more question before I resolve the merge conflicts -- several functions that calculate child tax credits in `calcfunctions.py` rely on `nu05`. Can @andersonfrailey or @MattHJensen confirm that we are good to switch this to `nu06` on a conceptual level?

MattHJensen:
@Peter-Metz, this is a good point about `nu05`. Currently it is only used for the computation of `ACTC_rt_bonus_under5family` and `CTC_new_c_under5_bonus`, both of which are 0 under current law but are used in the `Clinton2016.json` reform file and may be used elsewhere by users. 

I think it means we have two options: 

1. Keep `nu05` and `nu06` on both PUF and CPS, and leave the under5 bonus params as they are. This would ensure backwards compatibility for users of those params, but it would increase the size fo the file marginally and perhaps slow us down due to the extra work required to add `nu05` back to this CPS file. 
2. Drop `nu05`, replace the under5_bonus params with new under6_bonus params, adding an error message for under5 usage like we have for cpi_offset, and update Clinton2016.json to use `nu06` while noting the discrepancy from her stated policy proposal in the file notes. (This would also mean that the PUF and CPS file updates would likely need to be combined in the same PR.)

I am inclined towards (1), but we could get away w (2) given that we are coming up on a major release. I think (1) is better since the Clinton reform is important as a historical artifact, and, if she proposed a reform depending on `nu05`, it seems likely that others will do the same.

MattHJensen:
(I am seeing that I suggested (2) previously. https://github.com/PSLmodels/taxdata/pull/329#issuecomment-549367220)

MattHJensen:
Kindergarten entrance is most typically based on whether you are 5 in September (3/4 of the way through the year), so I can see a stronger case for policies to use nu06 rather than nu05. https://nces.ed.gov/programs/statereform/tab5_3.asp

MattHJensen:
I'm also seeing that @martinholmer suggested replacing nu05 with nu06 and moving Clinton2016.md to `taxcalc/reforms/archive.` https://github.com/PSLmodels/taxdata/issues/311#issuecomment-482065337

Upon further reconsideration, I suggest we go ahead and do that as part of following (2) from https://github.com/PSLmodels/Tax-Calculator/pull/2444#issuecomment-669981122.

Peter-Metz:
@MattHJensen I agree with your reasoning to replace `nu05` with `nu06`. The latest commits implement that change

Peter-Metz:
The GH action to build the jb docs is currently failing while trying to execute `recipe05`. Traceback is below. In a different PR, I hit this same error locally and was able to fix it by updating jupyter-book.

Is it possible we need to activate the `taxcalc-dev` environment in the GH action? Or perhaps install jupyter-book with `pip install -U jupyter-book`? Any other ideas?

cc @jdebacker @MaxGhenis @hdoupe 

```
ValueError                                Traceback (most recent call last)
<ipython-input-2-b85009722bd6> in <module>
    113 
    114 # construct reform-vs-baseline difference table with results for income deciles
--> 115 diff_table = calc1.difference_table(calc2, 'weighted_deciles', 'iitax')
    116 assert isinstance(diff_table, pd.DataFrame)
    117 diff_extract = pd.DataFrame()

~/work/Tax-Calculator/Tax-Calculator/taxcalc/calculator.py in difference_table(self, calc, groupby, tax_to_diff, pop_quantiles)
    524         calc_var_dframe = calc.dataframe(DIFF_VARIABLES)
    525         diff = create_difference_table(self_var_dframe, calc_var_dframe,
--> 526                                        groupby, tax_to_diff, pop_quantiles)
    527         del self_var_dframe
    528         del calc_var_dframe
~/work/Tax-Calculator/Tax-Calculator/taxcalc/utils.py in create_difference_table(vdf1, vdf2, groupby, tax_to_diff, pop_quantiles)
    531         dframe = add_quantile_table_row_variable(
    532             df2, baseline_expanded_income, 10,
--> 533             pop_quantiles=pop_quantiles, decile_details=True)
    534     elif groupby == 'standard_income_bins':
    535         dframe = add_income_table_row_variable(

~/work/Tax-Calculator/Tax-Calculator/taxcalc/utils.py in add_quantile_table_row_variable(dframe, income_measure, num_quantiles, pop_quantiles, decile_details, weight_by_income_measure)
    229     labels = range(1, (num_bins + 1))
    230     dframe['table_row'] = pd.cut(dframe['cumsum_temp'], bin_edges,
--> 231                                  right=False, labels=labels)
    232     dframe.drop('cumsum_temp', axis=1, inplace=True)
    233     return dframe

/usr/share/miniconda/envs/taxcalc-dev/lib/python3.7/site-packages/pandas/core/reshape/tile.py in cut(x, bins, right, labels, retbins, precision, include_lowest, duplicates, ordered)
    271         # GH 26045: cast to float64 to avoid an overflow
    272         if (np.diff(bins.astype("float64")) < 0).any():
--> 273             raise ValueError("bins must increase monotonically.")
    274 
    275     fac, bins = _bins_to_cuts(

ValueError: bins must increase monotonically.
```

MaxGhenis:
Looks more likely to be a `pandas` upgrade issue.

Peter-Metz:
The latest commit changes `recipe05` so that it groups the difference table by standard income bins instead of weighted deciles. More about why this change was necessary for the docs to build can be found in #2460. 

cc @MattHJensen

MattHJensen:
@Peter-Metz, could you remove the WIP tag when this is ready for another review?

Peter-Metz:
@MattHJensen done

MattHJensen:
Thanks @Peter-Metz. Merging.

Issue #2481, opened 2020-09-02
--tables and --graphs arguments causing errors


aelsibaie:
Steps to replicate:

Run `tc cps.csv 2022 --reform ptaxes0.json --graphs` or `tc cps.csv 2022 --reform ptaxes0.json --tables`
with the `ptaxes0.json` containing
```
{
    "FICA_ss_trt": {
        "2021": 0.130,
        "2022": 0.140
    },
    "FICA_mc_trt": {
        "2021": 0.030,
        "2022": 0.032
    }        
}
```

Associated errors:
```
(psl) C:\Users\Amir>tc cps.csv 2022 --reform ptaxes0.json --tables
You loaded data for 2014.
Tax-Calculator startup automatically extrapolated your data to 2022.
Traceback (most recent call last):
  File "C:\Users\Amir\miniconda3\envs\psl\lib\site-packages\pandas\core\indexes\base.py", line 2889, in get_loc
    return self._engine.get_loc(casted_key)
  File "pandas\_libs\index.pyx", line 70, in pandas._libs.index.IndexEngine.get_loc
  File "pandas\_libs\index.pyx", line 97, in pandas._libs.index.IndexEngine.get_loc
  File "pandas\_libs\hashtable_class_helper.pxi", line 1675, in pandas._libs.hashtable.PyObjectHashTable.get_item
  File "pandas\_libs\hashtable_class_helper.pxi", line 1683, in pandas._libs.hashtable.PyObjectHashTable.get_item
KeyError: 0

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "C:\Users\Amir\miniconda3\envs\psl\Scripts\tc-script.py", line 10, in <module>
    sys.exit(cli_tc_main())
  File "C:\Users\Amir\miniconda3\envs\psl\lib\site-packages\taxcalc\cli\tc.py", line 186, in cli_tc_main
    output_sqldb=args.sqldb)
  File "C:\Users\Amir\miniconda3\envs\psl\lib\site-packages\taxcalc\taxcalcio.py", line 464, in analyze
    self.write_tables_file()
  File "C:\Users\Amir\miniconda3\envs\psl\lib\site-packages\taxcalc\taxcalcio.py", line 546, in write_tables_file
    TaxCalcIO.write_decile_table(distdf, tfile, tkind='Reform Totals')
[...]

PR #2482, opened 2020-09-03
Update Conda installation instructions

Peter-Metz:
Adds the `conda-forge` channel to the conda installation instructions.

If the user doesn't have `conda-forge` in their default set of conda channels, `conda install -c pslmodels taxcalc` will install `taxcalc` 2.9.

This issue was reported by @aelsibaie in #2481

### DIFF
--- a/docs/usage/starting.md
+++ b/docs/usage/starting.md
@@ -26,7 +26,7 @@ Download a Tax-Calculator `taxcalc` package for your computer by executing
 this command from the command prompt:
 
 ```
-conda install -c PSLmodels taxcalc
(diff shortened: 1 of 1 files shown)


hdoupe:
LGTM, thanks for updating the docs @Peter-Metz!

MattHJensen:
Thanks @Peter-Metz. Merging.

Peter-Metz:
@MattHJensen what do you think about issuing a patch release to get this change into the docs? I could see this being a problem for a lot of users trying to update to 3.0

PR #2486, opened 2020-09-08
MTR for partnership income

codykallen:
In the the PUF, we use the variable `e26270` as holding partnership and S corporation income, and we use `k1bx14p` and `k1bx14s` for partnership income for the primary filer and spouse. The `Calculator.mtr` function allows for the computation of the MTR on `e26270`. Because this computation updates `e27260` but not `k1bx14p` or `k1bx14s`, this is technically only computing the MTRs on S corporation income, not partnership income. This PR adds MTR calculations for `k1bx14p`. 

The distinction between these in Tax-Calculator logic is that partnership income is hit by the self-employment tax, whereas S corporation income is not.

### DIFF
--- a/taxcalc/calculator.py
+++ b/taxcalc/calculator.py
@@ -536,7 +536,8 @@ class Calculator():
                            'e02400', 'p22250',
                            'p23250', 'e18500',
                            'e19200', 'e26270',
-                           'e19800', 'e20100']
+                           'e19800', 'e20100',
+                           'k1bx14p']
 
     def mtr(self, variable_str='e00200p',
             negative_finite_diff=False,
@@ -622,7 +623,8 @@ class Calculator():
         'e19200',  Schedule A interest paid;
         'e26270',  S-corporation/partnership income (also included in e02000);
         'e19800',  Charity cash contributions;
-        'e20100',  Charity non-cash contributions.
+        'e20100',  Charity non-cash contributions;
+        'k1bx14p', Partnership income (also included in e26270 and e02000).
         """
         # pylint: disable=too-many-arguments,too-many-statements
         # pylint: disable=too-many-locals,too-many-branches
@@ -649,6 +651,9 @@ class Calculator():
             divincome_var = self.array('e00600')
         elif variable_str == 'e26270':
             scheincome_var = self.array('e02000')
+        elif variable_str == 'k1bx14p':
+            scheincome_var = self.array('e02000')
+            scorpincome_var = self.array('e26270')
         # calculate level of taxes after a marginal increase in income
         self.array(variable_str, variable + finite_diff)
         if variable_str == 'e00200p':
@@ -661,6 +666,9 @@ class Calculator():
             self.array('e00600', divincome_var + finite_diff)
         elif variable_str == 'e26270':
             self.array('e02000', scheincome_var + finite_diff)
+        elif variable_str == 'k1bx14p':
+            self.array('e02000', scheincome_var + finite_diff)
+            self.array('e26270', scorpincome_var + finite_diff)
         if self.__consumption.has_response():
             self.__consumption.response(self.__records, finite_diff)
         self.calc_all(zero_out_calc_vars=zero_out_calculated_vars)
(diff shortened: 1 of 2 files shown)


codykallen:
This PR adds calculations for the MTR on partnership income. @jdebacker, @kpomerleau, this may be useful for you, especially for calculating MTRs on business income.

@martinholmer @MattHJensen, this passes almost all the tests, except that it doesn't pass the codecov test because:

> Merging #2486 into master will decrease coverage by 0.00%.

MattHJensen:
Thanks for the PR @codykallen! This looks good, but I'll review again tomorrow before merging.

jdebacker:
@codykallen Thanks for helping Tax-Calc produce accurate combined MTRs on pass-through income.

You note:

> Because this computation updates `e27260` but not `k1bx14p` or `k1bx14s`, this is technically only computing the MTRs on S corporation income, not partnership income. This PR adds MTR calculations for `k1bx14p`.

`e27260` include partnership and S corp income, so why isn't computing the MTR on the combined variable not including the MTR on partnership income?  `e27260` figures into the calculations of total tax liability as far as I can see, but perhaps it's missing some logic that distinguishes between that taxation of S corp and partnership income?

> The distinction between these in Tax-Calculator logic is that partnership income is hit by the self-employment tax, whereas S corporation income is not.

This is where I think your contribution is more relevant - SECA taxes are ignored when computing the MTR on `e27260` because they are not in the [`EI_PayrollTax` function](https://github.com/PSLmodels/Tax-Calculator/blob/3b9f4642ba53ed8e5d01155f543ad5d0e6f2f526/taxcalc/calcfunctions.py#L92), but `k1bx14p` and `k1bx14s` are.

However, from my read of your additions, it looks like you consider incremental changes in `e27260` and `e02000` when computing the MTR from `k1bx14p`:

```
elif variable_str == 'k1bx14p':
             scheincome_var = self.array('e02000')
             scorpincome_var = self.array('e26270')
.
.
.
elif variable_str == 'k1bx14p':
             self.array('e02000', scheincome_var + finite_diff)
             self.array('e26270', scorpincome_var + finite_diff)
```

Wouldn't you want to apply the finite diff to `k1bx14p` itself?  Changing `e02200` and `e26270` have no effect on the SECA calculation.  

But if you do this, then I think you miss other, non-SECA changes because `k1bx14p` [doesn't seem to be referenced in other parts of the tax logic](https://github.com/PSLmodels/Tax-Calculator/search?q=k1bx14p&unscoped_q=k1bx14p).

I could likely be missing something, but it seems more changes are needed to model the taxation of partnership income properly.

codykallen:
@jdebacker asked:

> e27260 include partnership and S corp income, so why isn't computing the MTR on the combined variable not including the MTR on partnership income? e27260 figures into the calculations of total tax liability as far as I can see, but perhaps it's missing some logic that distinguishes between that taxation of S corp and partnership income?

When we compute the MTR on `e26270`, we increase `e26270` by `finite_diff`, but we don't also increase `k1bx14p`. Therefore, this is an increase in S corporation income, but not an increase in partnership income. The only real distinction between partnership and S corporation income in Tax-Calculator is that partnership income gets hit by the self-employment tax and S corporation income does not. 

Also, when computing the MTR on `k1bx14p`, we need to increase `k1bx14p` by `finite_diff`, but we also need to increase every variable that `k1bx14p` is included in, which are `e26270` and `e02000`. Otherwise, we would be calculating the marginal tax effect of switching a unit of income from being classified as S corporation income to being classified as partnership income. 

@jdebacker also asked:

> Wouldn't you want to apply the finite diff to k1bx14p itself? 

That gets done automatically by the command:
`self.array(variable_str, variable + finite_diff)`

jdebacker:
@codykallen Thanks for explaining - I forgot about the `self.array(variable_str, variable + finite_diff)`.  All looks good to me.

MattHJensen:
@codykallen, thanks again for this PR, and @jdebacker, thanks a lot for helping with the review! Merging.

PR #2453, opened 2020-08-11
TAXSIM-32 validation update 

chusloj:
This PR updates the validation code for [TAXSIM-27](https://users.nber.org/~taxsim/taxsim27/) to use the new [TAXSIM-32](https://users.nber.org/~taxsim/taxsim32/) model.

Issue #2385, opened 2019-11-26
20% QBI (Pass-through income) deduction revenue estimate


kpomerleau:
Currently, the revenue effect of repealing the 20% qualified business income deduction raises about $16 billion each year. However, this is far lower than the revenue effect estimated by the Joint Committee on Taxation, which finds a revenue effect of about $50-$60 billion per year. Link to JCT estimate: https://www.cbo.gov/system/files/2018-06/51138-2018-04-revenueprojections.xlsx (tab 7).

My first hypothesis is that it may interact with the AMT, but it looks like this is not the case.

My second hypothesis is that there is an issue in how the deduction itself is being calculated. Given that tax brain does not have switches for the W-2 and K limits, it's hard to tell what's going on there and I am not sure what sort of data is driving that, so I don't have a suggestion there besides trying to run the model without the limitation code to see if the revenue effect is closer to other known estimates. 

Another thought, based on experience: JCT likely assumed some amount of general avoidance in the pass-through sector in response to the deduction. It might be worth while to have some sort of parameter for general avoidance for this provision (which could simply be a factor increasing qualified pass-through income) to baseline the model to the revenue loss that the JCT thinks would occur.

jdebacker:
@kpomerleau I'm not sure exactly how your are getting the revenue estimate form TC, but one thing to be wary of is comparing runs when there is a different stacking order when evaluating a several policy changes at once.  

E.g., if you were to do a revenue estimate of repealing the 20% QBI deduction by itself (i.e., keeping the lower individual tax rates, etc from the TCJA), you'd get a smaller number than if you estimated the 20% QBI deduction after repealing the lower rates under TCJA (since the deduction would be worth more with the higher rates).   

The CBO numbers your reference include the 20% QBI extension as part of a stack of policies. It looks like they come after the rate changes, but there are other important provisions that interact with the 20% QBI such as bonus deprecation, which appear to be stacked after the QBI extension.

But you are right that behavioral responses are important.  I know the Behavioral Responses package will allow you to apply elasticities that can capture avoidance behavior.  I'm not sure if  you can specify separate elasticities for pass-through income, though.  @martinholmer or @MattHJensen might be able to answer that.

MattHJensen:
Thanks for opening this issue. I will start looking into this, and I very much appreciate any headway others make as well. 

[...]

PR #2468, opened 2020-08-20
Require Paramtools >= 0.14.2

Peter-Metz:
Paramtools 0.14.2 resolves error thrown by `make_uguide.py`.

This PR takes the place of #2423. I accidentally opened #2423 from a branch of `PSLmodels/Tax-Calculator` as opposed to `Peter-Metz/Tax-Calculator`.

PR #2497, opened 2020-10-07
[WIP] Add switch for QBI deduction wage/capital limitations

Peter-Metz:
This PR adds a new parameter, `PT_qbid_limit_switch`, that turns on/off the QBI phaseout.

Above a certain taxable income threshold, the QBI deduction is subject to limitations based on wage expenses and capital income. Neither the PUF nor CPS contain the wage/capital data necessary to compute the limitations, so Tax-Calculator treats all taxpayers as having zero wage expenses and no capital income, and therefore fully subject to the phaseout.  This issue has been identified in multiple places, including #2385 and PSLmodels/taxdata#319.

In #2453 ([comment](https://github.com/PSLmodels/Tax-Calculator/pull/2453#issuecomment-688902747)), @feenberg told us that TAXSIM makes a different assumption, that all taxpayers have sufficient wage expense/ capital income to avoid the limitation.

This PR will give Tax-Calculator users the option to choose between the two approaches. The revenue implications are fairly large:

With full wage/capital limitations:
Screen Shot 2020-10-07 at 9 17 10 AM

Without wage/capital limitations:
Screen Shot 2020-10-07 at 9 17 57 AM


feenberg:
Does Matt or anyone else have a contact at JCT, OTA or CBO that can tell 
us what they are doing about the QBI wage and capital features?

Dan

MattHJensen:
Background: 

From a CBO modeler in response to a request from @feenberg, as reported in  https://github.com/PSLmodels/Tax-Calculator/pull/2453#issuecomment-688902747. 

> I asked a couple of colleagues about the binary choice between
assuming 1) no taxpayer is subject to the wage/capital restriction or
whether 2) all taxpayers are fully subject to the limitations (i.e, treat
them as having zero wage expense and no capital).  The consensus here is
that underestimate from option #2 would be larger than the overestimate
from option #1.

Discussion: 

@Peter-Metz and I discussed these ideas in chat. Interested in others' feedback. 

1. This proposed tax law parameter is really filling in for a data limitation. 
2. Using a tax law parameter in this way would be somewhat confusing to users. They would be required set a tax law parameter to a value that is at odds with current law (i.e., PT_qbid_limit_switch = false) under both baseline and reform scenario in order to better approximate what we think is happening under current law, namely that the non-binding limits are more accurate than fully-binding limits. 
3. Instead of adding a tax law parameter, this issue could be resolved in TaxData, either by imputing PT_binc_w2_wages and PT_ubia_property that mean no taxpayer will be subject to the wage/property limitation by or imputing values that bring the deduction amounts in line with SOI statistics. 
4. In the meantime, Tax-Calculator, Tax-Brain, CCC, OG-USA could focus on documenting the data limitation. (cc @jdebacker, @rickecon, @andersonfrailey.) 

@Peter-Metz, I'm interested in whether you still agree with these points. In the meantime, I've added a WIP tag here.

Peter-Metz:
@MattHJensen I agree that it would be conceptually nicer to handle this in taxdata, although it seems that it would be functionally the same. Two outstanding questions on my end:
1. Do we impute any other parameters in taxdata (cc @andersonfrailey)? We might want to have a separate conversation in taxdata regarding what the process would look like and any other considerations.
2. Regarding your second point, if we stick with a policy parameter switch, couldn't we make the default such that taxpayers are not subject to the limitations? If users wanted to fully impose wage/capital restrictions in their analysis, they could use the Python API to implement the restrictions in both the base and reform calculators.

MattHJensen:
> Do we impute any other parameters in taxdata (cc @andersonfrailey)? We might want to have a separate conversation in taxdata regarding what the process would look like and any other considerations.

See, e.g.,  `impute_itemxp.py`, `impute_pencon.py`, and a smaller imputations in `finalprep.py`. https://github.com/PSLmodels/taxdata/tree/master/puf_data. 

 > Regarding your second point, if we stick with a policy parameter switch, couldn't we make the default such that taxpayers are not subject to the limitations? If users wanted to fully impose wage/capital restrictions in their analysis, they could use the Python API to implement the restrictions in both the base and reform calculators.

This is true, we could set the default under baseline so that taxpayers aren't subject to the limitations. And that does seem like the natural solution if we want to solve this issue in Tax-Calculator. But I am quite hesitant because it means purposefully changing baseline to be out of step with tax law. Any downstream users that do have better data, are willing to do their own imputations, or rely on their users for data (like Tax-Cruncher) would be disadvantaged. Short version: I'm concerned about introducing an intentional error in Tax-Calculator's representation of tax law to support puf.csv users at the expense of other users.

donboyd5:
The QBI deduction wage/capital limitations issue has come up in #2513 as a possible explanation for why puf.csv shows net TCJA tax increases for AGI millionaires. In rough terms, I'm guessing I'm finding $12b more TCJA tax for agi millionaires than it looks like JCT has: I find a $3.8b net increase vs. very-very rough not-quite comparable JCT estimate of -$8b, or about a $12b difference. Of course JCT could be wrong, the numbers could be not comparable enough, data could have changed, etc. (And for reasons mentioned in #2513 related to SALT, the non-SALT related difference might be larger than $12b.)

@Peter-Metz (cc @MattHJensen, @feenberg, @andersonfrailey), as I read your tables above, it looks like the `PT_qbid_limit_switch` approach results in a tax change of about $35b ($1,557b - $1,522b) which seems consistent with the implied $34-44b JCT estimate mentioned in #2385. And a $35b difference across all taxpayers certainly seems like it could be large enough to help explain a $12b+ difference in tax for agi millionaires.

I'm wondering if your $35b falls in the right income ranges to help explain this. Is there any chance that you have, or could run at some point, your `PT_qbid_limit_switch` approach by agi range or some reasonably close approximation to agi? It would be nice to see how much of the difference is for agi millionaires. If so, that would be really great. Additional comments below.

donboyd5:
@Peter-Metz  @MattHJensen, @feenberg, @andersonfrailey,

My own thoughts on what would be useful is to have the issue well documented and leave placeholders or something like that in taxdata for variables needed to compute the limitation:

- specialized service business indicator (PT_SSTB_income)
- W-2 wages the business pays (PT_binc_w2_wages), and
- filer's share of qualified properties (PT_ubia_property)

and to have tax-calculator implement that logic ([which I understand that it does, I think](https://github.com/PSLmodels/taxdata/issues/319)).

Then, in the short term users can add these variables to puf.csv if they believe they have a way to impute them, and tax-calculator can use them if available. Over the longer term, perhaps taxdata could impute them.

MattHJensen:
@Peter-Metz,  I've overlooked the elegance of this approach -- sorry about that. I plan to merge this PR soon, pending a request below. 

The capability to turn off the limitation is a reasonable policy parameter to add in its own right, to help people designing and conducting policy reforms. If some users want to update the parameter in the baseline to accommodate data limitations, that's fine too. 

When I run the full test suite, I see new MTR results. Would you mind doing the same to confirm?

donboyd5:
Great! Thanks. That would really help in what we're doing re NY.

MattHJensen:
@donboyd5 It has certainly helped to clarify my thinking! 


By the way, do you know that you can check out this branch locally? 

`git fetch upstream pull/2497/head:pr-2497`

or

image

Peter-Metz:
@MattHJensen said:

>When I run the full test suite, I see new MTR results. Would you mind doing the same to confirm?

Hmm are you referring to the errors that say `IndexError: too many indices for array`? At the moment I'm hitting a few of those errors with the full test suite. These same errors also occur on the master branch, so it's not an issue with this PR per say. My only guess at the moment is that it has something to do with #2468 

@donboyd5 Normally I'd be happy to paste a distribution table, but I'm currently hitting the error I mentioned above. I'll check back in with this.

donboyd5:
Thanks @MattHJensen and @Peter-Metz.

Now that I know I can check it out from GH (something I've never done), after you resolve issues to your satisfaction, I'll do that and run it locally and do distribution tables.

If there are any pointers you can give after you've resolved issues - for example, showing how to incorporate the new parameters into a json file, or really just a bit of guidance on what the values should be - that would be great.

Thanks!

MattHJensen:
@Peter-Metz, I'm just getting this: 

```
E           ValueError: PUFCSV MTR RESULTS DIFFER
E           -------------------------------------------------
E           --- NEW RESULTS IN pufcsv_mtr_actual.txt FILE ---
E           --- if new OK, copy pufcsv_mtr_actual.txt to  ---
E           ---                 pufcsv_mtr_expect.txt     ---
E           ---            and rerun test.                ---
E           -------------------------------------------------

tests/test_pufcsv.py:273: ValueError
```

Peter-Metz:
@MattHJensen You're totally right. I'm using a new computer and didn't have things set up properly. I think #2498 fixes that MTR test.

@donboyd5 Now that I actually have things set up, here is the difference table (in $b), grouped by expanded income. Note the $24B tax cut for millionaires.

Screen Shot 2020-11-25 at 6 37 18 PM

donboyd5:
Perfect, thanks!  Don

MattHJensen:
> I think #2498 fixes that MTR test.

@Peter-Metz, I may be missing something -- could you expand? Shouldn't the test pass here since #2498 is already merged and this PR doesn't influence MTRs?

Peter-Metz:
@MattHJensen I hadn't merged in those changes to this branch yet. Test passes now

MattHJensen:
Thanks very much @Peter-Metz. Merging now.

PR #2500, opened 2020-10-27
Modify base for top capital gains tax rate (CG_rt4)

Peter-Metz:
Currently, the capital gains function inadvertently stacks the second (`CG_brk2`) and third (`CG_brk3`) brackets before applying the top capital gains rate (`CG_rt4`), instead of applying the top rate to income above `CG_brk3`. In current law, `CG_brk3` is set to `9e99` and `CG_rt4` is set to `1.0` so this change has no bearing on current law.

The current behavior can be visualized with a Tax-Cruncher skyline chart of MTR's wrt long-term capital gains. In the chart below, the top bracket is set to $1,000,000 but the top rate doesn't kick in until ~$1,450,000:

Screen Shot 2020-10-27 at 2 17 02 PM

I've also added a test that checks MTR's on the edges of the top bracket.

### DIFF
--- a/taxcalc/calcfunctions.py
+++ b/taxcalc/calcfunctions.py
@@ -994,7 +994,9 @@ def GainsTax(e00650, c01000, c23650, p23250, e01100, e58990, e00200,
         dwks30 = dwks22 + dwks28
         dwks31 = dwks21 - dwks30
         dwks32 = CG_rt3 * dwks31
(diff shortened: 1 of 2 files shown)


MattHJensen:
Thanks @Peter-Metz. Glad to see the new test, too. 

I also confirmed that the top rate under current law kicks in at ~$435 (single) as expected. 

image

PR #2502, opened 2020-10-30
[WIP] Add parameters for refundable CDCC

Peter-Metz:
This PR adds the tax logic and parameters for computing a fully refundable child and dependent care credit, as proposed by Joe Biden:

Screen Shot 2020-10-30 at 8 39 57 AM

([source](https://joebiden.com/caregiving/#))

Here are the distributional effects in 2021 of implementing Biden's CDCC provision (assuming that the new CDCC replaces the original, non-refundable CDCC):

Screen Shot 2020-10-30 at 8 44 47 AM

### DIFF
--- a/docs/guide/output_vars.md
+++ b/docs/guide/output_vars.md
@@ -529,6 +529,13 @@ _IRS Form Location:_
 2013-20??: calculated variable  
 
 
+##  `cdcc_new`  
+_Description_: search taxcalc/calcfunctions.py for how calculated and used  
+_Datatype_: float  
+_IRS Form Location:_  
+2013-20??: calculated variable  
+
+
 ##  `odc`  
 _Description_: Other Dependent Credit  
 _Datatype_: float  
--- a/docs/guide/policy_params.md
+++ b/docs/guide/policy_params.md
@@ -3763,6 +3763,94 @@ _Valid Range:_ min = 0 and max = 1
 _Out-of-Range Action:_ error  
 
 
+### Refundable Child and Dependent Care Credit
+
+####  `CDCC_new_c`  
+_Description:_ The maximum refundable credit allowed for each qualifying dependent.  
+_Has An Effect When Using:_ _PUF data:_ True _CPS data:_ True  
+_Can Be Inflation Indexed:_ True _Is Inflation Indexed:_ True  
+_Value Type:_ float  
+_Known Values:_  
+2013: 0.0  
+2014: 0.0  
+2015: 0.0  
+2016: 0.0  
+2017: 0.0  
+2018: 0.0  
+2019: 0.0  
+_Valid Range:_ min = 0 and max = 9e+99  
+_Out-of-Range Action:_ error  
+
+
+####  `CDCC_new_deps_max`  
+_Description:_ The refundable child & dependent care credit will be applied up to this many dependents.  
+_Has An Effect When Using:_ _PUF data:_ True _CPS data:_ True  
+_Can Be Inflation Indexed:_ False _Is Inflation Indexed:_ False  
+_Value Type:_ int  
+_Known Values:_  
+2013: 10  
+2014: 10  
+2015: 10  
+2016: 10  
+2017: 10  
+2018: 10  
+2019: 10  
+_Valid Range:_ min = 0 and max = 10  
+_Out-of-Range Action:_ error  
+
+
+####  `CDCC_new_frac`  
+_Description:_ Child care costs are multiplied by this fraction to calculate the refundable child & dependent care credit amount.  
+_Has An Effect When Using:_ _PUF data:_ True _CPS data:_ True  
+_Can Be Inflation Indexed:_ False _Is Inflation Indexed:_ False  
+_Value Type:_ float  
+_Known Values:_  
+2013: 1.0  
+2014: 1.0  
+2015: 1.0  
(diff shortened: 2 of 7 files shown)


Peter-Metz:
The test failure is because `docs/guide/policy_params.md` is outdated. I thought the docs were updated during releases?

hdoupe:
@Peter-Metz oh yeah, they should have been: https://github.com/PSLmodels/Tax-Calculator/blob/master/.github/workflows/deploy_parameters_docs.yml

Peter-Metz:
@hdoupe yeah, as far as I know the GH action is working. But I'm failing a test because I added parameters that don't show up in `policy_params.md` (because the docs haven't been built yet)

Peter-Metz:
I built the docs to get the tests to pass, but we should consider what to do with `test_policy.py::test_section_titles` given that the docs are now built when a release is issued.

MattHJensen:
@Peter-Metz, as I understand it, this parameter will offer misleading results with the PUF or CPS input files because qualifying expenses, `e32800`, is capped at $3,000 under current law for 1 dependent and $6,000 for 2 or more. 

From the 2011 PUF documentation: http://users.nber.org/~taxsim/gdb/gdb11.pdf

image

Has this been resolved in some way that I am missing/forgetting? 

If not, we could add the parameter but mark it incompatible with the CPS and PUF datafiles. 

-----

Separately, why the cap on the number of eligible dependents? I don't see that in Biden's proposal, and it adds complexity to the implementation.

Peter-Metz:
Thanks @MattHJensen. I didn't realize `e32800` was capped, so marking these parameters as incompatible with the puf and cps makes sense.

>Separately, why the cap on the number of eligible dependents?

Biden's proposal has a max credit of "$8,000 for one child and $16,000 for two or more children." I'm reading this as a cap of two eligible dependents.

Peter-Metz:
The latest commit makes `CDCC_new_*` incompatible with the PUF and CPS. I've also added the following note to `policy_current_law.json`:

>Child/dependent care expenses (e32800) reported in the PUF and CPS files are capped. Therefore, CDCC_new_* should not be used with the PUF or CPS files.

MattHJensen:
> Biden's proposal has a max credit of "$8,000 for one child and $16,000 for two or more children." I'm reading this as a cap of two eligible dependents.

Ah, I see. The current cdcc_c also has this feature (at $3,000 for one or $6,000 for two or more). It is modeled w/o a cap like this: 

    max_credit = min(f2441, 2) * CDCC_c
    c32800 = max(0., min(e32800, max_credit))

Peter-Metz:
The logic is the same, right? The only difference is that the two is hardcoded in the cdcc function.

MattHJensen:
> The logic is the same, right? The only difference is that the two is hardcoded in the cdcc function.

Yes, I am seeing now that the logic is implemented the same, but the hardcoded 2 doesn't represent represent the number of qualifying dependents, it represents the maximum number of qualifying dependents that will increase the deductible allowance. That allowance can be applied for qualified CDC expenses for an unlimited number of dependents. 

So I think the new parameter short name, title, and description are misstated. Given that they aren't necessary for the Biden proposal, one option is (a) to drop them and just hardcode a 2. Another option is (b) to think of a better way to describe them. At first I was leaning towards (a), but as part of writing this up I noticed that the documentation for existing parameter `CDCC_c`, is also similarly mislabeled, and figuring out a better way to describe that will probably require the same thinking as (b). So I am relatively indifferent between (a) and (b). 

```
"CDCC_c": {
        "title": "Maximum child & dependent care credit per dependent",
        "description": "The maximum amount of credit allowed for each qualifying dependent.",
```

I'll stew on this and reply if I think of a better way to describe `CDCC_c` and the new params. Interested in anything else you (@Peter-Metz) or others come up with too.

Peter-Metz:
@MattHJensen I see what you mean. One option would be to add an additional Paramtools "label" to `CDCC_c` and `CDCC_new_c` for number of qualifying dependents, similar to the `EITC_c` parameter. So, current law might look like:

```python
{CDCC_c: {2021: [0, 3000, 6000, 6000, 6000]}}
```

This approach would allow us to get rid of both the hardcoded 2 and any parameter for maximum number of dependents.

MattHJensen:
@Peter-Metz, that approach makes sense, but the parameter redesigns would break backwards compatibility. I suppose we could just go for it and tick up the version number. Do you think it is worth it, or am I missing a simpler approach?

Peter-Metz:
>@Peter-Metz want me to add a help-wanted tag here?

Thanks, @MattHJensen. Some help would be greatly appreciated!

martinholmer:
Why is PR #2502 --- Add parameters for refundable CDCC --- still open?

PR #2502 was opened on October 30, 2020.  But since PR #2573 was merged on March 15, 2021, the model has had a `CDCC_refundable` parameter and associated logic.  And the `CDCC_refundable` parameter is True for 2021 as part of current-law policy.

To an outside observer, PR #2502 seems redundant.  If PR #2502 would add something not now available in Tax-Calculator, this should be spelled out in the discussion of the pull request.

@Peter-Metz @MattHJensen @jdebacker

jdebacker:
I agree with Martin's comment above. This PR is no longer relevant since PR #2573 was merged. Closing.

PR #2508, opened 2020-11-02
Add parameters for QBI deduction phaseout

Peter-Metz:
This PR adds two parameters to model a phaseout of the QBI deduction, `PT_qbid_ps` and `PT_qbid_prt`, based on taxable income.

Biden has proposed a phaseout of the QBI deduction for those with taxable income > $400,000, although he has not specified the phaseout rate.

Since the wage and capital restrictions are fully enforced in Tax-Calculator for all taxpayers (see #2385 and PSLmodels/taxdata#319), implementing Biden's phaseout has little revenue effect.

### DIFF
--- a/taxcalc/calcfunctions.py
+++ b/taxcalc/calcfunctions.py
@@ -839,9 +839,8 @@ def TaxInc(c00100, standard, c04470, c04600, MARS, e00900, e26270,
         qbided = min(qbided, taxinc_cap)
 
         # apply qbid phaseout
-        modAGI = c00100
-        if qbided > 0. and modAGI > PT_qbid_ps[MARS - 1]:
-            excess = modAGI - PT_qbid_ps[MARS - 1]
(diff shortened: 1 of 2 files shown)


MattHJensen:
@Peter-Metz, thanks a lot for this PR. There are conflicts now, so I added back the WIP label. 

Do you think compatible_data should be set to false since these don't really work with puf and cps files?

Peter-Metz:
@MattHJensen the latest commit should resolve the conflicts.

>Do you think compatible_data should be set to false since these don't really work with puf and cps files?

Hmm, I'm not sure why the phaseout wouldn't work with the puf and cps. The only inputs needed to compute the phaseout are the amount of qbid and pre-qbid taxable income. What do you think?

MattHJensen:
> Hmm, I'm not sure why the phaseout wouldn't work with the puf and cps. The only inputs needed to compute the phaseout are the amount of qbid and pre-qbid taxable income. What do you think?

Yes, that makes sense to me now. Thanks a lot for the PR! Merging.

Issue #2521, opened 2020-12-24
Update to Python 3.9


MaxGhenis:
Python 3.9 was launched on 2020-10-05, conda only has up to 3.8: https://anaconda.org/pslmodels/taxcalc/files

MaxGhenis:
@MattHJensen could this get pushed to conda with taxcalc's next release?

MattHJensen:
> @MattHJensen will the next release be pushed to conda?

Yes, it will be pushed to conda.

jdebacker:
Closed with PR #2522.

PR #2522, opened 2020-12-24
Update to Python 3.9

MaxGhenis:
Addresses #2521 (once launched to conda)

If these are the correct six files to update when adding support for a new Python version, and if the policy is to support the three latest versions, I'd be happy to add that information to some part of the contributor documentation. Would this belong in the [contributor guide](https://pslmodels.github.io/Tax-Calculator/contributing/contributor_guide.html), [`RELEASING.md`](https://github.com/PSLmodels/Tax-Calculator/blob/master/RELEASING.md), or elsewhere?


MattHJensen:
Thanks a lot, @MaxGhenis. The only other file I could find that needs an update is `docs/usage/starting.md`

What do you think about a new /docs/contribution/dependencies.md? This Python management info could start things off, and later any other useful info on dependency management could go there.

MattHJensen:
Thanks a lot @MaxGhenis. Merging.

Issue #2518, opened 2020-12-20
Throw informative error when read_json_param_objects points to nonexistent json file


MaxGhenis:
I was helping another user who was confused by the error message shown in https://github.com/PSLmodels/Tax-Calculator/issues/2455 when `read_json_param_objects` points to a nonexistent json file:
```
ValueError: Text below contains invalid JSON:
Expecting value: line 1 column 1 (char 0)
Above location of the first error may be approximate.
The invalid JSON text is between the lines:
XXXX----.----1----.----2----.----3----.----4----.----5----.----6----.----7
0001_static/reformA.json
XXXX----.----1----.----2----.----3----.----4----.----5----.----6----.----7
```
It'd be good to check if the file exists and tell users when it doesn't.

MattHJensen:
Resolved by #2524

MattHJensen:
@MaxGhenis, thanks a lot for opening the issue as well as helping the other tax calculator user!

PR #2524, opened 2020-12-28
Check for existence of reform and assumption json files

chusloj:
This PR adds checks in the `Parameters` method `_read_json_revision` to check for the existence and validity of reform and assumption files. Addresses #2518.

cc @MaxGhenis

### DIFF
--- a/taxcalc/calculator.py
+++ b/taxcalc/calculator.py
@@ -1087,7 +1087,7 @@ class Calculator():
         The 'growdiff_response' subdictionary of the returned dictionary is
         suitable as input into the GrowDiff.update_growdiff method.
         """
-        # check for existence of reform
+        # check for existence of reform and assumption files
         if not os.path.isfile(reform):
             raise ValueError("This .json reform file does not exist or this is a misspecified directory.")
         if not os.path.isfile(assump):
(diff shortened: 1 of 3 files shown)


MattHJensen:
Thanks, @chusloj! 

@MaxGhenis, could you take a look at this to see if it resolves #2518 to your liking before I merge?

MaxGhenis:
LGTM, one other nice-to-have might be a test to ensure that passing a nonexistent json throws the specified error. Thanks @chusloj!

chusloj:
@MaxGhenis Good suggestion, I'll implement that and then this PR will be ready to merge.

chusloj:
Implemented new test. Now a non-existent file throws an `OSError`. A line in ` test_read_bad_json_assump_file()` already tests a non-existent assumption file, so it's been changed to scan for an `OSError` as well.

hdoupe:
@chusloj What do you think about throwing a [`FileNotFoundError`](https://docs.python.org/3/library/exceptions.html#FileNotFoundError) instead of an `OSError`? That way the user / program using the Tax-Calculator API has a little more info about what is going on.

chusloj:
@hdoupe `FileNotFoundError` would be more appropriate.

MattHJensen:
Thanks @chusloj for the contribution and @hdoupe and @MaxGhenis for reviews. Merging.

PR #2528, opened 2021-01-11
[WIP] Update tax-calc data

andersonfrailey:
This PR updates all of the files from `taxdata` that `taxcalc` uses. The corresponding `taxdata` PRs are [#367](https://github.com/PSLmodels/taxdata/pull/367) and [#369](https://github.com/PSLmodels/taxdata/pull/369). Details for the changes made can be found in those PRs and the effects on `taxcalc` projections of each can be found [here](https://github.com/PSLmodels/taxdata/blob/master/history/reports/taxdata_report_2020-12-20.pdf) and [here](https://github.com/PSLmodels/taxdata/blob/master/history/reports/taxdata_report_2021-01-02.pdf).

In addition, I had to make a few changes to other parts of `taxcalc`. Specifically I added a few variables to the PUF, updated the expected totals and MTRs for the CPS and PUF, and modified some of the tests to work with the new fils.


MattHJensen:
@andersonfrailey could you tell me a bit more about the provenance of the new puf variables?

andersonfrailey:
@MattHJensen, yes. With the taxdata refactor, we now use the same set of code used to create `cps.csv.gz` to create the CPS-based tax units that are used in our statistical matching process. These tax units include the identifying information needed to map them back to the CPS tax unit they came from, indicators for blind head and spouse, and, when C-TAM benefits aren't available, the benefit information that can be found in the CPS. So since we're now using that for the matching, they've all made their way to the final file. But if we want to move slower in adding these files, it's easy for me to remove them.

MattHJensen:
@andersonfrailey, thanks for the explanation of those variables. Could you leave them off for now so I can investigate further first?

MattHJensen:
The puf.csv, cps.csv, and associated files will be distributed directly from the Tax-Data project going forward, starting with the latest update. Consequently, I am closing this PR. 

@andersonfrailey, thanks for this and the many preceding PRs over the years! Looking forward to seeing the new distributions directly from the Tax-Data project.

PR #2536, opened 2021-01-22
[Review Ready] Update use cases as of 01-22-2021

chusloj:
This PR updates Tax-Calculator use cases as of January 22, 2021.

cc @MattHJensen

### DIFF
--- a/docs/use_cases.md
+++ b/docs/use_cases.md
@@ -1,268 +1,272 @@
 Recorded use cases
 ==================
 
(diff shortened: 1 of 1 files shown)


MattHJensen:
Thanks @chusloj. I'm thinking it would be helpful to flip the chronology so all of the numbers don't need to be updated in the future. Does that make sense to you?

chusloj:
@MattHJensen Makes perfect sense.

chusloj:
@MattHJensen Use cases are now sorted by earliest mention.

Issue #2520, opened 2020-12-23
Additional standard deduction for blind and aged


jdebacker:
The "Additional standard deduction for blind and aged", `STD_Aged` for 2013 for widowed is set to 1,500 [here](STD_Aged).

But the way I read [IRS Publication 501](https://www.irs.gov/pub/irs-prior/p501--2013.pdf), `STD_Aged`= 13,400 (from Table 7) - 12,200 (from Table 6) = 1,200.

My question is: am I doing the calculation of this parameter correctly (in which case we need to change the default value)?

cc @MattHJensen

MattHJensen:
@jdebacker, thanks a lot for finding and reporting this. Your calculation appears correct to me, and it looks like the problem for widowed+aged/blind taxpayers extends to other years than 2013. I am adding a bug label here.

jdebacker:
Addressed in PR #2537.

PR #2537, opened 2021-01-26
[Review Ready] Fix standard deduction for blind and aged

chusloj:
This PR updates `STD_Aged`,  "Additional standard deduction for blind and aged" in the current law. @jdebacker found an issue with miscalculated values for widowed individuals in #2520.


MattHJensen:
Thanks @chusloj, this looks good to me. Resolves #2520.

PR #2540, opened 2021-02-01
Update reforms.json 41 to drop ACTC_c to 0 with CTC_c

### LINKED ISSUE #2539: ACTC_c / CTC_c relationship
I just noticed that tests are failing on `master` when running reform 41 in `reforms.json`:

https://github.com/PSLmodels/Tax-Calculator/blob/0a4bc0613d0a2f2cb72d4c33f0431ee8ca6ac350/taxcalc/tests/reforms.json#L412-L419

image

This error makes sense because `ACTC_c` must be less than `CTC_c`:

https://github.com/PSLmodels/Tax-Calculator/blob/0a4bc0613d0a2f2cb72d4c33f0431ee8ca6ac350/taxcalc/policy_current_law.json#L13809-L13814

But `CTC_c` doesn't have a similar range validator which is why the error wasn't triggered before:

https://github.com/PSLmodels/Tax-Calculator/blob/0a4bc0613d0a2f2cb72d4c33f0431ee8ca6ac350/taxcalc/policy_current_law.json#L13457-L13462

The reason this error is being thrown now is that all parameter values are validated after #2532 -- not just the parameters that were updated. Before, the validation logic didn't check whether the value of `ACTC_c` was valid after `CTC_c` was validated. Now it does.

To fix this, we can update reform 41 to be:
```json
"41": {
    "baseline": "2017_law.json",
    "start_year": 2015,
    "value": {"CTC_c": 0, "ACTC_c": 0},
    "name": "Eliminate child tax credit",
    "output_type": "iitax",
    "compare_with": {"Tax Expenditure": [57.3, 57.0, 57.1, 56.8]}
},
```

hdoupe:
Resolves #2539

### DIFF
--- a/taxcalc/tests/reforms.json
+++ b/taxcalc/tests/reforms.json
@@ -412,7 +412,7 @@
   "41": {
       "baseline": "2017_law.json",
       "start_year": 2015,
-      "value": {"CTC_c": 0},
+      "value": {"CTC_c": 0, "ACTC_c": 0},
       "name": "Eliminate child tax credit",
       "output_type": "iitax",
       "compare_with": {"Tax Expenditure": [57.3, 57.0, 57.1, 56.8]}


MattHJensen:
Thanks for the explanation over in #2539. Merging.

PR #2567, opened 2021-03-02
Add testing for Mac OS in GH Actions

chusloj:
This PR adds the os `macos-latest` to the testing suite in GH actions. I figure we can make testing fully robust by including all OSs.


chusloj:
Due to the setup of Mac OS virtual machines in GH actions (see [this PR](https://github.com/actions/virtual-environments/issues/602)), testing on Mac OS won't work properly without a good bit of security-related additions to our GH action for testing, so I think it's best to leave Mac OS out. Closing.

Issue #2530, opened 2021-01-12
Add performance tests


MattHJensen:
It would be know and track how long it takes for taxcalc to finish common tasks and how long the test suite takes.

chusloj:
@MattHJensen 

`pytest` itself already has a way to profile individual tests within a test session using the `--durations` flag, but the output is quite clunky. Here's the output of `pytest test_decorators.py --durations=0`:

``` bash
=================================================== slowest durations ====================================================
0.28s call     taxcalc/tests/test_decorators.py::test_magic_apply_jit
0.21s call     taxcalc/tests/test_decorators.py::test_function_no_parameters_listed
0.18s call     taxcalc/tests/test_decorators.py::test_ret_everything_iterate_jit
0.15s call     taxcalc/tests/test_decorators.py::test_magic_iterate_jit
0.14s call     taxcalc/tests/test_decorators.py::test_function_takes_kwarg
0.14s call     taxcalc/tests/test_decorators.py::test_function_parameters_optional
0.12s call     taxcalc/tests/test_decorators.py::test_faux_function_iterate_jit
0.01s call     taxcalc/tests/test_decorators.py::test_force_no_jit
0.01s call     taxcalc/tests/test_decorators.py::test_iterate_jit_raises_on_unknown_return_argument
0.01s call     taxcalc/tests/test_decorators.py::test_iterate_jit_raises_on_no_return

(38 durations < 0.005s hidden.  Use -vv to show these durations.)
```

---
For benchmarking individual functions within a given test, we can use the [`pytest-benchmark`](https://pytest-benchmark.readthedocs.io/en/latest/) plugin. Example from `test_calcfunctions.py`:

``` python
def test_DependentCare(skip_jit, benchmark):
    """
    Tests the DependentCare function
    """

    test_tuple = (3, 2, 100000, 1, [250000, 500000, 250000, 500000, 250000],
                  .2, 7165, 5000, 0)
    test_value = calcfunctions.DependentCare(*test_tuple)
    expected_value = 25196

    benchmark(calcfunctions.DependentCare, (*test_tuple))
    assert np.allclose(test_value, expected_value)
```

Output:
``` bash
----------------------------------------------- benchmark: 1 tests -----------------------------------------------
Name (time in us)         Min      Max    Mean  StdDev  Median     IQR  Outliers  OPS (Kops/s)  Rounds  Iterations
------------------------------------------------------------------------------------------------------------------
[...]

PR #2570, opened 2021-03-05
Local test profiling

chusloj:
Addresses #2530.

This PR introduces test profiling locally using [`pytest_harvest`](https://smarie.github.io/python-pytest-harvest/). Whenever any subset of tests run, statistics from those test(s) will be compared from statistics in the provided `test_stats_benchmark.csv` file. If the difference between new and old tests exceeds a given tolerance value (for now it's 1.0s), the difference is printed to the user.

I'm thinking that a new GH action should be created to update the benchmark file when commits are made to the master branch.

**One caveat**: Accurate statistics can only be collected if the test suite is not parallelized.

cc @MattHJensen


MattHJensen:
Thanks very much @chusloj! 

> I'm thinking that a new GH action should be created to update the benchmark file when commits are made to the master branch.

Do you still think this is the right approach to keeping the benchmark file maintained? If so, could you add that to this PR?

chusloj:
@MattHJensen The new GH action is already committed to this PR as `build_new_test_benchmark.yml`. I forgot to mention that.

MattHJensen:
Thanks @chusloj, merging!

PR #2576, opened 2021-03-17
Fix testing benchmark GH action

chusloj:
Attempts to fix GH action to build testing benchmark file. @MattHJensen I believe `persist_credentials: false` must be turned off so that the `github-actions` bot can be authenticated for making commits.


MattHJensen:
Thanks @chusloj

Issue #2582, opened 2021-04-05
Pandas>=1.2.0 doesn't support Python 3.6


chusloj:
From `Pandas>=1.2.0` Python 3.6 is [no longer supported](https://pandas.pydata.org/pandas-docs/dev/whatsnew/v1.2.0.html#increased-minimum-version-for-python). I believe this is causing current tests to fail on windows using 3.6. The `conda-forge` package already ships with `python>=3.7` so I think testing for 3.6 should be dropped.

cc @MattHJensen

chusloj:
Addressed with #2584. Closing.

PR #2512, opened 2020-11-19
Update setup-miniconda action

MaxGhenis:
See https://github.com/conda-incubator/setup-miniconda/issues/101

PR #2584, opened 2021-04-05
[Ready] Remove Python 3.6 from tests in GH Actions

chusloj:
This PR removes Python 3.6 from the GH action's test suite. Addresses #2582.

cc @MattHJensen

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -1,4 +1,4 @@
-name: Build Package and Test Source Code [Python 3.6, 3.7, 3.8, 3.9]
+name: Build Package and Test Source Code [Python 3.7, 3.8, 3.9]
 
 on: [push, pull_request]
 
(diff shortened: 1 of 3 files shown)


jdebacker:
@chusloj Thanks for taking care of this!  Do you want to try to see if we can get code coverage working now (uncommenting the upload report lines)?  Also, we might want to add `pycodestyle` testing back, which was dropped with the GH Action, I think.

jdebacker:
Code coverage is back!  Thanks @chusloj!

chusloj:
@jdebacker Code coverage is back, but `pycodestyle` tests interfere with the test suite profiling, so that has been left out.

@MattHJensen In the environment, `fsspec` has to be temporarily pinned to `<=0.8.7` due to an issue in reading files from the web which is causing Paramtools to fail to read JSON files from URLs. cc @hdoupe

MattHJensen:
@chusloj could you help me understand that 1.45% reported reduction in code coverage?

chusloj:
@MattHJensen Looks like coverage now is compared to coverage from the last PR merged before codecov was shut off, which looks to be #2512. Regarding this patch, codecov says that [this `ValueError` message](https://github.com/PSLmodels/Tax-Calculator/blob/46dcd32a61c3b6c037a34905b0c4e5d13ec25eda/taxcalc/data.py#L104) is not covered.

PR #2585, opened 2021-04-06
Vectorize all tax calculation functions

chusloj:
This PR does the following:
1. Stores all tax variables as columns of a `pandas` dataframe attached to the underlying `Records` object
2. Decouples all functions in `calcfunctions.py` from the `iterate_jit` decorator and implements either (a) a function that takes advantage of `numpy`/`pandas` broadcasting or (b) a function that `numpy` will manually vectorize.

The tax logic within some functions may be less clear due to the introduction of `np.where` and `np.select`.

Thanks @hdoupe for changing the `Records` and `Data` classes from being `numpy`- to `pandas`- dependent

---
TODO:
- [ ] Clean up `calcfunctions` by creating a function that collections each tax function's relevant arguments, parameters and outputs in order to save space.
- [ ] Make the test suite run.


chusloj:
I may not get the chance to finish this PR, so here's the current status of the test suite (last thing to take care of before merging). The following tests are still failing:


``` python
test_calcfunctions.py::test_calc_and_used_vars
test_calcfunctions.py::test_calc_and_used_vars
test_cpscsv.py::test_agg
test_utils.py::test_create_tables
test_utils.py::test_table_columns_labels
```

---

The most common problems have been the following:
1. Missing or duplicate `MARS` indexing. There might be missing terms of `[MARS - 1]` in `calcfunctions.py` or there might be an unnecessary addition of the same term in the function calls within `calculator.py`.
2. Misspecification of `np.where` statements. For example, in a previous commit the following was added:

https://github.com/PSLmodels/Tax-Calculator/blob/0bc4d2b4f59e0f2e130a3d9ac8f9665644649818/taxcalc/calcfunctions.py#L2186-L2188

In line 2188, `eitc` is specified as the return value whereas before it was mistakenly specified as `0.` – line 2188 previously reset the value of `eitc` to `0.`.

3. Issues with conditions in `np.select`. For example, a previous commit made the following change (previous code is commented out):

https://github.com/PSLmodels/Tax-Calculator/blob/4698c29699f888058fa63d6821dda2990e6286c8/taxcalc/calcfunctions.py#L835-L848

Line 846 previously did not specify the nested `if` statement using a `DSI != 1` like so:
``` python
np.logical_and(DSI != 1, MIDR == 1)
```

chusloj:
@MattHJensen @rickecon @jdebacker I think that `test_calc_and_used_vars` and `test_function_args_usage` can be removed from the test suite. I believe these just check that all arguments, parameters, and returned variables are accounted for in functions used by the `@iterate_jit` decorator. With the decorator's removal, Python can know normally scan for misspecified function calls.

Additionally, some additional arguments have been added to certain functions (they're in the function body but weren't previously specified in function arguments) and all output variables have been removed from function arguments.

PR #2572, opened 2021-03-13
Tests for AfterTaxIncome, ExpandIncome, LumpSumTax, and FairShareTax

amshoulders:
Please see the attached file for the test values I used. This PR is ready to be reviewed.

[ASTests.xlsx](https://github.com/PSLmodels/Tax-Calculator/files/6137120/ASTests.xlsx)

PR #2587, opened 2021-04-21
Fix bugs in profiling suite

chusloj:
This PR fixes in a bug in the profiling suite noticed in #2572. The profiling suite will no longer throw an error if new tests are introduced.

Thanks @jdebacker @amshoulders 
cc @MattHJensen

### DIFF
--- a/taxcalc/tests/conftest.py
+++ b/taxcalc/tests/conftest.py
@@ -167,5 +167,4 @@ def pytest_sessionfinish(session):
 
     # Save new test stats to disk including time diff
     new_stats_df['time_diff'] = merge_df['time_diff'].values
-    print(new_stats_df['time_diff'])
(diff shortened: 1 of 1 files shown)


chusloj:
@MattHJensen Loss of code coverage is for the same reasons described in #2584.

MattHJensen:
> @MattHJensen Loss of code coverage is for the same reasons described in #2584.

Thanks for addressing this @chusloj. Could you help me understand why the codcov baseline wasn't updated with #2584?

chusloj:
> why the codcov baseline wasn't updated

@MattHJensen I'm not sure – I'm not able to see the full codecov report on their website. I'm receiving different errors when I visit the report saying that I'm either not allowed to view the report or that there's an error in Codecov's operations.

Issue #2595, opened 2021-05-31
Rename SS_thd50 and SS_thd85 to SS_thd1 and SS_thd2


MaxGhenis:
I think `SS_thd50` and `SS_thd85` are the only policy parameters with their default values in their name. This also diverges from `SS_percentage1` and `SS_percentage2` which are set to 0.5 and 0.85, respectively. Renaming the thresholds to `SS_thd1` and `SS_thd2` would be more consistent.

The variable descriptions also specify 50% and 85% in them, unlike other thresholds.

TishaJhabak1014:
@jdebacker Is this issue solved??

jdebacker:
@TishaJhabak1014 This issue is still open.  It will make a change that is backwards incompatible, so I will create new 4.0 branch to merge this change into whenever someone decides to tackle it.

PR #2596, opened 2021-05-31
Documentation changes to Social Security taxability policy parameters

MaxGhenis:
1. Populates `section_1` and `section_2` for `SS_percentage1` and `SS_percentage2` (I think this will add it to the Tax-Brain web app; see https://github.com/PSLmodels/Tax-Brain/issues/180).
2. Removes reference to 50% and 85% from the `SS_thd50` and `SS_thd85` policy parameter descriptions, for consistency with other threshold descriptions (see #2595).
3. Collapses `section_2` to "Social Security Benefit Taxability" for all four parameters.
4. Runs `make_uguide.py`, which looks to have been outdated prior to this, so a couple changes separate from Social Security taxability are included.


jdebacker:
@MaxGhenis Thanks for this PR clarifying documentation.  

Things look good -- just one question: did you want to go ahead and give better names to the `SS_thd50` and `SS_thd85` parameters in this PR?

MaxGhenis:
> Things look good -- just one question: did you want to go ahead and give better names to the `SS_thd50` and `SS_thd85` parameters in this PR?

Since this would be backward-incompatible, I was thinking of saving it for a major release. If you agree, should I make a PR with that change into a new `4.0.0` branch? And/or, is there a way to duplicate parameter names to provide both as an option until the major release?

jdebacker:
> Since this would be backward-incompatible, I was thinking of saving it for a major release. If you agree, should I make a PR with that change into a new 4.0.0 branch? 

Ah - I forgot about thist.  Let's save these changes for a later release.  

Thanks for this PR.  Merging now.

PR #2607, opened 2021-07-21
Update compare data for taxdata 0.3.2

jdebacker:
This PR updates expected results to match what I'm getting from testing with the latest PUF and CPS weights + the newly release PUF with `taxdata` 0.3.2.

Differences between model results and SOI targets shrink overall (and the match with the CPS file appear to be much better). However, a couple variables see increased differences with the data updates:
* SETAX differences are larger when using the PUF or CPS.
* FCTC differences are larger when using the PUF

cc @MattHJensen @andersonfrailey


MattHJensen:
Thanks @jdebacker. Merging.

Issue #2608, opened 2021-07-26
Child Tax Credit bonus for children under 6 years old


JackLandry:
The Child Tax Credit gives an additional [$600 for children under 6](https://www.taxpolicycenter.org/briefing-book/what-child-tax-credit) that doesn't seem to be implemented in [policy_current_law.json](https://github.com/PSLmodels/Tax-Calculator/blob/master/taxcalc/policy_current_law.json). Should that be added, or am I missing something?

jdebacker:
@JackLandry Thanks for looking into this.  

The CTC changes that occurred through the ARPA required the creating of new CTC parameters that allowed one to model the system under ARPA and the system before (and after) the ARPA changes.  Thus, the place to look is at the `CTC_new_` parameters.

[`CTC_new_c_under6_bonus` is set to a non-zero value](https://github.com/PSLmodels/Tax-Calculator/blob/297931b7c3cb5d7af5859ac7e6d54b3f255ddaa9/taxcalc/policy_current_law.json#L16164), but looking at this, it does appear to be incorrect.  It has a value of $1,000 instead of $600, as was specified in the [APRA.json](https://github.com/PSLmodels/examples/blob/main/psl_examples/taxcalc/ARPA.json) file and [in a unit test](https://github.com/PSLmodels/Tax-Calculator/blob/297931b7c3cb5d7af5859ac7e6d54b3f255ddaa9/taxcalc/tests/test_reforms.py#L130).

Thirdhuman:
If I do 2 dry runs and take the difference, one with "CTC_new_c_under6_bonus" set to $0 and the other with it set to $600, I get no difference between them.

jdebacker:
@Thirdhuman Thanks for checking this.  So that I can confirm, can I get some details:

* Are you using Tax-Calculator 3.2.0?
* What does your "reform" JSON (or dictionary) look like?
* What year(s) are you looking at?

Thirdhuman:
Is there a way to check confirm the version that's being used via the command line?  
It should the most recent one, unless I've somehow got 2 versions on my computer and am calling the wrong one.

![Screen Shot 2021-07-26 at 5 15 13 PM](https://user-images.githubusercontent.com/20157774/127060282-4090b4d9-5ad9-4a01-9791-f4980dbe571c.png)

![Test](https://user-images.githubusercontent.com/20157774/127059834-d3221c02-6775-470f-a557-e3a87be6d637.png)

[...]

PR #2609, opened 2021-07-27
Fix value of CTC_new_c_under6_bonus

jdebacker:
This PR corrects the value for the `CTC_new_c_under6_bonus` in the `policy_current_law.json` file.  The value is currently set at $1,000 in 2021, but @JackLandry correctly points out that the value should be $600 (see Issue #2608).

With this correction, I update the `test_reforms.py::test_round_trip_tcja_reform` to test years 2019-2022, rather than just 2020, which would have caught this error.


jdebacker:
@MattHJensen All tests pass for me locally.  This PR is ready for your review.

MattHJensen:
This looks good to me. Thanks to @jacklandry for helping w/ identification and @jdebacker for fixing.

Issue #2621, opened 2021-09-14
ARPA CTC credit amounts with TCJA phase in?


JackLandry:
Is it possible to model the status-quo ARPA child tax credit benefit amounts and phase-outs, while replacing full refundability with the TCJA phase in structure? There are so many CTC parameters I could have my understanding wrong, but I think you can get the TCJA phase-in with the ARPA bonuses for low income households, but you'd have to do a separate run for higher AGI units to get the phase outs? Here is my attempt to write the .json, which gives me an idiosyncratic error... Any help on the current error, as well as if my .json looks correct for what I'm trying to do, would be much appreciated 

.json
```
{
    "ACTC_c": {"2021": 2100.0},
    "ACTC_rt_bonus_under6family": {"2021": 420.0},
    "ACTC_rt": {"2021": 0.15}, 
    "ACTC_Income_thd": {"2021": 2500.0},
    "ACTC_ChildNum": {"2021":  3}, 
    "CTC_c": {"2021": 3000.0},
    "CTC_c_under6_bonus": {"2021": 600.0},
    "CTC_include17": {"2021": true},
    "CTC_refundable": {"2021": false},
    "CTC_ps": {"2021": [200000.0, 400000.0, 200000.0, 200000.0, 400000.0]},
    "CTC_prt": {"2021": 0.05}
}
```

Error:
```
file "/opt/anaconda3/envs/test/bin/tc", line 10, in <module>
    sys.exit(cli_tc_main())
  File "/opt/anaconda3/envs/test/lib/python3.9/site-packages/taxcalc/cli/tc.py", line 156, in cli_tc_main
    tcio.init(input_data=inputfn, tax_year=taxyear,
  File "/opt/anaconda3/envs/test/lib/python3.9/site-packages/taxcalc/taxcalcio.py", line 282, in init
    self.errmsg += pol.parameter_errors
TypeError: can only concatenate str (not "dict") to str
```

jdebacker:
@JackLandry These are great questions. 

The error is you are receiving is because `ACTC_rt_bonus_under6family` is a rate and is bounded by 0 and 1.  If I change 420 to 1, I can run the reform.

PR #2625, opened 2021-09-23
Fix parameter error parsing in taxcalcio module

### LINKED ISSUE #2622: Handling ParamTools error messages
@hdoupe From the errors reported in Issue #2621, it looks like there has been change in the `paramtools` errors from being strings to dictionaries and therefore an error is occurring on line 282 of `taxcalcio.py` and users don't see the intended validation error message.  No current `taxcalc` unit tests seem to catch this.  Can  you confirm that that this is the case and perhaps propose a fix (and ideally a test to catch this type of thing going forward)?

hdoupe:
Resolves #2622.

### DIFF
--- a/taxcalc/parameters.py
+++ b/taxcalc/parameters.py
@@ -689,11 +689,11 @@ class Parameters(pt.Parameters):
 
     @property
     def parameter_warnings(self):
-        return self.errors or ""
+        return self.errors or {}
 
     @property
     def parameter_errors(self):
-        return self.errors or ""
+        return self.errors or {}
 
     @staticmethod
     def _read_json_revision(obj, topkey):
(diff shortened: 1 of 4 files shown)


taxcalc/parameters.py
```diff
@@ -689,11 +689,11 @@ def num_years(self):
 
     @property
     def parameter_warnings(self):
-        return self.errors or ""
+        return self.errors or {}
```
hdoupe, line 692: The parameter errors are a dictionary that looks like this:

```python
{'AMEDT_rt': ['AMEDT_rt[year=2021] 1.8 > max 1 ']}
```

This change sets the default value to be a dictionary to keep the type consistent.

jdebacker:
@hdoupe Looks very good and I was able to run all tests locally, plus get the proper error message from the json file given in Issue #2621.

But before merging, can you get that new line of code covered?  I think it just involves passing a reform file to the baseline argument in the new test you wrote, which does catch all the new lines for the reform file error messages.

Thanks for your work on this!

hdoupe:
@jdebacker Looks like the code coverage test is passing now! Thanks for reviewing this.

jdebacker:
@hdoupe Thanks for getting that line covered -- and for the PR.  Merging.

Issue #2663, opened 2022-09-07
Incorrect EITC amount for two kids in 2022


JackLandry:
The EITC credit amount for two kids (EITC_c) in [current_law.json](https://github.com/PSLmodels/Tax-Calculator/blob/master/taxcalc/policy_current_law.json) is incorrect for two kids. It's coded as 6064, but it should be 6164. The IRS reference is [here](https://www.irs.gov/pub/irs-drop/rp-21-45.pdf).

jdebacker:
Resolved with PR #2673

PR #2673, opened 2023-01-25
Fix incorrect value for max EITC in 2022

jdebacker:
This PR fixes the incorrect max EITC value in 2022 for filers with 2 kids, pointed out by @JackLandry in Issue #2663.


jdebacker:
This PR is ready for review, @MattHJensen

PR #2694, opened 2023-08-14
File format for validation files from TAXSIM

jdebacker:
This PR updates the `input_setup.py` file in the TAXSIM 35 validation directory to remove all spaces from the file and only use commas to separate values.  Without this, random spaces maybe inserted into the output files which cause problems when they are read into DataFrames.


jdebacker:
Simple change here, merging.

PR #2696, opened 2023-09-11
Better mapping of taxsim s corp income to taxcalc

jdebacker:
When doing the validation between TAXSIM and `taxcalc`, one needs to add the `scorp` income (which under TAXSIM is defined as active s-corp income) and `otherprop` (which in TAXSIM includes passive s corp income + some other items on Sch E) together to get `e02000`, which represents all Sch E income in `taxcalc`.


bodiyang:
@jdebacker thanks Jason, looks good to me

jdebacker:
Thanks for review @bodiyang.  Merging

Issue #2639, opened 2021-12-26
Move license to root of repo


MaxGhenis:
[GitHub recognizes licenses](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/adding-a-license-to-a-repository) as `LICENSE` or `LICENSE.md` in the root of a repo and puts it in the repo sidebar. For example, here's how openfisca-core looks given their [license](https://github.com/openfisca/openfisca-core/blob/master/LICENSE):
![image](https://user-images.githubusercontent.com/6076111/147398970-2ac21539-4260-42b0-b5fd-56bc6f083d5b.png)

The Tax-Calculator license is currently in [docs/about/license.md](https://github.com/PSLmodels/Tax-Calculator/blob/master/docs/about/license.md), so it doesn't show up.

martinholmer:
@MaxGhenis,  Good idea!  Why don't you submit a pull request that fixes the problem you've identified in issue #2639.

jdebacker:
Completed with PR #2706

PR #2706, opened 2023-11-17
Move license file

jdebacker:
This PR moves the license file out of the Jupyter Book documentation files to the root directory of the repo.  Resolves Issue #2639.


jdebacker:
@MattHJensen This PR is ready for your review.

The change is to move `license.md` out of the Jupyter Book documentation and to the root directory so that it shows up on the main page of the GH repo. 

If we still feel like it should be in those docs (e.g., at https://taxcalc.pslmodels.org) then I am happy to duplicate the file and have it in both places. But Jupyter Book can't handle compiling with files above the root directory for the book.

MattHJensen:
@jdebacker, I think duplicate the license and then good to merge. Nice to have both places. License doesn't change, so duplication isn't a maintenance issue.

jdebacker:
@MattHJensen Do this PR look good now?

Issue #2632, opened 2021-11-16
Parameterize maximum net capital losses 


MaxGhenis:
[Capital losses are currently limited](https://www.irs.gov/taxtopics/tc409) as described by the IRS here:
>If your capital losses exceed your capital gains, the amount of the excess loss that you can claim to lower your income is the lesser of $3,000 ($1,500 if married filing separately) or your total net loss shown on line 21 of Schedule D (Form 1040).

This is currently hard-coded as `-3000` in the `ALD_InvInc_ec_base` function:

https://github.com/PSLmodels/Tax-Calculator/blob/cf452aaeb0bf7a5e13c11581963c3fafe86edb16/taxcalc/calcfunctions.py#L432

and the `CapGains` function:

https://github.com/PSLmodels/Tax-Calculator/blob/cf452aaeb0bf7a5e13c11581963c3fafe86edb16/taxcalc/calcfunctions.py#L541

Where
https://github.com/PSLmodels/Tax-Calculator/blob/cf452aaeb0bf7a5e13c11581963c3fafe86edb16/taxcalc/calcfunctions.py#L413-L414

jdebacker:
@MaxGhenis I think this is a duplicate of Issue #2624.  If you agree, would you mind closing this issue (any moving any additional detail you think is necessary over to 2624)?

MaxGhenis:
Oops, yes, I only searched for 3000 :)

PR #2707, opened 2023-11-18
Parameterize the limitation on capital losses

jdebacker:
This PR addresses Issue #2632 by parameterizing the capital loss limitation.

### DIFF
--- a/docs/guide/policy_params.md
+++ b/docs/guide/policy_params.md
@@ -14,4750 +14,4766 @@ are placed in an Other Parameters group at the end of the section.
 
 ### Offsets
 
-####  `parameter_indexing_CPI_offset`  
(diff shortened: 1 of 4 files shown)


jdebacker:
@MattHJensen This PR is ready for your review. Thanks!

MattHJensen:
@jdebacker, do you think we should protect puf and cps users from setting this parameter outside of the ranges supported by their data? Or say this param isn't compatible with those data?

jdebacker:
@MattHJensen Good point about data limitations affecting this parameter's usefulness.  To address this concern, I've added a note about this in the "notes" metadata field for the `capital_loss_limitation` parameter.

Note that I am reluctant to set `"puf": false` in the `compatible_data` field because the parameter will work with the PUF as expected, so long as the values is less than or equal to the current law value (as the PUF does not provide information on capital losses beyond the current law limit).

Do you think this is satisfactory?

PR #2713, opened 2024-02-04
A few minor changes to make input data testing easier

martinholmer:
This pull request contains a few changes that will make it easier to conduct validation tests of new datasets being developed for use by Tax-Calculator.  These changes have no effect on the taxes calculated by the model.

* In the `Makefile` change the logic of the `clean` target to correspond to the existing logic in the `package` target.

* In the `Makefile` add the `--disable-warnings` option to just the `pytest-cps` target.

* Add a `SS_all_in_agi` boolean parameter to the `policy_current_law.json` file and use it in the `SSBenefits` function in the `calcfunctions.py` module.  This parameter permits the estimation of the tax expenditure associated with the less-than-complete taxation of social security benefits.

* Promote a local variable in the `GainsTax` function in the `calcfunctions.py` module to be a global variable defined in the `records_variables.json` file.  This variable makes it easier to check that the existing `CG_nodiff` parameter is doing what it is designed to do.

* Suppress a pandas deprecation warning in the `taxcalcio.py` module.

* Break up several overly long lines in the `pytest_sessionfinish` function in the `conftest.py` module.


jdebacker:
Thanks for this PR @martinholmer!  Merging.

PR #2725, opened 2024-03-12
Revise ppp.py script so that it does not crash

martinholmer:
This pull request refactors the `ppp.py` script so that it does not stop with a fatal error.  The refactoring does not change the logic of the `ppp.py` script.

Also, I added decimal points to the new float parameter recently added because ParamTools writes this parameter's values as floats (instead of integers).   These changes are purely cosmetic, but will avoid spurious changes to the `policy_current_law.json` file when the `ppp.py` script is executed.


jdebacker:
Thanks for the updates, @martinholmer , merging.

PR #2730, opened 2024-03-26
Correct a few historical income tax bracket parameter values

martinholmer:
A year or two ago, there were a few mistakes made in entering by hand the values of some `II_brk?` and `PT_brk?` amounts.   This pull request fixes those incorrect tax parameter values.  It also fixes the `reforms/rounding2022.json` reform used in the round-trip-reform test so that the round-trip-reform test does not fail.  Also, the expected aggregate test results for 2020 and 2021 changed slightly given the correction in the tax bracket values.


jdebacker:
@martinholmer Thanks for this update.  I was surprise to see how off some of those values were.  From some spot checking, I found no errors in your updated values.  Merging.

PR #2731, opened 2024-03-27
Add script to automate use of projected CBO tax policy parameters

martinholmer:
This pull request adds a high-level script called `cbo2pcl.py` that reads a downloaded CBO spreadsheet and incorporates its tax policy parameters into the `policy_current_law.json` file.   This makes the future values of several major Tax-Calculator policy parameters be closer to official CBO parameter values than was the case in prior versions of Tax-Calculator.

These improvements in tax policy parameter values cause some changes in the `cpscsv_agg_expect.csv` and `pufcsv_agg_expect.csv` files.


martinholmer:
@hdoupe,  I would appreciate your review of Tax-Calculator PR #2731, which adds some CBO projected values to the `policy_current_law.json` file.   After adding those values, I find that any attempt to change the value of the `parameter_indexing_CPI_offset` parameter generates a model crash.  The easiest way to see the details of the crash is to comment out the `@pytest.mark.skip` statement just before the new `test_change_cpi_offset` (which is located at the bottom of the `test_policy.py` file) and execute that test.

Am I doing something wrong in PR #2731?
Or is there a problem in `paramtools`?

jdebacker:
@martinholmer, I know @hdoupe is super busy at the moment so might not be able to respond quickly.  Because of that, let me offer my understanding of the issue in case it helps you find a solution.

The traceback from the error in `test_policy.py::test_change_cpi_offset` suggests a missing key value at [line 368 of `parameters.py`](https://github.com/PSLmodels/Tax-Calculator/blob/9c633fcc402e4c78ceb0bdfda4888002319c9219/taxcalc/parameters.py#L368). The `_auto` key that is missing is generated if the parameter values is created "automatically" by extending previously known values (see [this notebook](https://github.com/PSLmodels/ParamTools/blob/0107eceb4a732e3c0eea32dd83b56ae253d009cb/docs/api/indexing.ipynb) for some details and examples on extending indexed parameter values with ParamTools).

In this PR, you are adding known values for many parameters up through 2034 in the `current_law_policy.json` file, but the docstring of the `parameters.adjust_with_indexing` function notes: 
```
The current values of parameters that are not being adjusted
(i.e. are not in params) are deleted after the last known year,
with the exception of parameters that revert to their pre-TCJA
values in 2026. Instead, these (2026) parameter values are
recalculated using the new inflation rates.
```
With admitting a less than complete understanding of what is happening, I think you are running into a conflict between the now hardcoded values for 2026-2034 of several parameters that were expected to revert to pre-2018 values (but updated for inflation indexing) after 2026 and the implicit assumption in Tax-Calculator that, unless they are explicitly adjusted, those values would be inferred from their 2017 values (and the inflation index values).

If this is true, one solution is to remove the bit of `parameters.py` that makes this assumption about the value of reverting TCJA parameters for years 2026 and beyond. However:
1. Such a change could have effects backwards compatibility of the project (perhaps not a big deal, since we are likely to release a version 4.0 soon that will do that anyway).
2. While the hardcoded CBO values from 2026-2034 provide the best estimates of those values under current law, and therefore make the updated `current_law_policy.json` useful for baseline analysis, there might be implications for certain reforms that would like to, e.g., keep the "revert to 2017 values" of current law, but want to provide an different set of CPI index values (which maybe exactly what the failing test is trying to test).

So while I'm not sure of the best path forward, I hope this helps inform the cause of the issue and potential solutions.

martinholmer:
@jdebacker,  Thank you for your thoughtful commentary on the changes in PR #2731.  

After further investigation, it seems as if the structure of Tax-Calculator does not support policy parameters having different last known years.   And it does not support changing the value of the `parameter_indexing_CPI_offset` more than once.   So, the changes proposed in this pull request are far from ideal.   As a result, this pull request is being closed in favor of pull request #2732.

PR #2538, opened 2021-01-29
Remove direct reliance on PUF and CPS files: Stage 1 (PUF)

chusloj:
This PR is the first in a pair of PRs that will remove Tax-Calculator's direct reliance on `puf.csv` and `cps.csv` files. This PR focuses on all content regarding the PUF, and the next will focus on the CPS file.

Each PR will be updated in stages:
1. Functions (and docstrings) and the CLI
2. Testing
3. Documentation

PR #2740, opened 2024-04-18
Add capability to handle tmd input

martinholmer:
Add code to handle new private `tmd.csv` in a manner like the private `puf.csv` is handled.

The private `tmd.csv` file contains 2021 data and has its own public `tmd_weights.csv.gz` file containing weights for 2021 through 2074.  Also, the new public `growfactors.csv` file has factors extending through 2074.

These three files are created in the [tax-microdata repository](https://github.com/PSLmodels/tax-microdata-benchmarking) using the following scripts:
* `create_taxcalc_input_variables.py` produces the private `tmd.csv` file
* `create_taxcalc_sampling_weights.py` produces the public `tmd_weights.csv.gz` file
* `create_taxcalc_growth_factors.py` produces the public `growfactors.csv` file

The code in this pull request still specifies the `Policy.LAST_BUDGET_YEAR` to be 2034; a subsequent pull request will increase it to 2074 (which has already been tested).


conda.recipe/meta.yaml
```diff
@@ -8,10 +8,10 @@ build:
 
 requirements:
   build:
-    - "python<3.12"
-    - "numpy>=1.14"
-    - "pandas>=1.2.0"
-    - "bokeh>=1.4.0, <3.0.0"
+    - "python>=3.9, <3.12"
+    - "numpy>=1.20, <2.0"
```
jdebacker, line 12: Is it important to provide an upper bound on the versions for certain packages, like `numpy` and `pandas`?  Historically, Tax-Calculator has tried to remain compatible with the most recent versions of the dependencies.

jdebacker:
@martinholmer I've reviewed this PR and had one small question left in my review.

But there is a larger issue that this PR touches on that might be worth discussion before merging.  There has been work in the past (see PR #2538) to disentangle the tax calculator logic from the data used for specific calculations.  Things like the `cps_constructor` method and the default use of the `puf.csv` are a convenience, but come at a cost of making implicit assumptions about the quality of certain data files that are outside the scope of Tax-Calculator.  This PR adds additional methods for convenience with a(nother) particular data file.  I can see the advantage of these conveniences, but on the other hand it does push in the opposite direction that the project was hoping to go.  cc @MattHJensen thoughts?

martinholmer:
@jdebacker said in PR #2740:

> But there is a larger issue that this PR touches on that might be worth discussion before merging. There has been work in the past (see PR https://github.com/PSLmodels/Tax-Calculator/pull/2538) to disentangle the tax calculator logic from the data used for specific calculations. Things like the cps_constructor method and the default use of the puf.csv are a convenience, but come at a cost of making implicit assumptions about the quality of certain data files that are outside the scope of Tax-Calculator. This PR adds additional methods for convenience with a(nother) particular data file. I can see the advantage of these conveniences, but on the other hand it does push in the opposite direction that the project was hoping to go.

There has been no work on PR #2538 in **over three years**.    Seems like there is little or no momentum in the "direction that the project was hoping to go".

jdebacker:
Thanks of the PR @martinholmer. Merging.

PR #2741, opened 2024-05-06
Python version

jdebacker:
This PR bumps the Python version used in some of the GitHub Actions and in the docs, making it consistent with those versions we test against, 3.9-3.11.


martinholmer:
@jdebacker,  Thanks so much for your quick response in fixing these Python versions.

PR #2762, opened 2024-06-18
Remove TAXSIM-27 validation tools

jdebacker:
This PR removes the TAXSIM-27 validation tools, which are obsolete now that TAXSIM-35 is used.


jdebacker:
@martinholmer @MattHJensen:

I do not see [`puf_fuzz.py`](https://github.com/PSLmodels/Tax-Calculator/blob/master/taxcalc/validation/puf_fuzz.py) or [`csv_taxdiffs.py`](https://github.com/PSLmodels/Tax-Calculator/blob/master/taxcalc/validation/csv_taxdiffs.py) called anywhere in this repository. 

I'm also not sure what `puf_fuzz.py` has to do with validation.

Would it be ok to remove these files? Or is there an use case that I am not aware of.

martinholmer:
@jdebacker asked in PR #2762:
> Would it be ok to remove these files?

Yes, as long as it is true that they are not called anywhere in the Tax-Calculator repository.

PR #2777, opened 2024-07-22
Add growfactors file for a current policy baseline

jdebacker:
This PR adds a growth factors file for use when using a current policy (i.e., TCJA permanence) baseline.


martinholmer:
@jdebacker said:
> This PR adds a growth factors file for use when using a current policy (i.e., TCJA permanence) baseline.

This doesn't make sense to me (neither does the title).

**Current-law policy** calls for the end of many TCJA provisions.

It is a **reform** that would extend those provision after 2025.

PR #2785, opened 2024-07-28
Fix qualified business income calculation

martinholmer:
The 4.2.0 version of Tax-Calculator does not calculate qualified business income (QBI) correctly when the business income is from self-employment (that is, reported on Schedule C).  The reason for the miscalculation is that the above-the-line deduction for the self-employment (SECA) tax on Schedule C income is not being subtracted from Schedule C gross income to get "Net QBI".   Here are two IRS forms as generated by 2022 Turbo Tax for a tax unit in which the head has $200,000 in Schedule C income and qualified business income of $188,208 with the difference being the $11,792 above-the-line deduction for the self-employment (SECA) tax.

First the form that computes the self-employment tax:

Screenshot 2024-07-28 at 1 39 10 PM

.

Then the QBID (Form 8995) summary:

Screenshot 2024-07-28 at 1 44 20 PM


jdebacker (approved):

jdebacker:
@martinholmer Thanks for finding and fixing this bug.  Merging changes.

PR #2818, opened 2024-09-29
Fix ALD_BusinessLosses_c parameter values after 2025

### LINKED ISSUE #2812: Values of ALD_BusinessLosses_c in policy_current_law.json and ext.json
(I'm assuming this param is about the excess business loss provision on which TCJA put a limit.)

If that's correct, I have a few questions:

1. It seems the IRA extended the provision through the end of 2028. (See [here](https://answerconnect.cch.com/topic/491200287cf010008520d8d385ad169402/excess-business-losses).) So the limit in `policy_current_law.json` should bounce back to infinity starting 2029, not 2026. Accordingly, `ext.json` should specify an inflation-adjusted limit for the year 2029 (and onward), not 2026.

2.  As things stand now, the limit is $270,000 for 2022, $289,000 for 2023, and, according to the article linked in point 1, $305,000 for 2024. However, the 2026 value in `ext.json` is at $319,821.19, which seems a bit too small even if that's when the provision will expire.

3. More generally, is `ext.json` automatically refreshed when relevant underlying params are changed—for example, when PR #2519 is merged?

martinholmer:
Fixes #2812.


martinholmer:
@towashington,  Would you like to review Tax-Calculator PR #2818?

towashington:
PR #2818 looks good to me. Thanks, @martinholmer!

martinholmer:
@towashington, Thanks for looking over PR #2818.

Issue #144, opened 2015-02-24
Extend taxcalc to accomodate CBO Budget Options


MattHJensen:
We should extend functions.py and params.json to accommodate several of the individual income tax revenue options from the [CBO Budget Options](https://www.cbo.gov/sites/default/files/cbofiles/attachments/49638-BudgetOptions.pdf). 

The options we should start with are:

- [x] Option 51: Eliminate the deduction for state and local taxes
- [x] Option 52: Curtail the deduction for charitable giving
- [ ] Option 53: Limit the value of itemized deductions
- [x] Option 61: Increase the payroll tax rate for medicare hospital insurance by 1 percentage point

MattHJensen:
Closing in favor of #229

PR #2880, opened 2025-03-31
drop Python 3.9 in docs

bodiyang:
This PR is to drop Python 3.9 in the documentation. 

I've seen errors coming up when trying to run Tax-Calc under Python 3.9 locally. 

This also matches with the Python version requirement in [taxcalc-dev environment](https://github.com/PSLmodels/Tax-Calculator/blob/master/environment.yml).

### DIFF
--- a/docs/usage/starting.md
+++ b/docs/usage/starting.md
@@ -2,7 +2,7 @@ Getting started
 ===============
 
 Tax-Calculator packages are available for Windows, Mac, and Linux computers
-that have the free Anaconda Python 3.9, 3.10, 3.11, or 3.12 distribution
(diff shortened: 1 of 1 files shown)


jdebacker:
@bodiyang Test failures here are due to updates in the `marshmallow` package upon which `paramtools` depends.  A short term solution (before `paramtools` is updated) is to pin `marshmallow<4.0.0` in the `taxcalc-dev` environment.

martinholmer:
@bodiyang,  Thanks for the documentation fix in PR #2880.
I've included your doc fix in PR #2885, which avoids the ParamTools bugs you experienced in  #2880.  For details on that see [ParamTools issue #144](https://github.com/PSLmodels/ParamTools/issues/144).

PR #2882, opened 2025-04-08
Bokeh package update

bodiyang:
To update Bokeh package usage in the model: 

* new grammar 
* package requirement in environment file >= 3.7


taxcalc/utils.py
```diff
@@ -1295,8 +1295,9 @@ def pch_graph_plot(data,
     fig.title.text_font_size = '12pt'
     line = data['line']
     fig.line(line.index, line.pch, line_color='blue', line_width=3)
-    fig.circle(0, 0, visible=False)  # force zero to be included on y axis
-    zero_grid_line_range = range(0, 101)
+    fig.scatter(0, 0, visible=False)  # force zero to be included on y axis
+    zero_grid_line_range = line.index
+    zero_grid_line_height = [0] * len(line.index)
     zero_grid_line_height = [0] * len(zero_grid_line_range)
```
jdebacker, line 1301: If the above changes give the same result (i.e., replacing the `range(0, 101)` with `line.index`, then this list and line 1299 can be removed as they are redundant (you can just as easily call `line.index` and `zero_grid_line_range`).

bodiyang: have made a cosmetic change, keeping `zero_grid_line_range` to be consistent with `zero_grid_line_height` at line 1301

Issue #2922, opened 2025-06-24
Add `PT_qbid_limited` boolean policy parameter


martinholmer:
In the discussion of issue #2917, @martinholmer suggested that there was little (if any) cost involved in removing the troublesome policy parameters `PT_rt?`, `PT_brk?`, and the associated `PT_*` (for example, `PT_top_stacking`) because a user interested in analyzing the effects of differential taxation of pass-through income could specify, using the `PT_qbid_*` parameters, the complete range from taxing pass-through income the same as other income to not taxing pass-through income at all.

The basic idea is to use existing parameters to remove any QBID limitation and then vary `PT_qbid_rt` from zero (same tax treatment) to any value up to one (no taxation of pass-through income).   Here is an JSON reform file example:

```
    "PT_qbid_rt": {"2030": 1.0},
    "PT_qbid_taxinc_thd": {"2030": [9e99, 9e99, 9e99, 9e99, 9e99]},
    "PT_qbid_prt": {"2030": 0.0}
```

The second and third lines of the above reform file remove all the major QBID limitations except for the limitation in this [code fragment](https://github.com/PSLmodels/Tax-Calculator/blob/1b54c855200b949fb216645f61de57d68805ae7c/taxcalc/calcfunctions.py#L1248-L1251).

What is needed is a new boolean policy parameter, `PT_qbid_limited`, that is true when QBID is being simulated (as under TCJA) and can be set to false when the `PT_qbid_rt` parameter is being used to conduct an analysis of the effects of differential taxation of pass-through income.   The new parameter will be used to either execute or skip the code fragment.  This enhancement will be done as soon as PR #2921 is merged and Tax-Calculator 5.0.0 packages are released.

PR #2923, opened 2025-06-24
Add `PT_qbid_limited` policy parameter and associated logic

martinholmer:
Resolves issue #2922.

Use of the new parameter is more extensive than envisioned in issue #2922, so the way to use the `PT_qbid_*` parameters to simulated differential taxation of pull-through income is to use a JSON reform file like the following:
```
{
    "PT_qbid_rt": {"2025": 1.0},  // 1.0 implies pass-through income is completely untaxed
    "PT_qbid_limited": {"2025": false}
}
```
For example, changing the value of `PT_qbid_rt` to 0.5 in the JSON reform file above, produces an intermediate policy in which pass-through income is taxed even more lightly than it is under TCJA policy.


martinholmer:
@jdebacker, Now that the `5-0-0` branch has been merged into the master branch and packages have been released for Tax-Calculator 5.0.0 version, PR #2823 is ready for your review.

PR #2921, opened 2025-06-22
Release Tax-Calculator 5.0.0 version

martinholmer:
This pull request represents a major release of Tax-Calculator, with
changes that make Tax-Calculator 5.0.0 incompatible with earlier
versions of Tax-Calculator.

First, the changes are described.  Then a list of the benefits and
costs of these changes is presented.

This benefit-cost analysis indicates that the benefits of these
changes are substantial and will be experienced by all model users and
developers, while the costs of these changes are quite small because
few (if any) users will be affected by these changes and there are
straightforward ways for affected users to work around any limitations
caused by these changes.

### Model Changes

- **Rename two OASDI benefit taxation policy parameters** (PR 2918)
- **Remove eight `PT_rt?` and seven `PT_brk?` and four related `PT_*` parameters** (PR 2919)
- **Remove eight policy parameters that limit itemized deductions** (PR 2920)

The first change simply renames existing parameters with no change in
tax logic.  The other two changes remove infrequently used parameters
in order to achieve the benefits listed below.  Despite the removal of
these 27 policy parameters and associated tax-calculation logic,
Tax-Calculator 5.0.0 still has ways of taxing pass-through income more
lightly than other income is taxed, and it still has several ways of
limiting itemized deductions (see this [discussion](https://github.com/PSLmodels/Tax-Calculator/issues/2917#issuecomment-2977321460) for details).

### Benefits and Costs

**Benefits for Model Users** 

All model users experience the benefit of faster model execution
described below.  These faster-execution benefits are particularly
significant for users who use Tax-Calculator to compute marginal tax
rates for each tax unit.

In addition, the removal of the `PT_rt?` and `PT_brk?` parameters
eliminates a long-standing problem that these parameters cause users
difficulty because users often forget to change **both** the `PT_*`
and `II_*` parameters when they want to simulate a simple change in
the income tax rate schedule.  Many users don't know anything about
the `PT_*` parameters, and are puzzled about the results they get
when they change just the `II_*` income tax rate schedule.

**Benefits for Model Developers** 

All model developers experience the benefit of faster model execution
described below.  These faster-execution benefits accelerate model
development because model testing takes less time.

In addition, the removal of these 27 policy parameters and associated
code simplifies tax-calculation logic making it easier to add new
parameters that will be needed to characterize current-law policy
after the expiration of TCJA.

And furthermore, the removal of these 27 policy parameters will make
it easier to develop a LLM interface to Tax-Calculator that can assist
[...]

PR #2924, opened 2025-06-24
Refactor Taxes function logic to reduce model execution time

martinholmer:
No change in tax-calculation logic, but the refactoring of the Taxes function reduces Tax-Calculator execution time by about 2.3 percent (using the same ten-year run as used for the speed test reported in PR #2921).


jdebacker:
@martinholmer Thanks for these simplifications of the code and the initial tests of the new OBBBA parameters.

PR #2930, opened 2025-07-09
Add new SeniorDed policy parameters and associated tax-calculation logic

martinholmer:
Represents the OBBBA deduction (available to both itemizers and non-itemizers) for seniors.


jdebacker:
@martinholmer Looks good to me. Thanks for these updates.

PR #2931, opened 2025-07-13
Add new minimum QBID policy parameters and associated QBID-calculation logic

martinholmer:
OBBBA adds a minimum QBID for those with qualified business income above a minimum level.


martinholmer:
@jdebacker,  Thanks for the review of PR #2931.   I'll merge after the tests run again.

Issue #2949, opened 2025-08-09
Change Indexed `FST_AGI_thd` Parameters to Non-Indexed


maxkossek:
There are two parameters in [policy_current_law.json](https://github.com/PSLmodels/Tax-Calculator/blob/master/taxcalc/policy_current_law.json) which don't seem to be inflation-indexed based on previous policy. We should change these parameters such that `Indexed=False`. Once set to false, we could remove the parameter from the [rounding.json](https://github.com/PSLmodels/Tax-Calculator/blob/master/taxcalc/reforms/rounding.json).

Examples of these non-indexed, but set to indexed, parameters include:
- `FST_AGI_thd_hi` stays at 2,000,000.
- `FST_AGI_thd_lo` stays at 1,000,000.

martinholmer:
@maxkossek,  Thanks for raising this issue.   I ran into this in PR #2948, so this would be a simplification.

I'll do this after both PR #2943 and PR #2948 are merged.

PR #2951, opened 2025-08-10
Make the two `FST_AGI_thd_*` parameters be non-indexed

martinholmer:
Resolves issue #2949.
Also adds `CTC_c` to the `taxcalc/reforms/rounding.json` reform.


martinholmer:
@maxkossek,  Are the changes in PR #2951 what you had in mind when you raised issue #2949?

PR #2957, opened 2025-08-26
Add documentation on using sub-national TMD weights with CLI tool

martinholmer:
Documents environmental variable used to generate sub-national estimates of federal taxes using TMD input data files.

### DIFF
--- a/docs/usage/data.md
+++ b/docs/usage/data.md
@@ -61,7 +61,7 @@ file.
 
 The [tax-microdata
 repository](https://github.com/PSLmodels/tax-microdata-benchmarking)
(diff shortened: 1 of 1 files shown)


maxkossek:
> @maxkossek, is this new documentation clear enough for you to understand how to generate NM estimates of federal taxes with the CLI tool?

@martinholmer Yes, this looks easy to understand. An example use case of a congressional district and adding the name of the environmental variable `TMD_AREA` in this sentence: "supply the CLI tool with an environmental variable [insert TMD_AREA] that indicates" would make it even clearer.

I've been using the subnational weights with the CLI and it has worked well.

Issue #2959, opened 2025-09-09
Remove PUF-related files and tests from Tax-Calculator repository


martinholmer:
For some years (for example, see the incomplete 2021 PR #2538), there has been a discussion about streamlining the code in Tax-Calculator so that it is less reliant on non-public input data.  The latest thinking on this matter is in the [discussion of PR 2925](https://github.com/PSLmodels/Tax-Calculator/pull/2925#issuecomment-3267335764), which was merged on September 8, 2025.

The plan is to do what @jdebacker said in the above-cited discussion:
```
I am in agreement with [Martin Holmer] that it would be 
useful to keep the publicly available CPS, but remove the
PUF files from Tax-Calculator.
```

The plan is to treat PUF input files in exactly the same way as the non-public TMD files are treated.

PR #2961, opened 2025-09-11
Remove PUF-related files and tests from Tax-Calculator repository

martinholmer:
Resolves issue #2959 and replaces PR #2538.

This PR removes the [taxdata](https://github.com/PSLmodels/taxdata)-produced `puf_weights.csv.gz` and `puf_ratios.csv` files.

So now users who want to use the PUF-related files with Tax-Calculator should put the `puf.csv` file, along with these other two PUF-related files in the same directory where `taxcalc` will find them when using the `tc` CLI or when using the Python API's new `Records.puf_constructor(...)` static method.  Avoid using the revised `Records()` constructor in the Python API unless using custom-data as Tax-Calculator input.   When using CPS, PUF, or TMD input data with the Python API, use the `Records.cps_constructor()`, the `Records.puf_constructor(...)`, or the `Records.tmd_constructor(...)` as appropriate.

Many of the tests have been converted to use the publicly-available CPS data that is still in the Tax-Calculator repository.  Other tests that simply tabulate the PUF data have been deleted because they belong in the [taxdata](https://github.com/PSLmodels/taxdata) repository.  After these test changes, there are only two tests that require non-CPS input data.   These `test_records.py` tests cross-check the data availability information in the `records_variables.json` file with the content of the `puf.csv` file and with the content of the `tmd.csv` file.   Those tests expect those files to be located in the repository's top-level directory (that is, the folder in which the `Makefile` is located).

### DIFF
--- a/docs/recipes/md_src/recipe00.md
+++ b/docs/recipes/md_src/recipe00.md
@@ -45,16 +45,7 @@ from bokeh.io import show, output_notebook
 
 ## Setup
 
-Use publicly-available CPS input file.
-
-NOTE: if you have access to the restricted-use IRS-SOI PUF-based input file
-and you have that file (named ‘puf.csv’) located in the directory
-where this script is located, then you can substitute the following
-statement for the prior statement:
-
-``
-recs = tc.Records()
-``
+Use publicly-available CPS input data included in Tax-Calculator.
 
 ```{code-cell} ipython3
 :hide-output: false


taxcalc/tests/reforms_expect.csv
jdebacker, line 1: Why are these expected values changing?  Is different data being used for the test?  Different parameters?

martinholmer: Yes, the reforms test now use the CPS data (not the PUF data).

taxcalc/tests/test_compare.py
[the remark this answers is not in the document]
jdebacker: My understanding of `test_compare.py` is that it's test of the Tax-Calculator logic and parameter values (for one year of tax law) by validating it model's output against aggregate statistics (those statistic appear to be only for 2015).  

Maybe this isn't the intent of the test.  But my thought was that if we remove it, what is there to test the current law parameters in any year?  And since there aren't unit tests for each function in `calcfunctions.py`, what is validating the logic?

As I explain this, I think I've talked myself into the following: drop `test_compare.py` and open an issue to discuss including at least some of the TAXSIM validation in the CI tests.

jdebacker:
@martinholmer This PR looks good to me.  Thanks for making these changes that had been discussed for so long.

PR #2971, opened 2025-10-25
Update Policy.LAST_KNOWN_YEAR to 2026

martinholmer:
This PR uses the `update_pcl.py` script, which was changed in PR #2967 to include 2026 policy parameter values, to update `taxcalc/policy_current_law.json`.   After doing that, this PR changes `Policy.LAST_KNOWN_YEAR` from 2025 to 2026.   After making those changes, there are seven test failures.  This PR changes the expected test results for six (out of the seven) tests.  The remaining test, which uses `taxcalc/reforms/OBBBA.json` as updated for the new IRS 2026 parameter values, should, as I understand it, produce a value of 0.00, but on this branch produces a value of -10.72.

@maxkossek, as the author of this test, is my understanding correct that it should produce a zero result?   If so, do you see what I have done wrong?


maxkossek:
> This PR uses the `update_pcl.py` script, which was changed in PR #2967 to include 2026 policy parameter values, to update `taxcalc/policy_current_law.json`. After doing that, this PR changes `Policy.LAST_KNOWN_YEAR` from 2025 to 2026. After making those changes, there are seven test failures. This PR changes the expected test results for six (out of the seven) tests. The remaining test, which uses `taxcalc/reforms/OBBBA.json` as updated for the new IRS 2026 parameter values, should, as I understand it, produce a value of 0.00, but on this branch produces a value of -10.72.
> 
> @maxkossek, as the author of this test, is my understanding correct that it should produce a zero result? If so, do you see what I have done wrong?

@martinholmer For the roundtrip test the reform should recreate the values of policy_current_law.json and hence produce the same results since OBBBA is current law. If I'm seeing this correctly, there is a discrepancy for `ALD_BusinessLosses_c` between the `OBBBA.json` and the `policy_current_law.json` for the year 2025. For example, the value for `MARS=single` is 313000 in `policy_current_law.json` while it is set to 256000 in OBBBA.json. The values for `STD` also don't coincide. It looks like you might have placed the 2026 values in policy_current_law.json for these parameters in the year 2025 in OBBBA.json. The 2026 values in `OBBBA.json` looks like they are all good.

martinholmer:
@maxkossek,  Thanks for looking at Tax-Calculator PR #2971.
Turns out that `taxcalc/reforms/OBBBA.json` needed an update to add correct 2026 values for the `STD`, `CTC_c`, `ACTC_c`, and `ALD_BusinessLosses_c` parameters.

PR #3010, opened 2026-04-02
Require behavioral responses with install

jdebacker:
With other packages that depend on `taxcalc`, I get the following error unless I manually install the Behavioral-Responses package:
```
packages/taxcalc/taxcalcio.py", line 16, in <module>
    import behresp
ModuleNotFoundError: No module named 'behresp'
```

I believe adding `behresp` to the list of packages in `install_requires` of `setup.py` will ensure that this critical dependency is installed with `taxcalc`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,6 +26,7 @@ config = {
         "bokeh>=2.4",
         "numba",
         "paramtools>=0.20.0",
+        "behresp"
     ],
     "classifiers": [
         "Development Status :: 4 - Beta",


martinholmer:
@jdebacker said in PR #3010:
> With other packages that depend on taxcalc, I get the following error unless I manually install the Behavioral-Responses package:

That's because these "other packages" are not using the `taxcalc-dev` environment.
Why do we even have a `taxcalc-dev` environment if people don't use it?

It seems like the logical conclusion to the sort of change in PR #3010 is to eliminate completely the `environment.yml` file.  @jdebacker, why don't you just go ahead and do that in PR #3010?

jdebacker:
@martinholmer writes:
> That's because these "other packages" are not using the taxcalc-dev environment.
Why do we even have a taxcalc-dev environment if people don't use it?

The `taxcalc-dev` environment is for developing Tax-Calculator.  When you install `taxcalc` as a dependency, you don't want all the dependencies in the `taxcalc-dev` environment.  Hence the `setup.py` specifying just want is needed for install.  Before you added `behresp` to `taxcalcio.py`, `behresp` was only needed in the development environment.  But now that it `behresp` is imported when loading the package, we need it as an install requirement.

martinholmer (approved):

Issue #3090, opened 2026-07-04
Negative AGI can reduce standard deduction through nonitemizer charitable cap


MaxGhenis:
## Summary

In Tax-Calculator 6.7.1 and current `master`, a filing unit with negative AGI can receive a smaller standard deduction when `STD_allow_charity_ded_nonitemizers` is true, even with zero charitable contributions.

The nonitemizer charitable deduction cap is computed as `ID_Charity_crt_all * c00100`. When `c00100` is negative, this creates a negative cap, and that negative amount is added to `standard`.

## Minimal reproducer

```python
import pandas as pd
from taxcalc import Records, Policy, Calculator

records = Records(data=pd.DataFrame({
    "RECID": [1],
    "MARS": [1],
    "XTOT": [1],
    "age_head": [40],
    "age_spouse": [0],
    "e00200": [0.0],
    "e00200p": [0.0],
    "e00200s": [0.0],
    "p23250": [-3000.0],
    "e19800": [0.0],
}), start_year=2026, weights=None)

calc = Calculator(policy=Policy(), records=records)
calc.advance_to_year(2026)
calc.calc_all()

print(calc.array("c00100")[0])   # -3000.0
print(calc.array("standard")[0]) # observed: 14300.0
```

For the same filing unit with `p23250 = 0.0`, the standard deduction is `16100.0`.

## Observed

On `taxcalc==6.7.1` and current `master` (`40726984aaa36357208df0514aa3eebfe00fd1bc`):

```text
{'p23250': -3000.0, 'c00100': -3000.0, 'e19800': 0.0, 'standard': 14300.0}
{'p23250': 0.0,     'c00100': 0.0,     'e19800': 0.0, 'standard': 16100.0}
```

## Expected

The nonitemizer charitable contribution add-on should not be negative. Negative AGI should cap the charitable add-on at zero, not reduce the standard deduction. In the example above, the expected 2026 standard deduction is `16100.0`.

This also matches the itemized charitable deduction path, which uses positive AGI for the percentage cap.

PR #3091, opened 2026-07-04
Prevent negative AGI from reducing standard deduction

MaxGhenis:
## Summary

An initial approach to resolving the bug reported in issue #3090.

This prevents a negative AGI from turning the nonitemizer charitable deduction cap into a negative amount that reduces the standard deduction.

## Root Cause

`StdDed` computes the reform/legacy CARES nonitemizer charitable contribution cap as:

```python
ID_Charity_crt_all * c00100
```

When `c00100` is negative, that cap is negative. The subsequent `min(...)` calls can then add a negative amount to `standard`, reducing the standard deduction even when cash charitable contributions are zero.

The itemized charitable deduction path already applies charitable percentage limits to positive AGI. This PR applies the same positive-AGI floor in the nonitemizer standard-deduction add-on.

## Validation

- `env -u UV_FROZEN uv run --python 3.13 --with pytest --with-editable . python -m pytest taxcalc/tests/test_calcfunctions.py -q -m stded`
- `env -u UV_FROZEN uv run --python 3.13 --with pytest --with-editable . python -m pytest taxcalc/tests/test_calcfunctions.py -q`
- `env -u UV_FROZEN uv run --python 3.13 --with pycodestyle --with-editable . python -m pycodestyle taxcalc/calcfunctions.py taxcalc/tests/test_calcfunctions.py`
- Manual one-row calculator reproducer: 2026 single filer with `c00100 = -3000` now has `standard = 16100`, matching the same filer with `c00100 = 0`.

### DIFF
--- a/taxcalc/calcfunctions.py
+++ b/taxcalc/calcfunctions.py
@@ -1689,7 +1689,7 @@ def StdDed(DSI, earned, STD, age_head, age_spouse, STD_Aged, STD_Dep,
     # current law; active in 2020-2021 and under reforms).
     # ----------------------------------------------------------------
     if STD_allow_charity_ded_nonitemizers:
-        capped_ded = min(e19800, ID_Charity_crt_all * c00100)
+        capped_ded = min(e19800, ID_Charity_crt_all * max(0., c00100))
         standard += min(capped_ded, STD_charity_ded_nonitemizers_max[MARS - 1])
     return standard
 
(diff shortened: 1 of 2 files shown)


martinholmer:
@MaxGhenis,  Thanks for the #3090 bug report.   As I mentioned in that issue, I think the fix in this pull request does not go far enough.   My research suggests there are no AGI-related caps on cash charity deductions for nonitemizers.  A more comprehensive bug fix is in pull request #3092.    Do you agree with my assessment?

MaxGhenis:
Closing in favor of #3092, which fixes #3090 the right way — removing the AGI-percentage cap from the nonitemizer path entirely (§170(p) has no percent-of-AGI limit) rather than just flooring my narrower fix here.

PR #3092, opened 2026-07-05
Fix logic that calculates cash-charity-deduction amounts for nonitemizers

### LINKED ISSUE #3090: Negative AGI can reduce standard deduction through nonitemizer charitable cap

martinholmer:
Fixes #3090 by correcting the cash-charity-deduction logic for nonitemizers in the `StdDed` function in the `calcfunctions.py` module.  This logic fix causes no aggregate income tax changes that show up in the tests (which typically round tax liability amounts to the nearest one-tenth of a billion dollars).

This logic fix removes the AGI-related capping of cash-charity deductions for nonitemizers because we have not been able to find any evidence that this AGI-based capping is part of the IRS code for the nonitemizer deduction.

One Google search produced the following account at https://skyig.com/blog/charitable-donation-limits/
```
For non-itemizers, the universal charitable deduction is $1,000 for
single filers and $2,000 for married couples filing jointly. Since
this above-the-line deduction is a flat limit applied to cash
donations, it does not rely on your Adjusted Gross Income (AGI) and is
fully available regardless of having a negative AGI.
```

Another Google search produced a link to a [Turbo Tax discussion](https://turbotax.intuit.com/tax-tips/tax-deductions-and-credits/one-big-beautiful-bill-charitable-deduction-changes-for-2026/c10OQHJpK)
that makes it clear that the AGI-based cap on cash contributions applies only to **itemized** deductions.

And finally, asking Claude Code if there are caps on nonitemizer cash charity deductions other than the $1000/$2000 single/married caps, the answer was no there are no other caps.

The key logic change in this pull request is in `calcfunctions.py`:

Screenshot 2026-07-05 at 5 37 10 PM


MaxGhenis (approved):
Reviewed the logic, parameter wiring, reforms, and test deltas (agent-assisted) — this is a clean fix for #3090, and I'm glad it goes further than my #3091. Removing the AGI cap entirely rather than just flooring it at zero is the statutorily correct call, since the nonitemizer above-the-line deduction (§170(p)) has no percent-of-AGI limit — that cap only belongs on the itemized path. I'll close #3091 in favor of this.

What I verified:

- `StdDed` (calcfunctions.py): the add-on is now `standard += min(e19800, STD_charity_ded_nonitemizers_max[MARS-1])`. With the max floored at 0 by its range validator and `e19800` non-negative, the add-on can't go negative, which is exactly the #3090 failure mode.
- Gating now rides entirely on `STD_charity_ded_nonitemizers_max`. Confirmed in policy_current_law.json: 0 for 2013–2019, [300, 600, 300, 300, 300] for 2020–2021, 0 for 2022–2025, [1000, 2000, 1000, 1000, 1000] for 2026+, `indexed: false`. The 2026 amounts and the no-indexing both match OBBBA.
- Removing the bool is handled cleanly: policy.py adds it to the removed-params map (clear "removed in 6.7.2" error), and CARES/OBBBA/NoOBBBA plus the input_data_tests dumps all drop it. The only surviving reference is the intentional redirect.
- cpscsv_agg_expect.csv: the only rows that move are Standard Deduction, Standard Deduction Filers, and AMT Income, and only in the 2020–2021 and 2026–2035 columns — the active years. No tax-liability rows change, so "no aggregate income tax changes" holds. The AMT-income dip is correct rather than a defect: for nonitemizers `c62100 = c00100 - e00700 - qbided - standard`, so a larger standard lowers AMTI, and liability is untouched because the affected filers (cash gifts above 60% of AGI, i.e. very low or negative AGI) are already at zero taxable income and under the AMT exemption.
- test_calcfunctions.py: `expected` is unchanged because the old tuples made the AGI cap non-binding, so old and new code agree there; the new `Charity_max_zero` case reproduces the old off behavior.

Two small things, neither blocking:

1. docs/guide/policy_params.md still carries the `STD_allow_charity_ded_nonitemizers` entry. #3082 stripped its removed params from this file, so regenerating via docs/guide/make/make_params.py would keep the convention — CI won't catch it since test_policy.py only checks section titles.
2. Pre-existing and out of scope: `STD_charity_ded_nonitemizers_max` sets 2020 to [300, 600, 300, 300, 300], but the CARES 2020 cap was $300 per return for every status ($150 MFS); the $600 MFJ amount started in 2021. Doesn't touch this fix — noting it for a possible separate issue.

Approving.

martinholmer:
@MaxGhenis,  Thanks for your detailed review of PR #3092.

Issue #3119, opened 2026-08-07
A new use case to highlight


jdebacker:
The Congressional Research Services has written up a report comparing Tax-Calculator and the microsimulation model at Yale Budget Lab: https://www.congress.gov/crs-product/R49052

martinholmer:
This issue was resolved by PR #3121.

PR #3121, opened 2026-08-10
Add recent CRS use case of TC

jdebacker:
Addresses Issue #3119, adding new CRS use case to the docs.


martinholmer (approved):
@jdebacker ,  PR #3121 wrt CRS use of Tax-Calculator looks fine to me.

Issue #2419, opened 2020-06-17
Unit testing of calcfunctions.py


jdebacker:
While is full code coverage for tests of `calcfunctions.py`, the unit tests essentially only confirm that all function arguments are passed properly and that those passed are used.  There ought to be tests that validate the logic of each function (i.e., that given certain values for inputs, the expected output values are produced).

jdebacker:
cc @MattHJensen This might be a good beginner issue to work on (at least for most of the functions in `calcfunctions.py`).

chusloj:
See #2489 and #2515.

MattHJensen:
Evidence in favor: https://github.com/PSLmodels/Tax-Calculator/issues/2551

jdebacker:
Checklist of functions in `calcfunctions.py` module that have unit tests written specifically for them (or in progress through open PRs):
- [ ] `BenefitPrograms`
- [ ] `EI_PayrollTax`
- [ ] `DependentCare`
- [ ] `Adj`
- [ ] `ALD_InvInc_ec_base`
- [ ] `CapGains`
- [ ] `SSBenefits`
- [ ] `UBI`
- [ ] `AGI`
- [ ] `ItemDedCap`
- [ ] `ItemDed`
- [ ] `AdditionalMedicareTax`
- [ ] `StdDed`
- [ ] `TaxInc`
- [ ] `SchXYZ`
- [ ] `SchXYZTax`
- [ ] `GainsTax`
- [ ] `AGIsurtax`
- [ ] `AMT`
- [ ] `NetInvIncTax`
- [ ] `F2441`
- [ ] `EITCamount`
- [ ] `EITC`
- [ ] `RefundablePayrollTaxCredit`
- [ ] `ChildDepTaxCredit`
- [ ] `PersonalTaxCredit`
- [ ] `AmOppCreditParts`
- [ ] `SchR`
- [ ] `EducationTaxCredit`
- [ ] `CharityCredit`
- [ ] `NonrefundableCredits`
- [ ] `AdditionalCTC`
- [ ] `C1040`
- [ ] `CTC_new`
- [ ] `IITAX`
- [ ] `Taxes`
- [ ] `ComputeBenefit`
- [ ] `BenefitSurtax`
- [ ] `BenefitLimitation`
- [ ] `FairShareTax`
- [ ] `LumpSumTax`
- [ ] `ExpandIncome`
- [ ] `AfterTaxIncome`

[...]

PR #3148, opened 2026-09-22
Add unit tests for CapGainsLoss, AdditionalMedicareTax and NetInvIncTax

akshita317:
Refs #2419 — three of the functions on the checklist in that issue.

## Why

#2419 observes that the `calcfunctions.py` tests "essentially only confirm that all function arguments are passed properly and that those passed are used", and asks for tests that validate each function's logic — that given inputs produce the expected outputs. `CapGainsLoss`, `AdditionalMedicareTax` and `NetInvIncTax` had no such test.

Expected values are derived from the tax forms, not from the functions' own output, so the tests fail if the logic changes.

## What is covered

**`CapGainsLoss`** — Schedule D Part III netting and the line 21 cap:
a net gain, a net loss under the cap, a loss above it for a single filer, the same loss for married-filing-separately (`$1,500`), and a short-term loss netted against a long-term gain before the cap applies.

**`AdditionalMedicareTax`** — Form 8959:
Part I on wages above the threshold; no tax at the threshold; Part II where wages consume part of the threshold (line 11) so only self-employment earnings above the remainder are taxed.

A fourth case pins the **per-spouse zero floor**. Schedule SE is filed separately by each spouse, so a spouse's loss must not offset the other's profit. With `$300,000` of taxpayer Schedule C profit, a `$100,000` spouse loss and wages at the `$250,000` joint threshold:

| | line 8 | tax |
|---|---|---|
| flooring per spouse (current behaviour) | `300000 * 0.9235` | **$2,493.45** |
| summing first, flooring after | `200000 * 0.9235` | $1,662.30 |

The test asserts the first and explicitly asserts the result is *not* the second, so it fails if that floor is ever removed.

**`NetInvIncTax`** — Form 8960 lines 12-17:
investment income below the MAGI excess; the `NIIT_PT_taxed` reform keeping active partnership income in the base; MAGI exactly at the threshold; a MAGI excess smaller than the investment income; negative investment income floored at zero by line 12; and the joint threshold.

## Verification

```
$ pytest taxcalc/tests/test_calcfunctions.py -q
84 passed
```

(69 before this change.) `pycodestyle` is clean and `pylint --disable=locally-disabled` still rates the file **10.00/10**, matching the rating before the change.

The new tests follow the existing style in the module: a single argument tuple plus `expected_value`, parametrized with descriptive ids, and the `skip_jit` fixture.


akshita317:
A note on the empty check list: the two workflow runs for this branch are sitting at `action_required`, which is the approval gate for a first-time contributor rather than a failure — so no CI has actually executed yet. A maintainer approving the runs would start them.

In the meantime, here is what I ran locally against this branch (Python 3.13, taxcalc installed editable):

```
$ pytest taxcalc/tests/test_calcfunctions.py -q
84 passed

$ pycodestyle taxcalc/tests/test_calcfunctions.py
(no output)

$ pylint --disable=locally-disabled taxcalc/tests/test_calcfunctions.py
Your code has been rated at 10.00/10
```

84 passing versus 69 before the change, and the file keeps the 10.00/10 rating it had beforehand.

martinholmer:
@akshita317,  Thank you for your Tax-Calculator contribution in PR #3148.
I'm approving the CI workflows and marking it a DRAFT now, while I review the substance of the unit tests you've added.  I hope to finish the review sometime later this week.    Thanks again for your work on this.

martinholmer:
@akshita317, I realize that the `test_calcfunctions.py` code is not very consistent and not well ordered in the module, so I will undertake a refactoring of the code after your PR #3148 changes are merged into the master branch.

But before I do that merge, I'd like you to make a few changes in the PR #3148 code.   Since it appears that you used a coding agent to help you prepare your PR, I thought I would try using a coding agent to conduct my initial PR review.   I used Claude Code (running Opus 5 with medium effort) to generate a response to the following prompt:

```
Conduct an adversarial code review of the changes on the git branch
pr-3148 (relative to the last commit on the master branch).  Be sure
to highlight the need for comments (on how the expected function
results are calculated) to specify (a) the tax year of the policy
parameter values, (b) the tax year of the IRS form used to calculate
expected results, and (c) form line numbers using a consistent notation.
Mention any other concerns you have.
Write your review to the ./review.md file.
Make no code changes.
```

I'm satisfied with Claude's final suggestions (that is, leaving some things for my followup refactoring), but would like your reactions to the AI-generated PR revision suggestions.

_Here is the complete response generated by Claude:_

# Review of branch `pr-3148`

Scope reviewed: `git diff master...pr-3148` — a single commit
(`e9e7f23f`) adding 176 lines to `taxcalc/tests/test_calcfunctions.py`.
No production code is touched.  Verified locally:

- `pytest taxcalc/tests/test_calcfunctions.py -k "CapGainsLoss or
  AdditionalMedicare or NetInvIncTax"` — 15 passed.
- `pycodestyle` (ignoring W503, W504, E712) — clean.
- `pylint` (with the CI disables) — 10.00/10.
- Every expected value in the diff was re-derived by hand and matches
  the code in `calcfunctions.py` and the current-law values in
  `policy_current_law.json`.

So the branch is correct as written.  The concerns below are about
whether the tests *document* and *discriminate* as well as they appear
to.

---

## 1. Comments must state the tax year of the policy parameter values

Every hard-coded parameter list in the new code is labeled only
"under current law", with no year:

- line 1129: `# Schedule D line 21 net-capital-loss limits by MARS under current law:`
- line 1169: `# Form 8959 line 5/9 thresholds by MARS and the line 7/13 rate under current law`
- line 1173: `# Current-law FICA rates; ...`
- line 1258: `# Form 8960 line 14 thresholds by MARS and the line 17 rate under current law`

"Current law" is a moving target: it is whatever
`policy_current_law.json` says *for some year*, and the reader of the
test has no way to know which year the author had in mind.  Today all
five parameters happen to be year-invariant — `Capital_loss_limitation`,
`AMEDT_ec`, `AMEDT_rt`, `NIIT_thd`, `NIIT_rt`, and the four `FICA_*`
rates each have a single 2013 `value` entry and `"indexed": false` —
which is *why* the tests pass with no `Policy` object anywhere.  That
is exactly the fact the comments should record, because it is the
assumption the tests silently depend on.  If a future baseline adds a
year-specific value (e.g. indexing the $3,000 Schedule D cap, which
has been proposed repeatedly), these tests keep passing while their
comments become false.

Recommended form, one line per constant block:

```python
# Schedule D (2025) line 21 net-capital-loss limits by MARS.
# Values are Capital_loss_limitation under current law for 2025;
# the parameter is not inflation-indexed and is unchanged since 2013.
CAP_LOSS_LIM = [3000., 3000., 1500., 3000., 3000.]
```

Name the Tax-Calculator parameter explicitly (`Capital_loss_limitation`,
`AMEDT_ec`/`AMEDT_rt`, `NIIT_thd`/`NIIT_rt`, `FICA_ss_trt_*`/
`FICA_mc_trt_*`) so a reader can `grep` the JSON.  `CAP_LOSS_LIM` is
currently the only block that never names its parameter.

Stronger alternative, worth considering: assert the linkage instead of
asserting it in prose, e.g. once per block

```python
assert CAP_LOSS_LIM == Policy().to_array('Capital_loss_limitation', year=2025).tolist()
```

That converts a stale comment into a test failure.  If that is judged
too heavy for a unit test of a calcfunction, the year-naming comment is
the minimum.

## 2. Comments must state the tax year of the IRS form used

No new comment or docstring names a form year, yet all of them cite
line numbers, and IRS line numbers are not stable across years.  A
reader in 2029 cannot check "Form 8960 line 14" without knowing which
year's Form 8960 is meant, and the very line the code maps to
(`c01000` → "Form 1040 line 7") was line 13 on the 2018 Form 1040 and
line 6 on the 2019 Form 1040.

Note also that the pre-existing docstrings in `calcfunctions.py` are
inconsistent on this point, so the new tests cannot simply inherit a
year: `NetInvIncTax` says "per Form 8960 (2025)", `AdditionalMedicareTax`
gives 2025 only for the thresholds ("for 2025"), and `CapGainsLoss`
gives no year at all.  The new tests should state the year explicitly
and should use the same year as the function they test — and the
mismatch above is worth fixing in a follow-up so that the two files
agree.

Recommendation: state the year once at the top of each new block and
again in the docstring's first sentence, e.g. "Tests the
`CapGainsLoss` function against the 2025 Schedule D (Form 1040)."

## 3. Form line-number notation is not consistent

The 176 new lines use at least six notations for the same kind of
reference:

| Location | Text |
|---|---|
| 1129 | `Schedule D line 21` |
| 1169 | `Form 8959 line 5/9 ... the line 7/13 rate` |
| 1173 | `Schedule SE line 4c` |
| 1185 | `Schedule C profit or loss` |
| 1194 | `Form 8959 Part I` |
| 1199 | `# line 6 = 300000 - 200000; line 7 = 0.009 * line 6` (no form named) |
| 1224-1226 | prose `line 8` alongside the variable `line8` |
| 1236 | `joint Form 8959 line 8 total` |
| 1241 | `the $250,000 joint threshold` (vs. `$250k` at 1170 and 1259) |

`calcfunctions.py` additionally mixes `Sch D`, `Sch SE`, `Sch E` with
the spelled-out `Schedule`.  Pick one convention and apply it
mechanically in the new code, for example:

- Always `Form NNNN (YYYY) line N`, and `Sch X (YYYY) line N` for
  schedules — never a bare `line N`, even in a one-line inline comment.
- Line ranges as `lines 1-7`, alternative lines as
  `line 5 (= line 9)` rather than `line 5/9`, since `5/9` reads as a
  fraction or a date.
- Dollar amounts spelled the same way throughout (`$200,000`, not a
  mix of `$200k` and `$200,000`).
- Keep the `lineN` variable names, but always pair them with the
  qualified form reference on first use in each test.

The bare `# line 6 = ...` / `# line 8 = ...` comments at lines 1199 and
1224-1225 are the worst offenders: those are the comments a future
reader will use to re-derive the expected numbers, and they name
neither the form nor its year.

---

## Other concerns

### 4. The `joint threshold` NIIT case does not test the joint threshold

`NIIT_JOINT` uses `c00100=300000` with `MARS=2`, so line 15 is
$50,000 while line 12 is only $30,000 — line 12 binds, and the
expected $1,140 is identical to the `nii below excess` case.  I
confirmed by mutation that shifting the threshold index by +1 or -1
(the classic `NIIT_thd[MARS]` vs `NIIT_thd[MARS-1]` bug) still yields
1140.0 for this case.  The test id promises coverage the assertion does
not provide.  Any MAGI below $280,000 fixes it: `c00100=270000` gives
line 15 = $20,000 and an expected $760, and the same tuple read with
the single threshold would give $1,140.  (`magi at threshold` is
insensitive for the same reason: 0.0 either way.)

### 5. Dead assertion in `test_AdditionalMedicareTax_floors_each_spouse`

```python
unfloored = AMEDT_RT * ((300000. - 100000.) * SECA_FRAC)
assert not np.allclose(actual_value, unfloored), ...
```

`expected_value` is 2493.45 and `unfloored` is 1662.30, both computed
in the test.  Once the preceding `assert np.allclose(actual_value,
expected_value)` passes, this assertion cannot fail — it is unreachable
as a failure and adds no coverage.  The useful content is the
explanation, which is already in the docstring.  Either delete the
assertion (keeping the docstring paragraph) or turn it into a real
check of the contrast, e.g. assert that the MFJ result for
`(sch_c_p=300000, sch_c_s=-100000)` equals the result for
`(sch_c_p=300000, sch_c_s=0.)`, which *is* the property being claimed
and would fail if the per-spouse floor were removed.

### 6. `sey_p` / `sey_s` composition is untested

`amedt_tuple` hard-wires `e02100p`, `e02100s`, `k1bx14p`, `k1bx14s` to
zero, so only the Schedule C components ever reach the function.
Dropping a term from either sum in `AdditionalMedicareTax` — e.g.
writing `sey_s = e00900s + e02100s` and losing `k1bx14s` — passes every
new test.  This is the single largest coverage gap in the diff, and it
is in the one piece of logic the branch's docstring flags as a
deliberate departure from `sey`.  Add one case that puts a distinct
nonzero value in each of the six per-spouse inputs so that a dropped or
duplicated term changes the answer.

### 7. Parts I and II are never both nonzero

`wages_only` has line 13 = 0; `se_uses_remaining_threshold` and
`floors_each_spouse` both have line 7 = 0 (wages at or below the
threshold).  So `ptax_amc = line7 + line13` is never exercised with
both addends positive, and a bug such as returning `max(line7, line13)`
would pass.  Add a case with wages above the threshold *and* SE income
(then line 11 = 0 by construction, which is also the Part II branch not
otherwise covered with nonzero wages).

### 8. No MFS or QSS coverage for the two threshold parameters

`AdditionalMedicareTax` and `NetInvIncTax` are tested only at MARS 1
and 2.  MARS 5 is the highest-risk value in the whole diff, because it
is the one place the two parameter lists legitimately disagree:
`AMEDT_ec[4]` is $200,000 (Form 8959 treats QSS like single) while
`NIIT_thd[4]` is $250,000 (Form 8960 treats QSS like MFJ).  A
copy-paste of one list over the other is undetectable today.  MARS 3
($125,000 on both forms) is also untested for these two functions,
though `CapGainsLoss` does cover MARS 3 and that case does discriminate
the index correctly.

### 9. Form 8960 line 4b is only tested where the two inputs are equal

`NIIT_BELOW_EXCESS` and `NIIT_PT_IN_BASE` both set
`e02000 == e26270 == 20000.`, so `e02000 - e26270` and the reversed
`e26270 - e02000` are both 0.  The sign convention survives only
because `NIIT_NEGATIVE` happens to catch a reversal (it would give
$1,786 instead of 0) — i.e. via a case whose stated purpose is the
zero-floor, not the line 4b sign.  Use distinct values
(`e02000=20000., e26270=5000.`) so the `pt taxed` / `nii below excess`
pair differs by exactly `NIIT_RT * e26270` and tests the adjustment
directly.

### 10. Two AMEDT expectations restate the implementation

```python
expected_value = AMEDT_RT * max(0., line8 - (AMEDT_EC[0] - 150000.))
```

re-derives the code's own formula, including the `max(0., ...)`, so a
shared misconception cannot be caught.  This is mitigated — well — by
the follow-up `assert np.allclose(expected_value, 381.15)`, but the
mitigation makes the computed expression redundant.  The
`CapGainsLoss`/`NetInvIncTax` tests get this right: literal expected
values, derivation in the comment.  Prefer that pattern uniformly:
assert against the literal, and put `0.009 * (92350 - 50000)` in the
comment.  (Note the slightly odd construct of asserting on
`expected_value` rather than on `actual_value`; with a literal
expectation that assertion disappears.)

While there: `SECA_FRAC` is derived from the four FICA rates and its
comment claims "i.e. 0.9235" without checking it.  One
`assert np.allclose(SECA_FRAC, 0.9235)` next to the definition makes
the claim self-verifying.

### 11. Missing boundary cases in `CapGainsLoss`

No case has a net loss exactly equal to the cap (-3000, or -1500 for
MARS 3), and no case pairs a long-term loss with a short-term gain
(`p22250 > 0, p23250 < 0`) — the netting test only exercises the
opposite sign pattern.  `CGL_SMALL_LOSS` sets `p23250 = 0.`, so the
long-term argument is zero in two of the five cases.  Cheap additions,
given the function is two lines.

### 12. Module-level constant names are generic in a shared 1,300-line file

`CAP_LOSS_LIM`, `AMEDT_EC`, `AMEDT_RT`, `NIIT_THD`, `NIIT_RT`,
`FICA_SS_ER`, `FICA_SS_EE`, `FICA_MC_ER`, `FICA_MC_EE`, `SECA_FRAC`
are added at module scope, where the file already has collision-prone
names (`tuple1` alone is redefined nine times at module scope in this
file, which is precisely the hazard).
`FICA_SS_ER` / `FICA_SS_EE` are also opaque: `_ER`/`_EE` for
employer/employee is not a convention used elsewhere in the repo, which
spells them `FICA_ss_trt_employer` / `FICA_ss_trt_employee`.  Consider
naming the constants after the parameters they mirror, or scoping them
inside the test/helper.

### 13. Minor comment wording

- 1283 `# MAGI at the threshold: no tax however large the income` —
  "however large the *investment* income"; AGI is pinned at the
  threshold in this case, so "the income" is misleading.
- 1136/1284 `# net loss larger than the cap: limited to $3,000 when
  single` — the $3,000 cap applies to every MARS except MFS, so
  "when single" suggests a filing-status specificity that is not there;
  "when not filing separately" is accurate.
- 1170 `$200k single/HoH/QSS, $250k MFJ, $125k MFS` is correct for
  `AMEDT_ec`, and 1259 `$200k single/HoH, $250k MFJ/QSS, $125k MFS` is
  correct for `NIIT_thd`; because these two lists differ in exactly one
  slot, it is worth saying so explicitly in both comments rather than
  leaving the difference to be noticed.

### 14. Structural inconsistency (observation, not a defect)

`CapGainsLoss` and `NetInvIncTax` are parametrized with ids;
`AdditionalMedicareTax` is four separate functions.  Both styles exist
in the file, and the four-function form buys per-case docstrings, so
this is defensible — but the branch would read more uniformly if the
three simple AMEDT cases were parametrized and only
`floors_each_spouse` kept its own function for the long explanation.

### 15. JIT path untested (pre-existing convention)

All 15 new tests take `skip_jit`, so the `@iterate_jit` wrappers and the
list-vs-array handling of `Capital_loss_limitation`, `AMEDT_ec`, and
`NIIT_thd` are exercised only in pure Python.  This matches the rest of
the file and is not a branch defect; flagging only because the new
tuples pass Python lists where the JIT'd code would see NumPy arrays.

---

## Summary

The diff is correct, style-clean, and better documented than most of the
file it joins — the docstrings explaining *why* the per-spouse floor
matters are a genuine improvement in local convention.  Before merge I
would ask for: the year-and-form-notation cleanup in items 1-3; the two
non-discriminating assertions in items 4-5; and the three real coverage
gaps in items 6-8 (per-spouse SE components, Part I + Part II together,
MARS 5).  Items 9-15 are smaller and can be batched or deferred.

akshita317:
Thanks for this — the review is substantive and I've pushed `9b467ba` addressing items 1-13.

I didn't take the findings on faith; I reproduced each of the discrimination claims before changing anything, and they hold:

**Item 4 (joint NIIT case) — correct, and worse than stated.** I checked all three index choices, not just the off-by-one:

```
c00100=300000, MARS=2:  NIIT_thd[0] -> 1140.0   NIIT_thd[1] -> 1140.0   NIIT_thd[2] -> 1140.0
c00100=270000, MARS=2:  NIIT_thd[0] -> 1140.0   NIIT_thd[1] ->  760.0   NIIT_thd[2] -> 1140.0
```

The case proved nothing about the index because line 12 ($45,000) bound rather than line 15. Fixed with modified AGI of $270,000.

**Item 7 — correct.** With the proposed case, line 18 is $1,731.15 while `max(line7, line13)` would give $900.00, so the addition is now pinned. Previously one addend was always zero in all three cases.

**Item 9 — correct.** With `e02000 == e26270` the line 4b adjustment and its reverse are both zero, so the sign was untested. Now `e02000=20000`, `e26270=5000`, and the `pt taxed` / `nii below excess` pair differs by exactly `0.038 * 5000`.

**Item 6 — correct and, I agree, the largest gap.** All six per-spouse components now carry distinct nonzero values: dropping `k1bx14s` changes the result from $548.559 to $523.6245.

**Item 8 — correct, and MARS 5 was the right one to single out.** I confirmed against `policy_current_law.json` that `AMEDT_ec[4]` is $200,000 while `NIIT_thd[4]` is $250,000. The new surviving-spouse NIIT case gives $760; read against the `AMEDT_ec` list it would give $1,710. Both comment blocks now state the divergence explicitly.

**Item 5 — correct, the assertion was unreachable.** Replaced with the property actually being claimed: the result for a $100,000 spouse loss must equal the result for no spouse self-employment income at all. That one does fail if the per-spouse floor is removed.

**Items 1-3.** Done. I also verified the assumption the tests rest on: `Capital_loss_limitation`, `AMEDT_ec`, `AMEDT_rt`, `NIIT_thd`, `NIIT_rt` and the four `FICA_*_trt_*` rates are each `"indexed": false` with a single 2013 value entry. The comments now say so, name the parameter, and give the year for both the parameter values and the form.

One suggestion I'd rather not take, and I'd welcome your view: the "stronger alternative" of asserting `CAP_LOSS_LIM == Policy().to_array('Capital_loss_limitation', year=2025)`. It would convert a stale comment into a failure, but it makes a unit test of a calcfunction fail when a *parameter* changes rather than when the *logic* does, which is the opposite of what #2419 asks these tests to catch — and `policy_current_law.json` is already guarded elsewhere. It would also pull `Policy` construction into tests that currently need nothing but the function. I took the year-and-parameter naming instead, which the review calls the minimum.

**Items 10-13** are in as well: literal expected values with the derivation in a comment, `SECA_FRAC` asserting its own 0.9235, loss-exactly-at-cap cases for both the $3,000 and $1,500 caps plus a long-term loss against a short-term gain, constants renamed to mirror the parameters, and the misleading "when single" and "however large the income" wording corrected.

**Items 14-15** I've left alone, as you suggested. The parametrized-vs-separate-function split is worth settling in your refactoring rather than by me guessing at the target convention, and the JIT path matches the rest of the module.

The module goes from 84 to 95 passing tests (69 before this PR). `pycodestyle` is clean and `pylint --disable=locally-disabled` still rates the file 10.00/10.

One note: the workflow runs on this branch are still sitting at `action_required`, so CI hasn't executed yet on either commit.

martinholmer:
@akshita317,  Thanks so much for you enhanced-per-review testing code in PR #3148.   As Claude said in the original review, your initial code was better than most of the existing code in ~/test_calcfunctions.py`, which has been written by multiple people over the past five or six years.  Now, after your post-review enhancements, the code in this PR is even better.

As I said before, I'm now going to undertake a refactoring of the `test_calcfunctions.py` code to make it more consistent, etc.

Thanks again for you help in adding the tests and providing the inspiration to streamline the existing `test_calcfunctions.py` code and make those tests more complete.

I will merge PR #3148 after the CI workflow on Github finishes.

